PR #843, opened 2012-06-12
Feature/new tz handling

nickvergessen:
RFC http://area51.phpbb.com/phpBB/viewtopic.php?f=108&t=32826

### DIFF
--- a/phpBB/includes/acp/acp_board.php
+++ b/phpBB/includes/acp/acp_board.php
@@ -901,7 +901,9 @@ class acp_board
 	*/
 	function timezone_select($value, $key)
 	{
-		$timezone_select = phpbb_timezone_select($value, true);
+		global $user;
+
+		$timezone_select = phpbb_timezone_select($user, $value, true);
 		$timezone_select['tz_select'];
 
 		return '<select name="config[' . $key . ']" id="' . $key . '">' . $timezone_select['tz_select'] . '</select>';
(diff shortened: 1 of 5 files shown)


phpBB/includes/datetime.php
```diff
@@ -0,0 +1,158 @@
...
+		{
+			/*
+			* Check the delta is less than or equal to 1 hour
+			* and the delta not more than a minute in the past
+			* and the delta is either greater than -5 seconds or timestamp
+			* and current time are of the same minute (they must be in the same hour already)
+			* finally check that relative dates are supported by the language pack
+			*/
+			if ($delta <= 3600 && $delta > -60 &&
+			  ($delta >= -5 || (($now_ts / 60) % 60) == (($timestamp / 60) % 60))
```
rxu, line 75: @nickvergessen Hey Joas 😃 Sorry for disturbing after a long time, but do you probably remember the logic behind those `-5` and `-60` values to handle timestamps ahead from `now`. Cheers.

nickvergessen: Phew it's a long time, but I guess it was related to your computer time being out of sync which would otherwise should that to post will be in the future?

rxu: Ah, so kind of server time misconfiguration in case it's within a minute with `now`? I'm just about patching this file for a small bug and was thinking about use of `DateInterval` class so trying to figure out what happens here :) Thanks.

PR #2153, opened 2014-03-17
[ticket/7187] Quote smilies error

sant0shg:
If similes is quoted, no space is there. So now added space. PHPBB3-7187

### DIFF
--- a/phpBB/posting.php
+++ b/phpBB/posting.php
@@ -1337,7 +1337,7 @@ if ($mode == 'quote' && !$submit && !$preview && !$refresh)
 {
 	if ($config['allow_bbcode'])
 	{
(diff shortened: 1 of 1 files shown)


marc1706:
@s9e can you confirm that this no longer happens with your new parser? I was unable to reproduce it on the current master branch.

In that case we should be able to close the ticket.

s9e:
@marc1706 Yes, I can confirm that. There's a bunch of regression tests in `tests/text_processing/tickets_data` and for most of them I left a comment on the corresponding ticket but I must have missed this one.

https://github.com/phpbb/phpbb/blob/master/tests/text_processing/tickets_data/PHPBB3-7187.txt
https://github.com/phpbb/phpbb/blob/master/tests/text_processing/tickets_data/PHPBB3-7187.html

PR #2168, opened 2014-03-21
[3.2][ticket/12288] Add ability to create profile urls containing full board url

Elsensee:
http://tracker.phpbb.com/browse/PHPBB3-12288

### DIFF
--- a/tests/functions/obtain_online_test.php
+++ b/tests/functions/obtain_online_test.php
@@ -27,6 +27,11 @@ class phpbb_functions_obtain_online_test extends phpbb_database_test_case
 		$db = $this->db = $this->new_dbal();
 		$config = array(
 			'load_online_time'	=> 5,
+			'force_server_vars'	=> 1,
+			'server_protocol'	=> 'http://',
+			'server_name'		=> 'www.example.org',
+			'server_port'		=> 80,
+			'script_path'		=> '/phpBB',
 		);
 	}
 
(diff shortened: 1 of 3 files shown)


tests/functions/obtain_online_test.php
```diff
@@ -27,6 +27,11 @@ protected function setUp()
 		$db = $this->db = $this->new_dbal();
 		$config = array(
 			'load_online_time'	=> 5,
+			'force_server_vars'	=> 1,
```
marc1706: @nickvergessen @bantu I'm pretty sure this hasn't been answered?

marc1706:
Needs rebase

Nicofuma:
status? IIRC one question is still pending @nickvergessen @bantu @marc1706 @Elsensee 

Note: needs to be rebased onto master

Elsensee:
@Nicofuma I am not sure if I remember it correctly but I think this function should be splitted into several functions inside a new/existing class.

Nicofuma:
which one?

Elsensee:
I think we could put it into user_loader class which currently has a wrapper function for get_username_string();
That way we can even get rid of these ugly static variables in the function. :)

Nicofuma:
I see

Nicofuma:
anyway you need to rebase on master and send a new PR :)

marc1706:
IMHO this is a rather minor change that could be added after feature freeze.

PR #2247, opened 2014-04-05
[3.2][ticket/neufke/12143] Make it possible to use translation in group names

Elsensee:
Original PR #1979 by neufke who is apparently not longer active. So I changed the things which were mentioned by @nickvergessen.

http://tracker.phpbb.com/browse/PHPBB3-12143


marc1706:
Should be modified as outlined above and rebased onto master

Elsensee:
New PR: #3652

PR #2585, opened 2014-06-13
[ticket/12684] Add command user:add

ptitlazy:
Add a command for adding a user.

https://tracker.phpbb.com/browse/PHPBB3-12684

### DIFF
--- a/phpBB/phpbb/console/command/user/add.php
+++ b/phpBB/phpbb/console/command/user/add.php
@@ -1,0 +1,167 @@
+<?php
+/**
+*
+* This file is part of the phpBB Forum Software package.
+*
+* @copyright (c) phpBB Limited
+* @license GNU General Public License, version 2 (GPL-2.0)
+*
+* For full copyright and license information, please see
+* the docs/CREDITS.txt file.
+*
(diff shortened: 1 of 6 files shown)


marc1706:
This also should send emails as user_add() triggers that.

:+1:

marc1706:
Please add the dedicated option I mentioned above. Otherwise it looks good to me.

phpBB/phpbb/console/command/user/add.php
```diff
@@ -0,0 +1,196 @@
...
+			'user_lang'				=> $this->config['default_lang'],
+			'user_type'				=> USER_NORMAL,
+			'user_regdate'			=> time(),
+		);
+
+		if (!function_exists('user_add'))
+		{
+			require_once dirname(__FILE__) . '/../../../../includes/functions_user.php';
+		}
+		user_add($user_row);
```
RMcGirr83: I'm not seeing where any sort of validation is occurring when using this.  EG, for the use of a username and a possible duplicate, or email for that matter (password I guess is minutia).  Are we concerned with this?

PR #2629, opened 2014-06-20
[ticket/12747] Drop support for Firebird

nickvergessen:
https://tracker.phpbb.com/browse/PHPBB3-12747

What should we do on update? Nothing?

### DIFF
--- a/tests/test_framework/phpbb_database_test_case.php
+++ b/tests/test_framework/phpbb_database_test_case.php
@@ -226,18 +226,6 @@ abstract class phpbb_database_test_case extends PHPUnit_Extensions_Database_Test
 		return new phpbb_database_test_connection_manager($config);
 	}
 
-	/**
-	* Converts a match in the middle of a string to uppercase.
-	*
-	* @param $matches The array of matches from a regular expression
-	*
-	* @return string The string with the specified match converted to uppercase
-	*/
-	static public function to_upper($matches)
-	{
-		return $matches[1] . strtoupper($matches[2]) . $matches[3];
-	}
-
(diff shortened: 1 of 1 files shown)


madorin:
Hello!
Just trying to upgrade my board to v3.1 and surprise, no Firebird support. Very sad ..  :( 
I fixed that PHP PDO bug in 7.x branch some time ago.
Can you reconsider adding Firebird support?
Thanks!

nickvergessen:
Sorry to here, but since you are the first one in over 2 years, just shows that it's not worth a lot of work :(

madorin:
It may be just because a lot of users simply does not know how to handle http 500 error code, in case they have specified firebird driver in config on phpbb 3.1 upgrade process.
Firebird probably is most SQL standard compilant dbms from open source world. I saw commit removing FB support, and it does not show a lot of work.

DavidIQ:
As also mentioned in a reply [to your forum post](https://area51.phpbb.com/phpBB/viewtopic.php?f=84&p=297541#p297541):

> Currently there are plans/discussions for dropping our whole database abstraction layer and moving to utilizing Doctrine as the DBAL layer (probably in 3.3), which unfortunately doesn't support Firebird neither. Because of this I see basically no chance of adding support for Firebird, unless we find the time to work upstream, however, having very limited capacities it is most likely not gonna happen.
 I hope someone proves me wrong though.

Even if we undid these changes, which involved 26 files, FB support would still be dropped once we go with Doctrine since [they don't support FB either](http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/introduction.html#introduction).

PR #2794, opened 2014-08-03
[Ticket/12911] convert imageset to sprite

hanakin:
convert all the forum/topic imageset icons into a single sprite to reduce server requests and overall file size

####TODO:
- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] IE
- [x] IOS
- [x] Android
- [x] Fix topic icon

PSD: https://www.dropbox.com/s/l5dq1kdmx7486zt/imageset.psd
Demo: http://12911.midaym.com/viewforum.php?f=2&sid=4d651a2a4d14afd7d883392bbc4985cc
Ticket: https://tracker.phpbb.com/browse/PHPBB3-12911

PHPBB3-12911


Louis7777:
I've used an [online image optimizer](http://optimizilla.com/) to reduce the size of the imageset.png down to 5,02 KB

imageset-min

callumacrae:
Looks good :tada: 

Doesn't this make it really tricky for other people to add images? I'd prefer it if we had a tool that we could use to automatically generate the spritesheet and CSS.

hanakin:
lets not rehash that conversation again. lets just convert to svgs for these in 3.2 or use font-awsome...

callumacrae:
Haha, I wasn't there for the first conversation. Close the PR, then?

hanakin:
@callumacrae i know we said use svgs and I think thats still the better approach. but that will still make the imageset difficult to use as the best method would be to use the ````<svg><use></svg>````

PR #3055, opened 2014-10-22
[ticket/13192] Add method for generating valid user page links based on mod_rewrite

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-13192

### DIFF
--- a/phpBB/includes/functions.php
+++ b/phpBB/includes/functions.php
@@ -2396,26 +2396,7 @@ function build_url($strip_vars = false)
 {
 	global $config, $user, $phpbb_path_helper;
 
-	$php_ext = $phpbb_path_helper->get_php_ext();
-	$page = $user->page['page'];
-
-	// We need to be cautious here.
-	// On some situations, the redirect path is an absolute URL, sometimes a relative path
-	// For a relative path, let's prefix it with $phpbb_root_path to point to the correct location,
-	// else we use the URL directly.
-	$url_parts = parse_url($page);
-
-	// URL
-	if ($url_parts === false || empty($url_parts['scheme']) || empty($url_parts['host']))
-	{
-		// Remove 'app.php/' from the page, when rewrite is enabled
-		if ($config['enable_mod_rewrite'] && strpos($page, 'app.' . $php_ext . '/') === 0)
-		{
-			$page = substr($page, strlen('app.' . $php_ext . '/'));
-		}
-
-		$page = $phpbb_path_helper->get_phpbb_root_path() . $page;
-	}
+	$page = $phpbb_path_helper->get_valid_user_page($user->page['page'], $config['enable_mod_rewrite']);
 
 	// Append SID
 	$redirect = append_sid($page, false, false);
--- a/phpBB/includes/functions.php
+++ b/phpBB/includes/functions.php
@@ -2396,7 +2396,7 @@ function build_url($strip_vars = false)
 {
 	global $config, $user, $phpbb_path_helper;
 
-	$page = $phpbb_path_helper->get_valid_user_page($user->page['page'], $config['enable_mod_rewrite']);
+	$page = $phpbb_path_helper->get_valid_page($user->page['page'], $config['enable_mod_rewrite']);
 
 	// Append SID
 	$redirect = append_sid($page, false, false);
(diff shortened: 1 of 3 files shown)


phpBB/phpbb/path_helper.php
```diff
@@ -445,4 +445,38 @@ public function append_url_params($url, $new_params, $is_amp = true)
...
+		{
+			// Remove 'app.php/' from the page, when rewrite is enabled.
+			// Treat app.php as a reserved file name and remove on mod rewrite
+			// even if it might not be in the phpBB root.
+			if ($mod_rewrite && ($app_position = strpos($page, 'app.' . $this->php_ext . '/')) !== false)
+			{
+				$page = substr($page, 0, $app_position) . substr($page, $app_position + strlen('app.' . $this->php_ext . '/'));
+			}
+
+			// Remove preceding slashes from page name and append to root path
```
marc1706: That's the same thing basically? Do you prefer your version?

nickvergessen: well you do not modify root path ;) so you do not append it to the root path ;)

marc1706: Of course I'm not modifying the root path. I'm appending the modified page name to the root path.

michaelcullum: This is correct although 'append to **the** root path'

@nickvergessen You are `prepending the root path to the page name` or `appending the page name to the root path`.

marc1706: The "the" was left out on purpose.

tests/path_helper/path_helper_test.php
```diff
@@ -421,4 +421,27 @@ public function test_get_web_root_path_from_ajax_referer($referer_url, $board_ur
 	{
 		$this->assertEquals($this->phpbb_root_path . $expected, $this->path_helper->get_web_root_path_from_ajax_referer($referer_url, $board_url));
 	}
+
+	public function data_get_valid_page()
+	{
+		return array(
+			// array( current page , mod_rewrite setting , expected output )
+			array('index', true, 'index'),
+			array('index', false, 'index'),
```
Nicofuma: it's still mixed

marc1706: Hm, seems like I didn't push the changes O.o

marc1706:
@nickvergessen Updated test cases

phpBB/phpbb/path_helper.php
```diff
@@ -445,4 +445,38 @@ public function append_url_params($url, $new_params, $is_amp = true)
...
+			// Remove 'app.php/' from the page, when rewrite is enabled.
+			// Treat app.php as a reserved file name and remove on mod rewrite
+			// even if it might not be in the phpBB root.
+			if ($mod_rewrite && ($app_position = strpos($page, 'app.' . $this->php_ext . '/')) !== false)
+			{
+				$page = substr($page, 0, $app_position) . substr($page, $app_position + strlen('app.' . $this->php_ext . '/'));
+			}
+
+			// Remove preceding slashes from page name and prepend root path
+			$page = $this->get_phpbb_root_path() . preg_replace('@^(?:([\\/\\\])?)@', '', $page);
```
nickvergessen: `$page = $this->get_phpbb_root_path() . ltrim($page, '/');`

marc1706:
@nickvergessen updated

phpBB/includes/functions.php
```diff
@@ -2719,8 +2700,8 @@ function confirm_box($check, $title = '', $hidden = '', $html_body = 'confirm_bo
 	}
 
 	// re-add sid / transform & to &amp; for user->page (user->page is always using &)
-	$use_page = ($u_action) ? $phpbb_root_path . $u_action : $phpbb_root_path . str_replace('&', '&amp;', $user->page['page']);
-	$u_action = reapply_sid($use_page);
+	$use_page = ($u_action) ? $u_action : str_replace('&', '&amp;', $user->page['page']);
+	$u_action = reapply_sid($phpbb_path_helper->get_valid_page($use_page, $config['enable_mod_rewrite']));
```
nickvergessen: We must only to that when `$u_action` is not used.
`$u_action` is a parameter of `confirm_box()` that allows using custom URLs,
and we should not modify this further?

marc1706: We should only apply the session ID if u_action is not used? That doesn't make any sense. What would reapplying the session ID do if we're not using it? And why shouldn't we make sure to apply the sid to the u_action path that might have been modified using the user page?

marc1706:
@bantu @naderman Could you maybe review this one to make sure confirm_box() and build_url() aren't broken in any way by these changes?

PR #3128, opened 2014-11-11
[ticket/13306] add error level to collector

omniError:
https://tracker.phpbb.com/browse/PHPBB3-13306

PHPBB3-13306

### DIFF
--- a/phpBB/phpbb/error_collector.php
+++ b/phpBB/phpbb/error_collector.php
@@ -16,15 +16,27 @@ namespace phpbb;
 class error_collector
 {
 	var $errors;
+	var $error_types;
 
-	function __construct()
+	/**
+	 * Constructor.
+	 *
+	 * The variable $error_types may be set to a mask of PHP error types that
+	 * the collector should keep, e.g. `E_ALL`. If unset, the current value of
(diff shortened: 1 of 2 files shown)


omniError:
@bantu is anything else keeping this from being merged? thanks

phpBB/phpbb/error_collector.php
```diff
@@ -16,15 +16,28 @@
...
-	function __construct()
+	/**
+	 * Constructor.
+	 *
+	 * The variable $error_types may be set to a mask of PHP error types that
+	 * the collector should keep, e.g. `E_ALL`. If unset, the current value of
+	 * the error_reporting() function will be used to determine which errors
+	 * the collector will keep.
+	 *
+	 * @see PHPBB3-13306
```
Nicofuma: We never include this kind of references, so I'm not sure that we should start

nickvergessen: From time to time we do (but more in tests), and I don't think it's bad.

omniError: The tracker page has a longer explanation of the reason for this, I see other projects do it that way and didn't know if there was a style convention for this.

PR #3138, opened 2014-11-12
[ticket/12642] Ensure CPF type specific options are set when editing booleans.

prototech:
prepare_hidden_fields is expected to return null if the option is not sent
in the request. The boolean method returns false instead, which results in
the options being set as false in hidden fields when accessing the first edit
step.

When checking the "Default value" option, there is also a failure to check
whether the "Field type" option is set to checkbox, thus resulting in this
option getting lost as well.

[PHPBB3-12642](https://tracker.phpbb.com/browse/PHPBB3-12642)

### DIFF
--- a/phpBB/phpbb/profilefields/type/type_bool.php
+++ b/phpBB/phpbb/profilefields/type/type_bool.php
@@ -367,29 +367,29 @@ class type_bool extends type_base
 	*/
 	public function prepare_hidden_fields($step, $key, $action, &$field_data)
 	{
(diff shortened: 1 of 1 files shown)


nickvergessen:
@prototech does this also fix https://tracker.phpbb.com/browse/PHPBB3-13282 ?

nickvergessen:
meh, boolean profile fields are so broken again.
Checkboxes display 1 independent from user selection, when its a checkbox with default being checked.
It should react on the user input and it should display the language option name.

PR #3173, opened 2014-11-22
[ticket/13361] Improve the exception listener

Nicofuma:
https://tracker.phpbb.com/browse/PHPBB3-13361

PHPBB3-13361

### DIFF
--- a/phpBB/phpbb/controller/helper.php
+++ b/phpBB/phpbb/controller/helper.php
@@ -184,6 +184,8 @@ class helper
 	* @param string $message The error message
 	* @param int $code The error code (e.g. 404, 500, 503, etc.)
 	* @return Response A Response instance
+	*
+	* @deprecated 3.1.3 (To be removed: 3.3.0) Use exceptions instead.
 	*/
 	public function error($message, $code = 500)
 	{
--- a/phpBB/phpbb/event/kernel_exception_subscriber.php
+++ b/phpBB/phpbb/event/kernel_exception_subscriber.php
@@ -94,7 +94,7 @@ class kernel_exception_subscriber implements EventSubscriberInterface
 				$data['trace'] = $exception->getTrace();
 			}
 
-			$response = new JsonResponse($message, 500);
+			$response = new JsonResponse($data, 500);
 		}
 
 		if ($exception instanceof HttpExceptionInterface)
--- a/phpBB/phpbb/event/kernel_exception_subscriber.php
+++ b/phpBB/phpbb/event/kernel_exception_subscriber.php
@@ -14,6 +14,7 @@
 namespace phpbb\event;
 
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpKernel\Exception\HttpExceptionInterface;
 use Symfony\Component\HttpKernel\KernelEvents;
 use Symfony\Component\HttpKernel\Event\GetResponseForExceptionEvent;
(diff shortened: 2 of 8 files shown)


Nicofuma:
bump

VSEphpbb:
No tests?

phpBB/phpbb/exception/http_exception.php
```diff
@@ -0,0 +1,69 @@
...
+*
+* @copyright (c) phpBB Limited <https://www.phpbb.com>
+* @license GNU General Public License, version 2 (GPL-2.0)
+*
+* For full copyright and license information, please see
+* the docs/CREDITS.txt file.
+*
+*/
+
+namespace phpbb\exception;
```
nickvergessen:
```
FILE: /home/travis/build/phpbb/phpbb/phpBB/phpbb/exception/http_exception.php
--------------------------------------------------------------------------------
FOUND 1 ERROR(S) AFFECTING 1 LINE(S)
--------------------------------------------------------------------------------
14 | ERROR | There must be one blank line after the namespace declaration
| | (PSR2.Namespaces.NamespaceDeclaration.BlankLineAfter)
--------------------------------------------------------------------------------
```

Nicofuma: fixed

phpBB/phpbb/event/kernel_exception_subscriber.php
```diff
@@ -53,23 +54,55 @@ public function __construct(\phpbb\template\template $template, \phpbb\user $use
...
+
+			if (defined('DEBUG'))
+			{
+				$data['trace'] = $exception->getTrace();
+			}
 
-		$this->template->set_filenames(array(
-			'body'	=> 'message_body.html',
-		));
+			$response = new JsonResponse($message, 500);
```
nickvergessen: Do you mean `$data` here, because $data is never used ptherwise?

phpBB/phpbb/exception/exception.php
```diff
@@ -0,0 +1,52 @@
...
+*/
+
+namespace phpbb\exception;
+
+/**
+ * Class exception
+ *
+ * Define an exception which support a language var as message.
+ */
+class exception extends \RuntimeException implements exception_interface
```
naderman: can we call this runtime_exception too then?

tests/event/exception_listener.php
```diff
@@ -0,0 +1,100 @@
+<?php
```
naderman: These tests are not being run as the file does not end in _test.php?

Nicofuma:
Updated

naderman:
@Nicofuma you removed the tests?

Nicofuma:
forgot to add the file.... should be done in 2 minutes

naderman:
Those tests appear to actually fail, so you'll need to fix them.

phpBB/phpbb/controller/helper.php
```diff
@@ -184,6 +184,8 @@ public function route($route, array $params = array(), $is_amp = true, $session_
 	* @param string $message The error message
 	* @param int $code The error code (e.g. 404, 500, 503, etc.)
 	* @return Response A Response instance
+	*
+	* @deprecated 3.1.3 (To be removed: 3.3.0) Use exceptions instead.
```
VSEphpbb: It may be helpful to have a simple example shown here as the alternative to using this method, other than just saying Use exceptions, and which exceptions we can/should use

Nicofuma:
Updated...

phpBB/phpbb/controller/helper.php
Nicofuma: throw new \phpbb\exception("LANG_KEY"); ?

nickvergessen:
you forgot to push?

Nicofuma:
now I do

Nicofuma:
Everything should be good now

phpBB/phpbb/controller/helper.php
VSEphpbb: I meant in the comment, like we do in this cases: https://github.com/phpbb/phpbb/blob/develop/phpBB/includes/functions.php#L3212
And also maybe explain which ones to use...http_exception, runtime_exception, phbb_exception, etc. in which cases, like which one for AJAX or non-AJAX ?

naderman: Don't think that belongs in the depreciation notice, we should document exceptions in general though and when to use which.

nickvergessen: Also what I wondered is, what happened to the status code?

Nicofuma: You need to use the \phpbb\exception\http_exception for that. The runtime_exception will always return an error 500.
And for the ajax requests, it should return a json response automatically.

VSEphpbb: Using helper->error is dictated by extension policies (https://www.phpbb.com/extensions/rules-and-policies/validation-policy/#php #12) so the way this PR intends to replace that with these exceptions should be documented in Ext dev or writers forums with specific examples.

PS: I assume these won't replace trigger_error's current functionality in the ACP yet?

Nicofuma: We can't replace trigger_error yes, this page is only for the "Symfony's Controllers"

About the policy.... the ext team can update the phpbb.com policy if they want. And this system will be explained in the dev doc (so the phpbb.com wiki for now) when I'll find the time.

PR #3176, opened 2014-11-24
[ticket/13346] Fix indentation, spacing and missing divs

Senky:
[PHPBB3-13346](https://tracker.phpbb.com/browse/PHPBB3-13346)

### DIFF
--- a/phpBB/styles/prosilver/template/posting_editor.html
+++ b/phpBB/styles/prosilver/template/posting_editor.html
@@ -183,9 +183,6 @@
 			</dl>
 			<!-- ENDIF -->
 		</fieldset>
-
-		</div>
-	</div>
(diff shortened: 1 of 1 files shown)


phpBB/styles/prosilver/template/posting_editor.html
```diff
@@ -183,6 +183,9 @@
 			</dl>
 			<!-- ENDIF -->
 		</fieldset>
-		<!-- ENDIF -->
+
+		</div>
```
nickvergessen: Well with your patch I get
```
Line 417, Column 7: End tag div seen, but there were open elements.
	</div>
```
on http://validator.w3.org/#validate_by_input
So it seems like you are wrong ;)

PR #3182, opened 2014-11-25
[ticket/12632] Add twig.debug and twig.auto_reload in config.yml

Nicofuma:
https://tracker.phpbb.com/browse/PHPBB3-12632
https://tracker.phpbb.com/browse/PHPBB3-13206

PHPBB3-13206
PHPBB3-12632


Nicofuma:
rebased

phpBB/phpbb/di/extension/core.php
```diff
@@ -68,6 +68,19 @@ public function load(array $configs, ContainerBuilder $container)
 			}
 		}
 
+		// Set the Twig options if defined in the environment
+		$definition = $container->getDefinition('template.twig.environment');
+		$twig_environment_options = $definition->getArgument(6);
```
marc1706: Should probably not use a magic number?

Nicofuma: I don't think a constant will have any sense (and where do you want to define it?)

marc1706: At least add an explanation what 6 means.

PR #3206, opened 2014-11-30
[ticket/12890] Search by custom profile fields for members

Dragooon:
Previous PR (#2776) was for 3.1 branch, this is for 3.2 branch

[PHPBB3-12890](https://tracker.phpbb.com/browse/PHPBB3-12890)


Nicofuma:
@nickvergessen do you remember if something was missing?

By the way, it needs to be rebased onto master

nickvergessen:
I don't think this was done and iirc it also does not work for all types of CPF

marc1706:
Doesn't seem like this will make it into alpha?

PR #3207, opened 2014-11-30
[ticket/13407] Update Symfony to 2.7.*@dev

Nicofuma:
https://tracker.phpbb.com/browse/PHPBB3-13407

PHPBB3-13407


Nicofuma:
bump

marc1706:
Looks good to me.

bantu:
@Nicofuma Looks like we could use `2.7.*@dev` now. What do you think?

marc1706:
I guess we'll have to remove the 5.3.3 build to be compatible with this?

Nicofuma:
php >=5.3.9

right....

Nicofuma:
updated

nickvergessen:
Should probably be discussed and announced prior to the change

bantu:
> Should probably be discussed and announced prior to the change

Using 2.7 LTS is the natural choice. If this implies that we have to update to another minor version, that is just fine. I don't think this needs any discussion. Surely, this needs to be communicated to our users as soon as possible.

bantu:
@Nicofuma Slightly more work required.

bantu:
"Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. Run update to update them."

bantu:
Do you have an RFC for this that needs to be moved?

PR #3226, opened 2014-12-10
[ticket/13420] Correctly filter by lt/gt/equal to 0 posts when pruning users.

prototech:
[PHPBB3-13420](https://tracker.phpbb.com/browse/PHPBB3-13420)

### DIFF
--- a/phpBB/includes/acp/acp_prune.php
+++ b/phpBB/includes/acp/acp_prune.php
@@ -397,7 +397,7 @@ class acp_prune
 			$joined_after = request_var('joined_after', '');
 			$active = request_var('active', '');
 
-			$count = request_var('count', 0);
+			$count = ($request->variable('count', '') === '') ? false : $request->variable('count', 0);
 
 			$active = ($active) ? explode('-', $active) : array();
 			$joined_before = ($joined_before) ? explode('-', $joined_before) : array();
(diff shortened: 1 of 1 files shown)


nickvergessen:
@prototech would be nice to have it ;)

phpBB/includes/acp/acp_prune.php
```diff
@@ -397,7 +397,7 @@ function get_prune_users(&$user_ids, &$usernames)
 			$joined_after = request_var('joined_after', '');
 			$active = request_var('active', '');
 
-			$count = request_var('count', 0);
+			$count = ($request->variable('count', '') === '') ? false : $request->variable('count', 0);
```
Elsensee: Why don't you just use type casting here?

nickvergessen: could also try `$request->variable('count', -1)` ? and then check for `($count >= 0)`

nickvergessen: okay, we use a similar mechanism already for the posts in queue:
```php
$posts_on_queue = (trim($request->variable('posts_on_queue', '')) === '') ? false : $request->variable('posts_on_queue', 0);
```
so I guess its fine?

PR #3227, opened 2014-12-10
[ticket/13366] Ensure plupload_last_gc config entry is dynamic.

prototech:
[PHPBB3-13366](https://tracker.phpbb.com/browse/PHPBB3-13366)


nickvergessen:
I had a similar problem with the dynamic config of the search index creation thing.
So I wonder if we should check that for all dynamic configs from time to time?

PR #3228, opened 2014-12-10
[ticket/13428] Add core events in memberlist.php for teampage

Wolfsblvt:
There are no actual events for the team page to modify the data that is displayed.

There should be two events for that, first one for modifying the query and second one for modifying the template vars for the users

[PHPBB3-13428] Add core events to memberlist.php for team - phpBB Tracker


nickvergessen:
Merged via https://github.com/phpbb/phpbb/compare/6fd8f1b835bb...64d880a15c3f

PR #3234, opened 2014-12-14
[ticket/13154] Event to edit user list for notification in notifications

brunoais:
https://tracker.phpbb.com/browse/PHPBB3-13154

PHPBB3-13154

### DIFF
--- a/tests/notification/base.php
+++ b/tests/notification/base.php
@@ -67,7 +67,7 @@ abstract class phpbb_tests_notification_base extends phpbb_database_test_case
 			$phpEx
 		);
 		
-		$phpbb_dispatcher = $this->phpbb_dispatcher = new phpbb_mock_event_dispatcher();
+		$this->phpbb_dispatcher = new phpbb_mock_event_dispatcher();
 
 		$phpbb_container = $this->container = new phpbb_mock_container_builder();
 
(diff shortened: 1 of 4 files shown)


tests/notification/base.php
```diff
@@ -66,6 +66,8 @@ protected function setUp()
 			$phpbb_root_path,
 			$phpEx
 		);
+		
+		$phpbb_dispatcher = $this->phpbb_dispatcher = new phpbb_mock_event_dispatcher();
```
nickvergessen: You don't need `$phpbb_dispatcher` so don't assign it?

brunoais: I did it to make it the same way as all other are doing. Should I still not assign it?

nickvergessen: Its a defined variable that is not needed, yes please remove it

nickvergessen: The others are needed due to `global $db, $config, $user, $auth, $cache, $phpbb_container;`

brunoais: I also added it to global, though...

nickvergessen: Not on this PR?!

brunoais: Look down @ submit_post_base.php

nickvergessen: submit_post_base.php itself creates `$phpbb_dispatcher = new phpbb_mock_event_dispatcher();`

PR #3236, opened 2014-12-20
[ticket/fix/hhvm] Output content in functional test status code issues

marc1706:
Literally just a PR for testing the HHVM failures.


marc1706:
Status right now: HHVM seems to have massive memory leaks that cause the tests to fail at some point. This usually happens when the memory size increases even more due to starting to run functional tests.

PR #3238, opened 2014-12-31
[ticket/13477] Update file cache driver for extension paths

lordlebrand:
For the extension system's version check, the filename of a cache file contains slashes. This breaks the file cache. This change simply replaces the slashes with underscores, in the _exists(), _read(), and _write() methods.

### DIFF
--- a/phpBB/phpbb/cache/driver/file.php
+++ b/phpBB/phpbb/cache/driver/file.php
@@ -279,6 +279,7 @@ class file extends \phpbb\cache\driver\base
 		if ($var_name[0] == '_')
 		{
 			global $phpEx;
+			$var_name = str_replace('/', '_', $var_name);
 			return file_exists($this->cache_dir . 'data' . $var_name . ".$phpEx");
 		}
 		else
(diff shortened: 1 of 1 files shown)


Nicofuma:
First of all, thanks for your contribution.

But your commit message doesn't follow our guidelines (https://wiki.phpbb.com/Git#Commit_Messages), their isn't any ticket related to your PR in our tracker (https://tracker.phpbb.com) and the PR will be perfect with a test.
So could you:
1. Create a ticket in our tracker
2. Fix your commit message
3. Add a test in the /tests/cache/common_test_case.php file

If you can't just say it and I'll do it for you.

phpBB/phpbb/cache/driver/file.php
```diff
@@ -279,6 +279,7 @@ function _exists($var_name)
 		if ($var_name[0] == '_')
 		{
 			global $phpEx;
+			$var_name = str_replace('/', '_', $var_name);
```
nickvergessen: I'd prefer to use `-` as replacement, because we use `_`  for our keys. 
Using a different character reduces the problem of potential conflicts.

marc1706:
Replaced by #3453

PR #3240, opened 2015-01-02
[ticket/13440] Add event core.login_box_failed

VSEphpbb:
https://tracker.phpbb.com/browse/PHPBB3-13440

PHPBB3-13440


VSEphpbb:
!set Event

PR #3243, opened 2015-01-02
[ticket/13437] Add template event viewtopic_body_post_author_before/after

kinerity:
PHPBB3-13437

### DIFF
--- a/phpBB/docs/events.md
+++ b/phpBB/docs/events.md
@@ -1463,7 +1463,7 @@ viewtopic_body_post_author_after
 * Since: 3.1.3-RC1
 * Purpose: Add content directly after the post author on the View topic screen
 
-viewtopic_body_post_author_before/after
+viewtopic_body_post_author_before
 ===
 * Locations:
     + styles/prosilver/template/viewtopic_body.html
--- a/phpBB/docs/events.md
+++ b/phpBB/docs/events.md
@@ -1455,6 +1455,22 @@ viewtopic_body_poll_question_prepend
 * Since: 3.1.0-b3
 * Purpose: Add content directly before the poll question on the View topic screen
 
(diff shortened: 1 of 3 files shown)


phpBB/docs/events.md
```diff
@@ -1455,6 +1455,22 @@ viewtopic_body_poll_question_prepend
 * Since: 3.1.0-b3
 * Purpose: Add content directly before the poll question on the View topic screen
 
+viewtopic_body_post_author_after
+===
+* Locations:
+    + styles/prosilver/template/viewtopic_body.html
+    + styles/subsilver2/template/viewtopic_body.html
+* Since: 3.1.3-RC1
+* Purpose: Add content directly after the post author on the View topic screen
```
nickvergessen: view should be lower case, same below

kinerity: I copied that from the above event line (viewtopic_body_poll_question_prepend). Should I still change it?

nickvergessen: yeah, change it here (we dont need to copy typos ;) )

nickvergessen:
Can you quickly explain, what you plan to do with these events?
It looks like you want to wrap the username with some HTML?
This is potentially dangerous, since another extension doing the same thing will cause the page to be invalid html

kinerity:
This is just an example, but the best I could come up with at the moment.

Assume an extension that wants to indicate a user as the original poster of the topic. WIth these events, you can place the code before or after the username (I included both even though I would only use the "after" event, as it seems standard prac). I could then create my template event file like so (prosilver example):

<div class="original_poster">{L_ORIGINAL_POSTER}</div>

My intention is NOT to wrapt the username in HTML (I'm a little smarter than that). *wink*

PR #3246, opened 2015-01-03
[ticket/13450] Type-hint return value of $phpbb_container->get()

MGaetan89:
https://tracker.phpbb.com/browse/PHPBB3-13450

PHPBB3-13450


MGaetan89:
I can send a PR for branch `develop` as well.

phpBB/app.php
```diff
@@ -28,7 +28,9 @@
 $auth->acl($user->data);
 $user->setup('app');
 
+/* @var $http_kernel Symfony\Component\HttpKernel\HttpKernel */
```
nickvergessen: Should they all have a leading slash?

phpBB/download/file.php
```diff
@@ -66,27 +66,35 @@
 	$phpbb_class_loader_ext->set_cache($phpbb_container->get('cache.driver'));
 
 	// set up caching
+	/* @var $cache phpbb\cache\service */
 	$cache = $phpbb_container->get('cache');
 
+	/* @var $phpbb_dispatcher phpbb\event\dispatcher */
 	$phpbb_dispatcher = $phpbb_container->get('dispatcher');
+	/* @var $request phpbb\request\request */
```
nickvergessen: request_interface

phpBB/download/file.php
```diff
@@ -66,27 +66,35 @@
...
 	$request	= $phpbb_container->get('request');
+	/* @var $db phpbb\db\driver\factory */
```
nickvergessen: driver_interface

phpBB/download/file.php
```diff
@@ -66,27 +66,35 @@
...
 	$db			= $phpbb_container->get('dbal.conn');
+	/* @var $phpbb_log phpbb\log\log */
```
nickvergessen: log_interface

nickvergessen:
We should use interfaces whereever possible.

Not really sure about 3.1 vs 3.2
I think it helps when going through the code, so I wouldn't really mind it.

phpBB/app.php
MGaetan89: It depends if the comment is in a namespace or not. I can add a leading slash everywhere to be sure.

MGaetan89:
I added a leading slash to all `@var` comment and used interface where possible.

MGaetan89:
Oops, I think I did something wrong here. I will try to fix it ASAP.

MGaetan89:
PR is clean again.

MGaetan89:
I just noticed that this PR is milestoned for 3.2, but it is based on `develop-ascraeus`.

phpBB/includes/acp/acp_groups.php
```diff
@@ -989,7 +991,6 @@ public function manage_position()
 		}
 		else if ($field && in_array($field, array('legend', 'teampage')))
 		{
-
 			$group_position = $phpbb_container->get('groupposition.' . $field);
```
Nicofuma: \phpbb\groupposition\groupposition_interface

phpBB/includes/acp/acp_profile.php
```diff
@@ -25,6 +25,10 @@ class acp_profile
 
 	var $edit_lang_id;
 	var $lang_defs;
+
+	/**
+	 * @var \phpbb\di\service_collection
+	 */
 	protected $type_collection;
```
Nicofuma: It's better to mark is as an ArrayAccess or an array I think.

Nicofuma:
memberlist.php:775 `/** @var $form \phpbb\message\form`
acp_groups.php:994 `/** @var $group_position\phpbb\groupposition\groupposition_interface`

Nicofuma:
About the comments you introduced, could you add an empty line before each of them? (It's hard to read when there isn't any empty lines between a comment and the previous variable)
By the way, I don't know if we should use `/* */` or `/** */` (IIRC we already use `/** */` in a few places) but it's a detail and doesn't really matter.

MGaetan89:
I used `/* */` when the comment is on one line. I can change to `/** */` if necessary.
I will update the PR later for your other comments.

phpBB/includes/acp/acp_profile.php
MGaetan89: If I use `ArrayAccess` or `ArrayObject` the method `service_collection#add()` won't be found by the IDE as it is added by phpBB.

MGaetan89:
> I just noticed that this PR is milestoned for 3.2, but it is based on `develop-ascraeus`.

Is this an issue?

marc1706:
That shouldn't be an issue. However, this might cause you to miss some of the spots for type hinting.

MGaetan89:
I can send the PR again against `develop` and check that all calls are documented.

marc1706:
In that case make sure to rebase the commit onto develop if this branch is also based on develop-ascraeus locally.

MGaetan89:
I rebased my local branch on `develop`. I found two more changes to do:
* https://github.com/phpbb/phpbb/blob/develop/phpBB/includes/acp/acp_attachments.php#L1349
* https://github.com/phpbb/phpbb/blob/develop/phpBB/phpbb/auth/provider/db.php#L158

I don't think that I can change the target branch of this PR? I have to close this one and open a new one?

PR #3250, opened 2015-01-04
[ticket/13453] sort param in canonical URL

tas2580:
Sense of the canonical URL is to provide a unique URL for pages with similar content like list with other search order. So the sort param shuld not be used in the canonical URL
See https://tools.ietf.org/html/rfc6596


marc1706:
Could you squash the the 4 commits into one and use the following commit message?
```
[ticket/13453] Sort param in Canonical URL

PHPBB3-13453
```

If you're not sure how to do that just let me know and I'll take care of it.

marc1706:
Also, please let me know what branch you based this off. This ticket looks like it can be merged into 3.1. However, your PR is against develop which is the branch for phpBB 3.2.

nickvergessen:
But https://www.phpbb.com/community/viewforum.php?f=14 and https://www.phpbb.com/community/viewforum.php?f=14&sd=a are totally different pages?

PR #3260, opened 2015-01-05
drop down options and Arabic letters in chrome

Scout4all:
https://tracker.phpbb.com/browse/PHPBB3-13464


nickvergessen:
@prototech mind having a look?

PayBas:
I don't really see any difference. Tested with FF, Chrome and IE11 with Arabic language.

@Scout4all could you provide screenshots so we can see what exactly the problem is (and what your solution attempts to solve)?

nickvergessen:
Related ticket:
https://tracker.phpbb.com/browse/PHPBB3-13464

PR #3261, opened 2015-01-05
Tahoma font for arabic letters

Scout4all:


nickvergessen:
Please add that to the css of your translation

PR #3263, opened 2015-01-05
[ticket/13468] Change add_log() calls with $phpbb_log->add()

MGaetan89:
https://tracker.phpbb.com/browse/PHPBB3-13468

PHPBB3-13468

### DIFF
--- a/phpBB/develop/create_search_index.php
+++ b/phpBB/develop/create_search_index.php
@@ -132,7 +132,7 @@ else
 
 $search->tidy();
 
-add_log('admin', 'LOG_SEARCH_INDEX_CREATED', $search_name);
+$phpbb_log->add('admin', $user->data['user_id'], $user->ip, 'LOG_SEARCH_INDEX_CREATED', false, array($search_name));
 
 echo $user->lang['SEARCH_INDEX_CREATED'] . "\n";
 echo 'Peak Memory Usage: ' . get_formatted_filesize(memory_get_peak_usage()) . "\n";
--- a/phpBB/includes/acp/acp_attachments.php
+++ b/phpBB/includes/acp/acp_attachments.php
@@ -43,7 +43,7 @@ class acp_attachments
 	function main($id, $mode)
 	{
 		global $db, $user, $auth, $template, $cache, $phpbb_container;
-		global $config, $phpbb_admin_path, $phpbb_root_path, $phpEx;
+		global $config, $phpbb_admin_path, $phpbb_root_path, $phpEx, $phpbb_log;
 
 		$this->id = $id;
 		$this->db = $db;
@@ -198,7 +198,7 @@ class acp_attachments
 
 				if ($submit)
 				{
-					add_log('admin', 'LOG_CONFIG_ATTACH');
+					$phpbb_log->add('admin', $user->data['user_id'], $user->ip, 'LOG_CONFIG_ATTACH');
 
 					// Check Settings
 					$this->test_upload($error, $this->new_config['upload_path'], false);
@@ -359,7 +359,7 @@ class acp_attachments
 									WHERE extension_id = ' . $row['extension_id'];
 								$db->sql_query($sql);
 
-								add_log('admin', 'LOG_ATTACH_EXT_UPDATE', $row['extension']);
+								$phpbb_log->add('admin', $user->data['user_id'], $user->ip, 'LOG_ATTACH_EXT_UPDATE', false, array($row['extension']));
 							}
 						}
 						$db->sql_freeresult($result);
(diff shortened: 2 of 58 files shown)


MGaetan89:
!set WIP

phpBB/develop/create_search_index.php
```diff
@@ -132,7 +132,7 @@
 
 $search->tidy();
 
-add_log('admin', 'LOG_SEARCH_INDEX_CREATED', $search_name);
+$phpbb_log->add('admin', $user->data['user_id'], $user->ip, 'LOG_SEARCH_INDEX_CREATED', time(), array($search_name));
```
Nicofuma: could be better to use false instead of time() here... I don't really know...

Nicofuma:
Seems to be good for now, don't forgot the other calls :)

phpBB/develop/create_search_index.php
MGaetan89: I didn't think about that. But I think `0` would be a better default value for `log_interface::add()`, as the documentation says it should be an int.

nickvergessen: although 0 is a valid timestamp, so maybe null instead?
Or we just fix the interface

nickvergessen:
```php
		if ($log_time == false)
		{
			$log_time = time();
		}
```
Should be false, also there should be 3 equals ;)
So mind fixing the interface and the implementation?

MGaetan89: So `null` by default and 3 equals?

nickvergessen: false by default and 3 equals

MGaetan89: Change is done.

MGaetan89:
Some tests failed because I used `$user->data['user_id']` instead of `ANONYMOUS` in some places.
Instead of checking the value of `$user->data['user_id']` everywhere, I added the method `$user->get_user_id()`.
If that solution is ok, I will keep the last commit, and use this method for the other `add_log()` that needs to be changed. Otherwise I will keep the checks where it was previously used.

nickvergessen:
I'd fix that in https://github.com/phpbb/phpbb/blob/develop/phpBB/phpbb/log/log.php#L232
`'user_id'	=> ($user_id) ? (int) $user_id : ANONYMOUS,`

MGaetan89:
So much easier... Thanks, I'll update the code.

MGaetan89:
I removed the `$user->get_user_id()` method. I will try to finish this PR tomorrow.

MGaetan89:
Is it valid to use a default value instead of a type?

MGaetan89:
I updated the remaining `add_log()` calls and move the function to `functions_compatibility.php`.
Next step is fixing the tests and then the PR is good to do.

MGaetan89:
!unset WIP

MGaetan89:
Test are all passing now. PR is ready for review/merge.

phpBB/develop/create_search_index.php
Nicofuma: You still need to use false instead of time() here and in a few other places

MGaetan89: But if I set `$log_time` to `false`, `time()` will be use. So it's the same, no?

nickvergessen:
needs a rebase (because of your type hint PR)

MGaetan89:
Done. Tests are running.
What is the final decision about using `false` vs. `time()` for the `$log_time` parameter?

Nicofuma:
I'll say `false` @nickvergessen ?

nickvergessen:
`false` and triple equals please as stated above :)

MGaetan89:
`===` is already there. I will change `time()` with `false` and update the PR :wink:

MGaetan89:
Updated and tests are passing.

Nicofuma:
Can you rebase again please?

MGaetan89:
Rebased. Tests are running.

MGaetan89:
Tests are passing :smile:

MGaetan89:
Rebased on `develop`.

Nicofuma:
Please, fix some comments by the way (add_log() will disappear one day ^^)

functions_user.php:line 317: `// Because these actions only fill the log unneccessarily we disable it`
phpbb\log\log.php:line 30: `* added when add() is called.`
phpbb\log\log_interface.php:line 35: `* page call. When add() is called and the type is disabled,`
test\dbal\migrator_tool_module_test.php: line 30: `// Disable the logs`

MGaetan89:
Good catch.
Changes are done.

Nicofuma:
Well, should be good now :) I let it open a few another hours and will merge it tomorrow :)

PR #3264, opened 2015-01-05
[ticket/13471] Add core.viewforum_assign_template_vars_before

omniError:
https://tracker.phpbb.com/browse/PHPBB3-13471

PHPBB3-13471

### DIFF
--- a/phpBB/viewforum.php
+++ b/phpBB/viewforum.php
@@ -320,6 +320,38 @@ if (!empty($_EXTRA_URL))
 	}
 }
 
+/**
+* Event to modify data before template variables are assigned
+*
+* @event core.viewforum_assign_template_vars_before
+* @var	int		forum_id			Forum ID
+* @var	array	forum_data			Array with forum data
+* @var	bool	s_display_active	Are active topics being displayed
+* @var	string	s_sort_dir			Sort direction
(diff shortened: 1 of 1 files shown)


omniError:
!set Event

omniError:
actually this should go to develop-ascraeus and not just develop
can you do this on your end or do I have to redo this?

nickvergessen:
Actually this event should be merged with the existing `core.viewforum_get_topic_data` (just some lines down).
You can then just reassign the variables in the template to overwrite the existing values.

omniError:
I disagree with merging them for a few reasons, the first being that it would put unexpected logic into that event (an event previously described to handle a single SQL query for topic data is now also changing unrelated template variables). Changing the event itself to be more consistent with what it would do breaks BC. Reassigning after that event would create a bunch of duplicate code since the values are determined inline in the array. This seemed like the least bad approach.

Nicofuma:
We can merge them into a new event with a better name and  description and mark the current event as deprecated and to be removed in phpBB 3.3 or 3.4 (or phpBB 4.0 maybe)

omniError:
I'm going to withdraw this for now, I saw at least another PR changing code near this area which would impact this proposal and I found another way to do what I wanted to do anyway.

PR #3265, opened 2015-01-06
[ticket/13453] Sort param in Canonical URL

tas2580:
https://tracker.phpbb.com/browse/PHPBB3-13453
PHPBB3-13453


bantu:
This change does not seem to be correct as the sorting parameters determine what content is shown or not shown on a multi page topic.

tas2580:
Yes, but it makes no sense to get different sorted pages into Googles index. Google seems to see that also, look at Googel Webmastertools there will Google set the sort parameter to ignore automatic after a while.

Also see: https://webmasters.stackexchange.com/questions/7136/pagination-duplicate-content-and-seo

nickvergessen:
Okay so the logic behind the canonical seems to be something like:
* If canonical is present and different from the current page url: ignore current page and crawl canonical
* If no canonical is present or the canonical is the same as the current page: crawl it

If that is the case, I guess the patch is fine

@tas2580 can you confirm?

PR #3266, opened 2015-01-07
[ticket/13474] Deny access on backup files of config.php

tas2580:
https://tracker.phpbb.com/browse/PHPBB3-13474

### DIFF
--- a/phpBB/.htaccess
+++ b/phpBB/.htaccess
@@ -39,6 +39,10 @@ RewriteRule ^(.*)$ app.php [QSA,L]
 			Order Allow,Deny
 			Deny from All
 		</Files>
+		<Files "config.*">
+			Order Allow,Deny
+			Deny from All
+		</Files>
 		<Files "common.php">
 			Order Allow,Deny
 			Deny from All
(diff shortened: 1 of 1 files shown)


tas2580:
Maybe you don't need <Files "config.php"> because <Files "config.*"> includes .php but maybe there are Apache versions with no support of .*

Nicofuma:
I don't think that we should try to match every backup file introduced by any editor because in this case this PR isn't enough (you are missing emacs and vim backup files per example)

Nicofuma:
well IIRC emacs and vim can create other backup files... no matter
But add the same changed to the web.config file please

nickvergessen:
commit message must have the following end:
```

PHPBB3-13474
```

bantu:
> I don't think that we should try to match every backup file introduced by any editor because in this case this PR isn't enough (you are missing emacs and vim backup files per example)

I agree. This change is pointless in my opinion. Some editors use a prefix, some use a suffix, etc. etc.

tas2580:
If an editor use a prefix its no problem because the file ends with .php and so the content will not be shown.

Why deny access to config.php its no problem to view config.php in a webbrowser, but its a problem if you can view config.bak or config.old in a webbrowser.
Of course this will not hide all possible config backups, there are to much editors and to much configurations, so you can never get all, there I agree with you. But you can get with a small change a lot of editors and configurations, so why do not do this?
Nicofuma is right in web.config should done the same as in htaccess and when you have Apache and IIS.

Maybe you will see the problem when I give you a list with config backups of phpBB, I will run a little script over tonight so that I can give you some examples tomorrow.

nickvergessen:
Closing as per ticket

PR #3269, opened 2015-01-08
Quote image as thumbnail

dragosvr92:
https://tracker.phpbb.com/browse/PHPBB3-13483

### DIFF
--- a/phpBB/styles/prosilver/theme/colours.css
+++ b/phpBB/styles/prosilver/theme/colours.css
@@ -414,6 +414,10 @@ blockquote {
 	border-color:#DBDBCE;
 }
 
+blockquote img {
+    max-height: 200px;
+}
+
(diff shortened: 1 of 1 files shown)


VSEphpbb:
Whether or not this change is worthwhile is debatable (you shouldn't really be augmenting post content just because it is quoted, and this change seems to be purely a subjective opinion and a-typical behavior, making it better as an extension), however, `colours.css` is for colors and background images. Height changes would go in `content.css`.

nickvergessen:
This needs a RFC first

dragosvr92:
We can add it to either content.css or common.css, wherever it needs to be. 
I was thinking the same thing, but area51 is still down.... :-\

nickvergessen:
Closing as per ticket

PR #3270, opened 2015-01-08
arrange css in a better order

dragosvr92:
While the browser renders this css the same, it prefers to arrange the properties in the order i used.
The bg class seemed to me that it was meant to be put in 3 lines instead of one...

### DIFF
--- a/phpBB/styles/prosilver/theme/colours.css
+++ b/phpBB/styles/prosilver/theme/colours.css
@@ -96,7 +96,9 @@ table.zebra-list tr:nth-child(even) td, ul.zebra-list li:nth-child(even) {
 	background-color: #e1ebf2;
 }
 
-.bg3	{ background-color: #cadceb; }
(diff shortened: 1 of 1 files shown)


phpBB/styles/prosilver/theme/colours.css
```diff
@@ -96,7 +96,9 @@ table.zebra-list tr:nth-child(even) td, ul.zebra-list li:nth-child(even) {
 	background-color: #e1ebf2;
 }
 
-.bg3	{ background-color: #cadceb; }
+.bg3	{
+	background-color: #cadceb;
+}
```
VSEphpbb: single css properties should be 1 line

nickvergessen:
Merged via https://github.com/phpbb/phpbb/commit/a1ecb10c5cc9a2bb4b7bab31cb11c958fd13986a

PR #3273, opened 2015-01-10
Travis/oracle

nickvergessen:


marc1706:
Based on travis output it looks like this takes up more resources than travis provides?

PR #3275, opened 2015-01-10
[ticket/13487] Add factory for db tool class

nickvergessen:
This allows us to split up the large file one by one easily.

https://tracker.phpbb.com/browse/PHPBB3-13487

### DIFF
--- a/phpBB/config/default/container/services_db.yml
+++ b/phpBB/config/default/container/services_db.yml
@@ -9,8 +9,13 @@ services:
         calls:
             - [sql_connect, [%dbal.dbhost%, %dbal.dbuser%, %dbal.dbpasswd%, %dbal.dbname%, %dbal.dbport%, false, %dbal.new_link%]]
 
+# ----- DB Tools -----
+    dbal.tools.factory:
+        class: phpbb\db\tools\factory
+
     dbal.tools:
         class: phpbb\db\tools\tools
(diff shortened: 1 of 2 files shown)


Nicofuma:
Actually here we have 2 ways to that on an "easier" (using the DIC to construct the real db tools object):
For both of them we have to define a private service per db tools type (ie per database) and then we can:
1. Using a factory and pass the container to it so instead of calling new we call $container->get('dbal.tools.<driver>')
2. We can  use a compiler pass to rename the real service or to create an alias.

I think I prefer the second solution (and by the way it could be nice to do the same think with the dbal driver)

nickvergessen:
Well im going to add an array implementation which is independent from the used db, does this still work (i don't know anything about compilerpass stuff)

Nicofuma:
When will this implementation be used and when will be the dbal driver dependent implementation?

nickvergessen:
It will be used for generating the schema.json and for tests,
the dbal driver version is used in code and the migrator

Nicofuma:
I think that the array implementation should be an independent public service and requested directly when needed.

nickvergessen:
Can we still ahve this little factory while we are splitting up the current db_tools file since it makes it a lot easier?

Nicofuma:
I think yes but if you do that this way you have to instantiate the class manually in the factory...

PR #3189, opened 2014-11-26
[ticket/12926] Support for IDN (IRI)

rxu:
Add international domain name support for URLs.
IDN for emails is unsupported yet.

<a href="https://tracker.phpbb.com/browse/PHPBB3-12926">PHPBB3-12926</a>.

PR #3278, opened 2015-01-11
[ticket/13492] Fix IDN support for custom BBCode URL tokens

rxu:
Custom BBCode URL tokens does not support IDN.
Custom BBCodes added earlier than IDN PR #3189 has been applied are unaffected.

<a href="https://tracker.phpbb.com/browse/PHPBB3-13492">PHPBB3-13492</a>.

### DIFF
--- a/phpBB/includes/acp/acp_bbcodes.php
+++ b/phpBB/includes/acp/acp_bbcodes.php
@@ -409,7 +409,9 @@ class acp_bbcodes
 	{
 		$bbcode_match = trim($bbcode_match);
 		$bbcode_tpl = trim($bbcode_tpl);
-		$utf8 = strpos($bbcode_match, 'INTTEXT') !== false;
+
+		// Allow unicode characters for URL|LOCAL_URL|RELATIVE_URL|INTTEXT tokens
+		$utf8 = preg_match('/\{(URL|LOCAL_URL|RELATIVE_URL|INTTEXT)\}/', $bbcode_match);
 
 		$utf8_pcre_properties = phpbb_pcre_utf8_support();
 
--- a/phpBB/includes/acp/acp_bbcodes.php
+++ b/phpBB/includes/acp/acp_bbcodes.php
@@ -411,7 +411,7 @@ class acp_bbcodes
 		$bbcode_tpl = trim($bbcode_tpl);
 
 		// Allow unicode characters for URL|LOCAL_URL|RELATIVE_URL|INTTEXT tokens
-		$utf8 = preg_match('/\{(URL|LOCAL_URL|RELATIVE_URL|INTTEXT)\}/', $bbcode_match);
+		$utf8 = preg_match('/(URL|LOCAL_URL|RELATIVE_URL|INTTEXT)/', $bbcode_match);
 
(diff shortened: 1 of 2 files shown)


rxu:
!set Blocker

Elsensee:
what about a migration which reparses the old custom bbcodes so they can benefit from this change?
@rxu and I already discussed this in irc and would like to hear another opinion :)

nickvergessen:
Well reparsing the posts is a no go.
As for redoing the magic in the bbcode table... that would be cool i guess?

rxu:
I believe that reparsing posts is not needed as it won't change any posts content.
As for the ```BBCODES_TABLE```, I'm working on the migration currently.

Elsensee:
To clarify: I did not say that I want reparsing the posts. But custom bbcodes needs to be parsed as well and formed into nice little regular expressions with the function acp_bbcodes->build_regexp(); :wink: 

However if posts are reparsed the content *may* change. For example one bbcode is with matching an URL. Then in the post in that spot there is an URL but with non-ASCI chars in it. So there would be an change.
But as nickvergessen already said, reparsing posts is a no go.

phpBB/phpbb/db/migration/data/v31x/update_custom_bbcodes_with_idn.php
```diff
@@ -0,0 +1,70 @@
...
+
+		$sql = 'SELECT bbcode_id, bbcode_match, bbcode_tpl
+			FROM ' . BBCODES_TABLE;
+		$result = $this->sql_query($sql);
+
+		$sql_ary = array();
+		while ($row = $this->db->sql_fetchrow($result))
+		{
+			$data = array();
+			if (preg_match('/\{(URL|LOCAL_URL|RELATIVE_URL)\}/', $row['bbcode_match']))
```
nickvergessen: Please remove the trailing `}`
We allow stuff like `{URL1},{URL2}` which is not matched correctly otherwise?

phpBB/includes/acp/acp_bbcodes.php
```diff
@@ -409,7 +409,9 @@ function build_regexp(&$bbcode_match, &$bbcode_tpl)
 	{
 		$bbcode_match = trim($bbcode_match);
 		$bbcode_tpl = trim($bbcode_tpl);
-		$utf8 = strpos($bbcode_match, 'INTTEXT') !== false;
+
+		// Allow unicode characters for URL|LOCAL_URL|RELATIVE_URL|INTTEXT tokens
+		$utf8 = preg_match('/\{(URL|LOCAL_URL|RELATIVE_URL|INTTEXT)\}/', $bbcode_match);
```
nickvergessen: Please remove the trailing `}`, or add a regex int check
We allow stuff like `{URL1},{URL2}` which is not matched correctly otherwise?

Elsensee: Then I guess remove both the `{` and `}` is better. (For this change and the change in migrations)

rxu: Indeed, fixing that.

rxu:
!unset WIP

PR #3279, opened 2015-01-11
[ticket/13493] $page_name missing from $base_url when URL ens with /

satanasov:
Test is $page_name is there - if not add it

PHPBB3-13493


phpBB/phpbb/controller/helper.php
```diff
@@ -142,6 +142,12 @@ public function route($route, array $params = array(), $is_amp = true, $session_
 
 		$base_url = $context->getBaseUrl();
 
+		// Regarding PHPBB3-13493 - if we request url ending on / there is no $page_name
+		if (!strpos($base_url, $page_name))
```
Elsensee: If `$page_name` is empty (what is true in your case), strpos will output a warning. Also you should check like `if (strpos($base_url, $page_name) === false)` because strpos can also return 0. (which of course isn't very likely in this case)

satanasov: $page_name is never empty. It is taken from https://github.com/satanasov/phpbb/blob/ticket/13493/phpBB/phpbb/controller/helper.php#L141

This means that it will always have  $page_name. But you are right for the check.

Elsensee: You are correct. It's taken from there. What it means afaik is: If the last char is a `/` then `$page_name` will be an empty string.

phpBB/phpbb/controller/helper.php
```diff
@@ -142,6 +142,12 @@ public function route($route, array $params = array(), $is_amp = true, $session_
 
 		$base_url = $context->getBaseUrl();
 
+		// Regarding PHPBB3-13493 - if we request url ending on / there is no $page_name
+		if (strpos($base_url, $page_name) === false)
+		{
+			$base_url = $base_url . '/' . $page_name;
```
nickvergessen: `$base_url .= '/' . $page_name;`

phpBB/phpbb/controller/helper.php
```diff
@@ -142,6 +142,12 @@ public function route($route, array $params = array(), $is_amp = true, $session_
 
 		$base_url = $context->getBaseUrl();
 
+		// Regarding PHPBB3-13493 - if we request url ending on / there is no $page_name
+		if (strpos($base_url, $page_name) === false)
```
marc1706: Check if $page_name is not empty.

satanasov:
This won't change the failing tests. 

Can somone do the tests - as I can't. I can do functional test if this is appropriete ...

phpBB/phpbb/controller/helper.php
nickvergessen: Replace with either:
`if (!$page_name || strpos($base_url, $page_name) === false)`
or
`if ($page_name && strpos($base_url, $page_name) === false)`

and the tests will pass.
I just don't know which one of the conditions is the right one.

nickvergessen:
@marc1706 want to have a look?

marc1706:
@nickvergessen See comment in the ticket: https://tracker.phpbb.com/browse/PHPBB3-13493?focusedCommentId=55669&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-55669

marc1706:
Replaced by: https://github.com/phpbb/phpbb/pull/3392

PR #3257, opened 2015-01-04
[ticket/13455] Change request_var() calls with $request->variable()

MGaetan89:
https://tracker.phpbb.com/browse/PHPBB3-13455

PHPBB3-13455

PR #3280, opened 2015-01-11
[ticket/13494] Change set_config() calls with $config->set()

MGaetan89:
https://tracker.phpbb.com/browse/PHPBB3-13494

PHPBB3-13494

### DIFF
--- a/phpBB/includes/acp/acp_forums.php
+++ b/phpBB/includes/acp/acp_forums.php
@@ -1962,7 +1962,7 @@ class acp_forums
 		$row = $db->sql_fetchrow($result);
 		$db->sql_freeresult($result);
 
-		set_config('num_posts', (int) $row['stat'], true);
+		$config->set('num_posts', (int) $row['stat'], false);
 
 		$sql = 'SELECT COUNT(topic_id) AS stat
 			FROM ' . TOPICS_TABLE . '
@@ -1971,7 +1971,7 @@ class acp_forums
 		$row = $db->sql_fetchrow($result);
 		$db->sql_freeresult($result);
 
-		set_config('num_topics', (int) $row['stat'], true);
+		$config->set('num_topics', (int) $row['stat'], false);
 
 		$sql = 'SELECT COUNT(attach_id) as stat
 			FROM ' . ATTACHMENTS_TABLE;
@@ -1979,7 +1979,7 @@ class acp_forums
 		$row = $db->sql_fetchrow($result);
 		$db->sql_freeresult($result);
 
-		set_config('num_files', (int) $row['stat'], true);
+		$config->set('num_files', (int) $row['stat'], false);
 
 		$sql = 'SELECT SUM(filesize) as stat
 			FROM ' . ATTACHMENTS_TABLE;
@@ -1987,7 +1987,7 @@ class acp_forums
 		$row = $db->sql_fetchrow($result);
 		$db->sql_freeresult($result);
 
-		set_config('upload_dir_size', (float) $row['stat'], true);
+		$config->set('upload_dir_size', (float) $row['stat'], false);
 
 		return array();
 	}
--- a/phpBB/includes/acp/acp_main.php
+++ b/phpBB/includes/acp/acp_main.php
@@ -121,8 +121,8 @@ class acp_main
 							trigger_error($user->lang['NO_AUTH_OPERATION'] . adm_back_link($this->u_action), E_USER_WARNING);
 						}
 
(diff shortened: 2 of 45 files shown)


phpBB/includes/acp/acp_forums.php
```diff
@@ -1960,7 +1960,7 @@ function delete_forum_content($forum_id)
 		$row = $db->sql_fetchrow($result);
 		$db->sql_freeresult($result);
 
-		set_config('num_posts', (int) $row['stat'], true);
+		$config->set('num_posts', (int) $row['stat'], false);
```
nickvergessen: ??

phpBB/includes/acp/acp_main.php
```diff
@@ -121,8 +121,8 @@ function main($id, $mode)
 							trigger_error($user->lang['NO_AUTH_OPERATION'] . adm_back_link($this->u_action), E_USER_WARNING);
 						}
 
-						set_config('record_online_users', 1, true);
-						set_config('record_online_date', time(), true);
+						$config->set('record_online_users', 1, false);
```
nickvergessen: ??

phpBB/includes/acp/acp_forums.php
nickvergessen: Oh okay, we inverted the parameter Oo,
quite bad, but okay then

MGaetan89: The parameter is not used anyway (as far as I could see). I don't know if it was something planned or removed.

nickvergessen: Yes we do use it Oo
https://github.com/phpbb/phpbb/blob/develop/phpBB/phpbb/config/db.php

MGaetan89: Missed that one. But I guess inverting the `is_dynamic` parameter is still correct, since `set_config()` does it?

nickvergessen: yes, its all okay in this PR

marc1706:
Seems like the PR can't be merged into develop anymore. Could you rebase this into latest develop?

MGaetan89:
Rebased on `develop`.

bantu:
This looks pretty good to me. Could you please rebase once more. The conflicts are rather easy to resolve. Thanks.

MGaetan89:
Done.

bantu:
Thanks. I assume you're doing set_config_count() next? :-P

MGaetan89:
I'm following rebase requests ;-)
I can do it, and then #3257

bantu:
@MGaetan89 Sounds perfect. Thanks a lot.

bantu:
Didn't realise you already had a PR for _count() at #3281.

phpBB/includes/compatibility_globals.php
```diff
@@ -44,7 +44,6 @@
 // Grab global variables, re-cache if necessary
 /* @var $config phpbb\config\db */
 $config = $phpbb_container->get('config');
-set_config(null, null, null, $config);
```
bantu: I think we actually need this somewhere. Otherwise set_config is not usable, although it is supposed to be still available.

MGaetan89: How about in `set_config()` directly? Adding this after `if ($set_config !== null)` (not sure if it works or not):

```php
if ($config === null)
{
	global $config;
}
```

This question is also valid for `set_config_count()` and `request_var()`.

bantu: I think this line should just be restored. And removed when the function is removed.

MGaetan89: Ok, I will add it back. Same for `set_config_count()` and `request_var()`.
Can I use the same ticket for `set_config()`?

nickvergessen: Since it's a fallback method now, I guess we could also use a global instead of keeping the call. :speak_no_evil:

bantu: Yes, I reopenend the ticket.

bantu: Just restoring that line might not actually work.

MGaetan89: The `$config` injection was used in three main places:
* The board itself
* The install
* In tests

Now it is no longer used at all. So maybe it would be easier to use `global $config;` and use the global config object? I don't know if some developers used the config injection.

For `request_var()` is still used in tests, but that's it. So I think it should use the same `global $request;`, no?

PR #3284, opened 2015-01-11
[ticket/13469] Fall back to soft_delete if no permissions for delete

Elsensee:
https://tracker.phpbb.com/browse/PHPBB3-13469

### DIFF
--- a/phpBB/posting.php
+++ b/phpBB/posting.php
@@ -330,14 +330,17 @@ switch ($mode)
 		{
 			$is_authed = true;
 		}
-	break;
+
+	// no break;
 
 	case 'soft_delete':
-		if ($user->data['is_registered'] && $phpbb_content_visibility->can_soft_delete($forum_id, $post_data['poster_id'], $post_data['post_edit_locked']))
+		if (!$is_authed && $user->data['is_registered'] && $phpbb_content_visibility->can_soft_delete($forum_id, $post_data['poster_id'], $post_data['post_edit_locked']))
 		{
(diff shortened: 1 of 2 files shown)


Crizz0:
I tried this in 3.1.2 and found after the changes one strange behavior:
- soft deleting one post at the end of the topic
Everything works. Post is gone for regular users. Moderators can see this posts.

BUT:
1. a regular user quotes a different post
2. a other regular user reads the quote, looks for the original posts --> cannot find it
3. moderators stil saw the "original post"
4. restore/permant delete of the soft deleted post the end of the topic
--> @2.) regular users can reads this original post again

:/ 

PS: Don't know exactly, if it is related to this code or to soft deleted in general.

gn36:
This modification changes the delete mode to soft_delete directly within the auth check, so the behavior you describe should relate to any deletion using soft_delete.

nickvergessen:
We have functional tests for deleting, can this be added somehow?

marc1706:
@Elsensee how are things looking on this one? Would be nice to see some tests for this behavior.

Elsensee:
Uhh, right. Give me some time. :)

Elsensee:
Updated. :)

Elsensee:
`[19:36:40] <Nicofuma> https://github.com/phpbb/phpbb/pull/3284 could you extract the new test to a new method please?`

nickvergessen:
Merged via http://git.io/ve4g7

PR #3291, opened 2015-01-13
Update content.css

dragosvr92:
The blockquote tags should be all near each other. Its cleaner...


VSEphpbb:
The original position made more sense because it is about defining the overall codebox styling first, then the necessary tweaks on that class for specific elements such as blockquote, code and p tags.

nickvergessen:
Closing as per above

PR #3292, opened 2015-01-13
PHPBB3-13502 - controller resolver should handle callable functions and objects

rfdy:
This change follows what Symfony does, by allowing the controller to be any type of callable.

### DIFF
--- a/phpBB/phpbb/controller/resolver.php
+++ b/phpBB/phpbb/controller/resolver.php
@@ -116,9 +116,18 @@ class resolver implements ControllerResolverInterface
 	*/
 	public function getArguments(Request $request, $controller)
 	{
-		// At this point, $controller contains the object and method name
-		list($object, $method) = $controller;
-		$mirror = new \ReflectionMethod($object, $method);
+		// At this point, $controller should be a callable
+		if (is_array($controller))
+		{
+			list($object, $method) = $controller;
+			$mirror = new \ReflectionMethod($object, $method);
+		} else if (is_object($controller) && !$controller instanceof \Closure)
+		{
+			$mirror = new \ReflectionObject($controller);
+			$mirror = $mirror->getMethod('__invoke');
+		} else {
+			$mirror = new \ReflectionFunction($controller);
+		}
 
 		$arguments = array();
 		$parameters = $mirror->getParameters();


nickvergessen:
I think this was not added intentionally since we want to help extension authors to use DI and therefor have the controllert as classes.

rfdy:
I think I see your point about making extensions easier to write, but I don't see how that is in conflict with DI. In symfony a controller is just any callable that returns a response. That leads to some very useful design patterns. Here is the code I'm using, of course there are other ways to achieve this, but I really don't want the controller to be available publicaly at all, which is why I used an anonymous function.

I use the kernel.controller event to alter the arguments to my method depending on its arguments, similar to how SensioFrameworkExtraBundle ParamConverter works. In my case its for an API I'm building and want to return a very specific response if the argument cannot be instantiated.

```
$event->setController(function() use ($e) {
    return new JsonErrorResponse($e->getMessage());
});
```

Making phpBB work similar to how you'd do things in Symfony seems like a good idea to me as it'll reduce the developer entry barrier. Anyone familiar with Symfony should be able to write phpBB extensions with ease. Changing simple conventions like this could hurt the process.

Thanks.

nickvergessen:
@Nicofuma

Nicofuma:
I see the point....
I'm not sure it's the best design (throwing an exception should do the same thing) but I agree about the Symfony "compliance" so... :+1: for this PR

rfdy:
Throwing an exception would work, but in this case, I really don't want to throw a global exception (ie. it can be thrown anywhere), and then adding another custom exception handler just for this.

Thanks for accepting the patch!

Nicofuma:
Please update the commit message.

By the way maybe you should add that this PR add the support for any callable in the arguments parser but not in the resolver itself. I think it's fine like that for now but maybe we should work on extending the base Resolver in the future (to support the service:method notation and to translate the exceptions)

marc1706:
Please update your commit message to fit the format outlined here: https://wiki.phpbb.com/Git#Commit_Messages

Nicofuma:
bump @rfdy

phpBB/phpbb/controller/resolver.php
```diff
@@ -116,9 +116,18 @@ public function getController(Request $request)
...
 	{
-		// At this point, $controller contains the object and method name
-		list($object, $method) = $controller;
-		$mirror = new \ReflectionMethod($object, $method);
+		// At this point, $controller should be a callable
+		if (is_array($controller))
+		{
+			list($object, $method) = $controller;
+			$mirror = new \ReflectionMethod($object, $method);
+		} else if (is_object($controller) && !$controller instanceof \Closure)
```
imkingdavid: In accordance with the phpBB coding guidelines (read: https://area51.phpbb.com/docs/31x/coding-guidelines.html#codelayout), to keep things consistent, brackets should go on their own line. Please change this one and the two around the `else` on line 128.

Nicofuma:
bump @rfdy

PR #3294, opened 2015-01-16
[ticket/13507] Fix horizontal scrollbar for large images in posts

PayBas:
https://tracker.phpbb.com/browse/PHPBB3-13507

PHPBB3-13507


PayBas:
!set Styles

PR #3296, opened 2015-01-16
[ticket/13431] Fix responsive RTL media queries for <IE11

PayBas:
https://tracker.phpbb.com/browse/PHPBB3-13431

PHPBB3-13431


PayBas:
!set Styles

PR #3297, opened 2015-01-16
[ticket/13511] Call the check() method with the phpcsFile object

Nicofuma:
https://tracker.phpbb.com/browse/PHPBB3-13511

PHPBB3-13511


bantu:
Replaced with #3299

PR #3302, opened 2015-01-17
[ticket/13512] Add template events before/after the post details

rxu:
Add template events to viewtopic_body.html to allow extensions adding content
before/after the post details.

<a href="https://tracker.phpbb.com/browse/PHPBB3-13512">PHPBB3-13512</a>.


rxu:
!set Event

PR #3306, opened 2015-01-19
[ticket/13519] Correctly validate imagick path as path and not string

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-13519


nickvergessen:
mind rebasing for passing tests?

bantu:
That should do the trick.

PR #3309, opened 2015-01-20
[ticket/13523] Remove PHP 5.2 builds from travis configuration

marc1706:
This is only for develop-olympus. Please make sure that this doesn't break the builds on develop-ascraeus and newer or ask me for another PR that reverts this in develop-ascraeus and newer.

Ticket: https://tracker.phpbb.com/browse/PHPBB3-13523


.travis.yml
```diff
@@ -1,6 +1,5 @@
 language: php
 php:
-  - 5.2
```
nickvergessen: Everything in this file, apart from this change here, should be enough to fix it?

marc1706: I'm not sure what you'e trying to say. Can you elaborate on this?

nickvergessen: Instead of removing 5.2 tests, we should try to fix them?
Or is there no phpunit version available for 5.2 anywhere?

marc1706: Please take a look at the ticket.

bantu:
Closing this in favour of #3310

PR #3312, opened 2015-01-20
[ticket/13527] Escape information received from version server

bantu:
https://tracker.phpbb.com/browse/PHPBB3-13527

### DIFF
--- a/phpBB/includes/acp/acp_update.php
+++ b/phpBB/includes/acp/acp_update.php
@@ -34,7 +34,7 @@ class acp_update
 		$this->page_title = 'ACP_VERSION_CHECK';
 
 		// Get current and latest version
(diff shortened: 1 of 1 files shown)


phpBB/includes/acp/acp_update.php
```diff
@@ -34,10 +34,7 @@ function main($id, $mode)
 		$this->page_title = 'ACP_VERSION_CHECK';
 
 		// Get current and latest version
-		$errstr = '';
-		$errno = 0;
-
-		$info = obtain_latest_version_info(request_var('versioncheck_force', false));
+		$info = htmlspecialchars(obtain_latest_version_info(request_var('versioncheck_force', false)));
```
cs278: You should set the charset argument.

PR #3315, opened 2015-01-20
[ticket/13530] Fix undefined variables

nickvergessen:
https://tracker.phpbb.com/browse/PHPBB3-13530


nickvergessen:
patch is based on dev-asc and should be merged there

PR #3316, opened 2015-01-20
[ticket/13531] Explicitly disallow trailing paths (e.g. PATH_INFO).

bantu:
https://tracker.phpbb.com/browse/PHPBB3-13531

### DIFF
--- a/phpBB/includes/startup.php
+++ b/phpBB/includes/startup.php
@@ -95,6 +95,40 @@ function deregister_globals()
 	unset($input);
 }
 
+/**
+ * Check if requested page uses a trailing path
+ *
+ * @param string $phpEx PHP extension
+ *
+ * @return bool True if trailing path is used, false if not
+ */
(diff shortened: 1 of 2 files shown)


phpBB/includes/startup.php
```diff
@@ -95,6 +95,40 @@ function deregister_globals()
...
+		return true;
+	}
+
+	return false;
+}
+
+// Check if trailing path is used
+if (phpbb_has_trailing_path($phpEx))
+{
+	exit('Trailing paths and path_info is not supported by phpBB 3.0');
```
cs278: Would it not be possible to perform a redirect? If not what about a proper status code?

bantu: Unfortunately send_status_line is not available here.

fgeek:
Please use CVE-2015-1431 for this issue, thanks.

PR #3321, opened 2015-01-21
[ticket/13282] Use 0 as default for integer type columns in postgresql

marc1706:
Needs different patch for develop.

Ticket: https://tracker.phpbb.com/browse/PHPBB3-13282

### DIFF
--- a/phpBB/phpbb/db/tools.php
+++ b/phpBB/phpbb/db/tools.php
@@ -1575,7 +1575,7 @@ class tools
 				else
 				{
 					// Integers need to have 0 instead of empty string as default
-					if (preg_match('/int/i', $column_type))
+					if (strpos($column_type, 'INT') === 0)
 					{
 						$default_val = '0';
 					}
--- a/phpBB/phpbb/db/tools.php
+++ b/phpBB/phpbb/db/tools.php
@@ -1574,7 +1574,15 @@ class tools
 				}
 				else
 				{
-					$default_val = "'" . $column_data[1] . "'";
+					// Integers need to have 0 instead of empty string as default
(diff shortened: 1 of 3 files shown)


marc1706:
@PayBas please check if this also fixes the issue for you

phpBB/phpbb/db/tools.php
```diff
@@ -1574,7 +1574,15 @@ function sql_prepare_column_data($table_name, $column_name, $column_data)
 				}
 				else
 				{
-					$default_val = "'" . $column_data[1] . "'";
+					// Integers need to have 0 instead of empty string as default
+					if (preg_match('/int/i', $column_type))
```
nickvergessen: I prefer `strpos() === 0`

PayBas:
Can confirm fix, nice job.

phpBB/phpbb/db/tools.php
marc1706: I thought we might also support TINT but as it turns out there is no such thing in postgresql. Right now this is case insensitive though. Changing it to strpos() will modify this. Again, this should be fine as we always use uppercase for the column type in postgres.
I'll change this when I add the tests later today.

marc1706:
@nickvergessen Added functional tests for this

nickvergessen:
Since you change the tools.php,
mind adding a test for that in https://github.com/phpbb/phpbb/blob/develop/tests/dbal/db_tools_test.php#L39 ?

marc1706:
Sure, will have to prepare a patch for develop anyway. If this PR looks fine to you I'll just base a branch on this and develop.

PR #3324, opened 2015-01-22
[ticket/13336] Add core events to the function user_active_flip()

rxu:
Add core event to the function user_active_flip() in
includes/functions_user.php to allow checking/modifying activated/deactivated
users data before submitting it to the database.
Event request: http://area51.phpbb.com/phpBB/viewtopic.php?f=111&t=44941

<a href="https://tracker.phpbb.com/browse/PHPBB3-13336">PHPBB3-13336</a>.

### DIFF
--- a/phpBB/includes/functions_user.php
+++ b/phpBB/includes/functions_user.php
@@ -703,7 +703,7 @@ function user_delete($mode, $user_ids, $retain_username = true)
 */
 function user_active_flip($mode, $user_id_ary, $reason = INACTIVE_MANUAL)
 {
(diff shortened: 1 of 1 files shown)


rxu:
!set Event

stevendegroote:
This may be nitpicking, but wouldn't it be better to rename the event to 'core.user_active_flip_before' , so that another one 'core.user_active_flip_after' can be added (similar to the events triggered on user delete ( 'core.delete_user_before' and 'core.delete_user_after')

rxu:
Makes sense, thanks.

PR #3329, opened 2015-01-24
[ticket/13414] Set Content-Length header only when 304 status is not sen...

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-13414

### DIFF
--- a/phpBB/download/file.php
+++ b/phpBB/download/file.php
@@ -509,11 +509,6 @@ function send_file_to_browser($attachment, $upload_dir, $category)
 		}
 	}
 
-	if ($size)
-	{
-		header("Content-Length: $size");
-	}
-
 	// Close the db connection before sending the file
 	$db->sql_close();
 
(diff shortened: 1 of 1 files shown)


phpBB/includes/functions_download.php
```diff
@@ -210,16 +210,16 @@ function send_file_to_browser($attachment, $upload_dir, $category)
...
-	{
-		header("Content-Length: $size");
-	}
-
 	// Close the db connection before sending the file etc.
 	file_gc(false);
 
 	if (!set_modified_headers($attachment['filetime'], $user->browser))
 	{
+		if ($size)
```
bantu: Move this block below Sendfile handling?

bantu:
I'd generally like to have this in 3.0.x too.

marc1706:
@bantu send another PR based on develop-olympus then?

phpBB/includes/functions_download.php
marc1706: sure.

marc1706:
Closing in favor of https://github.com/phpbb/phpbb/pull/3343

PR #3332, opened 2015-01-24
[ticket/10748] Split MS SQL from the db tools file into it's own

nickvergessen:
PHPBB3-10748

Surely the final implementation can be improved even more, but the first task is to reduce the size of tools.php

**PS:** Do NOT close the ticket on merge

### DIFF
--- a/phpBB/phpbb/db/tools/mssql.php
+++ b/phpBB/phpbb/db/tools/mssql.php
@@ -117,6 +117,8 @@ class mssql extends tools
 				$this->sql_layer = 'mssqlnative';
 			break;
 		}
+
+		$this->dbms_type_map = self::get_dbms_type_map();
 	}
 
 	/**
@@ -344,7 +346,7 @@ class mssql extends tools
 		}
 
 		// Get type
-		list($column_type, $orig_column_type) = $this->get_column_type($column_data[0]);
+		list($column_type, ) = $this->get_column_type($column_data[0]);
 
 		// Adjust default value if db-dependent specified
 		if (is_array($column_data[1]))
--- a/phpBB/phpbb/db/tools/mssql.php
+++ b/phpBB/phpbb/db/tools/mssql.php
@@ -1,0 +1,791 @@
+<?php
+/**
+*
+* This file is part of the phpBB Forum Software package.
+*
+* @copyright (c) phpBB Limited <https://www.phpbb.com>
+* @license GNU General Public License, version 2 (GPL-2.0)
+*
+* For full copyright and license information, please see
+* the docs/CREDITS.txt file.
+*
+*/
+
+namespace phpbb\db\tools;
+
+/**
+* Database Tools for handling cross-db actions such as altering columns, etc.
+* Currently not supported is returning SQL for creating tables.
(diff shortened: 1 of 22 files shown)


nickvergessen:
@Noxwizard please test when possible :)

Noxwizard:
@nickvergessen https://bamboo.phpbb.com/browse/PHPBB3-MSSQLNATIVE-BUILD-31/log

Nicofuma:
status? And you need to rebase

nickvergessen:
rebased

phpBB/phpbb/db/tools/tools.php
```diff
@@ -1593,6 +1354,7 @@ function sql_prepare_column_data($table_name, $column_name, $column_data)
 	*/
 	function get_column_type($column_map_type)
 	{
+		$column_type = '';
```
Nicofuma: I don't see where it is used

phpBB/phpbb/search/fulltext_sphinx.php
```diff
@@ -136,7 +136,8 @@ public function __construct(&$error, $phpbb_root_path, $phpEx, $auth, $config, $
 		$this->auth = $auth;
 
 		// Initialize \phpbb\db\tools\tools object
-		$this->db_tools = new \phpbb\db\tools\tools($this->db);
+		global $phpbb_container; // TODO inject into object
```
Nicofuma: TODO?

phpBB/phpbb/db/tools/mssql.php
```diff
@@ -0,0 +1,793 @@
...
+
+		return $existing_indexes;
+	}
+
+	/**
+	* Is the used MS SQL Server a SQL Server 2000?
+	*
+	* @return bool
+	*/
+	protected function mssql_is_sql_server_2000()
```
Nicofuma: we don't need the mssql_ prefix anymore

phpBB/phpbb/db/tools/mssql.php
```diff
@@ -0,0 +1,793 @@
...
+	* Get queries to drop the default constraints of a column
+	*
+	* We need to drop the default constraints of a column,
+	* before being able to change their type or deleting them.
+	*
+	* @param string $table_name
+	* @param string $column_name
+	* @return array		Array with SQL statements
+	*/
+	protected function mssql_get_drop_default_constraints_queries($table_name, $column_name)
```
Nicofuma: we don't need the mssql_ prefix anymore

phpBB/phpbb/search/fulltext_sphinx.php
nickvergessen: We can't do it in the moment, because all search classes need the same constructor...
so this is another task

phpBB/phpbb/db/tools/mssql.php
nickvergessen: By keeping it, we make it clear its only from this class?

Nicofuma: should be fine

PR #3334, opened 2015-01-24
[ticket/13542] Update core.ucp_prefs_personal_data to include $error

paul999:
Update the event so extensions will be able to validate (and deny) changes.

PHPBB3-13542

### DIFF
--- a/phpBB/includes/ucp/ucp_prefs.php
+++ b/phpBB/includes/ucp/ucp_prefs.php
@@ -69,7 +69,7 @@ class ucp_prefs
 				* @var	array	data		Array with current ucp options data
 				* @var	array	error		Array with list of errors
 				* @since 3.1.0-a1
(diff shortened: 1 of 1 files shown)


phpBB/includes/ucp/ucp_prefs.php
```diff
@@ -67,9 +67,11 @@ function main($id, $mode)
 				* @var	bool	submit		Do we display the form only
 				*							or did the user press submit
 				* @var	array	data		Array with current ucp options data
+				* @var	array	error		Array with list of errors
 				* @since 3.1.0-a1
+				* @changed 3.1.4-rc1
```
marc1706: Please also specify what was changed

paul999: Done :)

PR #3337, opened 2015-01-24
[ticket/9109] Properly document and calculate the group settings with va...

nickvergessen:
https://tracker.phpbb.com/browse/PHPBB3-9109


phpBB/includes/functions_privmsgs.php
```diff
@@ -2135,17 +2135,56 @@ function set_user_message_limit()
...
+}
+
+/**
+ * Get the maximum PM setting for the groups of the user
+ *
+ * @param \phpbb\db\driver\driver_interface $db
+ * @param int $user_id
+ * @param string $setting Only 'max_recipients' and 'message_limit' are supported
+ * @return int The maximum setting for all groups of the user, unless one group has '0'
+ */
```
marc1706: Maybe add @throws, too?

marc1706:
So now you're running 2 sql queries instead of one for this?

phpBB/includes/functions_privmsgs.php
```diff
@@ -2135,17 +2135,56 @@ function set_user_message_limit()
...
+function phpbb_get_max_setting_from_group(\phpbb\db\driver\driver_interface $db, $user_id, $setting)
+{
+	if ($setting !== 'max_recipients' && $setting !== 'message_limit')
+	{
+		throw new InvalidArgumentException('Setting "' . $setting . '" is not supported');
+	}
+
+	// Get maximum number of allowed recipients
+	$sql = 'SELECT MAX(g.group_' . $setting . ') as max_setting
```
marc1706: Select MAX() and MIN() and then check if min is 0 instead of running another query?

phpBB/language/en/acp/groups.php
```diff
@@ -83,7 +83,7 @@
 	'GROUP_MEMBERS'					=> 'Group members',
 	'GROUP_MEMBERS_EXPLAIN'			=> 'This is a complete listing of all the members of this usergroup. It includes separate sections for leaders, pending and existing members. From here you can manage all aspects of who has membership of this group and what their role is. To remove a leader but keep them in the group use Demote rather than delete. Similarly use Promote to make an existing member a leader.',
 	'GROUP_MESSAGE_LIMIT'			=> 'Group private message limit per folder',
-	'GROUP_MESSAGE_LIMIT_EXPLAIN'	=> 'This setting overrides the per-user folder message limit. A value of 0 means the user default limit will be used.',
+	'GROUP_MESSAGE_LIMIT_EXPLAIN'	=> 'This setting overrides the per-user folder message limit. The maximum for all groups of the user is used to determinate the actual value.<br />Set this value to 0 to overwrite the setting for all users of this group with the board-wide setting.',
```
marc1706: Should be determine and not determinate

phpBB/language/en/acp/groups.php
```diff
@@ -92,7 +92,7 @@
 	'GROUP_OPEN'					=> 'Open',
 	'GROUP_PENDING'					=> 'Pending members',
 	'GROUP_MAX_RECIPIENTS'			=> 'Maximum number of allowed recipients per private message',
-	'GROUP_MAX_RECIPIENTS_EXPLAIN'	=> 'The maximum number of allowed recipients in a private message. If 0 is entered, the board-wide setting is used.',
+	'GROUP_MAX_RECIPIENTS_EXPLAIN'	=> 'The maximum number of allowed recipients in a private message.  The maximum for all groups of the user is used to determinate the actual value.<br />Set this value to 0 to overwrite the setting for all users of this group with the board-wide setting.',
```
marc1706: See above

phpBB/includes/functions_privmsgs.php
nickvergessen: I dont query min, but only for 0 atm.
Since only 0 is a special value

marc1706: I meant you should run "SELECT MAX(g.group_' . $setting . ') as max_setting, MIN(g.group_' . $setting . ') as min_setting" instead of running another query. This should also cover the next query you are running. Also, your second query fails for the message_limit setting.

bantu: @marc1706 At least the WHERE clauses of the two queries are different. Not sure this works. Might involve a bit more work.

marc1706: The second WHERE clause is used to select the minimum setting 0 for g.group_max_recipients. This should be wrong for g_group_message_limit

phpBB/includes/functions_privmsgs.php
```diff
@@ -2135,17 +2135,56 @@ function set_user_message_limit()
...
+	if ($max_setting)
+	{
+		// If the user is limited by a group, check whether he is also set to
+		// unlimited in another group (value 0)
+		$sql = 'SELECT g.group_' . $setting . ' as max_setting
+			FROM ' . GROUPS_TABLE . ' g, ' . USER_GROUP_TABLE . ' ug
+			WHERE ug.user_id = ' . (int) $user_id . '
+				AND ug.user_pending = 0
+				AND ug.group_id = g.group_id
+				AND g.group_max_recipients = 0';
```
nickvergessen: Should be `g.group_' . $setting . '` here

marc1706:
PR against this PR branch: https://github.com/nickvergessen/phpbb/pull/14

nickvergessen:
i think we should do that for 3.2

marc1706:
Agreed.

bantu:
Size-wise, I think this is fine as a bug fix, assuming it is properly tested.

bantu:
As for behaviour, I think answers to https://tracker.phpbb.com/browse/PHPBB3-9109?focusedCommentId=35792&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-35792 need to be documented as comments in code somewhere.

nickvergessen:
Replaced by #3383

PR #3338, opened 2015-01-25
[ticket/12933] Handle *wildcard character in native search

dhruvgoel92:
https://tracker.phpbb.com/browse/PHPBB3-12933
PHPBB3-12933
Olympus Patch: https://github.com/phpbb/phpbb/pull/3346


nickvergessen:
Please rebase and throw away the first commit.
It has been merged with another ticket :)

tests/search/native_test.php
```diff
@@ -56,6 +58,30 @@ public function keywords()
...
+				true,
+				array('\'baaz%\''),
+				array(),
+				array(),
+			),
+			array(
+				'ba*az',
+				'all',
+				true,
+				array('\'ba%az\''),
```
dhruvgoel92: This looks weird, but split_keyword has been designed in such a way that it returns these strings in case of partial matches.

marc1706:
This patch should also be created for develop-olympus (currently searching for test* doesn't work).

PR #3341, opened 2015-01-26
[ticket/13544] Fixing wrong SQL query for permission_unset() for roles

brunoais:
https://tracker.phpbb.com/browse/PHPBB3-13544

PHPBB3-13544


phpBB/phpbb/db/migration/tool/permission.php
```diff
@@ -537,7 +537,8 @@ public function permission_unset($name, $auth_option, $type = 'role')
 				}
 
 				$sql = 'DELETE FROM ' . ACL_ROLES_DATA_TABLE . '
-					WHERE ' . $this->db->sql_in_set('auth_option_id', $to_remove);
+					WHERE ' . $this->db->sql_in_set('auth_option_id', $to_remove) . '
+						AND role_id = ' . $this->db->sql_escape($role_id);
```
nickvergessen: role_id is integer? if so cast to int instead of sql_escape()

brunoais: Yeah. That's also my favorite way.

PR #3342, opened 2015-01-27
[ticket/13549] Compare ORIG_PATH_INFO with SCRIPT_NAME for checking trailing paths

marc1706:
The ORIG_PATH_INFO on IIS also contains the script name. Only use that
for killing the script after removing the script name from ORIG_PATH_INFO.

Ticket: https://tracker.phpbb.com/browse/PHPBB3-13549

### DIFF
--- a/phpBB/includes/startup.php
+++ b/phpBB/includes/startup.php
@@ -104,8 +104,15 @@ function deregister_globals()
  */
 function phpbb_has_trailing_path($phpEx)
 {
+	// Check if web server is IIS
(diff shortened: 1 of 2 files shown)


marc1706:
Tested to be working on Microsoft IIS 8.5 & PHP 5.6.0

nickvergessen:
https://www.phpbb.com/community/viewtopic.php?p=13919786#p13919786
Seems to happen on non IIS aswell?

marc1706:
@nickvergessen Yeah, which is why this doesn't check for IIS anymore and should work for all server types. ;)

bantu:
Replaced by #3348

PR #3343, opened 2015-01-28
[ticket/13414] Set Content-Length header only if status is not 304

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-13414

### DIFF
--- a/phpBB/download/file.php
+++ b/phpBB/download/file.php
@@ -509,16 +509,18 @@ function send_file_to_browser($attachment, $upload_dir, $category)
 		}
 	}
 
-	if ($size)
-	{
-		header("Content-Length: $size");
-	}
-
 	// Close the db connection before sending the file
 	$db->sql_close();
 
 	if (!set_modified_headers($attachment['filetime'], $user->browser))
 	{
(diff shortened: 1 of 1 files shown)


bantu:
You're sending a header after sending the complete body.

marc1706:
Oh, should be before flush(). Then again, I guess the previous location was also fine (right after the beginning of the if)?

bantu:
@marc1706 Yes, that's where it has to go for olympus. ascraeus was a different story.

marc1706:
@bantu updated PR and also created an additional one for ascraeus

phpBB/download/file.php
```diff
@@ -509,16 +509,17 @@ function send_file_to_browser($attachment, $upload_dir, $category)
...
-	{
-		header("Content-Length: $size");
-	}
-
 	// Close the db connection before sending the file
 	$db->sql_close();
 
 	if (!set_modified_headers($attachment['filetime'], $user->browser))
 	{
+		// Send Content-Length after delivering file
```
bantu: "after delivering file" ?

marc1706:
@bantu Should be better now :)

PR #3350, opened 2015-01-29
Ticket/10388 Use Twig's Javascript escape filter instead of addslashes for LA_ keys

VSEphpbb:
This applies twigs javascript escaping to lang vars that are used in JS. Only caveat is if an LA_KEY is accidentally used in HTML context, it renders the escaping (ie: Hello\x20World). I have corrected the instances I found where an LA_KEY was used outside of javascript in this PR. LA_KEYS should always be in javascript only,  anyways.

https://tracker.phpbb.com/browse/PHPBB3-10388

PHPBB3-10388

### DIFF
--- a/phpBB/styles/prosilver/template/captcha_recaptcha.html
+++ b/phpBB/styles/prosilver/template/captcha_recaptcha.html
@@ -29,7 +29,7 @@
 		</div>
 		</noscript>
 
-		<a href="http://www.google.com/intl/{LA_RECAPTCHA_LANG}/policies/" target="_blank" class="recaptcha-responsive" style="display: none"><img alt="" width="71" height="36" src="{RECAPTCHA_SERVER}/img/clean/logo.png"></a>
+		<a href="http://www.google.com/intl/{L_RECAPTCHA_LANG}/policies/" target="_blank" class="recaptcha-responsive" style="display: none"><img alt="" width="71" height="36" src="{RECAPTCHA_SERVER}/img/clean/logo.png"></a>
 	</dd>
 	</dl>
 <!-- ELSE -->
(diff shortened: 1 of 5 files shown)


phpBB/styles/prosilver/template/captcha_recaptcha.html
```diff
@@ -29,7 +29,7 @@ <h3 class="captcha-title">{L_CONFIRMATION}</h3>
 		</div>
 		</noscript>
 
-		<a href="http://www.google.com/intl/{LA_RECAPTCHA_LANG}/policies/" target="_blank" class="recaptcha-responsive" style="display: none"><img alt="" width="71" height="36" src="{RECAPTCHA_SERVER}/img/clean/logo.png"></a>
+		<a href="http://www.google.com/intl/{L_RECAPTCHA_LANG}/policies/" target="_blank" class="recaptcha-responsive" style="display: none"><img alt="" width="71" height="36" src="{RECAPTCHA_SERVER}/img/clean/logo.png"></a>
```
nickvergessen: this was intentionally, because it is inside a URL.
Since it should only contain a few characters to identify the language, it should be fine to still use `LA_`?

nickvergessen:
Since this is a quite big behaviour change, I'd say 3.2?

Nicofuma:
I agree with 3.2

phpBB/styles/prosilver/template/captcha_recaptcha.html
VSEphpbb: is there any reason it needs escaping though? if it's just a few letters, and none would need escaping LA_ isn't needed. However, it ever did need escaping, it would be escaped to something not URL friendly, like `\x20` ?

marc1706:
I'd also say 3.2.

nickvergessen:
Since this is template related and we redid some template stuff, I'd say please do a new PR to avoid problems with tests after the merge. (Better save then sorry)

phpBB/styles/prosilver/template/captcha_recaptcha.html
nickvergessen: to be save against translators breaking stuff?
But I guess no

PR #3352, opened 2015-01-29
Ticket/10388 Use Twig's JS escape filter instead of addslashes on LA_ keys

VSEphpbb:
https://tracker.phpbb.com/browse/PHPBB3-10388

PHPBB3-10388

### DIFF
--- a/phpBB/phpbb/template/twig/extension.php
+++ b/phpBB/phpbb/template/twig/extension.php
@@ -71,7 +71,7 @@ class extension extends \Twig_Extension
 	{
 		return array(
 			new \Twig_SimpleFilter('subset', array($this, 'loop_subset'), array('needs_environment' => true)),
-			new \Twig_SimpleFilter('addslashes', 'addslashes'), // Deprecate addslashes in phpBB 3.1.4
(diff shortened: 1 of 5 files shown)


phpBB/phpbb/template/twig/extension.php
```diff
@@ -71,7 +71,7 @@ public function getFilters()
 	{
 		return array(
 			new \Twig_SimpleFilter('subset', array($this, 'loop_subset'), array('needs_environment' => true)),
-			new \Twig_SimpleFilter('addslashes', 'addslashes'),
+			new \Twig_SimpleFilter('addslashes', 'addslashes'), // Deprecated 3.2.0
```
marc1706: I'd prefer if you'd add the comment above this. I think it should say this:
```@deprecated 3.2.0 Uses twig's JS escape method instead of addslashes```

nickvergessen:
Do we have a release highlights page on the wiki already?
If so please add it, if not please ask nicofuma to create one and add it :)

marc1706:
Added it to a newly created release highlights page already:
https://wiki.phpbb.com/Release_Highlights/3.2.0-a1#Notable_Changes

PR #3358, opened 2015-02-02
[ticket/13348] Fetch complete rowset and return data row by row

marc1706:
Instead of keeping $result as a static pointer and calling sql_fetchrow()
on every iteration of get_time(), fetch the complete set ofrows upon querying
the database. Afterwards, return the rows one by one using array_shift().

Ticket: https://tracker.phpbb.com/browse/PHPBB3-13348

### DIFF
--- a/phpBB/phpbb/feed/base.php
+++ b/phpBB/phpbb/feed/base.php
@@ -68,7 +68,7 @@ abstract class base
 	var $separator_stats = "\xE2\x80\x94"; // &mdash;
 
 	/** @var array Feed database rows */
-	protected $feed_row;
+	protected $feed_rows;
 
 	/**
 	* Constructor
@@ -232,7 +232,7 @@ abstract class base
 
 	function get_item()
 	{
-		if (!is_array($this->feed_row))
+		if (!is_array($this->feed_rows))
 		{
 			if (!$this->get_sql())
 			{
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/feed/base.php
```diff
@@ -229,9 +229,9 @@ function get_passworded_forums()
 
 	function get_item()
 	{
-		static $result;
+		static $feed_row;
```
nickvergessen: use a member instead of a static variable for better testability?

phpBB/phpbb/feed/base.php
```diff
@@ -229,9 +229,9 @@ function get_passworded_forums()
...
 
-		if (!isset($result))
+		if (!isset($feed_row))
```
nickvergessen: then check `!is_array()` here

marc1706: You mean (!isset($feed_row) || !is_array($feed_row)) ?

marc1706: Oh, you meant just if it's an array

phpBB/phpbb/feed/base.php
```diff
@@ -67,6 +67,9 @@
 	*/
 	var $separator_stats = "\xE2\x80\x94"; // &mdash;
 
+	/** @var array Feed database rows */
+	protected $feed_row;
```
nickvergessen: rows?

bantu:
Why is https://github.com/marc1706/phpbb/blob/ticket/13348/phpBB/phpbb/feed/base.php#L125-134 insufficient?

marc1706:
I guess because it wasn't used at all yet? I guess because the class was called statically. Anyhow, didn't see this but it looks like we should be able to use that then instead.

bantu:
@marc1706 Use here. https://github.com/phpbb/phpbb/blob/develop/phpBB/feed.php#L151

marc1706:
@bantu I meant that $this->result wasn't used. Therefore calling the close() method didn't do anything.

bantu:
@marc1706 Ah. Ok. Very good. Is olympus affected too?

bantu:
It is.

marc1706:
Yeah, the olympus PR probably won't merge into dev-asc though. I'll create a new PR for olympus right away though.

marc1706:
Replacement PR against develop-olympus: https://github.com/phpbb/phpbb/pull/3365

PR #3360, opened 2015-02-02
[ticket/13569] Add missing sql_freeresults and remove unneeded results

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-13569


marc1706:
Meh, wrong target branch

PR #3364, opened 2015-02-03
[ticket/13564] Unlink user accounts upon deleting user

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-13564


phpBB/includes/functions_user.php
```diff
@@ -398,7 +398,7 @@ function user_add($user_row, $cp_data = false, $notifications_data = null)
  */
 function user_delete($mode, $user_ids, $retain_username = true)
 {
-	global $cache, $config, $db, $user, $phpbb_dispatcher;
+	global $cache, $config, $db, $user, $phpbb_container, $phpbb_dispatcher;
```
nickvergessen: needs rebasing since it was added by another PR

tests/functions_user/delete_user_test.php
```diff
@@ -38,6 +38,16 @@ protected function setUp()
 		$phpbb_dispatcher = new phpbb_mock_event_dispatcher();
 		$phpbb_container = new phpbb_mock_container_builder();
 		$phpbb_container->set('notification_manager', new phpbb_mock_notification_manager());
+		$phpbb_container->set(
+			'auth.provider.db',
+			new phpbb_mock_auth_provider()
```
nickvergessen: use a phpunit Mock from the MockBuilder and check whether the method is called?

phpBB/includes/functions_user.php
```diff
@@ -510,6 +513,38 @@ function user_delete($mode, $user_ids, $retain_username = true)
...
+			$provider_data = $auth_provider->get_auth_link_data($user_id);
+
+			if ($provider_data !== null)
+			{
+				$link_data = array(
+					'user_id' => $user_id,
+					'link_method' => 'user_delete',
+				);
+
+				// BLOCK_VARS might contain hidden fields necessary for unlinking accounts
```
nickvergessen: can't we just send this to the provider and let it decide what to do on it's own?

PR #3367, opened 2015-02-04
[ticket/13480] Add an event to viewtopic.php to modify template vars

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-13480


marc1706:
Grr, wrong target branch.

PR #3288, opened 2015-01-12
Added an event to viewtopic.php to modify template vars.

DerDave:
[feature/event]
(This is my first pull requst, I hope I don't screw this up)
I added an event to viewtopic.php tomodify the template vars as described in the ticket.

PHPBB3-13480

PR #3368, opened 2015-02-04
[ticket/13480] Add an event to viewtopic.php to modify template vars

marc1706:
https://tracker.phpbb.com/browse/PHPBB3-13480


marc1706:
Replaces https://github.com/phpbb/phpbb/pull/3288

nickvergessen:
There should already be an event that allows overwriting these values later on, not sure we need to add them to the event

PR #3372, opened 2015-02-05
[ticket/13579] Fix typo in embarrassed smilie language key

n-aleha:
https://tracker.phpbb.com/browse/PHPBB3-13579

PHPBB3-13579


nickvergessen:
Since this would break existing smilies, I think it's not worth it?
Only the language key is wrong, so shouldn't be too much trouble.

PR #3377, opened 2015-02-06
[ticket/13578] - Add ucp_register.php event

RMcGirr83:
PHPBB-13578

### DIFF
--- a/phpBB/includes/ucp/ucp_register.php
+++ b/phpBB/includes/ucp/ucp_register.php
@@ -30,7 +30,7 @@ class ucp_register
 	function main($id, $mode)
 	{
 		global $config, $db, $user, $auth, $template, $phpbb_root_path, $phpEx;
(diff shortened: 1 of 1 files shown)


RMcGirr83:
@nickvergessen I am not happy with this lingo.  Eg, for a new event to add to the $data array within the file I think it better to rename this to possibly core.ucp_register_data_check?  I completely forgot about the ability to have an event add stuffs to the data array.  Thoughts?

nickvergessen:
You can still rename the event until it was merged.
I think the name should be changed yes. Needs to include something like `check` and `after`/`before` so we know whether the event is before the user creation or after...

RMcGirr83:
Joas,

I'm having one hell of a time in merging my fix into this PR so am closing and will re-open a new one.  This github stuff confuses the crap out of me and I just can't get it to do what I want it to do.

PR #3384, opened 2015-02-08
[ticket/13592] Allow changing get_visibility_sql's result

brunoais:
https://tracker.phpbb.com/browse/PHPBB3-13592
PHPBB3-13592

### DIFF
--- a/phpBB/phpbb/content_visibility.php
+++ b/phpBB/phpbb/content_visibility.php
@@ -146,6 +146,7 @@ class content_visibility
 		$where_sql = '';
 
 		$get_visibility_sql_overwrite = false;
+
 		/**
 		* Allow changing the result of calling get_visibility_sql
 		*
--- a/phpBB/phpbb/content_visibility.php
+++ b/phpBB/phpbb/content_visibility.php
@@ -150,7 +150,7 @@ class content_visibility
 		/**
 		* Allow changing the result of calling get_visibility_sql
 		*
-		* @event core.phpbb_content_get_visibility_sql_visibility_before
+		* @event core.phpbb_content_visibility_get_visibility_sql_before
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/content_visibility.php
```diff
@@ -143,12 +143,42 @@ public function get_count($mode, $data, $forum_id)
...
+		* 													It must be either boolean or string
+		* @since 3.1.4-RC1
+		*/
+		$vars = array(
+			'where_sql',
+			'mode',
+			'forum_id',
+			'table_alias',
+			'get_visibility_sql_overwrite',
+		);
```
nickvergessen: remove `forums`

```
LogicException: The event 'core.phpbb_content_visibility_get_forums_visibility_before' from file 'phpbb/content_visibility.php:204' already exists in file 'phpbb/content_visibility.php'

brunoais: I feel like if this went back in time. I had changed all that O_O.
I wonder if it was git stash that borked it when I made a pull before 
the commit...
I'll recheck everything again

brunoais:
!unset WIP

phpBB/phpbb/content_visibility.php
brunoais: It was supposed to be `$forum_id` from the start. I think it should stay as it is one of the parameters of the method.
I think that keeping `$where_sql` is better than renaming to `$sql` to maintain the same name as used in the other methods of this object

brunoais:
I don't know what I was thinking but the name didn't make proper sense.
This one makes more sense and is on-par with the other event names in this object.

marc1706:
Merged by https://github.com/phpbb/phpbb/commit/0daf148aaa79a5ebd050afb978b7b8572b9d7ad3

PR #2473, opened 2014-05-17
[ticket/12559] Add forum setting to limit subforums legend to direct children only

PayBas:
https://tracker.phpbb.com/browse/PHPBB3-12559
http://area51.phpbb.com/phpBB/viewtopic.php?f=108&t=45585
PHPBB3-12559

PR #3386, opened 2015-02-09
[ticket/12559] Add forum setting to limit subforums legend to direct children only

PayBas:
https://tracker.phpbb.com/browse/PHPBB3-12559
http://area51.phpbb.com/phpBB/viewtopic.php?f=108&t=45585
PHPBB3-12559

>Actually, adding this as an option for categories in the ACP module is very tricky. There isn't really any place for a unique setting that is shared by forums & categories, but not links. We can't simply duplicate form components for obvious reasons, and introducing yet another table column for categories which does exactly the same thing as the forum version... that's just silly.

>So I'm going to leave it as is. If admins really want to limit subforums for categories, they can change the forum type to "forum", change the setting, and change the type back to "category".

>Not ideal, but it's the cleanest way, code-wise.

Copied from the original PR https://github.com/phpbb/phpbb/pull/2473

>please rebase onto develop and replace your request_var() calls by $request->variable()

Done


phpBB/phpbb/db/migration/data/v310/forums_legend_limit.php
```diff
@@ -0,0 +1,49 @@
...
+*
+* @copyright (c) phpBB Limited <https://www.phpbb.com>
+* @license GNU General Public License, version 2 (GPL-2.0)
+*
+* For full copyright and license information, please see
+* the docs/CREDITS.txt file.
+*
+*/
+
+namespace phpbb\db\migration\data\v310;
```
nickvergessen: `v31x`

phpBB/phpbb/db/migration/data/v310/forums_legend_limit.php
```diff
@@ -0,0 +1,49 @@
...
+class forums_legend_limit extends \phpbb\db\migration\migration
+{
+	public function effectively_installed()
+	{
+		return $this->db_tools->sql_column_exists($this->table_prefix . 'forums', 'display_subforum_limit');
+	}
+
+	static public function depends_on()
+	{
+		return array('\phpbb\db\migration\data\v310\beta3');
```
nickvergessen: use latest?

Nicofuma:
I don't remember why the PR wasn't merged, @nickvergessen any clue?

nickvergessen:
Needs proper testing, nothing more?

PR #3387, opened 2015-02-10
[ticket/13595] Remove unused instances of the bbcode class

s9e:
There are a few unused instances of the bbcode class created in various places. I think it's just leftover code from PHPBB3-11635.

https://tracker.phpbb.com/browse/PHPBB3-13595

### DIFF
--- a/phpBB/includes/functions_posting.php
+++ b/phpBB/includes/functions_posting.php
@@ -1036,7 +1036,7 @@ function load_drafts($topic_id = 0, $forum_id = 0, $id = 0, $pm_action = '', $ms
 */
 function topic_review($topic_id, $forum_id, $mode = 'topic_review', $cur_post_id = 0, $show_quote_button = true)
 {
-	global $user, $auth, $db, $template, $bbcode, $cache;
+	global $user, $auth, $db, $template, $cache;
 	global $config, $phpbb_root_path, $phpEx, $phpbb_container, $phpbb_dispatcher;
 
 	$phpbb_content_visibility = $phpbb_container->get('content.visibility');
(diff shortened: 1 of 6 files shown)


bantu:
It is not quite clear to me whether https://tracker.phpbb.com/browse/PHPBB3-13595 also affects 3.1 due to the various subtasks in that ticket.
@nickvergessen If this only removes dead code and applies to 3.1, it should go into 3.1 as it is a performance improvement.

marc1706:
I agree with what bantu said

s9e:
@bantu There's no subtask to PHPBB3-13595 so I guess you were talking about PHPBB3-11635? The patches that made this code redundant were applied a long time ago, it's not related to any other future changes.

It's easy enough to verify that no other file uses those `$bbcode` instances:
```
$ grep -R "\\\$bbcode->" phpBB
phpBB/includes/functions_content.php:                   $bbcode->bbcode($bitfield);
phpBB/includes/functions_content.php:           $bbcode->bbcode_second_pass($text, $uid);
```

bantu:
>  There's no subtask to PHPBB3-13595 so I guess you were talking about PHPBB3-11635? 

Correct. Updated.

nickvergessen:
@s9e so are the stray variables also in 3.1? if so, please fix there.

s9e:
Yes, it's in 3.0 onward, including 3.1. What branch should it be based on?

nickvergessen:
develop-ascraeus (sorry for the troubles, we should change the branch naming I think.

s9e:
Ok, I understand now. It was already based on develop-ascraeus, I just didn't pick the right branch for the PR.

PR #3391, opened 2015-02-11
[ticket/13597] Modify variable-variable syntax to be compatible with PHP7

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-13597


bantu:
https://wiki.php.net/rfc/uniform_variable_syntax lists other types of changes which may be required.

PR #3392, opened 2015-02-11
[ticket/13493] Append page name to base url if it doesn't contain it

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-13493

### DIFF
--- a/phpBB/phpbb/controller/helper.php
+++ b/phpBB/phpbb/controller/helper.php
@@ -142,6 +142,12 @@ class helper
 
 		$base_url = $context->getBaseUrl();
 
+		// Append page name if base URL does not contain it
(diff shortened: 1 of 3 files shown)


Nicofuma:
I'm not sure that I agree with this PR. We should have at least one test failing without it

marc1706:
The test I added should fail if you remove the change to the controller helper.

marc1706:
Right now this causes a rather bad issue. If you do not have mod rewrite enabled and access the website. for example via http://example.com, and click on an extension link the URL will be missing app.php. This will obviously causes 404 errors.

nickvergessen:
Merged via http://git.io/veRC0

PR #3396, opened 2015-02-14
[ticket/13524]Local Gallery avatars split every 50 images. Subcategories in gallery

kamijoutouma:
Local Gallery avatars split every 50 images. 
(scratch 50 images we want it to look right so 40 images instead)
Helps fix long loading times when gallery contains a large amount of avatars.
Local gallery avatars can have folders that are inside of a folder in gallery.
Example:
sub categories under one category like
/images/avatars/gallery/toaru/index 
/images/avatars/gallery/toaru/misaka
instead of just 
/images/avatars/gallery/toaru/

PHPBB3-13524

### DIFF
--- a/phpBB/adm/style/acp_groups.html
+++ b/phpBB/adm/style/acp_groups.html
@@ -124,8 +124,12 @@
 		</dl>
 		<!-- IF S_DISPLAY_GALLERY -->
 			<dl>
-				<dt><label>{L_AVATAR_GALLERY}:</label></dt>
-				<dd><input class="button2" type="submit" name="display_gallery" value="{L_DISPLAY_GALLERY}" /></dd>
+				<dt><label for="category">{L_AVATAR_CATEGORY}:</label></dt>
+				<dd><select name="category" id="category">{S_CAT_OPTIONS}</select>&nbsp;</dd>
+				<dt><label for="subcategory">{L_AVATAR_SUBCATEGORY}:</label></dt>
+				<dd><select name="subcategory" id="subcategory">{S_SUBCAT_OPTIONS}</select>&nbsp;</dd>
+				<dt><label for="subcategory">{L_AVATAR_PAGE}:</label></dt>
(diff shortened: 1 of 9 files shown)


kamijoutouma:
not done making changes

kamijoutouma:
somehow github is being dumb today https://github.com/phpbb/phpbb/pull/3397 is part of the commit here

kamijoutouma:
also cant find the spot for this in prosilver but the fix is applied to subsilver

kamijoutouma:
somehow github is being dumb today

DavidIQ:
Something like this wouldn't/shouldn't go into 3.0 which is approaching EOL sometime in the near future.

kamijoutouma:
ok

kamijoutouma:
well the rest of the changes are for 3.1.3 whenever i do them

kamijoutouma:
that was the last fix for 3.0.13 i fixed a counting issue working on 3.1.3

kamijoutouma:
uh doing this in 3.1.3 is a bit confusing i'm still learning OOP so this puts me in a bind if someone is willing to take the changes i made to 3.0.13 and make them work in 3.1.3 then by all means please do so 
because at the rate i'm going this wont make it to version 3.1.4-RC1 sorry

kamijoutouma:
i'm just now noticing this but gallery avatars in 3.1.3 do not show up i see the directory names as the categories but none of the images it just says None when i go to a category

kamijoutouma:
created an interesting bug in subcats fixed

kamijoutouma:
also noticed that phpbb 3.1.3 does part of this already it only needs pagination for images when its over 40-50

kamijoutouma:
this method doesn't have to be used as its technically a bad design 

its just a suggestion

nickvergessen:
I don't think we should fix that in 3.0, @bantu ?

kamijoutouma:
It's just an example doesn't have to be used 
also sorry I meant to close this

PR #3398, opened 2015-02-14
[ticket/13612] Run meta refresh until extension is fully enabled

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-13612

### DIFF
--- a/tests/test_framework/phpbb_functional_test_case.php
+++ b/tests/test_framework/phpbb_functional_test_case.php
@@ -410,7 +410,26 @@ class phpbb_functional_test_case extends phpbb_test_case
 		$form = $crawler->selectButton('Enable')->form();
 		$crawler = self::submit($form);
 		$this->add_lang('acp/extensions');
-		$this->assertContainsLang('EXTENSION_ENABLE_SUCCESS', $crawler->filter('div.successbox')->text());
+
+		$meta_refresh = $crawler->filter('meta[http-equiv="refresh"]');
+
+		// Wait for extension to be fully enabled
+		while (sizeof($meta_refresh))
+		{
+			preg_match('#url=.+/(adm+.+)#', $meta_refresh->attr('content'), $match);
+			$url = $match[1];
+			$crawler = self::request('POST', $url);
+			$meta_refresh = $crawler->filter('meta[http-equiv="refresh"]');
+		}
+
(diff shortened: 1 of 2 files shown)


tests/test_framework/phpbb_functional_test_case.php
```diff
@@ -410,7 +410,27 @@ public function install_ext($extension)
...
+		if (sizeof($meta_refresh))
+		{
+			do
+			{
+				preg_match('#url=.+/(adm+.+)#', $meta_refresh->attr('content'), $match);
+				$url = $match[1];
+				$crawler = self::request('POST', $url);
+				$meta_refresh = $crawler->filter('meta[http-equiv="refresh"]');
+			}
+			while (sizeof($meta_refresh));
```
nickvergessen: put `while (!empty($meta_refresh))` before the loop and remove the if ?

marc1706: Using empty() to check this does not work.

tests/test_framework/phpbb_functional_test_case.php
```diff
@@ -410,7 +410,26 @@ public function install_ext($extension)
...
+			$meta_refresh = $crawler->filter('meta[http-equiv="refresh"]');
+		}
+
+		if (!empty($meta_refresh))
+		{
+			$this->assertContainsLang('EXTENSIONS_ADMIN', $crawler->filter('.main > h1')->text());
+		}
+		else
+		{
+			$this->assertContainsLang('EXTENSION_ENABLE_SUCCESS', $crawler->filter('div.successbox')->text());
```
nickvergessen: shouldn't there always be a successbox at the end, because you loop, until `!empty($meta_refresh)` is true?

marc1706: That's the weird thing. There is no success box in that case.

nickvergessen: Then fix it? :p

marc1706: Done :P

marc1706:
BTW, this PR is based on develop-ascraeus ...

Nicofuma:
merged by  50b3c2b

PR #3399, opened 2015-02-14
[ticket/13614] Remove phpbb_pcre_utf8_support() and assume Unicode is supported by PCRE

s9e:
Some parts of phpBB check for Unicode support in PCRE using phpbb_pcre_utf8_support(), others such as get_preg_expression() don't. I assume that Unicode support was not a requirement in the past but now it is and the non-Unicode branches of those checks are now unreachable.

This PR removes those checks as well as dead code. You may want to keep phpbb_pcre_utf8_support() in 3.1 in case an extension relies on it.


phpBB/includes/functions.php
```diff
@@ -5448,22 +5436,6 @@ function phpbb_user_session_handler()
 }
 
 /**
-* Check if PCRE has UTF-8 support
-* PHP may not be linked with the bundled PCRE lib and instead with an older version
-*
-* @return bool	Returns true if PCRE (the regular expressions library) supports UTF-8 encoding
-*/
-function phpbb_pcre_utf8_support()
```
MGaetan89: The function should be deprecated and moved to `functions_compatibility.php` first.
Then removed in phpBB 3.3.

DavidIQ: Not sure that's needed. The main reason the functions_compatibility.php script exists is for extension or MOD compatibility. Can't think of a lot of cases this function would be used outside of phpBB in a MOD or extension, unless you know otherwise.

MGaetan89: I didn't even know that this function exists before this PR, so I can't argue against you on this.

phpBB/includes/functions_user.php
```diff
@@ -1626,89 +1626,37 @@ function validate_username($username, $allowed_username = false)
...
-			else if ($mbstring)
-			{
-				$regex = '[[:upper:][:lower:][:digit:]]+';
-			}
-			else
-			{
-				$pcre = true;
-				$regex = '[a-zA-Z0-9]+';
-			}
+			$regex = '[\p{Lu}\p{Ll}\p{N}]+';
```
nickvergessen: This value is written to the database, do we need to update it?

nickvergessen:
I think we should fix that in 3.2

phpBB/includes/functions_user.php
s9e: It shouldn't be necessary since the only installations that execute the else branches are those that lack PCRE support or Unicode support. And on those installations, other parts of phpBB already don't work anymore.

phpBB/includes/functions.php
bantu: It looks like we could easily add a dummy function returning true to functions_compatibility.php. @s9e What do you think?

bantu:
I agree. This patch is too big. Pleaes consider sending a new PR against master.

nickvergessen:
Please send against `3.1.x` instead of `develop-ascraeus`

phpBB/includes/functions.php
s9e: I have no opinion on that matter. There are no calls to `phpbb_pcre_utf8_support()` left in the source but if you think an extension may call that function it can be moved to `functions_compatibility.php` and optionally replaced with a stub.

bantu:
@nickvergessen I thought we wanted master?

nickvergessen:
yeah, hence my edit ;)

s9e:
New PR https://github.com/phpbb/phpbb/pull/3597

PR #3400, opened 2015-02-14
[ticket/13615] Avatar Gallery shows categories but no images in subsilver2

kamijoutouma:
Avatar Gallery shows categories but no images in subsilver2

you guys forgot to put the id

also has avatar_local_col.avatar_local_col

should be avatar_local_row.avatar_local_col

PHPBB3-13615

PR #3403, opened 2015-02-15
[ticket/13615] Correctly display images for gallery avatar in subsilver2

marc1706:
Avatar Gallery shows categories but no images in subsilver2.
The ID was not specified in the template. Also, the correct parent template
loop is called avatar_local_row and not avatar_local_col.

Ticket: https://tracker.phpbb.com/browse/PHPBB3-13615


marc1706:
Replaces #3400

PR #3404, opened 2015-02-15
[ticket/13622] Allow changing how get_visibility_sql() works

brunoais:
https://tracker.phpbb.com/browse/PHPBB3-13622

PHPBB3-13622

### DIFF
--- a/phpBB/phpbb/content_visibility.php
+++ b/phpBB/phpbb/content_visibility.php
@@ -146,6 +146,7 @@ class content_visibility
 		$where_sql = '';
 
 		$get_visibility_sql_overwrite = false;
+
 		/**
 		* Allow changing the result of calling get_visibility_sql
 		*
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/content_visibility.php
```diff
@@ -143,12 +143,42 @@ public function get_count($mode, $data, $forum_id)
 	*/
 	public function get_visibility_sql($mode, $forum_id, $table_alias = '')
 	{
+		$where_sql = '';
+
+		$get_visibility_sql_overwrite = false;
+		/**
```
marc1706: Please add a space before this.

brunoais: Added a new line.

marc1706: Which is what I meant, thanks. ;)

marc1706:
@nickvergessen comments?

nickvergessen:
I guess its fine?

nickvergessen:
There is already an event with the same name and same changes.
Only the variable comments are different. So I think this is not needed anymore?

PR #3407, opened 2015-02-16
[ticket/9457] Add WAI-ARIA landmarks to the Prosilver template files

Senky:
[PHPBB3-9457](https://tracker.phpbb.com/browse/PHPBB3-9457)

### DIFF
--- a/phpBB/styles/prosilver/template/navbar_footer.html
+++ b/phpBB/styles/prosilver/template/navbar_footer.html
@@ -1,4 +1,4 @@
-<div class="navbar">
+<div class="navbar" role="navigation">
 	<div class="inner">
 
 	<ul id="nav-footer" class="linklist bulletin" role="menubar">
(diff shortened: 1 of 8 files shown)


phpBB/styles/prosilver/template/navbar_footer.html
```diff
@@ -1,4 +1,4 @@
-<div class="navbar">
+<div class="navbar" role="navigation">
```
nickvergessen: what about navbar header? if it can only appear once, I'd put it there, since header also contains the forum parents list

Senky: navbar_header.html already includes role="navigation". It is allowed to include more navigation landmarks, so I think this should not be a problem. The only issue about this is, that footer navbar contains the same links, as header one does. So maybe only header navbar should be landmarked.

prototech:
Merged via https://github.com/phpbb/phpbb/commit/4276f49efaa290cf57936f782f26dbb0ab15ad2c

PR #3415, opened 2015-02-17
[ticket/13578] Add ucp_register.php core events

RMcGirr83:
https://tracker.phpbb.com/browse/PHPBB3-13578

### DIFF
--- a/phpBB/includes/ucp/ucp_register.php
+++ b/phpBB/includes/ucp/ucp_register.php
@@ -30,7 +30,7 @@ class ucp_register
 	function main($id, $mode)
 	{
 		global $config, $db, $user, $auth, $template, $phpbb_root_path, $phpEx;
-		global $request, $phpbb_container;
+		global $request, $phpbb_container, $phpbb_dispatcher;
 
 		//
 		if ($config['require_activation'] == USER_ACTIVATION_DISABLE ||
@@ -197,6 +197,19 @@ class ucp_register
 			'lang'				=> basename(request_var('lang', $user->lang_name)),
 			'tz'				=> request_var('tz', $timezone),
 		);
+		/**
+		* Add UCP register data before they are assigned to the template or submitted
+		*
+		* To assign data to the template, use $template->assign_vars()
+		*
+		* @event core.ucp_register_data_before
+		* @var	bool	submit		Do we display the form only
+		*							or did the user press submit
+		* @var	array	data		Array with current ucp registration data
+		* @since 3.1.4-RC1
+		*/
+		$vars = array('submit', 'data');
+		extract($phpbb_dispatcher->trigger_event('core.ucp_register_data_before', compact($vars)));
 
 		// Check and initialize some variables if needed
 		if ($submit)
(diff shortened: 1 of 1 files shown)


RMcGirr83:
!set event

RMcGirr83:
!set EVENT

RMcGirr83:
!set WIP

RMcGirr83:
!unset WIP

nickvergessen:
It's `!set Event` ;)

RMcGirr83:
!unset Event

RMcGirr83:
@nickvergessen will this commit be added?  I am writing an extension, actually two, and both need these events.

Sorry to bug you just asking.

phpBB/includes/ucp/ucp_register.php
```diff
@@ -319,6 +344,19 @@ function main($id, $mode)
...
+				*
+				* user_add allows adding more data into the users table
+				*
+				* @event core.ucp_register_user_row_after
+				* @var	bool	submit		Do we display the form only
+				*							or did the user press submit
+				* @var	array	user_row	Array with current ucp registration data
+				* @since 3.1.4-RC1
+				*/
+				$vars = array('submit', 'user_row');
```
nickvergessen: Please add `cp_data` aswell

phpBB/includes/ucp/ucp_register.php
```diff
@@ -257,6 +270,18 @@ function main($id, $mode)
...
+			* Check UCP registration data after they are submitted
+			*
+			* @event core.ucp_register_data_after
+			* @var	bool	submit		Do we display the form only
+			*							or did the user press submit
+			* @var	array data		Array with current ucp registration data
+			* @var	array error		Array with list of errors
+			* @since 3.1.4-RC1
+			*/
+			$vars = array('submit', 'data', 'error');
```
nickvergessen: Please add `cp_data` aswell

nickvergessen:
@RMcGirr83 if you add those two, I will merge it

RMcGirr83:
@nickvergessen done...waiting on Travis

RMcGirr83:
Thanks Joas

phpBB/includes/ucp/ucp_register.php
```diff
@@ -257,6 +270,19 @@ function main($id, $mode)
...
+			*
+			* @event core.ucp_register_data_after
+			* @var	bool	submit		Do we display the form only
+			*							or did the user press submit
+			* @var	array 	data		Array with current ucp registration data
+			* @var	array	cp_data		Array with custom profile fields data
+			* @var	array 	error		Array with list of errors
+			* @since 3.1.4-RC1
+			*/
+			$vars = array('submit', 'data', 'cp_data', 'error');
```
lavigor: It seems that `$submit` is always `true` here.
The same for `core.ucp_register_user_row_after`.

marc1706: This seems to be another case similar to this one: https://github.com/phpbb/phpbb/pull/3780/files
Would be nice if you could send a PR that takes care of this.

lavigor: OK, @marc1706.
Could you explain why phpBB uses `@change` and `@changed` invalid tags instead of several `@since` tags please?
Maybe there is a guideline on this case? Because there is still difference in tags usage when somebody modifies an event.
Official documentation http://www.phpdoc.org/docs/latest/references/phpdoc/tags/since.html suggests to add more than one `@since` tag.
> This tag can occur multiple times within a PHPDoc. In that case, each occurrence is treated as an entry to a change log. It is RECOMMENDED that you also provide a description to each such tag.

marc1706: I guess nobody every payed a close attention to what the correct approach for this would be. Could you do me the favor and create ticket for this on the tracker?

lavigor: OK.
https://tracker.phpbb.com/browse/PHPBB3-14092

PR #3424, opened 2015-02-19
[ticket/13632] Adding folder permission checking to database_update.php

MateBartus:
Checking if /store and /files directories are writable on the server, and 
adding error message when the conditions are not met. /cache folder is 
excluded from checking because if it is not writable, a fatal error occurs 
anyway.

PHPBB3-13632

### DIFF
--- a/phpBB/install/database_update.php
+++ b/phpBB/install/database_update.php
@@ -21,31 +21,36 @@ define('IN_INSTALL', true);
 $phpbb_root_path = (defined('PHPBB_ROOT_PATH')) ? PHPBB_ROOT_PATH : './../';
 $phpEx = substr(strrchr(__FILE__, '.'), 1);
 
+function render_footer()
+{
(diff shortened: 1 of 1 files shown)


nickvergessen:
I think we should display the error in the ACP, rather then on update, so the admin is warned anytime this happens, not only when he updated his installation.

MateBartus:
I can agree with you, however in that case the directory checks should be moved to a function, which raises the question which file would be the most appropriate place for it. My guess would be functions_install or functions_acp as it is used on installation and updates. Also I'm not sure if it is really necessary to write and remove 2+ files on every ACP dashboard load for example.

nickvergessen:
you could put it in the acp module file just like the version check is
We can then also cache the result for 24hours (just like with the version check aswell)

Anyway I'm closing this for now, since it needs a different patch anyway

PR #3426, opened 2015-02-20
[ticket/13637] Add core.ucp_pm_compose_alter_vars PHP event

kinerity:
Add a new PHP event to ucp_pm_compose to alter the default vars

PHPBB3-13637

### DIFF
--- a/phpBB/includes/ucp/ucp_pm_compose.php
+++ b/phpBB/includes/ucp/ucp_pm_compose.php
@@ -90,6 +90,32 @@ function compose_pm($id, $mode, $action, $user_folders = array())
 	// we include the language file here
 	$user->add_lang('viewtopic');
 
+	/**
+	* Modify the default vars before composing a PM
+	*
+	* @event core.ucp_pm_compose_alter_vars
+	* @var	int		msg_id					topic_id in the page request
+	* @var	int		to_user_id				The id of whom the message is to
+	* @var	int		to_group_id				The id of the group whom the message is to
+	* @var	bool	submit					Whether the user is sending the PM or not
+	* @var	bool	preview					Whether the user is previewing the PM or not
+	* @var	string	action					One of: post, reply, quote, forward, quotepost, edit, delete, smilies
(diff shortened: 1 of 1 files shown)


VSEphpbb:
"alter_vars" is weird and not consistent with our event naming. I'd suggest `core.ucp_pm_compose_modify_data`

phpBB/includes/ucp/ucp_pm_compose.php
```diff
@@ -90,6 +90,32 @@ function compose_pm($id, $mode, $action, $user_folders = array())
 	// we include the language file here
 	$user->add_lang('viewtopic');
 
+	/**
+	* Modify the default vars before composing a PM
+	*
+	* @event core.ucp_pm_compose_alter_vars
+	* @var	int		msg_id					topic_id in the page request
+	* @var	int		to_user_id				The id of whom the message is to
+	* @var	int		to_group_id				The id of the group whom the message is to
```
marc1706: whom shouldn't be needed in this case

phpBB/includes/ucp/ucp_pm_compose.php
```diff
@@ -90,6 +90,32 @@ function compose_pm($id, $mode, $action, $user_folders = array())
...
+	* @var	bool	submit					Whether the user is sending the PM or not
```
marc1706: Whether the form has been submitted

phpBB/includes/ucp/ucp_pm_compose.php
```diff
@@ -90,6 +90,32 @@ function compose_pm($id, $mode, $action, $user_folders = array())
 	// we include the language file here
 	$user->add_lang('viewtopic');
 
+	/**
+	* Modify the default vars before composing a PM
+	*
+	* @event core.ucp_pm_compose_modify_data
+	* @var	int		msg_id					topic_id in the page request
```
marc1706: It's actually the post_id and not the topic_id.

PR #3429, opened 2015-02-20
[ticket/13460] Distorted avatar shown to Dolphin users.

RushanGajanayake:
Added line #411 for fixed the max-width of the avatar image.

PHPBB3-13460


Crizz0:
Seems to be browser-bug, he has to keep the aspect ratio. All others i have (opera, chrome, firefox) have no problem. 

The new CSS property will cause images, which are wider than high, to look really ugly, because the aspect ration will be broken.

RushanGajanayake:
Is there anything else to do for solve this??

Crizz0:
I know nothing from the style-side. Thumb-Nails from this avatar-images would not lead to a solution, because it will setup a fix value for any style. And maybe some styles like to change the size...

PR #3433, opened 2015-02-22
[ticket/13643] Change the priority of kernel_terminate_subscriber to -10000

Nicofuma:
https://tracker.phpbb.com/browse/PHPBB3-13643

PHPBB3-13643

### DIFF
--- a/phpBB/phpbb/event/kernel_terminate_subscriber.php
+++ b/phpBB/phpbb/event/kernel_terminate_subscriber.php
@@ -35,7 +35,7 @@ class kernel_terminate_subscriber implements EventSubscriberInterface
 	public static function getSubscribedEvents()
 	{
 		return array(
-			KernelEvents::TERMINATE		=> 'on_kernel_terminate',
+			KernelEvents::TERMINATE		=> array('on_kernel_terminate', ~PHP_INT_MAX),
 		);
 	}
 }


phpBB/phpbb/event/kernel_terminate_subscriber.php
```diff
@@ -35,7 +35,7 @@ public function on_kernel_terminate(PostResponseEvent $event)
 	public static function getSubscribedEvents()
 	{
 		return array(
-			KernelEvents::TERMINATE		=> 'on_kernel_terminate',
+			KernelEvents::TERMINATE		=> array('on_kernel_terminate', -10000),
```
bantu: There is no case for not running this last, right? So make this something like `-PHP_INT_MAX` or `~PHP_INT_MAX` ?

bantu: There doesn't seem to be any tests for passing an array.

Nicofuma: updated to ~PHP_INT_MAX

Nicofuma: about the tests, you mean for `array('on_kernel_terminate', -10000)` ? It's part of the SYmfony api.

PR #3436, opened 2015-02-24
[ticket/13650] Validate ucp profile data changes

RMcGirr83:
You can change data on profile but can not generate an event to check input (eg, run it through the validate phpBB function) on that change.

PHPBB3-13650


RMcGirr83:
!set Event

PR #3437, opened 2015-02-24
[ticket/12368] Correctly check if avatars migration is installed

marc1706:
Try to set the avatar type to string value to check if new avatar
types are supported.

Ticket: https://tracker.phpbb.com/browse/PHPBB3-12368

### DIFF
--- a/phpBB/phpbb/db/migration/data/v310/avatars.php
+++ b/phpBB/phpbb/db/migration/data/v310/avatars.php
@@ -17,7 +17,29 @@ class avatars extends \phpbb\db\migration\migration
 {
 	public function effectively_installed()
 	{
-		return isset($this->config['allow_avatar_gravatar']);
+		// Get current avatar type of guest user
+		$sql = 'SELECT user_avatar_type
+			FROM ' . $this->table_prefix . 'users
+			WHERE user_id = ' . ANONYMOUS;
+		$result = $this->db->sql_query($sql);
+		$backup_type = $this->db->sql_fetchfield('user_avatar_type');
+		$this->db->sql_freeresult($result);
+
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/db/migration/data/v310/avatars.php
```diff
@@ -17,7 +17,29 @@ class avatars extends \phpbb\db\migration\migration
...
+
+		// Try to set avatar type to string
+		$sql = 'UPDATE ' . $this->table_prefix . "users
+			SET user_avatar_type = 'avatar.driver.upload'
+			WHERE user_id = " . ANONYMOUS;
+		$this->db->sql_return_on_error(true);
+		$effectively_installed = $this->db->sql_query($sql);
+		$this->db->sql_return_on_error();
+
+		// Return avatar type of guest user to previous state
```
Nicofuma: Restore*

Nicofuma:
As said on the ticket, I'd prefer a more general solution: 
> More preferably we set the cache driver to null on any CLI process / web request that uses the migrator, and then manually run a purge on the cache driver normally used afterwards, as our cache purge calls will otherwise become ineffective on the null cache.

marc1706:
This should be out of the scope of this issue though. Right now, we only check if the config exists. However, the gravatar mod for phpBB 3.0 already added that config entry. This will effectively prevent the migration from being installed just because of something a MOD already added. Using this effectively installed method makes sure that we don't run into this issue.l

Nicofuma:
well... you're right so it should be good like that for now but fix the comment please ^^

marc1706:
@Nicofuma done

PR #3440, opened 2015-02-24
[ticket/13636] Deleting inactive user in acp_users gives unexpected link

CHItA:
Success message did not provide the expected "previous page" link
(acp_inactive was expected, but the link took users to acp_users) if the
acp_users module was accessed by navigating from the inactive user list to
acp_users and then deleting inactive user.

PHPBB3-13636

### DIFF
--- a/phpBB/includes/acp/acp_users.php
+++ b/phpBB/includes/acp/acp_users.php
@@ -151,9 +151,9 @@ class acp_users
 		}
 
 		$template->assign_vars(array(
-			'U_BACK'			=> ( (empty($redirect)) ? $this->u_action : $redirect_url ),
+			'U_BACK'			=> (empty($redirect)) ? $this->u_action : $redirect_url,
 			'U_MODE_SELECT'		=> append_sid("{$phpbb_admin_path}index.$phpEx", "i=$id&amp;u=$user_id"),
-			'U_ACTION'			=> $this->u_action . '&amp;u=' . $user_id . ( (empty($redirect)) ? '' : '&amp;' . $redirect_tag ),
+			'U_ACTION'			=> $this->u_action . '&amp;u=' . $user_id . ((empty($redirect)) ? '' : '&amp;' . $redirect_tag),
 			'S_FORM_OPTIONS'	=> $s_form_options,
 			'MANAGED_USERNAME'	=> $user_row['username'])
 		);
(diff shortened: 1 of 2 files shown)


phpBB/includes/acp/acp_users.php
```diff
@@ -44,6 +44,10 @@ function main($id, $mode)
 		$user_id	= request_var('u', 0);
 		$action		= request_var('action', '');
 
+		$referer		= request_var('referer', '');
```
nickvergessen: we use mostly `redirect` for this purpose

MateBartus: I can rename it if you prefer that, I used referer because this only effects delete action (the back link on the success message more precisely and the back link in the top right corner), and not actually redirecting via headers.

phpBB/includes/acp/acp_users.php
```diff
@@ -221,19 +226,30 @@ function main($id, $mode)
...
-								confirm_box(false, $user->lang['CONFIRM_OPERATION'], build_hidden_fields(array(
+								$delete_confirm_hidden_fields = array(
 									'u'				=> $user_id,
 									'i'				=> $id,
 									'mode'			=> $mode,
 									'action'		=> $action,
 									'update'		=> true,
 									'delete'		=> 1,
-									'delete_type'	=> $delete_type))
+									'delete_type'	=> $delete_type
```
nickvergessen: Please add trailing `,`

phpBB/includes/acp/acp_users.php
```diff
@@ -146,9 +151,9 @@ function main($id, $mode)
 		}
 
 		$template->assign_vars(array(
-			'U_BACK'			=> $this->u_action,
+			'U_BACK'			=> ( (empty($redirect)) ? $this->u_action : $redirect_url ),
```
nickvergessen: remove most outer brackets

phpBB/includes/acp/acp_users.php
```diff
@@ -146,9 +151,9 @@ function main($id, $mode)
...
 			'U_MODE_SELECT'		=> append_sid("{$phpbb_admin_path}index.$phpEx", "i=$id&amp;u=$user_id"),
-			'U_ACTION'			=> $this->u_action . '&amp;u=' . $user_id,
+			'U_ACTION'			=> $this->u_action . '&amp;u=' . $user_id . ( (empty($redirect)) ? '' : '&amp;' . $redirect_tag ),
```
nickvergessen: Remove space between `( (`

marc1706:
It's kind of a hackish approach but I don't really see a better way of handling this?

CHItA:
Probably when we move the ACP into controllers, we could find a better (and general) solution for this kinds of things, however i think ATM that is the most amount of effort that we should put into solving this.

PR #3441, opened 2015-02-25
[feature/sql-bool-builder] Extending DBAL query builder for boolean generation

brunoais:
Improve DBAL's query builder by adding a boolean query builder using a structured stack of arrays.

This can be used everywhere in an SQL query where boolean operations are used.
Currently, this can only be used in the WHERE and the JOINs' ON clause.

The main and general use-cases are:
To flexibly the build of the WHERE clause in SQL queries
To ease, simplify and prevent errors when doing SQL query editing by phpBB's extensions

Topic: https://area51.phpbb.com/phpBB/viewtopic.php?f=81&t=47306
https://tracker.phpbb.com/browse/PHPBB3-13652

Documentation: https://github.com/phpbb/documentation/pull/51

PHPBB3-13652

### DIFF
--- a/phpBB/phpbb/db/driver/driver.php
+++ b/phpBB/phpbb/db/driver/driver.php
@@ -807,6 +807,8 @@ abstract class driver implements driver_interface
 
 	protected function _process_boolean_tree_first($operations_ary)
 	{
+		// In cases where an array exists but there is no head condition,
+		// it should be because there's only 1 WHERE clause. This seems the best way to deal with it.
 		if ($operations_ary[0] !== 'AND' &&
 			$operations_ary[0] !== 'OR')
 		{
--- a/phpBB/phpbb/db/driver/driver.php
+++ b/phpBB/phpbb/db/driver/driver.php
@@ -819,7 +819,7 @@ abstract class driver implements driver_interface
 	{
 		$operation = array_shift($operations_ary);
 
-		foreach ($operations_ary AS &$condition)
+		foreach ($operations_ary as &$condition)
 		{
 			switch ($condition[0])
 			{
--- a/phpBB/phpbb/db/driver/driver.php
+++ b/phpBB/phpbb/db/driver/driver.php
@@ -866,6 +866,24 @@ abstract class driver implements driver_interface
 
 								break;
 
+								case 'IS_NOT':
+
+									$condition[1] = 'IS NOT';
+
+								// no break
+								case 'IS':
+
+									// If the value is NULL, the string of it is the empty string ('') which is not the intended result.
+									// this should solve that
+									if ($condition[2] === null)
+									{
+										$condition[2] = 'NULL';
+									}
+
+									$condition = implode(' ', $condition);
+
+								break;
+
 								default:
 
 									$condition = implode(' ', $condition);
--- a/phpBB/phpbb/db/driver/driver.php
+++ b/phpBB/phpbb/db/driver/driver.php
@@ -831,7 +831,7 @@ abstract class driver implements driver_interface
 				break;
 				case 'NOT':
 
-					$condition = ' NOT ' . $this->_process_boolean_tree($condition);
+					$condition = ' NOT (' . $this->_process_boolean_tree($condition) . ') ';
 
 				break;
 
--- a/phpBB/phpbb/db/driver/driver.php
+++ b/phpBB/phpbb/db/driver/driver.php
@@ -792,11 +792,11 @@ abstract class driver implements driver_interface
 
 		return $sql;
 	}
-	
-	
+
+
 	protected function _process_boolean_tree_first($operations_ary)
 	{
-		if ($operations_ary[0] !== 'AND' && 
+		if ($operations_ary[0] !== 'AND' &&
 			$operations_ary[0] !== 'OR')
 		{
 			$operations_ary = array('AND', $operations_ary);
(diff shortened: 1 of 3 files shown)


brunoais:
!set WIP

brunoais:
!set WIP

phpBB/phpbb/db/driver/driver.php
```diff
@@ -793,6 +804,98 @@ function sql_build_query($query, $array)
...
+			{
+				case 'AND':
+				case 'OR':
+
+					$condition = ' ( ' . $this->_process_boolean_tree($condition) . ') ';
+
+				break;
+				case 'NOT':
+
+					$condition = ' NOT ' . $this->_process_boolean_tree($condition);
```
Nicofuma: I think you should add parentheses around the sub expression

brunoais: No need to add two layers of parenthesis ;).

phpBB/phpbb/db/driver/driver.php
```diff
@@ -793,6 +804,98 @@ function sql_build_query($query, $array)
 		return $sql;
 	}
 
+
+	protected function _process_boolean_tree_first($operations_ary)
+	{
+		if ($operations_ary[0] !== 'AND' &&
+			$operations_ary[0] !== 'OR')
+		{
+			$operations_ary = array('AND', $operations_ary);
```
nickvergessen: throw an error, if it something that is unknown
missing NOT from the allowed list.

phpBB/phpbb/db/driver/driver.php
nickvergessen: safty first, just add them, they dont hurt

phpBB/phpbb/db/driver/driver.php
nickvergessen: IN, NOT_IN?

brunoais: That's not what this is for. This is for a specific case where the user only specifies a single array with the single element for the WHERE clause. I didn't find a cleaner way of dealing with it, unfortunately...

I'll comment that.

phpBB/phpbb/db/driver/driver.php
brunoais: OK...

brunoais:
!unset WIP

brunoais:
Is there anything preventing this from being merged?

phpBB/phpbb/db/driver/driver.php
```diff
@@ -793,6 +804,130 @@ function sql_build_query($query, $array)
 		return $sql;
 	}
 
+
+	protected function _process_boolean_tree_first($operations_ary)
+	{
+		// In cases where an array exists but there is no head condition,
+		// it should be because there's only 1 WHERE clause. This seems the best way to deal with it.
+		if ($operations_ary[0] !== 'AND' &&
+			$operations_ary[0] !== 'OR')
```
Nicofuma: it could be a NOT

Nicofuma:
About the tests, instead of doing something complicated with a database test couldn't you make a simple unit test and compare the input array with the expected WHERE clause?

phpBB/phpbb/db/driver/driver.php
brunoais: Although it could. Part of the idea is to allow creating other connecting logic later if it makes sense.
Regardless, if you insist, I'll change it. I don't have any pressuring will to keep it like that. It was just an idea.

brunoais:
Please do proceed the explanation.
I know how to make tests to test something fully but I don't know how to test at best equilibrium between coverage and speed.

Nicofuma:
about the tests I had something like this in mind: (don't test the result of the SQL query but the result of the query generation)
```
class phpbb_boolean_processor_test extends phpbb_test_case
{
	private $db;

	public function setUp()
	{
		parent::setUp();

		$this->db = new \phpbb\db\driver\mysql();
	}

	public function data_where_builder()
	{
		return array(
			// BC
			array(
				'SELECT bar FROM (foo f) WHERE bar = 1',
				array(
					'SELECT' => 'bar',
					'FROM' => array('foo' => 'f'),
					'WHERE' => 'bar = 1',
				),
			),
			// single not_like
			array(
				"SELECT bar FROM (foo f) WHERE bar NOT LIKE 'gr%'",
				array(
					'SELECT' => 'bar',
					'FROM' => array('foo' => 'f'),
					'WHERE' => array('bar', 'NOT_LIKE', 'gr%'),
				),
			),
		);
	}

	/**
	 * @dataProvider data_where_builder
	 */
	public function test_where_builder($expected, $data)
	{
		$sql = $this->db->sql_build_query('SELECT', $data);
		self::assertEquals($expected, $sql);
	}
}
```

Also, I'm not very sure about the PR: shouldn't we use doctrine dbal and its query builder instead of rewriting everything ourselves?

brunoais:
> Also, I'm not very sure about the PR: shouldn't we use doctrine dbal and its query builder instead of rewriting everything ourselves?

I thought we had already discussed that O_O.
https://area51.phpbb.com/phpBB/viewtopic.php?p=275796#p275796

Don't forget to check the IRC logs of that time. There was some discussion (I believe, with you in it) about this subject and, I believe to be nickvergessen stating that both are not incompatible, they can actually complement each other.

brunoais:
@Nicofuma 
 > Anyway, the reason why I'm testing against the DB is just that the output syntax is not important, what is important is that it is valid SQL and has the same meaning as the one inputted
Nicofuma, does it still make sense to do the tests like that?

You still didn't answer

Nicofuma:
Well I don't know... the good thing with my proposal is that it is easy to have exhaustive tests But you are also right so... lets keep them like that I guess

phpBB/phpbb/db/driver/driver.php
```diff
@@ -793,6 +804,130 @@ function sql_build_query($query, $array)
...
+			$operations_ary = array('AND', $operations_ary);
+		}
+		return $this->_process_boolean_tree($operations_ary) . "\n";
+	}
+
+	protected function _process_boolean_tree($operations_ary)
+	{
+		$operation = array_shift($operations_ary);
+
+		foreach ($operations_ary as &$condition)
```
Nicofuma: Do you really need to use references here?

brunoais: If I want to keep the algorithm, yes.
If I want to keep the final result even if I use a different algorithm, no.
Regardless, If I don't use a reference, I'll have to change the algorithm significantly. How much exactly... I'm unsure.

phpBB/phpbb/db/driver/driver.php
```diff
@@ -793,6 +804,130 @@ function sql_build_query($query, $array)
...
+								default:
+
+									$condition = implode(' ', $condition);
+
+								break;
+							}
+
+						break;
+
+						case 5:
```
Nicofuma: write a test for this case please

Nicofuma:
Could you also draft a documentation (https://github.com/phpbb/documentation/tree/master/development) explaining the format and every case in it please?

phpBB/phpbb/db/driver/driver.php
```diff
@@ -793,6 +804,130 @@ function sql_build_query($query, $array)
...
+			$operations_ary[0] !== 'OR')
+		{
+			$operations_ary = array('AND', $operations_ary);
+		}
+		return $this->_process_boolean_tree($operations_ary) . "\n";
+	}
+
+	protected function _process_boolean_tree($operations_ary)
+	{
+		$operation = array_shift($operations_ary);
```
Nicofuma: couldn't you use arrays with a fixed length?

ex:
```
        'WHERE' => [
                'AND',
                ['t.topic_poster', '=', 1],
                ['f.forum_id', '>=', 'ALL', 'SELECT', [
                        'SELECT' => ['t.forum_id'],
                        'FROM' => [TOPICS_TABLE  => 't'],
                        'WHERE' => ['t.topic_poster', '=', 1],
                ],
        ],
```

Nicofuma:
Couldn't we have and use some constant like doctrine?
(ex: TYPE_AND, TYPE_OR, EQ, NEQ, LT, LTE, GT, GTE...)

I'm not sure but I think it could be useful.

Nicofuma:
doc: https://github.com/phpbb/documentation/pull/51

Nicofuma:
Thanks you for your contribution, but please if possible fix my two comments in another PR :)

phpBB/phpbb/db/driver/driver.php
brunoais: To do which work?
I think you are mistaking something here
That code extracts the `'AND'`, the  `'OR'` or the `'NOT'` from the array. It is supposed to be the first element of the array.

phpBB/phpbb/db/driver/driver.php
brunoais: Ups. Good catch. I'll do it ASAV (as soon as viable). Probably the earliest is this weekend...

PR #3444, opened 2015-02-25
[ticket/13617] Enforce column size limit for session_forum_id 

naderman:
https://tracker.phpbb.com/browse/PHPBB3-13617

Ascraeus version of https://github.com/phpbb/phpbb/pull/3443


phpBB/phpbb/session.php
```diff
@@ -130,6 +130,9 @@ static function extract_current_page($root_path)
 		$script_path .= (substr($script_path, -1, 1) == '/') ? '' : '/';
 		$root_script_path .= (substr($root_script_path, -1, 1) == '/') ? '' : '/';
 
+		$forum_id = request_var('f', 0);
```
MGaetan89: `$request->variable()`?

phpBB/phpbb/session.php
```diff
@@ -130,6 +130,9 @@ static function extract_current_page($root_path)
...
+		$forum_id = ($forum_id > 0 && $forum_id < 16777215) ? $forum_id : 0;
```
MGaetan89: Comment about `16777215`?

phpBB/phpbb/session.php
naderman: Indeed could change this, just kept request_var cause that was used before.

nickvergessen:
Should we have a function for this, maybe even on the request class?

marc1706:
I guess this is out of the scope of this PR though.

naderman:
Indeed, might be useful but let's open up a new ticket for that, and also to review if there are more cases like this.

marc1706:
For some reason the merge doesn't fast forward when merging the 3.0.x branch into the 3.1.x branch after having merged the different PRs for 3.0 and 3.1.

PR #3453, opened 2015-02-27
[ticket/13477] Update file cache driver for extension paths

RMcGirr83:
PHPBB3-13477

### DIFF
--- a/phpBB/phpbb/cache/driver/file.php
+++ b/phpBB/phpbb/cache/driver/file.php
@@ -279,6 +279,7 @@ class file extends \phpbb\cache\driver\base
 		if ($var_name[0] == '_')
 		{
 			global $phpEx;
+			$var_name = $this->replace_slashes($var_name);
 			return file_exists($this->cache_dir . 'data' . $var_name . ".$phpEx");
 		}
 		else
@@ -334,6 +335,7 @@ class file extends \phpbb\cache\driver\base
 	{
 		global $phpEx;
 
+		$filename = $this->replace_slashes($filename);
 		$file = "{$this->cache_dir}$filename.$phpEx";
 
 		$type = substr($filename, 0, strpos($filename, '_'));
@@ -516,6 +518,7 @@ class file extends \phpbb\cache\driver\base
 	{
 		global $phpEx;
 
+		$filename = $this->replace_slashes($filename);
 		$file = "{$this->cache_dir}$filename.$phpEx";
 
 		$lock = new \phpbb\lock\flock($file);
@@ -583,5 +586,16 @@ class file extends \phpbb\cache\driver\base
 		$lock->release();
 
 		return $return_value;
+	}
+
+	/**
+	* Replace slashes in the file name
+	*
+	* @param string $name filename
+	* @return string the corrected name
+	*/
+	function replace_slashes($name)
+	{
+		return str_replace('/', '-', $name);
 	}
 }
--- a/phpBB/phpbb/cache/driver/file.php
+++ b/phpBB/phpbb/cache/driver/file.php
@@ -279,7 +279,7 @@ class file extends \phpbb\cache\driver\base
 		if ($var_name[0] == '_')
 		{
 			global $phpEx;
-			$var_name = $this->replace_slashes($var_name);
+			$var_name = $this->clean_varname($var_name);
 			return file_exists($this->cache_dir . 'data' . $var_name . ".$phpEx");
 		}
 		else
@@ -335,7 +335,7 @@ class file extends \phpbb\cache\driver\base
 	{
 		global $phpEx;
 
-		$filename = $this->replace_slashes($filename);
+		$filename = $this->clean_varname($filename);
 		$file = "{$this->cache_dir}$filename.$phpEx";
 
 		$type = substr($filename, 0, strpos($filename, '_'));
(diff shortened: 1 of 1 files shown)


RMcGirr83:
https://github.com/phpbb/phpbb/pull/3238

Not sure on how to create tests.

phpBB/phpbb/cache/driver/file.php
```diff
@@ -279,6 +279,7 @@ function _exists($var_name)
 		if ($var_name[0] == '_')
 		{
 			global $phpEx;
+			$var_name = str_replace('/', '-', $var_name);
```
MGaetan89: Surround by blank lines for readability?

bantu:
We should probably make that a function/method instead of repeating this statement.

RMcGirr83:
Travis barfed on the commit but it doesn't appear to have anything to do with the actual commit?

phpBB/phpbb/cache/driver/file.php
```diff
@@ -584,4 +587,21 @@ function _write($filename, $data = null, $expires = 0, $query = '')
 
 		return $return_value;
 	}
+
+	/**
+	*
```
nickvergessen: Remove the space at the end of the line

phpBB/phpbb/cache/driver/file.php
```diff
@@ -584,4 +587,21 @@ function _write($filename, $data = null, $expires = 0, $query = '')
 
 		return $return_value;
 	}
+
+	/**
+	* 
+	* Change the cache path name for extensions
+	*
+	* {@inheritDoc}
```
nickvergessen: remove this line

phpBB/phpbb/cache/driver/file.php
nickvergessen: Actually remove this line aswell

phpBB/phpbb/cache/driver/file.php
```diff
@@ -584,4 +587,21 @@ function _write($filename, $data = null, $expires = 0, $query = '')
...
+	* @param string $name extension filename
+	* @return string the corrected path of extension
+	*/
+	function extension_path($name)
```
nickvergessen: This has nothing to do with extensions, its a general issue.
Maybe `get_file_name()` ?

phpBB/phpbb/cache/driver/file.php
```diff
@@ -584,4 +587,19 @@ function _write($filename, $data = null, $expires = 0, $query = '')
...
 		return $return_value;
 	}
+
+	/**
+	* Change the cache path name
+	*
+	* @param string $name filename
+	* @return string the corrected path
+	*/
+	function get_file_name($name)
```
marc1706: Please use a more descriptive name like like `replace_slashes()`.

phpBB/phpbb/cache/driver/file.php
```diff
@@ -584,4 +587,19 @@ function _write($filename, $data = null, $expires = 0, $query = '')
...
+	{
+		// remove slashes and replace with -
+		// as we can't write to a non-existent directory
+		$name = str_replace('/', '-', $name);
+
+		return $name;
```
marc1706: Just do `return str_replace('/', '-', $name);` instead of re-assigning the str_replace output above.

phpBB/phpbb/cache/driver/file.php
nickvergessen: The method should say what it is for, not what it does :P

marc1706: Well it's not for retrieving the filename but rather for replacing the slashes in the file name?

RMcGirr83:
Guess I was trying to make it more difficult than it actually was.

:offtopic:

When making PR's and needing to update the PR is it preferred to amend or make a new commit based off the original ticket/PR?

bantu:
> When making PR's and needing to update the PR is it preferred to amend or make a new commit based off the original ticket/PR?

It is basically up to the PR author, but for example 80 commits for changing a single line does not really make sense. Also, each commit should have a different/descriptive commit message if you want to go with multiple commits.

bantu:
Restarted the errored job.

phpBB/phpbb/cache/driver/file.php
```diff
@@ -584,4 +587,15 @@ function _write($filename, $data = null, $expires = 0, $query = '')
...
 		return $return_value;
 	}
+
+	/**
+	* Replace slashes in the file name
+	*
+	* @param string $name filename
+	* @return string the corrected name
+	*/
+	function replace_slashes($name)
```
bantu: We should have the "protected" keyword before "function" here since this function is only used internally in the file class.

bantu: Following the previous discussion in https://github.com/phpbb/phpbb/pull/3453#discussion_r25686975, this function should probably called "clean_varname($name)".

phpBB/phpbb/cache/driver/file.php
```diff
@@ -584,4 +587,15 @@ function _write($filename, $data = null, $expires = 0, $query = '')
 
 		return $return_value;
 	}
+
+	/**
+	* Replace slashes in the file name
+	*
+	* @param string $name filename
```
bantu: It takes the name of a cache variable ...

phpBB/phpbb/cache/driver/file.php
```diff
@@ -584,4 +587,15 @@ function _write($filename, $data = null, $expires = 0, $query = '')
...
+	* @return string the corrected name
```
bantu: ... and returns a name that is safe to use as a filename.

PR #3454, opened 2015-02-27
[ticket/13590] Headers of remember me login keys are not centered

RMcGirr83:
PHPBB3-13590

### DIFF
--- a/phpBB/styles/prosilver/template/mcp_logs.html
+++ b/phpBB/styles/prosilver/template/mcp_logs.html
@@ -22,10 +22,10 @@
 	<table class="table1">
 	<thead>
 	<tr>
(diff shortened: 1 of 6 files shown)


bantu:
I am not sure, but I think the general idea to not have CSS in HTML.

marc1706:
Yeah, definitely shouldn't be using the style HTML attribute.

RMcGirr83:
Suggest a class.  I was told once that tables shouldn't be used in prosilver?

bantu:
> I was told once that tables shouldn't be used in prosilver?

I think it is "do not use tables for styling", e.g. splitting the page layout into halfes. It is fine to use tables for actual tables.

PR #3457, opened 2015-03-01
[ticket/13661] Add event to allow editing the queries used to get the logs

brunoais:
https://tracker.phpbb.com/browse/PHPBB3-13661

### DIFF
--- a/phpBB/phpbb/log/log.php
+++ b/phpBB/phpbb/log/log.php
@@ -563,7 +563,7 @@ class log implements \phpbb\log\log_interface
 		*									is false, no entries will be returned.
 		* @var	string	sql_additional		Additional conditions for the entries,
 		*									e.g.: 'AND l.forum_id = 1'
-		* @since 3.1.0-a1
+		* @since 3.1.4-RC1
 		*/
 		$vars = array(
 			'get_logs_sql_ary',
--- a/phpBB/phpbb/log/log.php
+++ b/phpBB/phpbb/log/log.php
@@ -535,7 +535,7 @@ class log implements \phpbb\log\log_interface
 			'ORDER_BY' => $sort_by,
 		);
 
-		if($log_type){
+		if($log_time){
 			$get_logs_sql_ary['WHERE'] = 'l.log_time >= ' . (int) $log_time . '
 					AND ' . $get_logs_sql_ary['WHERE'];
 		}
--- a/phpBB/phpbb/log/log.php
+++ b/phpBB/phpbb/log/log.php
@@ -535,7 +535,8 @@ class log implements \phpbb\log\log_interface
 			'ORDER_BY' => $sort_by,
 		);
 
-		if($log_time){
+		if($log_time)
+		{
 			$get_logs_sql_ary['WHERE'] = 'l.log_time >= ' . (int) $log_time . '
 					AND ' . $get_logs_sql_ary['WHERE'];
 		}
--- a/phpBB/phpbb/log/log.php
+++ b/phpBB/phpbb/log/log.php
@@ -537,6 +537,49 @@ class log implements \phpbb\log\log_interface
 					AND ' . $get_logs_sql_ary['WHERE'];
 		}
 
+		/**
+		* Modify the query to obtain the logs data
+		*
+		* @event core.get_logs_main_query_before
+		* @var	array	get_logs_sql_ary	The array in the format of the query builder with the query
+		*									to get the log count and the log list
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/log/log.php
```diff
@@ -521,15 +521,73 @@ public function get_logs($mode, $count_logs = true, $limit = 0, $offset = 0, $fo
...
+
+		if($log_type){
+			$get_logs_sql_ary['WHERE'] = 'l.log_type = ' . (int) $log_type . '
+					AND ' . $get_logs_sql_ary['WHERE'];
+		}
+
+		/**
+		* Modify the query to obtain the logs data
+		*
+		* @event core.get_logs_main_query_before
```
nickvergessen: This event is duplicated above.
remove the one above please

brunoais: I think it's better to deprecate it first. There might be extensions already using and relying on it.
Additionally, there are 2 different operations that can no longer be properly done with this new event, if you remove the previous one:
- Change $log_type or even make a new valid $log_type.
- Change the $keywords list.

Instead, what I can do is to change how that code is organized so that those functionality is not lost.
Regardless, though. I believe the one being removed should be deprecated first.

nickvergessen: Ah didn't see the log_type validation there :/ hmm yeah I guess then 2 events are required.
Then its fine I guess we need two events then

brunoais: 2 events it is.

nickvergessen:
You need to adjust soem tests for this

brunoais:
That's unexpected. It's not supposed to affect the method's tests. I have a hard time believing I need to adjust tests and not that I made a mistake.

phpBB/phpbb/log/log.php
```diff
@@ -521,15 +521,76 @@ public function get_logs($mode, $count_logs = true, $limit = 0, $offset = 0, $fo
...
-			$sql = 'SELECT COUNT(l.log_id) AS total_entries
-				FROM ' . $this->log_table . ' l, ' . USERS_TABLE . ' u
-				WHERE l.log_type = ' . (int) $log_type . '
+		$get_logs_sql_ary = array(
+			'SELECT' => 'l.*, u.username, u.username_clean, u.user_colour',
+			'FROM' => array(
+						$this->log_table => 'l',
+						USERS_TABLE => 'u',
+					),
+			'WHERE' => 'l.log_type = ' . (int) $log_type . "
```
nickvergessen: Single quote at the end? your syntax is now broken

brunoais: No. The syntax is correct. There's no syntax error.

phpBB/phpbb/log/log.php
```diff
@@ -521,15 +521,76 @@ public function get_logs($mode, $count_logs = true, $limit = 0, $offset = 0, $fo
...
 					AND l.user_id = u.user_id
-					AND l.log_time >= ' . (int) $log_time . "
 					$sql_keywords
-					$sql_additional";
+					$sql_additional",
+
+			'ORDER_BY' => $sort_by,
+		);
+
+		if($log_type){
```
gooof: not detected by codesniffer?

brunoais: I thought I had it in $log_time Oo?

MGaetan89: I think he was talking about code style. It should be: 

```php
if ($log_time)
{
```

nickvergessen: @Nicofuma

marc1706:
Remove the [code style] from your last commit msg.

phpBB/phpbb/log/log.php
```diff
@@ -521,15 +521,77 @@ public function get_logs($mode, $count_logs = true, $limit = 0, $offset = 0, $fo
...
+		* @var	int		log_time			Limit maximum age of log entries
+		* @var	string	sort_by				SQL order option
+		* @var	string	keywords			Will only return entries that have the
+		*									keywords in log_operation or log_data
+		* @var	string	profile_url			URL to the users profile
+		* @var	int		log_type			Limit logs to a certain type. If log_type
+		*									is false, no entries will be returned.
+		* @var	string	sql_additional		Additional conditions for the entries,
+		*									e.g.: 'AND l.forum_id = 1'
+		* @since 3.1.4-RC1
```
marc1706: 3.1.5-RC1

PR #3458, opened 2015-03-01
[ticket/8494] Ignore errors from the database schema preparation

nickvergessen:
This allows installing phpBB twice in the same database with a different
prefix

https://tracker.phpbb.com/browse/PHPBB3-8494

### DIFF
--- a/phpBB/install/install_install.php
+++ b/phpBB/install/install_install.php
@@ -1164,12 +1164,10 @@ class install_install extends module
 
 			foreach ($sql_query as $sql)
 			{
-				//$sql = trim(str_replace('|', ';', $sql));
-				if (!$db->sql_query($sql))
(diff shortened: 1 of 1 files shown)


bantu:
Not sure silently ignoring errors is such a good idea.

nickvergessen:
This only applies to https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/install/schemas/postgres_schema.sql and https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/install/schemas/oracle_schema.sql btw (this is why we can fix it in 3.1 and not 3.0)

Not sure if something can go wrong with those functions any day

Nicofuma:
it's fine with me. 
@bantu ?

PR #1646, opened 2013-08-08
[3.2][ticket/11768] Integrate s9e\TextFormatter

s9e:
http://tracker.phpbb.com/browse/PHPBB3-11768
http://area51.phpbb.com/phpBB/viewtopic.php?f=108&t=44467

PR #3382, opened 2015-02-08
[ticket/8672] Add class for retrieving imagesize without download

marc1706:
getimagesize() always downloads the complete file before checking
the actual image dimensions. This class will be able to do the same
without having to download possibly large files.

Ticket: https://tracker.phpbb.com/browse/PHPBB3-8672

PR #3461, opened 2015-03-02
[3.2][ticket/11768] Integrate s9e\TextFormatter

s9e:
http://tracker.phpbb.com/browse/PHPBB3-11768
http://area51.phpbb.com/phpBB/viewtopic.php?f=108&t=44467
~~https://github.com/phpbb/phpbb/pull/1646~~


Nicofuma:
thanks

phpBB/phpbb/textformatter/cache.php
```diff
@@ -0,0 +1,37 @@
...
+*
+* For full copyright and license information, please see
+* the docs/CREDITS.txt file.
+*
+*/
+
+namespace phpbb\textformatter;
+
+/**
+* text_formatter.cache service
```
Nicofuma: don't mention the service in the doc block

phpBB/phpbb/textformatter/cache.php
```diff
@@ -0,0 +1,37 @@
...
+*
+* Currently only used to signal that something that could effect the rendering has changed.
+* BBCodes, smilies, censored words, templates, etc...
+*
+* @todo functionality should be moved to data_access
```
Nicofuma: todo?

phpBB/phpbb/textformatter/cache.php
```diff
@@ -0,0 +1,37 @@
...
+*
+* @package phpBB3
```
Nicofuma: we don't use @package

phpBB/phpbb/textformatter/data_access.php
```diff
@@ -0,0 +1,233 @@
...
+
+/**
+* text_formatter.data_access service
+*
+* Data access layer that fetchs BBCodes, smilies and censored words from the database.
+* To be extended to include insert/update/delete operations.
+*
+* Also used to get templates.
+*
+* @package phpBB3
```
Nicofuma: we don't use @package

phpBB/phpbb/textformatter/data_access.php
```diff
@@ -0,0 +1,233 @@
...
+*
+* For full copyright and license information, please see
+* the docs/CREDITS.txt file.
+*
+*/
+
+namespace phpbb\textformatter;
+
+/**
+* text_formatter.data_access service
```
Nicofuma: don't mention the service

phpBB/phpbb/textformatter/data_access.php
```diff
@@ -0,0 +1,233 @@
...
+
+		return false;
+	}
+
+	/**
+	* Return the list of censored words
+	*
+	* @return array
+	*/
+	public function get_words()
```
Nicofuma: use a more precise function name

phpBB/phpbb/textformatter/parser.php
```diff
@@ -0,0 +1,121 @@
...
+* For full copyright and license information, please see
+* the docs/CREDITS.txt file.
+*
+*/
+
+namespace phpbb\textformatter;
+
+/**
+* text_formatter.parser service
+* @package phpBB3
```
Nicofuma: remove service name and @package

phpBB/phpbb/textformatter/parser.php
```diff
@@ -0,0 +1,121 @@
...
+*/
+abstract class parser
```
Nicofuma: should be an interface

phpBB/phpbb/textformatter/renderer.php
```diff
@@ -0,0 +1,136 @@
...
+* For full copyright and license information, please see
+* the docs/CREDITS.txt file.
+*
+*/
+
+namespace phpbb\textformatter;
+
+/**
+* text_formatter.renderer service
+* @package phpBB3
```
Nicofuma: service / @package

phpBB/phpbb/textformatter/renderer.php
```diff
@@ -0,0 +1,136 @@
...
+*/
+abstract class renderer
+{
+	/**
+	* Render given text
+	*
+	* @param  string $text Text, as parsed by the text_formatter.parser service
```
Nicofuma: don't mention the service but the corresponding interface

phpBB/phpbb/textformatter/renderer.php
```diff
@@ -0,0 +1,136 @@
...
+	* @return string
+	*/
+	abstract public function render($text);
+
+	/**
+	* Automatically set the smilies path based on config
+	*
+	* Called by the service container
```
Nicofuma: to be removed

phpBB/phpbb/textformatter/renderer.php
```diff
@@ -0,0 +1,136 @@
...
+		$root_path = (defined('PHPBB_USE_BOARD_URL_PATH') && PHPBB_USE_BOARD_URL_PATH) ? generate_board_url() . '/' : $path_helper->get_web_root_path();
+
+		$this->set_smilies_path($root_path . $config['smilies_path']);
+	}
+
+	/**
+	* Configure this renderer as per the user's settings
+	*
+	* Should set the locale as well as the viewcensor/viewflash/viewimg/viewsmilies options.
+	* Called by the service container
```
Nicofuma: again

phpBB/phpbb/textformatter/renderer.php
```diff
@@ -0,0 +1,136 @@
...
+*
+*/
+
+namespace phpbb\textformatter;
+
+/**
+* text_formatter.renderer service
+* @package phpBB3
+*/
+abstract class renderer
```
Nicofuma: should be/have an interface

phpBB/phpbb/textformatter/s9e/factory.php
```diff
@@ -0,0 +1,517 @@
...
+*/
+
+namespace phpbb\textformatter\s9e;
+
+use s9e\TextFormatter\Configurator;
+use s9e\TextFormatter\Configurator\Items\AttributeFilters\Regexp as RegexpFilter;
+
+/**
+* Creates s9e\TextFormatter objects
+* @package phpBB3
```
Nicofuma: @package

phpBB/phpbb/textformatter/s9e/factory.php
```diff
@@ -0,0 +1,517 @@
...
+*/
+class factory implements \phpbb\textformatter\cache
+{
+	/**
+	* @var phpbb_cache_driver_interface $cache
```
Nicofuma: /phpbb/cache/driver/driver_interface

phpBB/phpbb/textformatter/s9e/factory.php
```diff
@@ -0,0 +1,517 @@
...
+		'list'  => array('{LIST_TYPE}' => '{HASHMAP}'),
+		'quote' => array('{USERNAME}' => '{TEXT1}'),
+		'size'  => array('{SIZE}' => '{FONTSIZE}'),
+		'url'   => array('{DESCRIPTION}' => '{TEXT}'),
+	);
+
+	/**
+	* @var \phpbb\textformatter\data_access
+	*/
+	protected $dal;
```
Nicofuma: use a name with more sense

phpBB/phpbb/textformatter/s9e/factory.php
```diff
@@ -0,0 +1,517 @@
...
+			);
+
+			try
+			{
+				$configurator->BBCodes->addCustom($row['bbcode_match'], $tpl);
+			}
+			catch (\Exception $e)
+			{
+				/**
+				* @todo log an error?
```
Nicofuma: todo?

phpBB/phpbb/textformatter/cache.php
s9e: I think it's a note from a previous iteration that I forgot to remove.

phpBB/phpbb/textformatter/s9e/factory.php
```diff
@@ -0,0 +1,517 @@
...
+		}
+
+		// Replace custom tokens and normalize templates
+		foreach ($templates as $bbcode_name => &$style_templates)
+		{
+			foreach ($style_templates as &$template)
+			{
+				if (isset($this->custom_tokens[$bbcode_name]))
+				{
+					$template = strtr($template, $this->custom_tokens[$bbcode_name]);
```
Nicofuma: I'm not sure about the references

phpBB/phpbb/textformatter/s9e/factory.php
```diff
@@ -0,0 +1,517 @@
...
+		return $bbcodes;
+	}
+
+	/**
+	* Extract and recompose individual BBCode templates from a style's template file
+	*
+	* @param  string $template Style template (bbcode.html)
+	* @return array Associative array matching BBCode names to their template
+	*/
+	protected function extract_templates($template)
```
Nicofuma: you need to extract all the bbcode templates for all the styles?

phpBB/phpbb/textformatter/s9e/parser.php
```diff
@@ -0,0 +1,344 @@
...
+*/
+
+namespace phpbb\textformatter\s9e;
+
+use s9e\TextFormatter\Parser\BuiltInFilters;
+use s9e\TextFormatter\Parser\Logger;
+
+/**
+* s9e\TextFormatter\Parser adapter
+* @package phpBB3
```
Nicofuma: @package

phpBB/phpbb/textformatter/s9e/parser.php
```diff
@@ -0,0 +1,344 @@
...
+*/
+class parser extends \phpbb\textformatter\parser
+{
+	/**
+	* @var s9e\TextFormatter\Parser
```
Nicofuma: \s9e\...

phpBB/phpbb/textformatter/s9e/parser.php
```diff
@@ -0,0 +1,344 @@
...
+	*/
+	protected $parser;
+
+	/**
+	* @var phpbb\user User object, used for translating errors
```
Nicofuma: \phpbb\user

phpBB/phpbb/textformatter/s9e/parser.php
```diff
@@ -0,0 +1,344 @@
...
+	*/
+	protected $user;
+
+	/**
+	* Constructor
+	*
+	* @param  phpbb\cache\driver_interface $cache
```
Nicofuma: same here, when you use the full namespace you have to include the first \

phpBB/phpbb/textformatter/s9e/parser.php
```diff
@@ -0,0 +1,344 @@
...
+	* @return null
+	*/
+	public function __construct(\phpbb\cache\driver\driver_interface $cache, $key, \phpbb\user $user, factory $factory)
+	{
+		$this->user = $user;
+
+		$parser = $cache->get($key);
+		if (!$parser)
+		{
+			extract($factory->regenerate());
```
Nicofuma: extract?

phpBB/phpbb/textformatter/s9e/renderer.php
```diff
@@ -0,0 +1,256 @@
...
+* For full copyright and license information, please see
+* the docs/CREDITS.txt file.
+*
+*/
+
+namespace phpbb\textformatter\s9e;
+
+/**
+* s9e\TextFormatter\Renderer adapter
+* @package phpBB3
```
Nicofuma: @package, class namespace

phpBB/phpbb/textformatter/s9e/renderer.php
```diff
@@ -0,0 +1,256 @@
...
+	*
+	* @param  \phpbb\cache\driver\driver_interface $cache
+	* @param  string $cache_dir Path to the cache dir
+	* @param  string $key Cache key
+	* @param  factory $factory
+	* @return null
+	*/
+	public function __construct(\phpbb\cache\driver\driver_interface $cache, $cache_dir, $key, factory $factory)
+	{
+		$renderer_data = $cache->get($key);
```
Nicofuma: what do you store in $renderer_data? (what kind/format of data and where they are generated, where are they used. Maybe the caching can be optimized here)

phpBB/phpbb/textformatter/s9e/utils.php
```diff
@@ -0,0 +1,67 @@
...
+* For full copyright and license information, please see
+* the docs/CREDITS.txt file.
+*
+*/
+
+namespace phpbb\textformatter\s9e;
+
+/**
+* Text manipulation utilities
+* @package phpBB3
```
Nicofuma: @package

phpBB/phpbb/textformatter/utils.php
```diff
@@ -0,0 +1,62 @@
...
+*/
+
+namespace phpbb\textformatter;
+
+/**
+* text_formatter.utils service
+*
+* Used to manipulate a parsed text
+*
+* @package phpBB3
```
Nicofuma: service / @package

phpBB/phpbb/textformatter/utils.php
```diff
@@ -0,0 +1,62 @@
...
+*/
+abstract class utils
```
Nicofuma: should be an interface

tests/text_formatter/s9e/default_formatting_test.php
```diff
@@ -0,0 +1,218 @@
+<?php
+/**
+*
+* @package testing
+* @copyright (c) 2013 phpBB Group
+* @license http://opensource.org/licenses/gpl-2.0.php GNU General Public License v2
```
Nicofuma: wrong header file

tests/text_formatter/s9e/factory_test.php
```diff
@@ -0,0 +1,178 @@
+<?php
+/**
+*
+* @package testing
+* @copyright (c) 2013 phpBB Group
+* @license http://opensource.org/licenses/gpl-2.0.php GNU General Public License v2
```
Nicofuma: wrong header file

phpBB/phpbb/textformatter/renderer.php
s9e: I don't know which part you want to remove.

tests/text_formatter/s9e/factory_test.php
```diff
@@ -0,0 +1,178 @@
...
+*
+*/
+require_once __DIR__ . '/../../../phpBB/includes/functions.php';
+require_once __DIR__ . '/../../../phpBB/includes/functions_content.php';
+require_once __DIR__ . '/../../mock/user.php';
+require_once __DIR__ . '/../../mock/cache.php';
+require_once __DIR__ . '/../../test_framework/phpbb_database_test_case.php';
```
Nicofuma: no need to include it, it's autoloaded

tests/text_formatter/s9e/factory_test.php
```diff
@@ -0,0 +1,178 @@
...
+*
+* @package testing
+* @copyright (c) 2013 phpBB Group
+* @license http://opensource.org/licenses/gpl-2.0.php GNU General Public License v2
+*
+*/
+require_once __DIR__ . '/../../../phpBB/includes/functions.php';
+require_once __DIR__ . '/../../../phpBB/includes/functions_content.php';
+require_once __DIR__ . '/../../mock/user.php';
+require_once __DIR__ . '/../../mock/cache.php';
```
Nicofuma: no need to include it, it's autoloaded

tests/text_formatter/s9e/parser_test.php
```diff
@@ -0,0 +1,175 @@
+<?php
+/**
+*
+* @package testing
+* @copyright (c) 2013 phpBB Group
+* @license http://opensource.org/licenses/gpl-2.0.php GNU General Public License v2
```
Nicofuma: wrong header file

tests/text_formatter/s9e/parser_test.php
```diff
@@ -0,0 +1,175 @@
...
+*
+*/
+require_once __DIR__ . '/../../../phpBB/includes/functions.php';
+require_once __DIR__ . '/../../../phpBB/includes/functions_content.php';
+require_once __DIR__ . '/../../mock/user.php';
+require_once __DIR__ . '/../../mock/cache.php';
```
Nicofuma: no need to include it, it's autoloaded

tests/text_formatter/s9e/renderer_test.php
```diff
@@ -0,0 +1,363 @@
...
+/**
+*
+* @package testing
+* @copyright (c) 2013 phpBB Group
+* @license http://opensource.org/licenses/gpl-2.0.php GNU General Public License v2
+*
+*/
+require_once __DIR__ . '/../../../phpBB/includes/functions.php';
+require_once __DIR__ . '/../../../phpBB/includes/functions_content.php';
+require_once __DIR__ . '/../../mock/cache.php';
```
Nicofuma: no need to include it, it's autoloaded

tests/text_formatter/s9e/renderer_test.php
```diff
@@ -0,0 +1,363 @@
+<?php
+/**
+*
+* @package testing
+* @copyright (c) 2013 phpBB Group
+* @license http://opensource.org/licenses/gpl-2.0.php GNU General Public License v2
```
Nicofuma: wrong header file

tests/text_formatter/s9e/utils_test.php
```diff
@@ -0,0 +1,157 @@
+<?php
+/**
+*
+* @package testing
+* @copyright (c) 2013 phpBB Group
+* @license http://opensource.org/licenses/gpl-2.0.php GNU General Public License v2
```
Nicofuma: wrong header file

tests/text_processing/decode_message_test.php
```diff
@@ -0,0 +1,87 @@
+<?php
+/**
+*
+* @package testing
+* @copyright (c) 2011 phpBB Group
+* @license http://opensource.org/licenses/gpl-2.0.php GNU General Public License v2
```
Nicofuma: wrong header file

tests/text_processing/generate_text_for_edit_test.php
```diff
@@ -0,0 +1,90 @@
+<?php
+/**
+*
+* @package testing
+* @copyright (c) 2013 phpBB Group
+* @license http://opensource.org/licenses/gpl-2.0.php GNU General Public License v2
```
Nicofuma: wrong header file

tests/text_processing/generate_text_for_edit_test.php
```diff
@@ -0,0 +1,90 @@
...
+*
+*/
+
+require_once __DIR__ . '/../../phpBB/includes/functions.php';
+require_once __DIR__ . '/../../phpBB/includes/functions_content.php';
+require_once __DIR__ . '/../mock/user.php';
+require_once __DIR__ . '/../mock/cache.php';
```
Nicofuma: no need to include it, it's autoloaded

tests/text_processing/generate_text_for_storage_test.php
```diff
@@ -0,0 +1,70 @@
...
+* @package testing
+* @copyright (c) 2013 phpBB Group
+* @license http://opensource.org/licenses/gpl-2.0.php GNU General Public License v2
+*
+*/
+
+require_once __DIR__ . '/../../phpBB/includes/functions.php';
+require_once __DIR__ . '/../../phpBB/includes/functions_content.php';
+require_once __DIR__ . '/../../phpBB/includes/utf/utf_tools.php';
+require_once __DIR__ . '/../mock/container_builder.php';
```
Nicofuma: no need to include it, it's autoloaded

tests/text_processing/generate_text_for_storage_test.php
```diff
@@ -0,0 +1,70 @@
+<?php
+/**
+*
+* @package testing
+* @copyright (c) 2013 phpBB Group
+* @license http://opensource.org/licenses/gpl-2.0.php GNU General Public License v2
```
Nicofuma: wrong header file

tests/text_processing/message_parser_test.php
```diff
@@ -0,0 +1,532 @@
+<?php
+/**
+*
+* @package testing
+* @copyright (c) 2011 phpBB Group
+* @license http://opensource.org/licenses/gpl-2.0.php GNU General Public License v2
```
Nicofuma: wrong header file

tests/text_processing/smilies_test.php
```diff
@@ -0,0 +1,48 @@
+<?php
+/**
+*
+* @package testing
+* @copyright (c) 2013 phpBB Group
+* @license http://opensource.org/licenses/gpl-2.0.php GNU General Public License v2
```
Nicofuma: wrong header file

tests/text_processing/strip_bbcode_test.php
```diff
@@ -0,0 +1,38 @@
+<?php
+/**
+*
+* @package testing
+* @copyright (c) 2011 phpBB Group
+* @license http://opensource.org/licenses/gpl-2.0.php GNU General Public License v2
```
Nicofuma: wrong header file

tests/text_processing/tickets_data/PHPBB3-7275.after.php
```diff
@@ -0,0 +1,15 @@
+<?php
+/**
+*
+* @package testing
+* @copyright (c) 2013 phpBB Group
+* @license http://opensource.org/licenses/gpl-2.0.php GNU General Public License v2
```
Nicofuma: wrong header file

tests/text_processing/tickets_test.php
```diff
@@ -0,0 +1,91 @@
...
+* @package testing
+* @copyright (c) 2013 phpBB Group
+* @license http://opensource.org/licenses/gpl-2.0.php GNU General Public License v2
+*
+*/
+
+require_once __DIR__ . '/../../phpBB/includes/functions.php';
+require_once __DIR__ . '/../../phpBB/includes/functions_content.php';
+require_once __DIR__ . '/../../phpBB/includes/utf/utf_tools.php';
+require_once __DIR__ . '/../mock/container_builder.php';
```
Nicofuma: no need to include it, it's autoloaded

tests/text_processing/tickets_test.php
```diff
@@ -0,0 +1,91 @@
+<?php
+/**
+*
+* @package testing
+* @copyright (c) 2013 phpBB Group
+* @license http://opensource.org/licenses/gpl-2.0.php GNU General Public License v2
```
Nicofuma: wrong header file

phpBB/phpbb/textformatter/s9e/factory.php
s9e: That's for you to decide. I'm glad leaving this question unanswered for now.

phpBB/phpbb/textformatter/s9e/factory.php
s9e: Yes.

phpBB/phpbb/textformatter/s9e/parser.php
s9e: Yes. Do you want to replace it with explicit var declarations?

phpBB/phpbb/textformatter/s9e/renderer.php
s9e: A serialized instance of a class that extends `s9e\TextFormatter\Renderer` and the name of the class. You need the name of the class to load its PHP file before unserializing it.

tests/text_processing/generate_text_for_edit_test.php
s9e: Which ones? I remember having tests fail because a class or function was missing.

phpBB/phpbb/textformatter/s9e/renderer.php
Nicofuma: in this case I think it's better to directly dump a php class (have a look at https://github.com/symfony/DependencyInjection/blob/master/Dumper/PhpDumper.php https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/Routing/Matcher/Dumper/PhpMatcherDumper.php and https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/Routing/Generator/Dumper/PhpGeneratorDumper.php)

Also you could save you some work by using \Symfony\Component\Config\ConfigCache (you can see an example in \phpbb\di\container_builder and \phpbb\routing\router)

Nicofuma: I'm right when I assume that the data change only when you invalidate the cache, isn't it?

tests/text_processing/generate_text_for_edit_test.php
Nicofuma: all the mock should be autoloaded

phpBB/phpbb/textformatter/s9e/renderer.php
s9e: I've thought about it and realized we didn't need to cache an instance of the renderer anymore, just the name of the class. 736c9ee1ff215f349cbe113b7a0366472baedb7f

phpBB/phpbb/textformatter/s9e/factory.php
s9e: On that note, I've just enabled the creation of unsafe BBCodes to match 3.1's behaviour but at some point in the future the ACP module should be modified to leverage the library's ability to detect unsafe BBCodes. 34add55b6acb48362eb11fb63e07f6eb0a528066

phpBB/phpbb/textformatter/s9e/renderer.php
Nicofuma: actually if the data doesn't change it's much more faster to dump it as a correct php code (especially if you have Opcache or something similar installed)

s9e: I don't know what you mean. The library already dumps the compiled templates and all of the data it needs for rendering. The file is in `cache/<environment>/s9e_rendrerer_<hash>.php` and it looks like [this](https://gist.github.com/s9e/6283645).

Nicofuma: you removed the serialize thing. tight

phpBB/phpbb/textformatter/data_access.php
```diff
@@ -0,0 +1,229 @@
...
+*/
+class data_access
+{
+	/**
+	* @var string Name of the BBCodes table
+	*/
+	protected $bbcodes_table;
+
+	/**
+	* @var \phpbb_db_driver
```
Nicofuma: \phpbb\db\driver\driver_interface

phpBB/phpbb/textformatter/renderer.php
```diff
@@ -0,0 +1,129 @@
...
+* @license GNU General Public License, version 2 (GPL-2.0)
+*
+* For full copyright and license information, please see
+* the docs/CREDITS.txt file.
+*
+*/
+
+namespace phpbb\textformatter;
+
+abstract class renderer
```
Nicofuma: make it an interface or at least add a renderer_interface

phpBB/phpbb/textformatter/parser.php
```diff
@@ -0,0 +1,111 @@
...
+* @license GNU General Public License, version 2 (GPL-2.0)
+*
+* For full copyright and license information, please see
+* the docs/CREDITS.txt file.
+*
+*/
+
+namespace phpbb\textformatter;
+
+interface parser
```
Nicofuma: parser_interface

phpBB/phpbb/textformatter/s9e/parser.php
```diff
@@ -0,0 +1,355 @@
...
+				$errors[] = $this->user->lang($msg, $context['max_' . strtolower($m[1])]);
+			}
+			else if ($msg === 'Tag is disabled')
+			{
+				$name = strtolower($context['tag']->getName());
+				$errors[] = $this->user->lang('UNAUTHORISED_BBCODE', '[' . $name . ']');
+			}
+			else if ($msg === 'UNABLE_GET_IMAGE_SIZE')
+			{
+				$errors[] = $this->user->lang[$msg];
```
Nicofuma: `$this->user->lang($msg);̀

phpBB/phpbb/textformatter/s9e/parser.php
```diff
@@ -0,0 +1,355 @@
...
+		$url = BuiltInFilters::filterUrl($url, $url_config, $logger);
+
+		if ($url === false)
+		{
+			return false;
+		}
+
+		if ($max_height || $max_width)
+		{
+			$stats = @getimagesize($url);
```
Nicofuma: You will have to use https://github.com/phpbb/phpbb/pull/3382 when it will be merged

phpBB/phpbb/textformatter/renderer.php
s9e: Made an interface. I removed configure_smilies_path() and configure_user(), not sure whether that's should be a feature. At the moment it's just an implementation detail.

phpBB/phpbb/textformatter/s9e/parser.php
s9e: Is there a reason for using a method call here? Array access through `$user->lang` is ubiquitous in the rest of the code.

phpBB/phpbb/textformatter/s9e/renderer.php
```diff
@@ -0,0 +1,285 @@
...
+			$text = $this->censor->censorHtml($text, true);
+		}
+
+		$html = $this->renderer->render($text);
+
+		/**
+		* @see bbcode::bbcode_second_pass_code()
+		*/
+		$html = preg_replace_callback(
+			'#(<code>)(.*?)(</code>)#is',
```
Nicofuma: can't we have something like `<code class="php"></code>` ?

phpBB/phpbb/textformatter/utils.php
```diff
@@ -0,0 +1,58 @@
...
+* the docs/CREDITS.txt file.
+*
+*/
+
+namespace phpbb\textformatter;
+
+/**
+* Used to manipulate a parsed text
+*/
+abstract class utils
```
Nicofuma: `interface utils_interface`

phpBB/posting.php
```diff
@@ -1525,9 +1525,13 @@
 			'L_MAX_VOTES'		=> $user->lang('MAX_OPTIONS_SELECT', (int) $post_data['poll_max_options']),
 		));
 
-		$parse_poll->message = implode("\n", $post_data['poll_options']);
-		$parse_poll->format_display($post_data['enable_bbcode'], $post_data['enable_urls'], $post_data['enable_smilies']);
-		$preview_poll_options = explode('<br />', $parse_poll->message);
+		$preview_poll_options = array();
```
Nicofuma: seems to be unrelated, there is a reason for this change?

phpBB/phpbb/textformatter/s9e/parser.php
Nicofuma: Array access is kind of legacy I think. Using a method call is definitively better and anyway I don't think the lang array will stay public forever

phpBB/phpbb/textformatter/s9e/renderer.php
s9e: Only by modifying prosilver's `bbcode.html` as well as the default `[code]` definition.

If you want to modify `bbcode.html` we might as well use `<pre><code>` and scrap the tabs-to-mixed-spaces as well.

Nicofuma: the text formatter should be generic and not tied to a single style. So yes I think you should handle the case where the code tag include other data such as a class

phpBB/posting.php
s9e: Each poll option has to be parsed/rendered individually. The old code relies on the assumption that newlines are replaced with `<br />` exactly, which may not always be true depending on the text formatter. And in this case, s9e\TextFormatter replaces newlines with `<br>`.

IIRC there are also some edge cases where you can start a BBCode in one poll option and end it in another, and end up with a broken layout.

Nicofuma: ok

phpBB/phpbb/textformatter/s9e/renderer.php
s9e: Sure. I already changed it.

phpBB/phpbb/textformatter/s9e/parser.php
s9e: Ok I subscribed to the GitHub issue.

phpBB/includes/acp/acp_styles.php
```diff
@@ -53,6 +53,9 @@ class acp_styles
 	/** @var \phpbb\auth\auth */
 	protected $auth;
 
+	/** @var \phpbb\textformatter\cache */
```
marc1706: cache_interface

phpBB/phpbb/textformatter/data_access.php
```diff
@@ -0,0 +1,229 @@
...
+*/
+class data_access
+{
+	/**
+	* @var string Name of the BBCodes table
+	*/
+	protected $bbcodes_table;
+
+	/**
+	* @var \phpbb_db_driver_interface
```
marc1706: \phpbb\db\driver\driver_interface

phpBB/phpbb/textformatter/data_access.php
```diff
@@ -0,0 +1,229 @@
...
+	/**
+	* Constructor
+	*
+	* @param  \phpbb\db\driver\driver_interface $db Database connection
+	* @param  string $bbcodes_table Name of the BBCodes table
+	* @param  string $smilies_table Name of the smilies table
+	* @param  string $styles_table  Name of the styles table
+	* @param  string $words_table   Name of the words table
+	* @param  string $styles_path   Path to the styles dir
+	* @return null
```
marc1706: Isn't the return type \phpbb\textformatter\data_access ?

phpBB/phpbb/textformatter/parser_interface.php
```diff
@@ -0,0 +1,111 @@
...
+	* @param  string $name
+	* @param  mixed  $value
+	* @return null
+	*/
+	public function set_var($name, $value);
+
+	/**
+	* Set multiple variables to be used by the parser
+	*
+	* @param  array Associative array of [name => value]
```
marc1706: @param array $vars Ass...

phpBB/phpbb/textformatter/s9e/factory.php
```diff
@@ -0,0 +1,511 @@
...
+use s9e\TextFormatter\Configurator\Items\AttributeFilters\Regexp as RegexpFilter;
+use s9e\TextFormatter\Configurator\Items\UnsafeTemplate;
+
+/**
+* Creates s9e\TextFormatter objects
+*/
+class factory implements \phpbb\textformatter\cache_interface
+{
+	/**
+	* @var \phpbb\cache\driver_interface $cache
```
marc1706: \phpbb\cache\driver\driver_interface

phpBB/phpbb/textformatter/s9e/factory.php
```diff
@@ -0,0 +1,511 @@
...
+
+	/**
+	* Constructor
+	*
+	* @param  \phpbb\textformatter\data_access $data_access
+	* @param  \phpbb\cache\driver\driver_interface $cache
+	* @param  string $cache_dir          Path to the cache dir
+	* @param  string $cache_key_parser   Cache key used for the parser
+	* @param  string $cache_key_renderer Cache key used for the renderer
+	* @return null
```
marc1706: Same as above, isn't the constructor basically returning the object itself?

phpBB/phpbb/textformatter/s9e/factory.php
```diff
@@ -0,0 +1,511 @@
...
+		// Group identical templates together
+		$grouped_templates = array();
+		foreach ($style_templates as $style_id => $style_template)
+		{
+			$grouped_templates[$style_template][] = $style_id;
+		}
+
+		if (count($grouped_templates) === 1)
+		{
+			return $style_template;
```
marc1706: Will return null if $style_templates is empty

phpBB/phpbb/textformatter/s9e/parser.php
```diff
@@ -0,0 +1,355 @@
...
+	protected $user;
+
+	/**
+	* Constructor
+	*
+	* @param  \phpbb\cache\driver_interface $cache
+	* @param  string $key Cache key
+	* @param  \phpbb\user $user
+	* @param  factory $factory
+	* @return null
```
marc1706: See above comments regarding constructor return type

phpBB/phpbb/textformatter/s9e/parser.php
```diff
@@ -0,0 +1,355 @@
...
+			}
+		}
+
+		return array_unique($errors);
+	}
+
+	/**
+	* Return the instance of s9e\TextFormatter\Parser used by this object
+	*
+	* @return s9e\TextFormatter\Parser
```
marc1706: @return \s9e\TextFormatter\Parser

phpBB/phpbb/textformatter/s9e/renderer.php
```diff
@@ -0,0 +1,285 @@
...
+	protected $viewsmilies = false;
+
+	/**
+	* Constructor
+	*
+	* @param  \phpbb\cache\driver\driver_interface $cache
+	* @param  string $cache_dir Path to the cache dir
+	* @param  string $key Cache key
+	* @param  factory $factory
+	* @return null
```
marc1706: See above comments regarding constructor return type

phpBB/phpbb/textformatter/s9e/renderer.php
```diff
@@ -0,0 +1,285 @@
...
+		}
+
+		// Set the style id
+		$this->renderer->setParameter('STYLE_ID', $user->style['style_id']);
+	}
+
+	/**
+	* Return the instance of s9e\TextFormatter\Renderer used by this object
+	*
+	* @return s9e\TextFormatter\Renderer
```
marc1706: \s9e\TextFormatter\Renderer

tests/test_framework/phpbb_test_case_helpers.php
```diff
@@ -298,4 +300,212 @@ public function empty_dir($path)
...
+
+		// Create and register the text_formatter.s9e.renderer service and its alias
+		$renderer = new \phpbb\textformatter\s9e\renderer(
+			$cache,
+			$cache_dir,
+			$cache_key_renderer,
+			$factory
+		);
+
+		$root_path = ($container->hasParameter('core.root_path'))
```
marc1706: $root_path is not used?

phpBB/phpbb/textformatter/data_access.php
s9e: I removed the annotation so that no one has to wonder.

Nicofuma:
I'll try to test and profile it intensively in the following days. But if anyone else can test it he is welcome

phpBB/includes/acp/acp_bbcodes.php
```diff
@@ -280,6 +281,7 @@ function main($id, $mode)
 							WHERE bbcode_id = ' . $bbcode_id;
 						$db->sql_query($sql);
 						$cache->destroy('sql', BBCODES_TABLE);
+						$phpbb_container->get('text_formatter.cache')->invalidate();
```
brunoais: Why are you not using your BBCodeMonkey in this class?
AFAIK, it is a superset of phpBB's BBCode definition for custom BBCode...

phpBB/includes/acp/acp_bbcodes.php
```diff
@@ -269,6 +269,7 @@ function main($id, $mode)
```
brunoais: Why are you not using your BBCodeMonkey in this class?
AFAIK, it is a superset of phpBB's BBCode definition for custom BBCode...

s9e: To what end? The BBCodeMonkey class is already used when `$configurator->BBCodes->addCustom()` is called during configuration. It could be used in the ACP to more accurately detect unsafe BBCodes or broken definitions but it would require a proper interface. If you have a suggestion, you can develop your idea in the A51 thread.

brunoais: Ah!
phpbb/textformatter/s9e/factory.php belongs to phpBB itself... Ok, then it makes sense.

brunoais:
https://github.com/phpbb/phpbb/pull/3461/files#diff-1bf578cdf0f9f031144aeddc58d34a12R553
I think you should change this to allow new lines in BBCode tags as your system allows it.

Nicofuma:
Well it looks good to me. I will give it another look soon and merge it tonight or in the following days

brunoais:
This is the QUOTE definition, for example
`[QUOTE author={TEXT1;optional} url={URL;optional} author={PARSE=/^\\[url=(?'url'.*?)](?'author'.*)\\[\\/url]$/i} author={PARSE=/^\\[url](?'author'(?'url'.*?))\\[\\/url]$/i}]{TEXT2}[/QUOTE]`

Do you think it is as easy to read as:
	[QUOTE
		author={TEXT1;optional}
		url={URL;optional}
		author={PARSE=/^\\[url=(?'url'.*?)](?'author'.*)\\[\\/url]$/i}
		author={PARSE=/^\\[url](?'author'(?'url'.*?))\\[\\/url]$/i}
		author={PARSE=/(?'url'https?:\\/\\/[^[\\]]+)/i}
		]{TEXT2}[/QUOTE]
?

That's just an example, though.

brunoais:
I used the above definition but I called the tag 'BLOAT'.
I tried to submit a post with this:

> [bloat="AEI!"]Yo world[/bloat]

phpBB entered in an infinite loop.
Bug?

s9e:
@brunoais Ok, so your point is that the library accepts whitespace in BBCodes definitions but phpBB's admin panel doesn't. That's correct. I tried to keep this PR as small as possible because I didn't want to tack on improvements that would delay it more than necessary. The BBCodes admin panel could certainly be improved, but I don't think this is the best time for it.

I can't reproduce this infinite loop. Can you post the content of your `phpbb_bbcodes` table into a Gist or on A51 as well as the text you tried submitting please? I changed the way attributes are parsed by the BBCodes plugin a few days ago, it's not impossible that it introduced a bug.

brunoais:
Use this:
https://gist.github.com/brunoais/3f0718c279a25096cf0b

Also, it seems that it also works if the BBCode isn't defined. I was able to reproduce it also in that state.

s9e:
I can't reproduce. What line does it give you in the error message?

brunoais:
There are no errors. It just becomes an infinite loop somewhere in the code.

s9e:
If you set a non-zero value for `max_execution_time` in your php.ini it should tell you where the execution stopped.

brunoais:
OK, I'll try that.

brunoais:
`Fatal error: Maximum execution time of 2 seconds exceeded in D:\Users\bruno\server2\htdocs\local\git\phpbb3WYSIWYG\phpBB\vendor\twig\twig\lib\Twig\Extension.php on line 11`
I'll try some more times.
It is quite hard to get it to timeout, though... It is locking in even if I lower the timeout number...

brunoais:
Meanwhile, I remembered to check if php had been updated... It had... And quite some sub-versions.
I upgraded to php v.5.4.39 (non-thread-safe) and the infinite loop disappeared.
Probably, the infinite loop was related to one of the libraries that textFormatter uses which was buggy in the version I had. What was the real cause, though... I don't think I'll or I would ever find it.

VSEphpbb:
@Nicofuma By working with @s9e  on how this affects extensions, I've now spent some days using the board with this and it has been working very well. Any issues we found have been resolved. :+1: It's good for me now too.

Nicofuma:
Actually I think we should add a column in the database stating which formatter was used for the post. This way it will make conversion easier.

A simple int column and two constants (for now) (eg: TEXTFORMATTER_PHPBB=0 and TEXTFORMATTER_S9E=1 should do the trick)

Elsensee:
Is it quickly possible to convert posts using the old bbcode parser to use your new stuff :wink: (for example just before displaying them) or would you prefer to do that in another PR?

Nicofuma:
I'm going to merge this PR but I want to write down what we still have to do:
- [ ] Add a cron job to silently reparse all the posts
- [ ] Design an API/Interface for the whole text formatting thingy (I mean: replace generate_text_for_*, includes/bbcode.php and includes/message_parser.php)
- [ ] At least make includes/message_parser.php autoloadable
- [ ] We don't want to fix message_parser to be PHP7 compatible, so we need to reparse the posts on the fly (if needed) when we try to display them.

brunoais:
@Nicofuma 
I think that TextFormater on phpBB should exist the same way that the search backend exists. There should be a base interface of communication for basics setup and then each one would do its job based on that.
What do you think of that? Does it make sense?

Nicofuma:
It make sense yes but it's a little more complicated because we need to always be able to reparse a post with another formatter (and so to clean the parsed post) and we also need generic events for the most common goal of the extensions.

brunoais:
The same way there's an "installer" for each FULLTEXT search, one can be made for TextFormatter.

PR #2462, opened 2014-05-15
[ticket/12505] Remove outdated media handling in attachments

n-aleha:
Remove code for handling media in RealPlayer, QuickTime, and Windows Media
Player.

[PHPBB3-12505](https://tracker.phpbb.com/browse/PHPBB3-12505)

PR #3478, opened 2015-03-08
[ticket/12505] Remove outdated media handling in attachments

n-aleha:
Remove code for handling media in RealPlayer, QuickTime, and Windows Media
Player.

[PHPBB3-12505](https://tracker.phpbb.com/browse/PHPBB3-12505)
~~https://github.com/phpbb/phpbb/pull/2462~~

### DIFF
--- a/phpBB/includes/constants.php
+++ b/phpBB/includes/constants.php
@@ -171,11 +171,11 @@ define('CONFIRM_REPORT', 4);
 // Categories - Attachments
 define('ATTACHMENT_CATEGORY_NONE', 0);
 define('ATTACHMENT_CATEGORY_IMAGE', 1); // Inline Images
-define('ATTACHMENT_CATEGORY_WM', 2); // Windows Media Files - Streaming
-define('ATTACHMENT_CATEGORY_RM', 3); // Real Media Files - Streaming
+define('ATTACHMENT_CATEGORY_WM', 2); // Windows Media Files - Streaming - @deprecated 3.1
+define('ATTACHMENT_CATEGORY_RM', 3); // Real Media Files - Streaming - @deprecated 3.1
 define('ATTACHMENT_CATEGORY_THUMB', 4); // Not used within the database, only while displaying posts
 define('ATTACHMENT_CATEGORY_FLASH', 5); // Flash/SWF files
-define('ATTACHMENT_CATEGORY_QUICKTIME', 6); // Quicktime/Mov files
+define('ATTACHMENT_CATEGORY_QUICKTIME', 6); // Quicktime/Mov files - @deprecated 3.1
 
 // BBCode UID length
 define('BBCODE_UID_LEN', 8);
--- a/phpBB/includes/constants.php
+++ b/phpBB/includes/constants.php
@@ -171,11 +171,11 @@ define('CONFIRM_REPORT', 4);
 // Categories - Attachments
 define('ATTACHMENT_CATEGORY_NONE', 0);
 define('ATTACHMENT_CATEGORY_IMAGE', 1); // Inline Images
-define('ATTACHMENT_CATEGORY_WM', 2); // Windows Media Files - Streaming - @deprecated 3.1
-define('ATTACHMENT_CATEGORY_RM', 3); // Real Media Files - Streaming - @deprecated 3.1
+define('ATTACHMENT_CATEGORY_WM', 2); // Windows Media Files - Streaming - @deprecated 3.2
+define('ATTACHMENT_CATEGORY_RM', 3); // Real Media Files - Streaming - @deprecated 3.2
 define('ATTACHMENT_CATEGORY_THUMB', 4); // Not used within the database, only while displaying posts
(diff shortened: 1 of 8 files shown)


marc1706:
@Nicofuma shouldn't we add a migration to update the existing board with the database changes?

phpBB/includes/constants.php
```diff
@@ -171,11 +171,11 @@
 // Categories - Attachments
 define('ATTACHMENT_CATEGORY_NONE', 0);
 define('ATTACHMENT_CATEGORY_IMAGE', 1); // Inline Images
-define('ATTACHMENT_CATEGORY_WM', 2); // Windows Media Files - Streaming
-define('ATTACHMENT_CATEGORY_RM', 3); // Real Media Files - Streaming
+define('ATTACHMENT_CATEGORY_WM', 2); // Windows Media Files - Streaming - @deprecated 3.1
```
Nicofuma: @deprecated 3.2 now :)

Nicofuma:
Updating the database could be a good idea but I don't remember if we store this information somewhere in the existing attachments. Do we?

n-aleha:
The main obstacle of this PR is how to migrate existing settings with respect to an admin's choices.
There is some talk about it in the previous PR #2462 and a migration attempt https://gist.github.com/n-aleha/8258c160cfd51f185304

Currently, we keep the attachment Special Categories as constants but the actual Extension Groups are in the db. We also keep in the attachment table, only the file extension but I don't think we need to touch the attachments table.

The most compromising solution will be to keep the Extension Groups for: Quicktime Media, Real Media and Windows Media and change their special category to None. 
If we adopt that, I should change the commit https://github.com/n-aleha/phpbb/commit/993cf9c42cb412b08d3483629050585f44c65750 and not move all those extensions to Downloadable Files.

I should probably squash the last commit https://github.com/n-aleha/phpbb/commit/a292f625d01092fec8a0d33eaf66d71680e7f0bd too.

Nicofuma:
Moving everything to download-able sounds like a good idea to me.
But....

- If a user hasn't enabled an outdated extension group, we could simply delete it
- If a board admin has deleted the donwloadable files group... I don't know

Or we can do what you describe for the existing installation and remove everything for the new one.

@marc1706 thoughts?

marc1706:
Yeah, just change the special category to none and move those types to downloadable files.

Btw, you currently get this in the acp:

```
[phpBB Debug] PHP Notice: in file [ROOT]/includes/acp/acp_attachments.php on line 895: Undefined offset: 6
[phpBB Debug] PHP Notice: in file [ROOT]/includes/acp/acp_attachments.php on line 895: Undefined offset: 3
[phpBB Debug] PHP Notice: in file [ROOT]/includes/acp/acp_attachments.php on line 895: Undefined offset: 2
```

marc1706:
@n-aleha How are things looking? Care to fix the mentioned issues?

n-aleha:
Let me summarize this.
For new installations we keep none of the extensions.
See here, https://github.com/n-aleha/phpbb/commit/993cf9c42cb412b08d3483629050585f44c65750 for the ones that we keep.

For existing installations, if the outdated extension groups exists, change all file extensions group (even the ones added by a user) to category none and move them to downloadable files.

And if the admin has deleted the downloadable files group?

marc1706:
In that case you should set the extension group ID to 0 which means that no group is assigned.

n-aleha:
I've updated this. It should be ready for a 1st review.

phpBB/phpbb/db/migration/data/v320/remove_outdated_media.php
```diff
@@ -0,0 +1,83 @@
...
+			array('custom', array(array($this, 'change_extension_group'))),
+		);
+	}
+
+	public function change_extension_group()
+	{
+		// select group ids of outdated media
+		$sql = 'SELECT group_id
+			FROM ' . EXTENSION_GROUPS_TABLE . '
+			WHERE ' . $this->db->sql_in_set('cat_id', $cat_id);
```
marc1706: Needs to be $this->cat_id

marc1706:
You can test a migration by checking out master, installing the board, merging your PR, and afterwards running install/database_update.php

n-aleha:
Updated.The migration file seems to work.
Thanks for the migration tip @marc1706 
With version upgrade, do we need to clear the extensions cache as we do when we delete an extension group from acp? See, https://github.com/phpbb/phpbb/blob/master/phpBB/includes/acp/acp_attachments.php#L658

marc1706:
Yes, please do that in the migration, too.

Elsensee:
IMHO, no. Cache is purged after an update. This includes the extensions cache too, of course.

Nicofuma:
you need to rebase

About the cache, it is purged when using database_update.php and when using db:migrate

n-aleha:
Rebased.
The conflict was in the following commit because of event addition: https://github.com/n-aleha/phpbb/commit/436b020f9e93eb19f52fa8ce65f6b596d4d7f20d

Nicofuma:
Looks good to me now, @marc1706 ?

PR #3482, opened 2015-03-11
[ticket/13684] comment attachment may only be resized vertically

dragosvr92:
More details are explained in the ticket bellow. 

https://tracker.phpbb.com/browse/PHPBB3-13684

PHPBB3-13684

### DIFF
--- a/phpBB/styles/prosilver/theme/forms.css
+++ b/phpBB/styles/prosilver/theme/forms.css
@@ -260,6 +260,7 @@ fieldset.submit-buttons input {
 .inputbox {
 	border: 1px solid transparent;
 	padding: 2px;
+	resize: vertical;
 }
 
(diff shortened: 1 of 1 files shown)


prototech:
Do not apply this to the `.inputbox` class as you're affecting the whole board, not simply these textareas. Target those textareas specifically. Also add the `.autowidth` class to it while you're at it seeing as they're limited to a 85% width at the moment.

dragosvr92:
So where apply it at..? Is " .inputbox .autowidth {resize: vertical;}" an okay path to use? 
Btw, the autowidth class seems to fix the issue. the resize attribute wouldnt really be needed anymore. It will only just add the vertical arrow instead of the corner one. 

I think i will need rxu's help with changing all this with GitTortoise...

marc1706:
Replace by #3959

PR #3479, opened 2015-03-10
[ticket/13685] Allow modifying the query for mysql native fulltext search

brunoais:
https://tracker.phpbb.com/browse/PHPBB3-13685

PR #3486, opened 2015-03-11
[ticket/13693] Allow modifying the MySQL author search query

brunoais:
- [ ] based on https://github.com/phpbb/phpbb/pull/3479

https://tracker.phpbb.com/browse/PHPBB3-13693

### DIFF
--- a/phpBB/phpbb/search/fulltext_mysql.php
+++ b/phpBB/phpbb/search/fulltext_mysql.php
@@ -628,6 +628,55 @@ class fulltext_mysql extends \phpbb\search\base
 
 		$m_approve_fid_sql = ' AND ' . $post_visibility;
 
+		/**
+		* Allow changing the query used to search for posts by author in fulltext_mysql
+		*
+		* @event core.search_mysql_author_query_before
+		* @var	int		result_count		The previous result count for the format of the query.
+		*									Set to 0 to force a re-count
+		* @var	string	sql_sort_table		CROSS JOIN'ed table to allow doing the sort chosen
+		* @var	string	sql_sort_join		Condition to define how to join the CROSS JOIN'ed table specifyed in sql_sort_table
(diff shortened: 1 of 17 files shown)


Nicofuma:
about this search events I think we should have the same event name for all of them (the same event for mysql, native and postgres) to set the common parameters and a dedicated one just to edit the stuff specific to one of the engine.

brunoais:
That is not possible without meddling with the code in such way that I believe it is better not to have it in a maintenance-version.
I believe I can get the time for that in the summer. There's not much code.
Anyway, doing this the best way requires a decent amount of code revamp.

Nicofuma:
lets do that later so

phpBB/phpbb/search/fulltext_mysql.php
```diff
@@ -620,6 +628,55 @@ public function author_search($type, $firstpost_only, $sort_by_sql, $sort_key, $
...
+		* @var	string	sort_dir			"a" for ASC or "d" dor DESC for the sort order used
+		* @var	string	sql_sort			The result SQL when processing sort_by_sql + sort_key + sort_dir
+		* @var	string	sort_days			Time, in days, that the oldest post showing can have
+		* @var	string	sql_time			The SQL to search on the time specifyed by sort_days
+		* @var	bool	firstpost_only		Wether or not to search only on the first post of the topics
+		* @var	array	ex_fid_ary			Forum ids that must not be searched on
+		* @var	array	sql_fora			SQL query for ex_fid_ary
+		* @var	string	m_approve_fid_sql	WHERE clause condition on post_visibility restrictions
+		* @var	int		start				How many posts to skip in the search results (used for pagination)
+		* @since 3.1.4-RC1
```
marc1706: 3.1.5-RC1

PR #3490, opened 2015-03-15
[ticket/13550] Improve checks for Plupload temp directory writability

prototech:
This prevents a PHP warning from being issued that results in an invalid JSON response being returned to the user.

[PHPBB3-13550](https://tracker.phpbb.com/browse/PHPBB3-13550)

### DIFF
--- a/phpBB/phpbb/plupload/plupload.php
+++ b/phpBB/phpbb/plupload/plupload.php
@@ -326,7 +326,7 @@ class plupload
 
 		$tmp_file = $this->temporary_filepath($upload['tmp_name']);
 
-		if (!move_uploaded_file($upload['tmp_name'], $tmp_file))
+		if (!phpbb_is_writable($this->temporary_directory) || !move_uploaded_file($upload['tmp_name'], $tmp_file))
 		{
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/plupload/plupload.php
```diff
@@ -326,7 +326,7 @@ protected function integrate_uploaded_file($form_name, $chunk, $file_path)
 
 		$tmp_file = $this->temporary_filepath($upload['tmp_name']);
 
-		if (!move_uploaded_file($upload['tmp_name'], $tmp_file))
+		if (!is_writable($this->temporary_directory) || !move_uploaded_file($upload['tmp_name'], $tmp_file))
```
MateBartus: phpbb_is_writable would might be better as PHP's is_writable is buggy on windows

prototech: Updated.

prototech:
Updated INSTALL.html to actually point out that the directory needs to be writable as there is no mention of it and updated the install script to check the permissions on it as part of the installation requirements.

PR #3491, opened 2015-03-15
Update startup.php

ajquick:
[3.1.3] ]While suppressing the output from the 'set_magic_quotes_runtime(0)' is sufficient with normal phpBB installs, using this file in association with the oauth plugin will cause an error on some versions of PHP. This fixes a potential bug where set_magic_quotes_runtime is depreciated in newer versions of PHP.

### DIFF
--- a/phpBB/includes/startup.php
+++ b/phpBB/includes/startup.php
@@ -90,7 +90,11 @@ if (version_compare(PHP_VERSION, '5.4.0-dev', '>='))
 }
 else
 {
-	@set_magic_quotes_runtime(0);
+	if(get_magic_quotes_runtime())
+	{
+		// Deactivate
+		@set_magic_quotes_runtime(0);
+	}
 
 	// Be paranoid with passed vars
 	if (@ini_get('register_globals') == '1' || strtolower(@ini_get('register_globals')) == 'on' || !function_exists('ini_get'))


marc1706:
@bantu Shouldn't have travis thrown a failure based on the commit message format?

@ajquick Could you maybe modify your commit message to fit this format? https://wiki.phpbb.com/Git#Commit_Messages

ajquick:
I have not created a ticket.

This is an error that is normally surpressed, but the Oauth code doesn't surpress the error and throws a code. Occurs in PHP 5.3.

marc1706:
Not sure how we should handle this as set_magic_quotes_runtime() has been deprecated since 5.3.0: http://php.net/manual/en/function.set-magic-quotes-runtime.php

ajquick:
Correct.

The code I submitted with this merge checks if set_magic_quotes_runtime is a valid function that can be used. If it is, it will run, if it isn't it won't. There are other ways to do this I'm sure, but what is recommended.

See here: http://php.net/manual/en/function.get-magic-quotes-runtime.php

Normally this is not a problem since phpBB has error handling and also suppresses any errors from this function. However, the Oauth plugin that is standard with phpBB 3.1.3 has its own error reporting. If you run the oauth on php v 5.3 or higher Oauth will throw an error and will not work.

This problem may only also affect php > 5.3 and php < 5.4 I'm not sure.

Nicofuma:
replaced by https://github.com/phpbb/phpbb/pull/3685

PR #3495, opened 2015-03-22
[ticket/13598] lock topic on creation patch

LMOspaceace:
[ticket/13598] lock topic on creation patch

When replying to a topic, there is a check box to lock the topic.
This option should simply also be available on the initial posting process

[PHPBB3-13598](https://tracker.phpbb.com/browse/PHPBB3-13598)

### DIFF
--- a/phpBB/posting.php
+++ b/phpBB/posting.php
@@ -616,6 +616,7 @@ $post_data['enable_urls'] = $post_data['enable_magic_url'];
 if ($mode != 'edit')
 {
 	$post_data['enable_sig']		= ($config['allow_sig'] && $user->optionget('attachsig')) ? true: false;
+	$post_data['topic_status']		= ($mode == 'post' && (isset($_POST['lock_topic']) && $auth->acl_get('m_lock', $forum_id))) ? true : false;
 	$post_data['enable_smilies']	= ($config['allow_smilies'] && $user->optionget('smilies')) ? true : false;
 	$post_data['enable_bbcode']		= ($config['allow_bbcode'] && $user->optionget('bbcode')) ? true : false;
 	$post_data['enable_urls']		= true;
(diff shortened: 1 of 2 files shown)


RMcGirr83:
You need to add the ticket number into the commit message, eg

[ticket/13598] lock topic on creation patch

You also don't need the link to the area51 topic that belongs in the ticket itself on the tracker.  You also need to link to the actual ticket in the commit comment using markdown (eg [PHPBB3-13598] (https://tracker.phpbb.com/browse/PHPBB3-13598)  https://guides.github.com/features/mastering-markdown/)

You should be able to edit your commit (amend previous commit in git) to fix it all

LMOspaceace:
ok. edited... hope that is correct now

prototech:
You've edited the PR title. The actual commit message is still the same. See https://help.github.com/articles/changing-a-commit-message/

phpBB/posting.php
```diff
@@ -616,6 +616,7 @@
 if ($mode != 'edit')
 {
 	$post_data['enable_sig']		= ($config['allow_sig'] && $user->optionget('attachsig')) ? true: false;
+	$post_data['topic_status']		= ($mode == 'post' && (isset($_POST['lock_topic']) && $auth->acl_get('m_lock', $forum_id))) ? true : false;
```
prototech: - Use `$request->is_set_post('lock_topic')` instead of `(isset($_POST['lock_topic'])`.
- Users can also lock their own topics. Use `$auth->acl_gets('m_lock', 'f_user_lock', $forum_id)))` to check for both permissions.
- topic_status is an integer, not a boolean. Use `ITEM_LOCKED` and `ITEM_UNLOCKED` to get the appropriate value.
- Move the this line after `$post_data['enable_urls']...`

prototech: Move this to its own `if ($mode == 'post')` block outside of the `if ($mode == 'edit')`. The way it is now is causing the "Lock topic" checkbox to be unset when replying/quoting in a locked topic (as a moderator).

LMOspaceace:
this whole process is just way too complicated to try to get something into phpbb that should already be in place :(

nickvergessen:
Someone wants to take this over (and also add a test for this)?

RMcGirr83:
I can take it over but at tests I fail.

prototech:
Replaced by https://github.com/phpbb/phpbb/pull/3528

PR #3498, opened 2015-03-23
[ticket/9982] Allow more unicode characters in INTTEXT

Elsensee:
Mostly the patch from @igorw ~~but I included other punctuation as well because I did see no reason for not including them.~~

[PHPBB3-9982](https://tracker.phpbb.com/browse/PHPBB3-9982)


nickvergessen:
Can you add a test to make sure it does not allow `"` which was the reason for this new field type?

Nicofuma:
:+1:

s9e:
What's the use case for INTTEXT? Is there a situation where it's preferable to TEXT?

nickvergessen:
Yes, inside HTML attributes, because it does not allow `"`

s9e:
From what I can tell, special characters `<>"&` are escaped either way and they can't be used to break out of attribute values in 3.1. In 3.2, attribute values are always escaped as well.

I looked into it for a while and I don't think there's a practical difference between TEXT and INTTEXT.

Elsensee:
Okay so what should we do now with INTTEXT?

s9e:
I'd recommend not touching it in 3.1 and phasing it out in 3.2 in favor of TEXT.

marc1706:
bump @nickvergessen @Nicofuma

Nicofuma:
I don't really knows about this one. I don't see where `<>"&` are escaped when using TEXT... But if it is the case in 3.2 I guess we can drop it.

And about 3.1, I think we can keep this patch or keep the current behaviour to avoid any risk

PR #3500, opened 2015-03-23
[ticket/13703] Update upload.php

Nicofuma:
[3.1.X] upload driver for avatar tweak

This change makes the uploaded avatars use the PHPBB_USE_BOARD_URL_PATH constant, which is needed when trying to use the get_user_avatar() function in a listener for index changes.

The reason for this is change is avatars when loaded thru events for index page when phpbb is is a subfolder situation would be set to ./../ instead of still being set to ./ for avatar this change permits that it's a simple line i saw in your smilies functions.

PHPBB3-13703

### DIFF
--- a/phpBB/phpbb/avatar/driver/upload.php
+++ b/phpBB/phpbb/avatar/driver/upload.php
@@ -48,8 +48,9 @@ class upload extends \phpbb\avatar\driver\driver
 	*/
 	public function get_data($row, $ignore_config = false)
 	{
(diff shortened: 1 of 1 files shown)


marc1706:
@Nicofuma need to fix the commit message

Nicofuma:
actually it's for @Reaper88 :)

If you can amend your commit to follow our guidelines (the first line should be prefixed by [ticket/13703] and the last line should be PHPBB3-13703) it would be nice

Reaper88:
It's done on my end

Reaper88:
For this one it's you nicofuma, I edited mine.

nickvergessen:
@marc1706

marc1706:
Replaced by #3518

PR #3502, opened 2015-03-24
[ticket/13564] Unlink user accounts upon deleting user

marc1706:
Previous PR: #3364 

Ticket: https://tracker.phpbb.com/browse/PHPBB3-13564


nickvergessen:
Deleting users has tests IIRC, can we add that easily there?

marc1706:
I could try to prepare a database fixture for testing this.

PR #3503, opened 2015-03-24
Update path_helper.php

ajquick:
[3.1.3] Prevent php function "str_repeat" from going negative, which can happen on some installations.

Common error:

"path_helper.php on line 232: str_repeat() [function.str-repeat]: Second argument has to be greater than or equal to 0"

### DIFF
--- a/phpBB/phpbb/path_helper.php
+++ b/phpBB/phpbb/path_helper.php
@@ -229,7 +229,7 @@ class path_helper
 
 		// Prepend ../ to the phpbb_root_path as many times as / exists in path_info
 		$this->web_root_path = $this->filesystem->clean_path(
-			'./' . str_repeat('../', $corrections) . $this->phpbb_root_path
+			'./' . str_repeat('../', $corrections >= 0 ? $corrections : 0) . $this->phpbb_root_path
 		);
 		return $this->web_root_path;
 	}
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/path_helper.php
```diff
@@ -229,7 +229,7 @@ public function get_web_root_path()
 
 		// Prepend ../ to the phpbb_root_path as many times as / exists in path_info
 		$this->web_root_path = $this->filesystem->clean_path(
-			'./' . str_repeat('../', $corrections) . $this->phpbb_root_path
+			'./' . str_repeat('../', $corrections >= 0 ? $corrections : 0) . $this->phpbb_root_path
```
MGaetan89: Or simply `max($corrections, 0)`? And same for the other changed line.

nickvergessen:
We shouldn't fight the result, but the symptom.
can you name the values which cause this for you, so we can add unit tests for this?

ajquick:
@nickvergessen typically the result of weird data in the URL, or in some cases accessing certain URLs with no url data at all. You can Google the error that shows up and find numerous sites with the error. Here is an example from phpBB.com:

https://www.phpbb.com/community/viewtopic.php%3Ff%3D46%26p%3D13545141

(Note that yes, this is not an output that phpBB would typically give, but since Google managed to find it, I'd assume that it is output somewhere.)

nickvergessen:
Closing since it needs more work anyway.
Ticket created at https://tracker.phpbb.com/browse/PHPBB3-13918

PR #3510, opened 2015-03-30
[ticket/13732] Update composer to be php7 compliant

Nicofuma:
PHPBB3-13732


marc1706:
Maybe add to which version we are upgrading in the commit message?

Nicofuma:
done

PR #3512, opened 2015-04-02
[ticket/13738] Use new branch names in sami-all.conf.php.

bantu:
https://tracker.phpbb.com/browse/PHPBB3-13738


Nicofuma:
shouldn't we generate the doc for master by the way?

bantu:
> shouldn't we generate the doc for master by the way?

There were some issues in Sami related to that.

PR #3487, opened 2015-03-12
[ticket/13697] Moving filesystem related functions to filesystem service

MateBartus:
https://tracker.phpbb.com/browse/PHPBB3-13697

 * Moving filesystem service to \phpbb\filesystem namespace
 * Wraping Symfony's Filesystem component
 * Moving filesystem related functions from includes/functions.php
   into \phpbb\filesystem\filesystem
   Functions moved (and deprecated):
     - phpbb_chmod
     - phpbb_is_writable
     - phpbb_is_absolute
     - phpbb_own_realpath
     - phpbb_realpath
 * Adding interface for filesystem service

PHPBB3-13697

PR #3514, opened 2015-04-02
[ticket/13740] Refactoring installer

CHItA:
Please note that this is a work in progress.

The description of this idea can be found in the [ticket](https://tracker.phpbb.com/browse/PHPBB3-13740).

**List of removed features**:
 - config.php has to be writable, the download option has been removed

**Dependencies that has to be merged before this**:
 - ~~#3487~~ - Moving filesystem related functions into a service (**merged**)
 - ~~#3535~~  - Moving language related functionality into a service (**merged**)
 - ~~#3577~~ -  Make router's extension manager dependency optional (**merged**)
 - ~~#3581~~ - Make template's user dependency optional (**merged**)
 - ~~#3608~~ - Don't fail if the cache isn't writable (**merged**)
 - ~~#3554~~ - Wither interface for container_builder (**merged**)
 - ~~#3552~~ - Moving module managing functionality into a service (**merged**)
 - ~~#3726~~ - Add orderable service collections (**merged**)

**TODO**:
 - [x] Moving and refactoring old installer
 - [ ] ~~Moving and refactoring old updater~~
 - [ ] ~~Moving and refactoring old converter~~
 - [x] Implement AJAX UI
 - [ ] ~~Implement no JS UI~~
 - [x] Add install config environment and install service config files
 - [x] Add/implement new installer helpers
 - [x] Add tests

**Screenshots**:
Please see the [ticket](https://tracker.phpbb.com/browse/PHPBB3-13740) for screenshots.


MateBartus:
!set WIP

Nicofuma:
For the second point (service container builder) I could work on it I already have a few ideas and worked on something earlier and I think there is place for a lot of improvements (about a safe/restricted mode which will not require any config file nor database connection per example. I still have to think about it and write it down on paper)

phpBB/install/database_task_base.php
```diff
@@ -0,0 +1,63 @@
...
+ * @license GNU General Public License, version 2 (GPL-2.0)
+ *
+ * For full copyright and license information, please see
+ * the docs/CREDITS.txt file.
+ *
+ */
+
+namespace phpbb\install;
+
+abstract class database_task_base extends task_base
```
Nicofuma: add a comment explaining what is the goal of the class please

MateBartus:
If you could do the DI container it would be great.

MateBartus:
I'm almost finished moving the old install code into new installer tasks, so feel free to start reviewing those (all the related files are in the *Installer's install tasks and modules* commit). There are still a few broken/unimplemented parts:
 - board, email and server data obtainer task's are not checking input and not generating forms
 - obtain_data or install_filesystem module should have another task for finding imagick path
 - building database index is not implemented, but I don't really get it why that is in the installer anyway
 - install_data's add_modules task is not implemented properly, probably the related functionality should be extracted to a service from acp_modules class
 - add_config_settings task will use obtain_board_data and also we could do password hashing the standard way. Also the update stuff could be built by the database driver.

Beyond the issues listed here the install part is almost there. I think when all the above issues will get fixed I will try to hack together a working installer for the AJAX frontend version, before moving the convert and update legacy code to the new installer.

phpBB/config/install/container/services.yml
```diff
@@ -0,0 +1,5 @@
+imports:
+    - { resource: services_installer.yml }
+    - { resource: services_install_database.yml }
+    - { resource: services_install_filesystem.yml }
+    - { resource: services_install_requirements.yml }
```
Nicofuma: you can include these sub-files in `services_installer.yml` I think

MateBartus:
Another question is that should we keep the install_update.php diff stuff. At first glance it seems like something to make updating easier with a tons of MODs. So the question is, how much we want to support updates for those who modify phpBB's source files.

phpBB/install/app.php
```diff
@@ -0,0 +1,80 @@
...
+require($phpbb_root_path . 'includes/utf/utf_tools.' . $phpEx);
+
+// Set PHP error handler to ours
+set_error_handler(defined('PHPBB_MSG_HANDLER') ? PHPBB_MSG_HANDLER : 'msg_handler');
+
+$phpbb_installer_container_builder = new \phpbb\di\container_builder($phpbb_root_path, $phpEx);
+$phpbb_installer_container = $phpbb_installer_container_builder
+	->with_environment('installer')
+	->without_extensions()
+	->without_cache()
```
Nicofuma: you can use the cache for the installer env I think

MateBartus: Yep, it would be nice, the problem is, at this point we don't know if cache is writable. I'm not sure what happens when cache is not writable and we try to write it, I'll look into that.

Nicofuma: It should fail silently but it needs to be tested.

MateBartus:
If anyone wants to see the installer in action, you can get the installer with its dependencies from [here](https://github.com/MateBartus/phpbb/tree/test/installer). For now, it only contains the controllers for the installer's static pages.

phpBB/install/helper/container_factory.php
```diff
@@ -0,0 +1,149 @@
...
+
+		// This is needed because container_builder::get_env_parameters() uses $_SERVER
+		if ($disable_super_globals)
+		{
+			$this->request->enable_super_globals();
+		}
+
+		$this->container = $phpbb_container = $phpbb_container_builder
+			->with_config($phpbb_config_php_file)
+			->without_cache()
```
MateBartus: @Nicofuma I guess I cannot use cache here right?

Nicofuma: Actually I think you can. At this point we already have the config.php file and nothing should change

MateBartus: My concern is that I need to inject the request service (few lines below) from the other container, and I'm not sure if we cache the container it will be usable in the "one container" environment as it would be cached under the main environment's cache path.

Nicofuma: it needs to be tested but I think that if the service is declared as synthetic you can inject it even with a cached container

Nicofuma: if not then keep it like that

Nicofuma:
What is the current status?

MateBartus:
I have exams currently so this got to the end of my priority list, but at least it is still on it. The installation works that only needs the log messages and the progress bar stuff for the AJAX UI.

BTW @marc1706: any opinion on the JS?

Nicofuma:
@callumacrae js?

phpBB/assets/javascript/installer.js
```diff
@@ -0,0 +1,221 @@
+/**
+ * Installer's AJAX frontend handler
+ */
+
+(function($) { // Avoid conflicts with other libraries
+    // Global variables
```
callumacrae: These are not global variables

phpBB/assets/javascript/installer.js
```diff
@@ -0,0 +1,221 @@
...
+    var pollTimer = null;
+    var nextReadPosition = 0;
+
+    // Template related variables
+    var contentWrapper = $('.install-body').find('.main');
```
callumacrae: Should be called `$contentWrapper`—jQuery objects should have variable names prefixed with dollar signs.

phpBB/assets/javascript/installer.js
```diff
@@ -0,0 +1,221 @@
...
+
+    // Intercept form submits
+    intercept_form_submit($('#install_install'));
+
+    function poll_content(xhReq)
+    {
```
callumacrae: braces on same line in the js

phpBB/assets/javascript/installer.js
```diff
@@ -0,0 +1,221 @@
...
+        }
+
+        form.find(':submit').bind('click', function(event) {
+            event.preventDefault();
+            submit_form(form, $(this));
+        });
+
+    }
+
+    function create_xhr_object()
```
callumacrae: use jquery!

phpBB/assets/javascript/installer.js
```diff
@@ -0,0 +1,221 @@
...
+    // Intercept form submits
+    intercept_form_submit($('#install_install'));
+
+    function poll_content(xhReq)
+    {
+        var messages = xhReq.responseText;
+
+        do {
+            var unprocessed = messages.substring(nextReadPosition);
+            var messageEndIndex = unprocessed.indexOf("}\n\n");
```
callumacrae: single quotes

phpBB/assets/javascript/installer.js
```diff
@@ -0,0 +1,221 @@
...
+
+            if (messageEndIndex != -1)
+            {
+                var endOfMessageIndex = messageEndIndex + 3; // 3 is the length of "}\n\n"
+                var message = unprocessed.substring(0, endOfMessageIndex);
+                parse_message(message);
+                nextReadPosition += endOfMessageIndex;
+            }
+        } while (messageEndIndex != -1);
```
callumacrae: You should use exact equality operators, you've done it in other places too. `!==` not `!=`

phpBB/assets/javascript/installer.js
```diff
@@ -0,0 +1,221 @@
...
+            add_form(responseObject.form);
+        }
+    }
+
+    function add_message(type, messages)
+    {
+        // Get message containers
+        var errorContainer = $('#error-container');
+        var warningContainer = $('#warning-container');
+        var logContainer = $('#log-container');
```
callumacrae: These variables should all have dollar signs. I won't comment on any of the other ones

phpBB/assets/javascript/installer.js
```diff
@@ -0,0 +1,221 @@
...
+
+        var title, description, msgElement, arraySize = messages.length;
+        for (var i = 0; i < arraySize; i++)
+        {
+            msgElement = $(document.createElement('div'));
```
callumacrae: Why not just `$('<div />')`?

phpBB/assets/javascript/installer.js
MateBartus: I have no idea how to proccess streamed response with jQuery, AFAIK it is impossible to access response text while connection is still alive.

phpBB/assets/javascript/installer.js
```diff
@@ -0,0 +1,221 @@
...
+        var formContainer = $('#content-container');
+        formContainer.html(formHtml);
+        var form = $('#install_install');
+        intercept_form_submit(form);
+    }
+
+    function start_polling(xhReq)
+    {
+        reset_polling();
+        pollTimer = setInterval(function() { poll_content(xhReq); }, 500)
```
callumacrae: new lines or bind

phpBB/assets/javascript/installer.js
```diff
@@ -0,0 +1,221 @@
...
+        clearInterval(pollTimer);
+        nextReadPosition = 0;
+    }
+
+    function submit_form(form, submitBtn)
+    {
+        form.css('display', 'none');
+
+        var xhReq = create_xhr_object();
+        xhReq.open("POST", form.attr('action'), true);
```
callumacrae: single quotes. but this entire function can be reduces to like two lines of jquery

phpBB/assets/javascript/installer.js
callumacrae: Ah, didn't realise that that was what you're doing.

phpBB/assets/javascript/installer.js
MateBartus: I meant global as global within the wrapper function or whatever it is called. Any suggestion for naming that?

callumacrae: Config variables? Installer variables?

phpBB/assets/javascript/installer.js
```diff
@@ -0,0 +1,205 @@
...
+        }
+
+        if (responseObject.hasOwnProperty('form')) {
+            add_form(responseObject.form);
+        }
+    }
+
+    function add_message(type, messages) {
+        // Get message containers
+        var errorContainer = $('#error-container');
```
callumacrae: There's lots more variables that need dollar signs, such as these ones :)

phpBB/assets/javascript/installer.js
```diff
@@ -0,0 +1,205 @@
...
+    function add_form(formHtml) {
+        var formContainer = $('#content-container');
+        formContainer.html(formHtml);
+        var form = $('#install_install');
+        intercept_form_submit(form);
+    }
+
+    function start_polling(xhReq) {
+        reset_polling();
+        pollTimer = setInterval(function () {
```
callumacrae: Would it be better to use setTimeout for this one, and fire the new request 500ms after the old one comes in?

MateBartus: That would require rewriting the PHP stuff, which is a possibility, but I wanted to use streaming response because by doing it that way I can avoid a lot of autoloading overhead and what not.

callumacrae: Wouldn't require rewriting anything

http://www.thecodeship.com/web-development/alternative-to-javascript-evil-setinterval/

MateBartus: If what you proposing is to replace this line with anything I'm fine with that. Also if I did something wrong, and the exact same behavior can be kept server side, don't ask, just command because I know nothing about javascript. :)

phpBB/install/controller/install.php
```diff
@@ -0,0 +1,109 @@
...
+			$this->iohandler_factory->set_environment('ajax');
+		}
+		else
+		{
+			$this->iohandler_factory->set_environment('nojs');
+		}
+
+		if ($this->request->is_ajax())
+		{
+			$installer = &$this->installer;
```
Nicofuma: no need for reference.

phpBB/assets/javascript/installer.js
callumacrae: It has exactly the same effect, it just behaves in an easier-to-understand way on slow networks:

```js
function interval() {
  ajaxRequest(function (data) {
    // do something with data
    setTimeout(interval, 500);
  });
}
```

Basically behaves exactly like setInterval, but waits for the previous request to finish before sending a new one

MateBartus: I don't really get it, this does not create a new request (at least it should not), I just want to check if the streaming response has pushed out a new JSON object.

naderman: @callumacrae can you propose the actual change you want to him to make? maybe then we understand what you mean exactly.

MateBartus: we discussed it on IRC, also this is not the final JS I just wanted Marc to look at it, as he had some issues with the version 0.0. So I guess we can solve the implementation issues later, when the everything is implemented.

naderman: Sounds good

naderman:
Can we please move all code into phpBB/phpbb/ - the only thing left in phpBB/install should be the app.php to launch it all. So one can potentially use the other code after installation from other places.

MateBartus:
I guess it is possible, however that would require some security hardenings on tasks IMO (as of now, there is no and there were no checks on the inputs provided during install, only the basic type castings to avoid sql injections).

Nicofuma:
Actually there is one point to discuss and solve: How to deal with the updater, especially with the diff part of it?

CHItA:
@callumacrae Could you have a look at the JS please? Luckily, there won't be much more

phpBB/assets/javascript/installer.js
```diff
@@ -0,0 +1,270 @@
+/**
+ * Installer's AJAX frontend handler
+ */
+
+(function($) { // Avoid conflicts with other libraries
+    // Installer variables
```
marc1706: Missing ```'use strict';```, see https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/assets/javascript/core.js#L8

phpBB/assets/javascript/installer.js
```diff
@@ -0,0 +1,270 @@
...
+
+    // Intercept form submits
+    intercept_form_submit($('#install_install'));
+
+    function poll_content(xhReq) {
+        var messages = xhReq.responseText;
+
+        do {
+            var unprocessed = messages.substring(nextReadPosition);
+            var messageEndIndex = unprocessed.indexOf('}\n\n');
```
marc1706: The variable is defined inside the do loop. However, you use it inside the while comparison. I'm pretty sure this might cause infinite loops since messageEndIndex is not defined outside the do {}.

phpBB/assets/javascript/installer.js
```diff
@@ -0,0 +1,270 @@
...
+
+    function parse_message(messageJSON) {
+        $('#loading_indicator').css('display', 'none');
+
+        messageJSON = messageJSON.trim();
+        var responseObject = JSON.parse(messageJSON);
+
+        // Parse object
+        if (responseObject.hasOwnProperty('errors')) {
+            add_message('error', responseObject.errors)
```
marc1706: Missing semi-colon

phpBB/assets/javascript/installer.js
```diff
@@ -0,0 +1,270 @@
...
+        }
+
+        if (responseObject.hasOwnProperty('warnings')) {
+            add_message('warning', responseObject.warnings)
```
marc1706: Missing semi-colon

phpBB/assets/javascript/installer.js
```diff
@@ -0,0 +1,270 @@
...
+    var progressTimer = null;
+    var currentProgress = 0;
+
+    // Template related variables
+    var $contentWrapper = $('.install-body').find('.main');
+
+    // Intercept form submits
+    intercept_form_submit($('#install_install'));
+
+    function poll_content(xhReq) {
```
marc1706: Please add docblocks for these functions.

phpBB/assets/javascript/installer.js
callumacrae: It's fine, JavaScript has function scoping, not block scoping.

Having said that, it duplicates the variable declaration because the variables are defined in the loop. Won't cause an error to be thrown or affect behaviour, but is still worng.

phpBB/assets/javascript/installer.js
```diff
@@ -0,0 +1,270 @@
...
+        var messages = xhReq.responseText;
+
+        do {
+            var unprocessed = messages.substring(nextReadPosition);
+            var messageEndIndex = unprocessed.indexOf('}\n\n');
+
+            if (messageEndIndex !== -1) {
+                var endOfMessageIndex = messageEndIndex + 3; // 3 is the length of "}\n\n"
```
callumacrae: Instead of a comment explaining the magic number, why not move `}\n\n` into a variable and add the `.length` property instead of 3?

phpBB/assets/javascript/installer.js
```diff
@@ -0,0 +1,270 @@
...
+            }
+        } while (messageEndIndex !== -1);
+
+        if (xhReq.readyState === 4) {
+            $('#loading_indicator').css('display', 'none');
+            reset_polling();
+        }
+    }
+
+    function parse_message(messageJSON) {
```
callumacrae: function names should be camelCase

phpBB/assets/javascript/installer.js
```diff
@@ -0,0 +1,270 @@
...
+        $('#loading_indicator').css('display', 'none');
+
+        messageJSON = messageJSON.trim();
```
callumacrae: Is this line necessary? `JSON.parse` will ignore the whitespace.

phpBB/assets/javascript/installer.js
```diff
@@ -0,0 +1,270 @@
...
+        }, 10);
+    }
+
+    function incrementFiller($progressText, $progressFiller, progressLimit) {
+        currentProgress++;
+        $progressText.text(currentProgress + '%');
+        $progressFiller.css('width', currentProgress + '%');
+
+        if (currentProgress >= progressLimit || currentProgress >= 100) {
+            console.log("In if; " + progressLimit + "; " + currentProgress);
```
callumacrae: This will throw an error in IE8

callumacrae:
I would recommend merging in / rebasing onto master and running JSHint on your javascript, it will probably pick up a few things that we missed :)

CHItA:
It was already rebased, btw JSHint is running on Travis or where should I have a look?

callumacrae:
It isn't running on Travis and I can't add node dependencies to the project, so you'll have to use an editor plugin.

CHItA:
Ohh, right, I just thought it because of the rebasing, that it would mean you don't see it among the tests. Thanks!

CHItA:
@Nicofuma (and anyone else who is interested), about the diff part, if we keep it, then it is an opportunity for outsourcing it. So does anyone knows a lightweight PHP diff lib which would get this job done. IIRC I saw an RFC for post diffs so in that case it could be reused. Or just extract ours into a service?

Nicofuma:
our diff is to limited (and actually it isn't really ours ^^)

maybe we could try to use https://packagist.org/packages/sebastian/diff

CHItA:
I guess with that we could assume that it is tested properly... :)

phpBB/adm/style/install_header.html
```diff
@@ -43,7 +43,7 @@
 							<li<!-- IF l_block1.S_SELECTED --> id="activemenu"<!-- ENDIF -->><a href="{l_block1.U_TITLE}"><span>{l_block1.L_TITLE}</span></a></li>
 						<!-- END l_block1 -->
 						<!-- BEGIN l_block2 -->
-							<li<!-- IF l_block2.S_SELECTED --> id="activemenu"<!-- ENDIF -->><span<!-- IF l_block2.S_COMPLETE --> class="completed"<!-- ENDIF -->>{l_block2.L_TITLE}</span></li>
+							<li<!-- IF l_block2.S_SELECTED --> id="activemenu"<!-- ENDIF -->><span<!-- IF l_block2.S_COMPLETE --> class="completed"<!-- ENDIF --> id="installer-stage-{l_block2.STAGE_NAME}">{l_block2.L_TITLE}</span></li>
```
Nicofuma: it can be used byy something else than the installer

CHItA: but it is not. Do you see any scenario where it would happen?

phpBB/common.php
```diff
@@ -52,7 +52,7 @@
 	}
 
 	// $phpbb_root_path accounts for redirects from e.g. /adm
-	$script_path = trim(dirname($script_name)) . '/' . $phpbb_root_path . 'install/index.' . $phpEx;
+	$script_path = trim(dirname($script_name)) . '/' . $phpbb_root_path . 'install/app.' . $phpEx;
```
Nicofuma: that's true only for the installer

phpBB/config/default/container/services_http.yml
```diff
@@ -0,0 +1,23 @@
...
+    request_stack:
+        class: Symfony\Component\HttpFoundation\RequestStack
+
+    request:
+        class: phpbb\request\request
+        arguments:
+            - null
+            - %core.disable_super_globals%
+
+    http_kernel:
```
Nicofuma: we try to keep an alphabetical order

phpBB/common.php
CHItA: but will be for all the others, or those should go into a separate PR?

phpBB/config/installer/container/services_installer.yml
```diff
@@ -0,0 +1,70 @@
...
+        arguments:
+            - @request
+            - %core.root_path%
+            - %core.php_ext%
+
+# -------- Installer --------------------------------
+    installer.module_base:
+        abstract: true
+        calls:
+            - [setup, [@service_container, @installer.helper.config, @installer.helper.iohandler]]
```
Nicofuma: I don't like injecting the container. Couldn't we use a service collection for the tasks of the modules

phpBB/config/installer/container/services_installer.yml
```diff
@@ -0,0 +1,70 @@
...
+
+    installer.installer.abstract:
+        class: phpbb\install\installer
+        abstract: true
+        arguments:
+            - @installer.helper.config
+            - @service_container
```
Nicofuma: same here, couldn't we use a service collection to list the modules?

phpBB/includes/compatibility_globals.php
```diff
@@ -1,62 +1,79 @@
...
+ *
+ * This file is part of the phpBB Forum Software package.
+ *
+ * @copyright (c) phpBB Limited <https://www.phpbb.com>
+ * @license GNU General Public License, version 2 (GPL-2.0)
+ *
+ * For full copyright and license information, please see
+ * the docs/CREDITS.txt file.
+ *
+ */
```
Nicofuma: remove the extra space in the whole docblock please

phpBB/config/installer/container/services_installer.yml
CHItA: I guess we could, if that initialize the tasks then it might be a bit more difficult, but in principle I don't see why we couldn't.

phpBB/phpbb/install/controller/helper.php
```diff
@@ -0,0 +1,240 @@
...
+ *
+ * This file is part of the phpBB Forum Software package.
+ *
+ * @copyright (c) phpBB Limited <https://www.phpbb.com>
+ * @license GNU General Public License, version 2 (GPL-2.0)
+ *
+ * For full copyright and license information, please see
+ * the docs/CREDITS.txt file.
+ *
+ */
```
Nicofuma: The file doc comment should not be indented.

phpBB/phpbb/install/controller/helper.php
```diff
@@ -0,0 +1,240 @@
...
+	 * @var string
+	 */
+	protected $phpbb_admin_path;
+
+	/**
+	 * @var string
+	 */
+	protected $phpbb_root_path;
+
+	public function __construct(\phpbb\language\language $language, \phpbb\language\language_file_helper $lang_helper, \phpbb\install\helper\navigation\navigation_provider $nav, \phpbb\template\template $template, \phpbb\path_helper $path_helper, \phpbb\symfony_request $request, \phpbb\routing\router $router, $phpbb_root_path)
```
Nicofuma: doc block

phpBB/phpbb/install/controller/installer_index.php
```diff
@@ -0,0 +1,79 @@
...
+	{
+		$this->helper = $helper;
+		$this->language = $language;
+		$this->template = $template;
+		$this->phpbb_root_path = $phpbb_root_path;
+	}
+
+	public function handle($mode)
+	{
+		switch ($mode)
```
Nicofuma: use different  routes

phpBB/phpbb/install/exception/installer_exception.php
```diff
@@ -0,0 +1,22 @@
...
+ * the docs/CREDITS.txt file.
+ *
+ */
+
+namespace phpbb\install\exception;
+
+/**
+ * Installer's base exception
+ */
+class installer_exception extends \Exception
```
Nicofuma: extends phpbb_runtime_exception

phpBB/phpbb/install/exception/invalid_service_name_exception.php
```diff
@@ -0,0 +1,69 @@
...
+	/**
+	 * Constructor
+	 *
+	 * @param string	$error	The name of the missing installer module
+	 * @param array		$params	Additional values for message translation
+	 */
+	public function __construct($error, $params = array())
+	{
+		$this->error = $error;
+		$this->params = $params;
```
Nicofuma: see comment above and don't reimplement the logic

phpBB/phpbb/install/helper/database.php
```diff
@@ -0,0 +1,459 @@
...
+				$errors[] = array(
+					'title' => 'INST_ERR_PREFIX',
+				);
+			}
+
+			// Check if database version is supported
+			switch ($dbms)
+			{
+				case 'mysqli':
+					if (version_compare(mysqli_get_server_info($db->get_db_connect_id()), '4.1.3', '<'))
```
Nicofuma: the dbal driver interface provides or should provide a get_version() method

phpBB/common.php
Nicofuma: we are not going to move the whole acp into controllers in this PR

phpBB/adm/style/install_header.html
Nicofuma: I was talking about the file :) it looks wrong to me to have an `id="installer-stage-{l_block2.STAGE_NAME}"` in the acp

phpBB/config/installer/container/services_installer.yml
Nicofuma: the services are lazy loaded and if you retrieve only the keys they shouldn't be loaded (currently it's not possible to retrieve an array of the keys but it can be done)

phpBB/common.php
CHItA: This is inside an `if (!defined('PHPBB_INSTALLED'))` block. And btw it wouldn't be that bad to move the whole ACP into controllers in this PR ;) so obsolete?

phpBB/adm/style/install_header.html
CHItA: It is `install_header.html` and it is only used by the installer style files. Not the ACP ones

phpBB/common.php
Nicofuma: good point

phpBB/config/installer/container/services_installer.yml
CHItA: then service collection it is then

phpBB/adm/style/install_header.html
Nicofuma: .... I don't know why but I read overall_header.html ....

phpBB/phpbb/install/controller/installer_index.php
CHItA: Actually these are different routes, just in one controller (as the controller logic is the same). So separate controllers or just leave it as it is?

phpBB/config/installer/container/services_installer.yml
CHItA: Maybe a specific service collection for the installer specific stuff? I was wondering, if that might would be a good idea as I could have a method for [this](https://github.com/phpbb/phpbb/pull/3514/files#diff-e36e2b7b4e9d4e2303234ed604175856R235)

phpBB/phpbb/install/controller/installer_index.php
Nicofuma: ok, fine like that

phpBB/config/installer/container/services_installer.yml
Nicofuma: you can specialized it if it makes sense yes

Nicofuma:
I have merged https://github.com/phpbb/phpbb/pull/3726 Tell me if something else is needed to merge this PR.

CHItA:
Thanks, I think I don't need anything else merged ATM.

phpBB/config/installer/container/services_install_data.yml
```diff
@@ -0,0 +1,44 @@
...
+        class: phpbb\install\module\install_data\task\add_bots
+        arguments:
+            - @installer.helper.config
+            - @installer.helper.iohandler
+            - @installer.helper.container_factory
+            - @language
+            - %core.root_path%
+            - %core.php_ext%
+        tags:
+            - { name: install_data_install, order: 2 }
```
Nicofuma: maybe we could use 10/20/30.... to be make our life easier if we want to add a new task later

phpBB/install-config.yml
```diff
@@ -0,0 +1,35 @@
+installer:
```
Nicofuma: this file is a sample, part of the (to rite) documentation. Please move it to phpBB/docs/install-config.sample.yml

phpBB/install_new/app.php
```diff
@@ -0,0 +1,79 @@
...
+// Let's do the common.php logic
+//
+require($phpbb_root_path . 'includes/startup.' . $phpEx);
+require($phpbb_root_path . 'phpbb/class_loader.' . $phpEx);
+
+$phpbb_class_loader = new \phpbb\class_loader('phpbb\\', "{$phpbb_root_path}phpbb/", $phpEx);
+$phpbb_class_loader->register();
+
+// In case $phpbb_adm_relative_path is not set (in case of an update), use the default.
+$phpbb_adm_relative_path = (isset($phpbb_adm_relative_path)) ? $phpbb_adm_relative_path : 'adm/';
```
Nicofuma: here `$phpbb_adm_relative_path` will never be set (we don't load any config file)

phpBB/phpbb/install/console/command/install/config/default_config.php
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	/**
+	 *
+	 * {@inheritdoc}
+	 */
+	protected function configure()
+	{
+		$this
+			->setName('install:config:default')
```
Nicofuma: I wonder if we really need this command.
I think it can be removed (it should be part of the documentation)

phpBB/phpbb/install/console/command/install/install.php
```diff
@@ -0,0 +1,194 @@
...
+
+		$this->installer->set_iohandler($iohandler);
+
+		$config_file = $input->getArgument('config-file');
+
+		if (!is_file($config_file))
+		{
+			$iohandler->add_error_message(array('MISSING_FILE', array($config_file)));
+
+			return;
```
Nicofuma: return 1

phpBB/phpbb/install/console/command/install/install.php
```diff
@@ -0,0 +1,194 @@
...
+
+		try
+		{
+			$config = Yaml::parse(file_get_contents($config_file), true, false);
+		}
+		catch (ParseException $e)
+		{
+			$iohandler->add_error_message('INVALID_YAML_FILE');
+
+			return;
```
Nicofuma: return 1

phpBB/phpbb/install/console/command/install/install.php
```diff
@@ -0,0 +1,194 @@
...
+
+		try
+		{
+			$config = $processor->processConfiguration($configuration, $config);
+		}
+		catch (Exception $e)
+		{
+			$iohandler->add_error_message('INVALID_CONFIGURATION', $e->getMessage());
+
+			return;
```
Nicofuma: return 1

phpBB/phpbb/install/console/command/install/install.php
```diff
@@ -0,0 +1,194 @@
...
+		$this->register_configuration($iohandler, $config);
+
+		try
+		{
+			$this->installer->run();
+		}
+		catch (installer_exception $e)
+		{
+			$iohandler->add_error_message($e->getMessage());
+			return;
```
Nicofuma: return 1

phpBB/phpbb/install/controller/install.php
```diff
@@ -0,0 +1,198 @@
...
+
+		// Set the appropriate input-output handler
+		$this->installer->set_iohandler($this->iohandler_factory->get());
+
+		// Set up navigation
+		$nav_data = $this->installer_config->get_navigation_data();
+		/** @var \phpbb\install\helper\iohandler\iohandler_interface $iohandler */
+		$iohandler = $this->iohandler_factory->get();
+
+		if ($this->install_helper->is_phpbb_installed())
```
Nicofuma: I think it should be the first thing in the method.

phpBB/phpbb/install/exception/cannot_build_container_exception.php
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * For full copyright and license information, please see
+ * the docs/CREDITS.txt file.
+ *
+ */
+
+namespace phpbb\install\exception;
+
+/**
+ * This exception should be thrown when
```
Nicofuma: when? :)

phpBB/phpbb/install/exception/installer_config_not_writable_exception.php
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * For full copyright and license information, please see
+ * the docs/CREDITS.txt file.
+ *
+ */
+
+namespace phpbb\install\exception;
+
+/**
+ * Exception for the event when installer config is not writable to disk
```
Nicofuma: `Thrown when installer config is not writable to disk`

phpBB/phpbb/install/exception/invalid_dbms_exception.php
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * For full copyright and license information, please see
+ * the docs/CREDITS.txt file.
+ *
+ */
+
+namespace phpbb\install\exception;
+
+/**
+ * This exception should be thrown when
```
Nicofuma: When?

phpBB/phpbb/install/helper/container_factory.php
```diff
@@ -0,0 +1,149 @@
...
+	/**
+	 * Returns the specified parameter from the container
+	 *
+	 * @param string	$param_name
+	 *
+	 * @return mixed
+	 */
+	public function get_parameter($param_name)
+	{
+		return $this->container->getParameter($param_name);
```
Nicofuma: you should check and build the container here too

phpBB/phpbb/install/helper/iohandler/ajax_iohandler.php
```diff
@@ -0,0 +1,274 @@
...
+	/**
+	 * {@inheritdoc}
+	 */
+	public function send_response()
+	{
+		$json_data_array = $this->prepare_json_array();
+		$json_data = json_encode($json_data_array);
+
+		// Try to push content to the browser
+		print (str_pad(' ', 4096) . "\n");
```
Nicofuma: no space after print please

phpBB/phpbb/install/console/command/install/config/default_config.php
CHItA: no opinion. Should I delete the file?

Nicofuma: yes please :)

By the way, in the doc the sample file has an extra dot in its name ;)

phpBB/adm/style/install_header.html
```diff
@@ -15,10 +15,14 @@
 		<h1>{L_INSTALL_PANEL}</h1>
 		<p id="skip"><a href="#acp">{L_SKIP}</a></p>
 		<!-- IF S_LANG_SELECT -->
-		<form method="post" action="#">
+		<form method="post" action="#" id="language_selector">
 			<fieldset class="nobg">
 				<label for="language">{L_SELECT_LANG}{L_COLON}</label>
-				{S_LANG_SELECT}
```
Nicofuma:
```
<!-- IF _loop.language_select_item -->
				<select  id="language" name="language">
                    <!-- BEGIN language_select_item -->
                    <option value="{language_select_item.VALUE}"<!-- IF language_select_item.SELECTED --> selected="selected"<!-- ENDIF -->>{language_select_item.NAME}</option>
                    <!-- END language_select_item -->
                </select>
<!-- ELSE -->
{S_LANG_SELECT}
<!-- ENDIF -->
```

Should work and solve the BC issue I think

Nicofuma: tested and it works

Nicofuma:
In case you didn't see it:
```
2015/07/23 00:09:30 [error] 4162#0: *1 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Class '\phpbb\db\migration\data\v310\notifications_board' not found in.../phpbb/phpBB/phpbb/db/migrator.php on line 804" while reading upstream, client: 127.0.0.1, server: phpbb.local, request: "POST /install_new/app.php/install HTTP/1.1", upstream: "fastcgi://unix:/var/run/php5-fpm.sock:", host: "phpbb.local", referrer: "http://phpbb.local/install_new/app.php/install"
```

Nicofuma:
I don't remember if it's already the case, but otherwise you should delete the config file in stores when the install is completed.

CHItA:
That's already done

phpBB/phpbb/install/module/obtain_data/task/obtain_server_data.php
```diff
@@ -0,0 +1,203 @@
...
+		$script_path = str_replace(array('\\', '//'), '/', $script_path);
+		$script_path = trim(dirname(dirname($script_path)));
+
+		// Server data
+		$cookie_secure		= $this->io_handler->get_input('cookie_secure', $cookie_secure);
+		$server_protocol	= $this->io_handler->get_input('server_protocol', $server_protocol);
+		$force_server_vars	= $this->io_handler->get_input('force_server_vars', 0);
+		$server_name		= $this->io_handler->get_input('server_name', $server_name);
+		$server_port		= $this->io_handler->get_input('server_port', $server_port);
+		$script_path		= $this->io_handler->get_input('script_path', $script_path);
```
CHItA: service path has `install_new` in it which it should not

CHItA: Fixed

CHItA:
At my end it is ready to be merged.
@Nicofuma, @callumacrae, @marc1706

phpBB/phpbb/install/controller/helper.php
```diff
@@ -0,0 +1,363 @@
...
+			$lang = $lang_cookie;
+			$this->language_cookie = $lang;
+		}
+
+		$lang = (!empty($lang)) ? $lang : null;
+		$this->render_language_select($lang);
+
+		if ($lang !== null)
+		{
+			$this->language->set_user_language($lang, true);
```
CHItA: This might be a vulnerability, it should be checked at least if the result does not contain any path changing characters. Even better that the directory is valid etc.

Nicofuma: should be done line 194 and throw an exception in case of error I think (create a ticket and lets do that later)

phpBB/assets/javascript/installer.js
```diff
@@ -0,0 +1,400 @@
...
+
+		var $title, $description, $msgElement, arraySize = messages.length;
+		for (var i = 0; i < arraySize; i++) {
+			$msgElement = $('<div />');
+			$title = $(document.createElement('strong'));
+			$title.text(messages[i].title);
+			$msgElement.append($title);
+
+			if (messages[i].hasOwnProperty('description')) {
+				$description = $(document.createElement('p'));
```
marc1706: Any reason why you are not using the same as here? https://github.com/phpbb/phpbb/pull/3514/files#diff-7bf94728366cca27f7f0afb9e259924bR147

phpBB/config/default/container/services.yml
```diff
@@ -8,13 +8,16 @@ imports:
     - { resource: services_db.yml }
     - { resource: services_event.yml }
     - { resource: services_feed.yml }
+    - { resource: services_files.yml }
```
marc1706: Will cause a nice conflict with my filename here: https://github.com/phpbb/phpbb/pull/3727
I personally don't care if I have to merge those two files.

phpBB/includes/compatibility_globals.php
```diff
@@ -12,51 +12,68 @@
 */
 
 /**
-*/
+ */
```
marc1706: Unnecessary change, shouldn't be in here ;)

phpBB/language/en/install_new.php
```diff
@@ -0,0 +1,307 @@
...
+
+	// License
+	'LICENSE_TITLE'		=> 'General Public License',
+
+	// Install page
+	'INSTALL_INTRO'			=> 'Welcome to Installation',
+	'INSTALL_INTRO_BODY'	=> 'With this option, it is possible to install phpBB3 onto your server.</p><p>In order to proceed, you will need your database settings. If you do not know your database settings, please contact your host and ask for them. You will not be able to continue without them. You need:</p>
+
+	<ul>
```
marc1706: I don't like that we still have such a large portion of HTML code inside the language file

phpBB/phpbb/console/application.php
```diff
@@ -35,7 +35,7 @@ class application extends \Symfony\Component\Console\Application
 	* @param string			$version	The version of the application
 	* @param \phpbb\user	$user		The user which runs the application (used for translation)
 	*/
-	public function __construct($name, $version, \phpbb\user $user)
+	public function __construct($name, $version, $user)
```
marc1706: Any reason to remove this?

CHItA: probably because we don't have the user object there, but that was done by @Nicofuma

phpBB/config/default/container/services.yml
CHItA: I think that if you merge the two, that would make sense, so I think we should only change the file names if the services don't go together logically.

phpBB/language/en/install_new.php
CHItA: I don't like that either, but i don't feel like anything about it if that's fine

phpBB/phpbb/install/controller/helper.php
CHItA: now it has a basic check

phpBB/phpbb/console/application.php
Nicofuma: the user is only used for the lang() function and in the installer we can't inject the user.

Anyway we need to change the constructor to `public function __construct($name, $version, \phpbb\language\language $language)`

CHItA:
I assume the remaining comments to be either resolved or won't fix (as of this PR), unless instructed otherwise.

phpBB/phpbb/install/controller/helper.php
```diff
@@ -0,0 +1,364 @@
...
+		{
+			$lang = $this->phpbb_request->variable('language', '');
+
+			if (!empty($lang))
+			{
+				$this->language_cookie = $lang;
+			}
+		}
+
+		// Retrive language from cookie
```
marc1706: Retrieve

phpBB/phpbb/install/helper/config.php
```diff
@@ -0,0 +1,386 @@
...
+		$file_content = @file_get_contents($this->install_config_file);
+		$serialized_data = trim(substr($file_content, 8));
+
+		$this->installer_config = array();
+		$this->progress_data = array();
+		$this->navigation_data = array();
+
+		if (!empty($serialized_data))
+		{
+			$unserialized_data = unserialize($serialized_data);
```
marc1706: I'd prefer it if we would be using the json format instead of serialize(). There are definitely some implications and possible issues with using serialize() and unserialize()

phpBB/phpbb/install/helper/container_factory.php
```diff
@@ -0,0 +1,171 @@
...
+	 * @param string	$param_name
+	 *
+	 * @return mixed
+	 *
+	 * @throws \phpbb\install\exception\cannot_build_container_exception	When container cannot be built
+	 */
+	public function get_parameter($param_name)
+	{
+		// Check if container was built, if not try to build it
+		if ($this->container === null)
```
marc1706: Duplicate of above code. Please use a method for doing this instead of duplicating code.

phpBB/phpbb/install/helper/database.php
```diff
@@ -0,0 +1,459 @@
...
+	 * @param \phpbb\filesystem\filesystem_interface	$filesystem			Filesystem interface
+	 * @param string									$phpbb_root_path	Path to phpBB's root
+	 */
+	public function __construct(\phpbb\filesystem\filesystem_interface $filesystem, $phpbb_root_path)
+	{
+		$this->filesystem		= $filesystem;
+		$this->phpbb_root_path	= $phpbb_root_path;
+
+		// DBMS supported by phpBB
+		$this->supported_dbms = array(
```
marc1706: Any reason why you are defining this in the constructor and not just by default for the class member?

phpBB/phpbb/install/module/install_data/task/add_bots.php
```diff
@@ -0,0 +1,240 @@
...
+		 * 'Susie [Sync]'				'!Susie (http://www.sync2it.com/susie)'
+		 * 'Tbot [Bot]'					'Tbot/'
+		 * 'Thumbshots [Capture]'		'thumbshots-de-Bot'
+		 * 'Vagabondo [Crawler]'		'http://webagent.wise-guys.nl/'
+		 * 'Walhello [Bot]'				'appie 1.1 (www.walhello.com)'
+		 * 'WissenOnline [Bot]'			'WissenOnline-Bot'
+		 * 'WWWeasel [Bot]'				'WWWeasel Robot v'
+		 * 'Xaldon [Spider]'			'Xaldon WebSpider'
+		 */
+		$this->bot_list = array(
```
marc1706: Same as above. Why are you defining this in the constructor and not by default for $this->bot_list?

phpBB/phpbb/install/controller/helper.php
CHItA: done

phpBB/phpbb/install/helper/container_factory.php
CHItA: done

phpBB/phpbb/install/schemas/schema_data.sql
```diff
@@ -0,0 +1,821 @@
+#
```
Nicofuma: I wonder if these files shouldn't sit in /phpBB/install/schemas (so /phpBB/install_new/schemas for now)

@phpbb/development-team thought?

CHItA: https://tracker.phpbb.com/browse/PHPBB3-14044

Nicofuma: I mean these files are resources and I don't think we want to have any resources under phpBB/phpbb, I'm seeing this directory more like src/ or lib/ in some packages.

phpBB/phpbb/install/module/obtain_data/task/obtain_server_data.php
```diff
@@ -0,0 +1,203 @@
...
+
+		$script_path = htmlspecialchars_decode($this->io_handler->get_server_variable('PHP_SELF'));
+
+		if (!$script_path)
+		{
+			$script_path = htmlspecialchars_decode($this->io_handler->get_server_variable('REQUEST_URI'));
+		}
+
+		$script_path = str_replace(array('\\', '//'), '/', $script_path);
+		$script_path = trim(dirname(dirname(dirname($script_path)))); // Because we are in install/app.php/route_name
```
Nicofuma: we should use the Symfony Request object to retrieve the default values. 
But it can be done in another PR. I'm just writing it for the record.

CHItA: it uses our request object i guess now.

phpBB/phpbb/install/helper/navigation/navigation_provider.php
```diff
@@ -0,0 +1,115 @@
+<?php
```
Nicofuma: For the record: revamp this file to not use references

CHItA: What is the problem with references?

phpBB/phpbb/install/helper/iohandler/cli_iohandler.php
```diff
@@ -0,0 +1,265 @@
+<?php
```
Nicofuma: For the record: revamp the progress bar to do like in the reparsers (always use a progress bar, just change the format)

phpBB/phpbb/install/controller/install.php
```diff
@@ -0,0 +1,200 @@
...
+
+			// Try to bypass any server output buffers
+			$response->headers->set('X-Accel-Buffering', 'no');
+
+			return $response;
+		}
+		else
+		{
+			// Determine whether the installation was started or not
+			if (true)
```
Nicofuma: purpose of this if?

CHItA: Relevant for the NO JS IO, the todo is below this whole block

phpBB/phpbb/install/controller/helper.php
```diff
@@ -0,0 +1,364 @@
...
+		$this->navigation_provider = $nav;
+		$this->template = $template;
+		$this->path_helper = $path_helper;
+		$this->phpbb_request = $phpbb_request;
+		$this->request = $request;
+		$this->router = $router;
+		$this->phpbb_root_path = $phpbb_root_path;
+		$this->phpbb_admin_path = $phpbb_root_path . 'adm/';
+
+		$this->handle_language_select();
```
Nicofuma: It shouldn't be done in the constructor but in a function called by the controllers...

CHItA: I think it is fine here

phpBB/install_new/phpbbcli.php
```diff
@@ -0,0 +1,68 @@
...
+include($phpbb_root_path . 'includes/functions_compatibility.' . $phpEx);
+require($phpbb_root_path . 'includes/functions_user.' . $phpEx);
+require($phpbb_root_path . 'includes/utf/utf_tools.' . $phpEx);
+
+// Set PHP error handler to ours
+set_error_handler(defined('PHPBB_MSG_HANDLER') ? PHPBB_MSG_HANDLER : 'msg_handler');
+
+$phpbb_installer_container_builder = new \phpbb\di\container_builder($phpbb_root_path, $phpEx);
+$phpbb_installer_container = $phpbb_installer_container_builder
+	->with_environment('installer')
```
Nicofuma: why not using `define('PHPBB_ENVIRONMENT', 'installer');` instead?

phpBB/install_new/app.php
```diff
@@ -0,0 +1,79 @@
...
+include($phpbb_root_path . 'includes/functions_compatibility.' . $phpEx);
+require($phpbb_root_path . 'includes/functions_user.' . $phpEx);
+require($phpbb_root_path . 'includes/utf/utf_tools.' . $phpEx);
+
+// Set PHP error handler to ours
+set_error_handler(defined('PHPBB_MSG_HANDLER') ? PHPBB_MSG_HANDLER : 'msg_handler');
+
+$phpbb_installer_container_builder = new \phpbb\di\container_builder($phpbb_root_path, $phpEx);
+$phpbb_installer_container = $phpbb_installer_container_builder
+	->with_environment('installer')
```
Nicofuma: why not using `define('PHPBB_ENVIRONMENT', 'installer');` instead?

phpBB/phpbb/install/helper/navigation/navigation_provider.php
Nicofuma: in some cases it can hurt performances very hard (when a performance is passed instead of a raw variable or the other side) and it often (always?) make the code more complicated to understand.

phpBB/phpbb/install/controller/install.php
Nicofuma: ok

phpBB/phpbb/install/controller/helper.php
Nicofuma: that's because for now the request lives in the container, but one time it will change and we will pass a request_stack object instead and the request will be set by the http_kernel. Having the request in the container can be easy at first but it's a non sense and it also make things a lot more difficult sometimes (it has been discussed a lot in Symfony)

phpBB/install_new/phpbbcli.php
CHItA: no particular reason, probably it should be fixed in an other PR if ever

phpBB/phpbb/install/helper/iohandler/cli_iohandler.php
CHItA: It can wait till another PR i guess

phpBB/phpbb/install/helper/navigation/navigation_provider.php
CHItA: Fixed in places where it was trivial, other stuff can be fixed latter if someone wants to

phpBB/phpbb/install/schemas/schema_data.sql
CHItA: It can be decided when deduplicating the two installers i guess

PR #3515, opened 2015-04-03
Ticket/13741 Remove outdated comments and references in css files

cyberalien:
https://tracker.phpbb.com/browse/PHPBB3-13741


bantu:
@nickvergessen I think this is fine for 3.1.

PR #3516, opened 2015-04-03
[Ticket/13742] Update local.php same as ticket/13703

Reaper88:
[Ticket/13742] Local Gallery driver for avatar tweak

This change makes the uploaded avatars use the PHPBB_USE_BOARD_URL_PATH constant, which is needed when trying to use the get_user_avatar() function in a listener for index changes.

The reason for this is change is avatars when loaded thru events for index page when phpbb is is a subfolder situation would be set to ./../ instead of still being set to ./ for avatar this change permits that it's a simple line i saw in your smilies functions.

PHPBB3-13742

### DIFF
--- a/phpBB/phpbb/avatar/driver/local.php
+++ b/phpBB/phpbb/avatar/driver/local.php
@@ -23,8 +23,9 @@ class local extends \phpbb\avatar\driver\driver
 	*/
 	public function get_data($row)
 	{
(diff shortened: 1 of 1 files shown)


nickvergessen:
@marc1706

marc1706:
Replaced by #3519

PR #3517, opened 2015-04-03
[ticket/13739] Add template core event posting_poll_body_options_after

Wolfsblvt:
Template event added for the poll options while posting a topic
so that you may add additional options for polls.

PHPBB3-13739

### DIFF
--- a/phpBB/styles/prosilver/template/posting_poll_body.html
+++ b/phpBB/styles/prosilver/template/posting_poll_body.html
@@ -44,8 +44,9 @@
 			</dl>
 		<!-- ENDIF -->
 		<!-- EVENT posting_poll_body_options_after -->
+	<!-- ENDIF -->
 
-	<!-- ELSEIF S_POLL_DELETE -->
+	<!-- IF S_POLL_DELETE -->
 		<dl class="fields1">
 			<dt><label for="poll_delete">{L_POLL_DELETE}{L_COLON}</label></dt>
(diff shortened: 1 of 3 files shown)


phpBB/styles/prosilver/template/posting_poll_body.html
```diff
@@ -43,6 +43,7 @@
 				<dd><label for="poll_vote_change"><input type="checkbox" id="poll_vote_change" name="poll_vote_change"{VOTE_CHANGE_CHECKED} /> {L_POLL_VOTE_CHANGE_EXPLAIN}</label></dd>
 			</dl>
 		<!-- ENDIF -->
+		<!-- EVENT posting_poll_body_options_after -->
```
nickvergessen: In subsilver2 the event is also used in case of `S_POLL_DELETE`, here it is not.
Mind fixing that, so both behave the same?

Wolfsblvt: You mean adding the event to the `ELSEIF S_POLL_DELETE` too?
I can do that.

nickvergessen: or closing the if and reopening it like in subsilver2

Wolfsblvt: Like this?
Seems to be more correct I would say. In this case it is working the same in both styles.

nickvergessen: Yes, now you just need to swap the order of EVENT and ENDIF ;)

Wolfsblvt: This gets me learning how to use git with pull requests :P Hope it is correct now.

PR #3520, opened 2015-04-05
[ticket/13743] Add $phpbb_root_path and $phpEx definitions in message_parser.php

rxu:
```message_parser.php``` performs include of bbcode.php at the very start using
```$phpbb_root_path``` and ```$phpEx``` vars, so there's a possibility those can be
undefined at that point, especially when message_parser.php is being included
in event listener or another extension file where ```$phpbb_root_path``` and ```$phpEx```
don't exist. This can be fixed by adding the appropriate vars definitions.

<a href="https://tracker.phpbb.com/browse/PHPBB3-13743">PHPBB3-13743</a>.

### DIFF
--- a/phpBB/includes/message_parser.php
+++ b/phpBB/includes/message_parser.php
@@ -21,6 +21,19 @@ if (!defined('IN_PHPBB'))
 
 if (!class_exists('bbcode'))
 {
+	// The following lines are for extensions which include message_parser.php
+	// while $phpbb_root_path and $phpEx are out of the script scope
+	// which may lead to the 'Undefined variable' and 'failed to open stream' errors
+	if (!isset($phpbb_root_path))
+	{
+		global $phpbb_root_path;
+	}
+
+	if (!isset($phpEx))
+	{
+		global $phpEx;
+	}
+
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: phpBB/includes/message_parser.php [hunk 1]
```diff
@@ -19,6 +19,19 @@
 	exit;
 }
 
+// The following lines are for extensions which include message_parser.php
+// while $phpbb_root_path and $phpEx are out of the script scope
+// which may lead to the 'Undefined variable' and 'failed to open stream' errors
+if (!isset($phpbb_root_path))
```
nickvergessen: Move inside the class check?

rxu: I thought about that, but if ```bbcode`` class does exist but the vars aren't defined, it will lead to errors as well.

MGaetan89: Wouldn't it be easier to simply add `global $phpbb_root_path, $phpEx;` in the class check?

rxu: Well, when including from the script where those vars are defined, this would be redundant.

PR #3521, opened 2015-04-05
[ticket/13745] Add viewtopic.php core event to allow manipulating poll data

rxu:
<a href="https://tracker.phpbb.com/browse/PHPBB3-13745">PHPBB3-13745</a>.

### DIFF
--- a/phpBB/viewtopic.php
+++ b/phpBB/viewtopic.php
@@ -804,6 +804,36 @@ if (!empty($topic_data['poll_start']))
 		($auth->acl_get('f_votechg', $forum_id) && $topic_data['poll_vote_change']))) ? true : false;
 	$s_display_results = (!$s_can_vote || ($s_can_vote && sizeof($cur_voted_id)) || $view == 'viewpoll') ? true : false;
 
+	/**
+	* Event to manipulate the poll data
+	*
+	* @event core.viewtopic_modify_poll_data
+	* @var	array	cur_voted_id				Array with options' IDs current user has voted for
+	* @var	int		forum_id					The topic's forum id
+	* @var	array	poll_info					Array with the poll information
+	* @var	bool	s_can_vote					Flag indicating if a user can vote
+	* @var	bool	s_display_results			Flag indicating if results or poll options should be displayed
+	* @var	int		topic_id					The id of the topic the user tries to access
+	* @var	array	topic_data					All the information from the topic and forum tables for this topic
+	* @var	string	viewtopic_url				URL to the topic page
+	* @var	array	vote_counts					Array with the vote counts for every poll option
+	* @var	array	voted_id					Array with updated options' IDs current user is voting for
+	* @since 3.1.5-RC1
+	*/
+	$vars = array(
+		'cur_voted_id',
+		'forum_id',
+		'poll_info',
+		's_can_vote',
+		's_display_results',
+		'topic_id',
+		'topic_data',
+		'viewtopic_url',
+		'vote_counts',
+		'voted_id',
+	);
+	extract($phpbb_dispatcher->trigger_event('core.viewtopic_modify_poll_data', compact($vars)));
+
 	if ($update && $s_can_vote)
 	{
 
@@ -937,6 +967,7 @@ if (!empty($topic_data['poll_start']))
 
 	$topic_data['poll_title'] = generate_text_for_display($topic_data['poll_title'], $poll_info[0]['bbcode_uid'], $poll_info[0]['bbcode_bitfield'], $parse_flags, true);
 
+	$poll_template_data = $poll_options_template_data = array();
 	foreach ($poll_info as $poll_option)
 	{
 		$option_pct = ($poll_total > 0) ? $poll_option['poll_option_total'] / $poll_total : 0;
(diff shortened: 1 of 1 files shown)


marc1706:
Typo in your commit message. It's viewtopic ;)

rxu:
:P Will fix after the tests finished.

javiexin:
Missing variable $s_display_results from the event interface.
$poll_options_template_data variable should be unset.
Also, the code in viewtopic that fills in the (poll) template variables could be also exposed in the event, see https://www.phpbb.com/community/viewtopic.php?p=14032241#p14032241

rxu:
The event should be ready for merge now.

javiexin:
With the latest change, the variables s_display_results and s_can_vote are now redundant in the event interface, as they are part of the poll_template_data array.  They can/should be removed.

rxu:
I think we can go with it. Some redundant vars don't hurt the event ;)

javiexin:
The problem with redundant vars is the fact that some extension might change one and not the other instance, and then a second extension using the same event might take the value from the unmodified one. That would create unreliable behaviour.  But yes, this would only matter when multiple extensions use the same event, and use/modify the two redundant values differently, unlikely.  So we can go with it if you think it is best.

rxu:
Well, that would be just a bug in extension which we can't really avoid by any event structure. I think that trying to workaround it here would be an overkill, so let's just leave it to the extension authors responsibility.

gn36:
Leaving the redundancy in there in will probably cause headaches for extension developers, though, because one would think that an option listed explicitly within an event will take preceedence when changed instead of changing a value within an array. I would suggest at least adding some sort of a "read only" note to the description, since that is the only thing those vars are good for.

rxu:
Currently I don't see the correct way to workaround this thing. taking those vars out of the event will make ext developers having to recalculate those and they still can assign this recalculated result to the template vars the same way. As I said above, that is actually the ext developers responsibility to create bug-free scripts.
Everything I can think of for now is adding a comment to the docblock stating that if you change the ```s_display_results``` and ```s_can_vote``` values you also have to change the related template vars values accordingly.

javiexin:
Recalculating those variables to get the same result as now (with the vars in the event) is as difficult as $s_can_vote=$poll_template_data['S_CAN_VOTE']; $s_display_results=$poll_template_data['S_DISPLAY_RESULTS']; because if they really need a different value, they will have to recalculate anyway.  Having to assign to both is just cumbersome, and asking for errors in my opinion. They only possible reason to keep them is what gn36 mentioned, to have a known-to-be-unchanged (by other extensions) value, and clearly mark them as read-only.  Personally, I would remove them: no benefit, multiple possible issues.

rxu:
This sounds pretty senseless, sorry. If so, we have to remove any vars assigned to the template to avoid same things, as you can also change any value of the ```$topic_data``` array, ```$poll_total```, ```$viewtopic_url``` and ```$poll_info``` array which values are previously assigned to the ```$poll_options_template_data```. So why do we have to consider ```$s_can_vote``` etc as something unique?

rxu:
Abstracting from this particular vars, the event can be actually splitted into 2 events like ```_before``` event, then assigning the template arrays, then ```_after``` event, then assigning arrays to the template.

javiexin:
Looked at it again. $poll_total and $s_display_results are NOT used anywhere after the event, so keeping them would just be error prone.  However, $s_can_vote and $viewtopic_url are used after the event, so modifying them will have consequences down the (execution) line.  So, I think that those NEED to be exposed, as you might want to have the effect prolong over the execution of the program.  But those that are not used any longer after the event are pretty much useless.  Of course, arrays like $topic_data or $poll_info include a lot of information that is not available through template variables (in an unmodified form), so those should be kept as well.
For me, the rule of thumb is if it may impact future execution, then it should stay; if it includes unique information/data, it should stay; otherwise, it is just redundant, and may be removed without impact.

javiexin:
No, I would not split in "before" and "after" here.  I think there is no need for that.

Maybe, there is a need for another event within the "ajax" execution block, around line 901, to be able to modify execution before ajax actually refreshes the page.  But I guess this would be a different request.

rxu:
Having 2 events would allow to split the 'main' variables from the template ones to avoid collisions of non-synced vars/template values.

javiexin:
Ok, if we go for the split, I would do it spreading the points more that immediately before and immediately after.  I would keep the "after" event where it is now, and would add the "before" event around line 806.

rxu:
For the current PR code, L806 seems to be too early to place event, I'd go with https://github.com/rxu/phpbb3/blob/ticket/13745/phpBB/viewtopic.php#L939.

javiexin:
In my opinion, that would be useless.  In case the event was place there, I would consider it mostly redundant. Everything that can be done in this potential new event could be done in the already existing one. If the position is there, I would not clutter the code more.

If you want a "before" event, make it so you can actually modify some behaviour, like the "updating" block. What I may agree with you is that it may not belong in this request, but in a different one.

rxu:
Well, adding another event is ok within this PR I think, as it's all about manipulating poll data :)

rxu:
!set WIP

gn36:
I would agree with L806 for another event. In that location, the modification of $s_can_vote makes actually sense because it will affect both the display and the update process itself.

rxu:
Updated #3521.

PR #3526, opened 2015-04-07
[ticket/13750] Add core event to allow modifying navlinks text

rxu:
Add generate_forum_nav() core event in includes/functions_display.php
to allow modifying navlinks text. Use case example: extension(s) implementing
multilanguage boards including translated forum names and so on.

<a href="https://tracker.phpbb.com/browse/PHPBB3-13750">PHPBB3-13750</a>.


rxu:
!set Event

PR #3527, opened 2015-04-07
[ticket/13709] Add fallback for email templates used by extensions

Elsensee:
Also add a fallback to english for the core like we do it in `$user->add_lang()`

[PHPBB3-13709](https://tracker.phpbb.com/browse/PHPBB3-13709)

### DIFF
--- a/phpBB/includes/functions_messenger.php
+++ b/phpBB/includes/functions_messenger.php
@@ -231,6 +231,13 @@ class messenger
 			$template_lang = basename($config['default_lang']);
 		}
 
+		$ext_template_paths = array(
+			array(
+				'name' 		=> $template_lang . '_email',
+				'ext_path' 	=> 'language/' . $template_lang . '/email' . $template_dir_prefix,
+			),
+		);
+
 		if ($template_path)
 		{
 			$template_paths = array(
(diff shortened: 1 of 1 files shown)


Elsensee:
Wanted to write tests for this by setting the email template in different languages and then checking for the right path to template with the function `get_filename_for_handle()` of the template class.

But the $template object of the messenger class is protected so I cannot access it.

Any other ideas on how to write a test for this? :)

dragosvr92:
I followed the instructions on my 3.1.3 board and i switched language to one secondary to English. I still got a blank page when i tried to use it, as i dont have its language files for all the installed extensions. :(

dragosvr92:
Nevermind.... This is only for email templates used by extensions, not html templates also :(
https://www.phpbb.com/community/viewtopic.php?f=466&t=2306361

Elsensee:
Correct, but what you're describing *should* already work. => https://github.com/phpbb/phpbb/blob/526329e6972914ac6d62c092da410e349d96d808/phpBB/phpbb/user.php#L656-L684
But you can of course create a ticket in our bugtracker and we will look into it. :)

VSEphpbb:
Template files (except for emails) DO properly fall back if extensions do not have translations...either to the board default if supported by the ext, or to english as a last resort.

dragosvr92:
Well, they do not seem to fall back for me. And seeing that you are wrong in claiming MODs also used to  fall back to en. I tend to think that is not for ext either. But we seem to be off-topic here, so i think we should speak on the phpbb.com link i left above :)

VSEphpbb:
@dragosvr92 As far as extensions go, English fallback is fully supported except email templates which this PR fixes. As for MODs, they fell back too except for MOD's info_acp lang files, which showed up as raw lang keys (and could break an ACP page if it was the first module, typically in the .MODs page)

Nicofuma:
:+1:

Elsensee:
Turns out my approach doesn't work, so I'm back at the point where I have no idea how to write tests for this. If anyone has an idea, please share it.
Otherwise go and merge this. ^^

marc1706:
Have you tried extending the class with a new test class and overwriting unneeded methods with basically more or less blank methods?

Nicofuma:
@Elsensee you need to rebase your branch. After that we could talk about "how to write" the test (don't hesitate to ask on IRC when you have time to work on that)

RMcGirr83:
bump @Elsensee

Elsensee:
Gotcha.

rmcgirr83:
bump @Elsensee again

Elsensee:
Rebased.

And in all honesty: I'd rather write tests for this if we refactor this class instead of now.
Writing tests for new (and/or refactored) code is pretty much the only option I see to have (almost) full code coverage regarding tests.
If you don't think so, I'll sit down and try to write a test for this method, but it'll really be a lot of hacking.

marc1706:
Is there already a ticket that would result in a refactoring of this file? If not, please add it.

PR #3530, opened 2015-04-09
[ticket/13753] Add template events to forum category row headers

VSEphpbb:
https://tracker.phpbb.com/browse/PHPBB3-13753

PHPBB3-13753

I did not include subsilver in this PR because it's design does not seem to lend itself well to this event, partly because it's tabular but mostly because unlike in prosilver, this category/header section of subsilver is broken into three different locations.


VSEphpbb:
!set Event

PR #3534, opened 2015-04-13
[ticket/13761] Use __DIR__ for phpbb_root_path

Wolfsblvt:
Adding `__DIR__` to `$phpbb_root_path`, so that the correct folder for
includes
is chosen when the test suites are executed in different environments.

PHPBB3-13761

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -12,8 +12,15 @@
 */
 
 define('IN_PHPBB', true);
-$phpbb_root_path = __DIR__ . '/../phpBB/';
+$phpbb_root_path = 'phpBB/';
 $phpEx = 'php';
+
+// We need to check if we got problems with relative paths
+if (!file_exists($phpbb_root_path . 'includes/startup.php'))
+{
+	$phpbb_root_path = __DIR__ . '/../phpBB/';
+}
+
 require_once $phpbb_root_path . 'includes/startup.php';
 
 $table_prefix = 'phpbb_';
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -12,7 +12,7 @@
 */
 
 define('IN_PHPBB', true);
-$phpbb_root_path = 'phpBB/';
+$phpbb_root_path = __DIR__ . '/../phpBB/';
 $phpEx = 'php';
 require_once $phpbb_root_path . 'includes/startup.php';
 


VSEphpbb:
This could break a lot of tests, especially for extensions. What was wrong with the current way of getting to includes folders?

VSEphpbb:
Maybe not, I could be wrong... ignore me :cry: 

but it does break some of phpBB's tests

Wolfsblvt:
Ah, I see. All tests are red :-/
Wrong with the current way is "phpBB/" as a totally relative path.
On my environment that does not work for testing, cause the files to include will be searched in the wrong folder.

And yeah, it breaks a LOT of tests. Also I have seen now that something is string. I am using __DIR__ there a different way like on the bottom with
```php
require_once __DIR__ . '/vendor/autoload.php';
```
So I think this __DIR__ pretens we are in phpbb root. How can that be? bootstrap.php is inside of \tests

Elsensee:
> So I think this `__DIR__` pretens we are in phpbb root.

That is not correct. Travis is set up to download facebook/webdriver via composer which then creates that vendor/autoload.php file inside the tests folder.
`__DIR__` will always return the same result. ;) (Well.. in that file at least)

Wolfsblvt:
Okay, please wait a bit before merge, I will check some things tomorrow first.

Wolfsblvt:
Nah, I think this is all we can do.
I don't think it will screw anything up, cause it won't be executed on travis or for most others.

VSEphpbb:
@Wolfsblvt I still don't see why we need this. I thought you made this to deal with an issue in your IDE, which you later solved because of a set up issue you had with your IDE and phpbb.

Wolfsblvt:
No, I haven't solved that. When I change this back, no phpunit test is working, not even extension tests.

This **is** for unit testing issues with all IDE wich have a different execution path than phpBB root.

nickvergessen:
I'm not sure, do you mind sharing your directory structure?

testing works quite nice and okay here:
```
$ phpBB/vendor/bin/phpunit --configuration phpBB/ext/$EXTNAME/travis/phpunit-$DB-travis.xml --bootstrap ./tests/bootstrap.php
```

Wolfsblvt:
Wich directory structure?
I am using a complete clone of phpbb repository, So I have the same structure.

The problem seems to be with the phpunit you use. When I run your command, all is working (as far as I remember).
But my IDE comes with its own PHPUnit integrated, so it uses that to run the tests. In this case phpunit is in another place, so we need to make sure that files are searched relative to phpbb, wich is done with that __DIR__

bantu:
I think this PR should be rejected.

Right now `$phpbb_root_path` needs to be relative, because it not only used for includes, but also for URL generation. As such, you just can not just change `$phpbb_root_path` to be absolute, which is what using `__DIR__` does.

You may change individual includes in tests to be absolute by adding `__DIR__`, however, the phpBB includes will remain relative, thus likely not solving your problem.

Bottom line: Just configure your IDE to use the correct PHPUnit.

marc1706:
I agree with bantu.

Wolfsblvt:
Understand. I am okay with that.

PR #3535, opened 2015-04-14
[ticket/13762] Moving language related functionality into a service

MateBartus:
[PHPBB3-13762](https://tracker.phpbb.com/browse/PHPBB3-13762)


MateBartus:
!set WIP

MateBartus:
!set WIP

Nicofuma:
Replacing `$user->lang[] with $user->lang()` is definitively something to do one day but not today I think (to much trouble).

Anyway I'd like to keep the BC with the old `$user->lang['key']` way. For that you can add a special treatment in the magic method `__get()`

MateBartus:
Just reverted that commit while you were typing that message. BC definitely should be kept and my RegEx replace turned out pretty badly anyhow.

phpBB/config/default/container/services.yml
```diff
@@ -116,6 +116,12 @@ services:
             - @controller.resolver
             - @request_stack
 
+    language:
```
MateBartus: This service part should be rewritten, but I'm not sure how exactly.

Probably this part should be renamed to language.parent, made abstract, and then language service could link to it, and we could specify passing extension manager into the setter, but that should be environment dependent, so  I'm not sure where it should be specified.

Nicofuma: I think that we should split it into at least two services:
- language_loader which will find the language files and load them when requested (basically it's a stateless service (minus caching) which as a `load(<component>, <locale>) : array('key' => 'value')` method or something similar).
- language which will take a loader as parameter. This service will be used by everyone.

Skouat:
Hi,
The filename of `phpBB/phpbb/language/loader/extenension_language_file_loader.php` should be `phpBB/phpbb/language/loader/extension_language_file_loader.php`

phpBB/phpbb/language/language_file_helper.php
```diff
@@ -0,0 +1,71 @@
...
+
+	/**
+	 * Returns available languages
+	 *
+	 * @return array
+	 */
+	public function get_available_languages()
+	{
+		// Find available language packages
+		$finder = new Finder();
```
nickvergessen: any reason why you don't use `phpbb\finder` ?

phpBB/phpbb/language/loader/core_language_file_loader.php
```diff
@@ -0,0 +1,82 @@
...
+			$this->locale = $locale;
+		}
+
+		// Determine path to language directory
+		$path = ($this->lang_dir !== false) ? $this->lang_dir : $this->phpbb_root_path . 'language/';
+
+		// Determine filename
+		if ($use_help && strpos($component, '/') !== false)
+		{
+			$filename = dirname($component) . '/help_' . basename($component);
```
nickvergessen: if we change all this, I think it's time to drop the help_ stuff.
I have a PR locally which turns the faq page to a normal language var based feature, so we don't need this hack anymore ;)

phpBB/phpbb/language/language_file_helper.php
MateBartus: `phpbb\finder` seems to lack the functionality I use here, also caching is not an issue here.

phpBB/phpbb/language/loader/core_language_file_loader.php
MateBartus: Sounds good!

phpBB/phpbb/language/language.php
```diff
@@ -0,0 +1,590 @@
...
+	}
+
+	/**
+	 * Function to set custom language path (able to use directory outside of phpBB)
+	 *
+	 * @param string $lang_path New language path used.
+	 *
+	 * @return null
+	 */
+	public function set_custom_lang_path($lang_path)
```
MateBartus: Does anyone know why this feature was implemented (i think it was introduced in 3.0.3). These path change only changes the loading path (for core files) and it is not used at language install or any other places really.

Nicofuma: no idea, but I think that we could nuke it.

phpBB/config/default/container/services.yml
Nicofuma: by the way, move this service definition to services_language.yml please

phpBB/config/default/container/services_language.yml
```diff
@@ -0,0 +1,26 @@
...
+        parent: language.loader.file_loader_base
+        arguments:
+            - %core.root_path%
+
+    language.loader.extension_file_loader:
+        class: phpbb\language\loader\extension_language_file_loader
+        scope: prototype
+        parent: language.loader.file_loader_base
+        arguments:
+            - @ext.manager
```
Nicofuma: maybe add and use a tag for the language loader?

phpBB/phpbb/language/exception/invalid_plural_rule_exception.php
```diff
@@ -0,0 +1,22 @@
...
+ * the docs/CREDITS.txt file.
+ *
+ */
+
+namespace phpbb\language\exception;
+
+/**
+ * Exception thrown by \phpbb\language\language::get_plural_form()
+ */
+class invalid_plural_rule_exception extends \phpbb\exception\runtime_exception
```
Nicofuma: would it make sense to have a base language_exception class?

phpBB/phpbb/language/language.php
```diff
@@ -0,0 +1,590 @@
...
+		$this->language_root_path	= null;
+		$this->lang					= array();
+		$this->loaded_language_sets	= array(
+			'core'	=> array(),
+			'ext'	=> array(),
+		);
+
+		// Common language files
+		$this->common_language_files = array(
+			'common',
```
Nicofuma: I think it could be nice to be able to configure it through an optional parameter in the DI

phpBB/phpbb/language/exception/invalid_plural_rule_exception.php
MateBartus: as of now, I don't really any good reason why we would need that

phpBB/phpbb/language/loader/extension_language_file_loader.php
```diff
@@ -0,0 +1,77 @@
...
+	 */
+	public function load($component, $locale, &$lang, &$help)
+	{
+		if (!is_array($locale))
+		{
+			$this->locale = array($locale);
+		}
+		else
+		{
+			$this->locale = $locale;
```
Nicofuma: we pass the locale via the parameters, so lets remain stateless here and pass them to the parent call aswell

phpBB/phpbb/user.php
```diff
@@ -540,42 +450,16 @@ function get_plural_form($number, $force_rule = false)
...
-			}
-			unset($lang_set);
-		}
-		else if ($lang_set)
-		{
-			$this->set_lang($this->lang, $this->help, $lang_set, $use_db, $use_help, $ext_name);
-		}
+		// Load lang and help for BC purposes
+		$this->lang = $this->language->get_lang_array();
+		$this->help = $this->language->get_help_array();
```
Nicofuma: I think we should remove that and use the `__get()` magic method instead

phpBB/config/production/container/services.yml
```diff
@@ -1,2 +1,8 @@
 imports:
     - { resource: ../../default/container/services.yml }
+
+services:
```
MateBartus: Reminder: Move this to a separate file, and import that instead.

Although if anyone has a better idea for handling this, that is certainly welcomed.

phpBB/phpbb/language/exception/invalid_plural_rule_exception.php
Nicofuma: (a base class or an interface) the idea being to be able to catch every language related exceotion and not the others.

phpBB/config/production/container/services.yml
Nicofuma: why you can't just put in default/services_language.yml ?

MateBartus: I need the language service in the installer, and probably I cannot have user there (I'm not sure but probably something has dependency of the from the user dependencies that I cannot have there)

phpBB/config/default/container/services_language.yml
```diff
@@ -0,0 +1,18 @@
...
+        arguments:
+            - %core.root_path%
+
+    language:
+        class: phpbb\language\language
+        arguments:
+            - @language.loader
+
+# -------- Language loader services --------
+    language.loader_parent:
```
Nicofuma: language.abstract_loader

phpBB/config/default/container/services_language.yml
```diff
@@ -0,0 +1,18 @@
...
+        abstract: true
+        class: phpbb\language\language_file_loader
```
Nicofuma: phpbb\language\abstract_file_loader

phpBB/phpbb/language/language_file_loader.php
```diff
@@ -0,0 +1,223 @@
...
+{
+	/**
+	 * Global fallback language
+	 *
+	 * ISO code of the language to fallback to when the specified language entries
+	 * cannot be found.
+	 *
+	 * @var string
+	 */
+	const FALLBACK_LANGUAGE = 'en';
```
Nicofuma: should be moved to language/language and handled by this class/service

phpBB/phpbb/user.php
```diff
@@ -512,24 +429,20 @@ function lang()
 	* @param $number        int|float   The number we want to get the plural case for. Float numbers are floored.
 	* @param $force_rule    mixed   False to use the plural rule of the language package
 	*                               or an integer to force a certain plural rule
-	* @return   int     The plural-case we need to use for the number plural-rule combination
+	* @return int|bool     The plural-case we need to use for the number plural-rule combination, false if $force_rule
+	* 					   was invalid.
 	*/
 	function get_plural_form($number, $force_rule = false)
```
Nicofuma: should be deprecated

Nicofuma:
https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/install/index.php#L243 needs to be updated

Nicofuma:
https://github.com/phpbb/phpbb/blob/3.1.x/tests/template/template_test_case.php#L124 you need to update the loader to set the variables

Nicofuma:
same here: https://github.com/phpbb/phpbb/blob/3.1.x/tests/user/lang_test.php#L21

Nicofuma:
please rebase

Nicofuma:
merged #3539 you can rebase

MateBartus:
!unset WIP

phpBB/phpbb/language/exception/invalid_plural_rule_exception.php
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * For full copyright and license information, please see
+ * the docs/CREDITS.txt file.
+ *
+ */
+
+namespace phpbb\language\exception;
+
+/**
+ * Exception thrown by \phpbb\language\language::get_plural_form()
```
Nicofuma: explain what mean this exception and not which function throw it

phpBB/phpbb/language/language.php
```diff
@@ -0,0 +1,571 @@
...
+
+	/**
+	 * @var \phpbb\language\language_file_loader Language file loader
+	 */
+	protected $loader;
+
+	/**
+	 * @var string	ISO code of the User's language
+	 */
+	protected $user_language;
```
Nicofuma: move it just after default_language please

phpBB/phpbb/language/language.php
```diff
@@ -0,0 +1,571 @@
...
+	 * <code>
+	 * $component = array('posting');
+	 * $component = array('posting', 'viewtopic')
+	 * $component = 'posting'
+	 * </code>
+	 *
+	 * @param string|array	$component		The name of the language component to load
+	 * @param string|null	$extension_name	Name of the extension to load component from, or null for core file
+	 *
+	 * @return null
```
Nicofuma: no need for this line

phpBB/phpbb/language/language.php
```diff
@@ -0,0 +1,571 @@
...
+
+		$this->set_fallback_array();
+	}
+
+	/**
+	 * Function to set the board's default language to display.
+	 *
+	 * @param string	$default_lang_iso	ISO code of the board's default language
+	 *
+	 * @return null
```
Nicofuma: no need for this line

phpBB/phpbb/language/language.php
```diff
@@ -0,0 +1,571 @@
...
+
+		$this->language_fallback = array(self::FALLBACK_LANGUAGE);
+	}
+
+	/**
+	 * Function to set user's language to display.
+	 *
+	 * @param string	$user_lang_iso	ISO code of the User's language
+	 *
+	 * @return null
```
Nicofuma: no need for this line

phpBB/phpbb/language/language.php
```diff
@@ -0,0 +1,571 @@
...
+	}
+
+	/**
+	 * Loads common language files
+	 *
+	 * @return null
+	 */
+	protected function load_common_language_files()
+	{
+		foreach ($this->common_language_files as $lang_file)
```
Nicofuma: wrap it under `if(!$this->common_language_files_loaded )`

phpBB/phpbb/language/language.php
```diff
@@ -0,0 +1,571 @@
...
+	 * Load core language file
+	 *
+	 * @param string	$component	Name of the component to load
+	 *
+	 * @return null
+	 */
+	protected function load_core_file($component)
+	{
+		// Check if the component is already loaded
+		if (in_array($component, $this->loaded_language_sets['core']))
```
Nicofuma: I prefer to store the information as key value: `if (isset($this->loaded_language_sets['core'][$component])` (this way it could be also possible to store the state of the component (not found or invalid file per example))

phpBB/phpbb/language/language.php
```diff
@@ -0,0 +1,571 @@
...
+		{
+			return;
+		}
+
+		$this->loader->load($component, $this->language_fallback, $this->lang);
+		$this->loaded_language_sets['core'][] = $component;
```
Nicofuma: so if we do what I said above: $this->loaded_language_sets['core'][$component] = true;

phpBB/phpbb/language/language.php
```diff
@@ -0,0 +1,571 @@
...
+	 *
+	 * @param string	$extension_name	Name of the extension to load language from
+	 * @param string	$component		Name of the component to load
+	 *
+	 * @return null
+	 */
+	protected function load_extension($extension_name, $component)
+	{
+		// Check if the component is already loaded
+		if (in_array(array($extension_name, $component), $this->loaded_language_sets['ext']))
```
Nicofuma: same here

phpBB/phpbb/language/language_file_helper.php
Nicofuma: and you want only the core language files, right?

MateBartus: Mostly. This service used at two places, the installer, and the ACP language install module, it was implemented both places so this is the deduplication of it. And the assumption behind my code (and probably the old one as well) that a language is only installable when the core files are there, therefore I don't really care for extension language packs here.

phpBB/phpbb/language/language_file_loader.php
```diff
@@ -0,0 +1,226 @@
...
+	 * @param string|array	$locale		ISO code of the language to load, or array of ISO codes if you want to
+	 * 									specify additional language fallback steps
+	 * @param array			$lang		Array reference containing language strings
+	 *
+	 * @return null
+	 */
+	public function load_extension($extension, $component, $locale, &$lang)
+	{
+		// Check if extension manager was loaded
+		if (is_null($this->extension_manager))
```
Nicofuma: === null

phpBB/phpbb/language/language_file_loader.php
```diff
@@ -0,0 +1,226 @@
...
+	/**
+	 * Loads language array for the given extension component
+	 *
+	 * @param string		$extension	Name of the extension
+	 * @param string		$component	Name of the language component
+	 * @param string|array	$locale		ISO code of the language to load, or array of ISO codes if you want to
+	 * 									specify additional language fallback steps
+	 * @param array			$lang		Array reference containing language strings
+	 *
+	 * @return null
```
Nicofuma: useless

phpBB/phpbb/language/language_file_loader.php
```diff
@@ -0,0 +1,226 @@
...
+	public function load_extension($extension, $component, $locale, &$lang)
+	{
+		// Check if extension manager was loaded
+		if (is_null($this->extension_manager))
+		{
+			// If not, let's return
+			return;
+		}
+
+		if (!is_array($locale))
```
Nicofuma: $locale = (array) $locale;

phpBB/phpbb/language/language_file_helper.php
Nicofuma: fine with me

phpBB/phpbb/language/exception/language_file_not_exists.php
```diff
@@ -0,0 +1,22 @@
...
+ * the docs/CREDITS.txt file.
+ *
+ */
+
+namespace phpbb\language\exception;
+
+/**
+ * This exception is thrown when the language file is not found
+ */
+class language_file_not_exists extends language_exception
```
Nicofuma: language_file_not_found

phpBB/phpbb/language/language.php
MateBartus: fixed (github doesn't show it as an outdated diff for some reason, at least not at my end)

phpBB/phpbb/language/language_file_loader.php
MateBartus: fixed (github doesn't show it as an outdated diff for some reason, at least not at my end)

Nicofuma: didn't you forget to commit or push?

phpBB/phpbb/language/language.php
Nicofuma: I guess it's because it's exactly the same line

MateBartus: moved from 71 to 51...

phpBB/phpbb/language/language_file_helper.php
Nicofuma: to be precise, each time we want to limit the depth we write the iterator chain manually

MateBartus: Is this an issue then?

Nicofuma: with our finder yes, but not with your PR :)

phpBB/includes/functions_compatibility.php
```diff
@@ -482,3 +482,22 @@ function phpbb_realpath($path)
...
+
+/**
+ * Determine which plural form we should use.
+ * For some languages this is not as simple as for English.
+ *
+ * @param $rule		int			ID of the plural rule we want to use, see http://wiki.phpbb.com/Plural_Rules#Plural_Rules
+ * @param $number	int|float	The number we want to get the plural case for. Float numbers are floored.
+ * @return	int		The plural-case we need to use for the number plural-rule combination
+ *
+ * @deprecated 3.2.0-dev (To be removed: 3.3.0)
```
MGaetan89: Shouldn't it be "To be removed: 3.4.0"?

MateBartus: I'm not even sure if anyone uses this function (as it was called by user::lang()). So 3.3.0 is more then too much time to adopt.

PR #3536, opened 2015-04-14
[ticket/12230] Make new users post limit retroactive

Elsensee:
I could imagine to do this the other way around as well, so users who weren't in the group will go into it (again) when the administrator sets the limit up.. (again)

I'm just not sure *how* because if I just do this for all users the administrator could get into that group too. We could exclude him with `user_type <> USER_FOUNDER` but what if there are other administrators? I don't think there is a simple way we could get all administrators with some SQL query to exclude them from this.

### DIFF
--- a/phpBB/includes/acp/acp_board.php
+++ b/phpBB/includes/acp/acp_board.php
@@ -520,6 +520,12 @@ class acp_board
 				continue;
 			}
 
+			if ($config_name == 'new_member_post_limit' && $submit)
+			{
+				// Now apply the new configuration to all current users
+				$this->set_new_post_limit($config[$config_name], $cfg_array[$config_name]);
+			}
+
 			$this->new_config[$config_name] = $config_value = $cfg_array[$config_name];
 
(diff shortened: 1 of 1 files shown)


rxu:
Actually if the newly registered membership limit is lowered/eliminated users will be removed from the related usergroup as soon as they visit the board. Is it really needed to be fixed this way?

gn36:
I would not add new users to the group if the setting is changed. This might cause a lot of administrators to accidentally restrict existing users, which, from my experience, will only cause trouble. 

The only benefit I would see from removing the users immediately from the group is that some random mod/extension might depend on the question whether a user is in the newly registered users group to determine whether or not to send them a notification or an E-Mail. I don't know such a mod, though.

Elsensee:
@rxu Right, but they are only removed from the group if the setting is not 0, because `empty(0)` returns true... I missed that too, so we should rather fix that if condition.

Elsensee:
Replaced by #3553.

PR #3531, opened 2015-04-09
Take into account $data['post_time'] in submit_post

mcallegari:
Without this, the $data['post_time'] mentioned in the Function.submit_post documentation is pretty useless.

This is fundamental when migrating non mySQL-based forums (e.g. Sourceforge) with posts back in time

PR #3538, opened 2015-04-16
[ticket/13759] Take post_time into account

RMcGirr83:
when using the submit_post function

PHPBB3-13759

### DIFF
--- a/phpBB/includes/functions_posting.php
+++ b/phpBB/includes/functions_posting.php
@@ -1542,7 +1542,14 @@ function submit_post($mode, $subject, $username, $topic_type, &$poll, &$data, $u
 		return false;
 	}
 
-	$current_time = time();
+	if (!empty($data['post_time']))
+	{
+		$current_time = $data['post_time'];
+	}
+	else
+	{
+		$current_time = time();
+	}
 
 	if ($mode == 'post')
 	{


RMcGirr83:
[PHPBB3-13759] (https://tracker.phpbb.com/browse/PHPBB3-13759)

https://github.com/phpbb/phpbb/pull/3531

Nicofuma:
:+1:

RMcGirr83:
:point_up:

nickvergessen:
Regression: https://area51.phpbb.com/phpBB/viewtopic.php?f=99&t=48056 please revert or fix

rxu:
Probably can be fixed just by excluding the ```quote``` mode
```php
if ($mode != 'quote' && !empty($data['post_time']))
```

nickvergessen:
and edit, and ... ?

nickvergessen:
Or just dont send in post_time to submit_post in our calls

RMcGirr83:
Editing passes the posts time as well.  Not sure about all the other modes though.  Reply doesn't seem to.  It's a bit odd that there is this line in posting.php prior to the submit_post function

`'post_time'				=> (isset($post_data['post_time'])) ? (int) $post_data['post_time'] : $current_time,`

but then $current_time is used as default in submit_post function and I can't see where $data['post_time'] is ever used.

RMcGirr83:
Any comment from anyone?

nickvergessen:
Well I guess if you make sure this is really not the case, you can remove it, so it is not injected by default?

RMcGirr83:
I tested pretty extensively if $data doesn't have post_time within it and ran across no issues.  So should I amend this ticket or start a new one (will require new ticket in tracker?).

Would you prefer the code be commented out or completely removed?

marc1706:
I'd recommend you remove the line if it's really not needed and properly explain why this was done in the commit message.

RMcGirr83:
https://github.com/phpbb/phpbb/pull/3734

PR #3541, opened 2015-04-19
[ticket/13767] Remove vendor/s9e/text-formatter/.git during build

s9e:
PHPBB3-13767

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -319,6 +319,8 @@
 		<delete file="${dir}/vendor/psr/log/.gitignore" />
 		<delete file="${dir}/vendor/psr/log/README.md" />
 
+		<delete dir="${dir}/vendor/s9e/text-formatter/.git" />
+
 		<delete dir="${dir}/vendor/symfony/config/Symfony/Component/Config/Tests" />
 		<delete dir="${dir}/vendor/symfony/config/Symfony/Component/Config/.git" />
 		<delete file="${dir}/vendor/symfony/config/Symfony/Component/Config/.gitignore" />


marc1706:
Couldn't you just add that to your .gitattributes for the text-formatter using export-ignore?

s9e:
I don't know anything about export-ignore. build.xml has a bunch of `<delete dir="${dir}/vendor/*/.git" />` commands, I'm just going with the flow.

marc1706:
I'm talking about something like this: https://github.com/marc1706/fast-image-size/blob/master/.gitattributes

That'll prevent hardcoded commands for cleaning the vendor directory.

s9e:
I don't know how that's supposed to work. The fact that there are several commands that delete .git directories leads me to believe that it doesn't.

bantu:
Considering that there are already delete statements for other .git directories, this is fine. We should rearrange this so all .git directories are removed before packaging.

bantu:
unfortunately, phing package fails with a segfault in php right now. not caused by this patch, though. https://gist.github.com/bantu/28ac9251ccf1c908b053

s9e:
@bantu I looked at your segfault out of curiosity and I know *exactly* what line of code triggers it: [this one](https://github.com/s9e/TextFormatter/blob/693fe939bd4e0e0a5b9020af5c8ace7b8606b5bc/src/Bundles/MediaPack.php#L29). Apparently PHP segfaults if a string has too many backslashes. [Here's a similar case from Symfony's issue tracker](https://github.com/symfony/symfony/issues/9462).

I think that https://github.com/s9e/TextFormatter/commit/559f3c4def03e5fbbf5301198ccd69c91d902659 will fix it but I'm waiting on Travis before sending a PR.

brunoais:
This reminds me a sentence somewhat like this:

> Those high level languages memory management... Even though they try to hide it hard that memory doesn't exist for the developer, the developer still needs to think about the memory behind. Worst of all, if they fail at memory management, there's nothing the programmer can do to solve it. He always has to solve the problem differently, if he wants the problem solved.

PR #3542, opened 2015-04-19
[ticket/10572] Unguarded includes in acp files

RMcGirr83:
PHPBB3-10572

### DIFF
--- a/phpBB/includes/acp/acp_attachments.php
+++ b/phpBB/includes/acp/acp_attachments.php
@@ -106,7 +106,10 @@ class acp_attachments
 		{
 			case 'attach':
 
-				include_once($phpbb_root_path . 'includes/functions_posting.' . $phpEx);
+				if (!function_exists('get_supported_image_types'))
+				{
+					include($phpbb_root_path . 'includes/functions_posting.' . $phpEx);
+				}
 
 				$sql = 'SELECT group_name, cat_id
 					FROM ' . EXTENSION_GROUPS_TABLE . '
--- a/phpBB/includes/acp/acp_ban.php
+++ b/phpBB/includes/acp/acp_ban.php
@@ -28,7 +28,10 @@ class acp_ban
 		global $user, $template, $request, $phpbb_dispatcher;
 		global $phpbb_root_path, $phpEx;
 
-		include($phpbb_root_path . 'includes/functions_user.' . $phpEx);
+		if (!function_exists('user_ban'))
+		{
+			include($phpbb_root_path . 'includes/functions_user.' . $phpEx);
+		}
 
 		$bansubmit	= $request->is_set_post('bansubmit');
 		$unbansubmit = $request->is_set_post('unbansubmit');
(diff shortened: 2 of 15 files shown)


RMcGirr83:
[PHPBB3-10572] (https://tracker.phpbb.com/browse/PHPBB3-10572)

callumacrae:
Why not use `include_once` instead?

If you see unnecessary includes, I'd say remove them.

RMcGirr83:
@callumacrae I was under the impression that include is faster than include_once

callumacrae:
Because it keeps track of the files that have been included; which is sort of the point.

Also, premature optimization blablabla

RMcGirr83:
That makes no sense

callumacrae:
No, you just don't understand it :)

Which bit doesn't make sense?

RMcGirr83:
Here's an idea if there is a more preferred way then state so and require
it. Contributors shouldn't have to "guess" what may or may not be the
preferred way.

Make it easier on contributors, not more convoluted than it needs to be.

No one else seems to have an issue with the edits.

cs278:
Some background, https://www.phpbb.com/community/viewtopic.php?f=71&t=565933, on PHP 4.3 `include_once` was an order of magnitude slower. A quick and dirty benchmark on PHP 5.6.9 suggests this is no longer the case, 369ns for `include_once` vs 216ns for a `!function_exists() { include }` construct.

In conclusion I'd say, just use `include_once`.

VSEphpbb:
include once is still expensive
https://robert.accettura.com/blog/2011/06/11/phps-include_once-is-insanely-expensive/

callumacrae:
9.74361634 nanoseconds is… not expensive.

VSEphpbb:
Those are seconds, not nanoseconds. I just wrote a quick similar script of my own, and include_once took 10 seconds to do what a regular include wrapped in a function exists condition did in less than a second.

Elsensee:
@callumacrae was talking about the time of *one* include_once. Also you don't include 1,000,000 files - phpBB doesn't even have that much.

I would still prefer to have just one way to include stuff (by using function_exists or similar) but not *necessarily* with such a PR.
I hope we refactored everything one day (even ACP, MCP, UCP) and can get rid of these include_once's that way.

callumacrae:
Yeah I was referring to `(10.020140171051 - 0.27652382850647) / 1000000`, a lot of which will be overhead from the loop itself anyway.

Also, there's no way that test was done scientifically with that level of precision. The test was either ran only once, or ran a few times and the results that best proved his points chosen. 14 decimal points is pointless, unless you're measuring to 17 decimal places and they're all within 1e-14 of each other!

Nicofuma:
Your branch needs to be rebased

RMcGirr83:
Seems there isn't consensus so  rebasing is premature

RMcGirr83:
@Nicofuma hope I did that correctly

RMcGirr83:
!unset WIP

PR #3545, opened 2015-04-21
[ticket/13771] Allow AJAX errors to support exceptions messages

VSEphpbb:
https://tracker.phpbb.com/browse/PHPBB3-13771
PHPBB3-13771

### DIFF
--- a/phpBB/assets/javascript/core.js
+++ b/phpBB/assets/javascript/core.js
@@ -262,10 +262,13 @@ phpbb.ajaxify = function(options) {
 				console.log('AJAX error. status: ' + textStatus + ', message: ' + errorThrown);
 			}
 			phpbb.clearLoadingTimeout();
-			var errorText = false;
-			var responseText = $.parseJSON(jqXHR.responseText);
-			if (typeof responseText.message === 'string' && responseText.message.length > 0) {
-				errorText = responseText.message;
+			var responseText, errorText = false;
+			try {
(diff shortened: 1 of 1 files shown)


phpBB/assets/javascript/core.js
```diff
@@ -263,7 +263,11 @@ phpbb.ajaxify = function(options) {
 			}
 			phpbb.clearLoadingTimeout();
 			var errorText = false;
-			if (typeof errorThrown === 'string' && errorThrown.length > 0) {
+			var responseText = $.parseJSON(jqXHR.responseText);
```
Nicofuma: if responseText isn't a valid json object what happen?

@callumacrae @prototech

MGaetan89: [From the doc](http://api.jquery.com/jQuery.parseJSON/):

> Passing in a malformed JSON string results in a JavaScript exception being thrown.

So I think the exception should be handled here somehow.

callumacrae: Yep, if it's an error handler it's pretty likely that this isn't JSON at all (for example, a 500 error). It should be wrapped in a try-catch block, and `JSON.parse` should be used, not `$.parseJSON`—phpBB doesn't support IE7!

VSEphpbb:
Updated

VSEphpbb:
bump

Nicofuma:
@callumacrae

callumacrae:
lgtm

PR #3551, opened 2015-04-26
[ticket/13779] Set new auth options to the role only if matching the role type

rxu:
Migrations' permission tool allows setting permissions to the role which
doesn't match the role type, e.g. m_ permissions for u_ role types and so on.
As one of side effects, this may lead to granting users moderative/admin
permissions silently.
With this patch the only new permissions matching the role type will be set.

<a href="https://tracker.phpbb.com/browse/PHPBB3-13779">PHPBB3-13779</a>.

### DIFF
--- a/phpBB/phpbb/db/migration/tool/permission.php
+++ b/phpBB/phpbb/db/migration/tool/permission.php
@@ -425,13 +425,27 @@ class permission implements \phpbb\db\migration\tool\tool_interface
 				$role_id = (int) $this->db->sql_fetchfield('auth_role_id');
 				if ($role_id)
 				{
-					$sql = 'SELECT role_name
+					$sql = 'SELECT role_name, role_type
 						FROM ' . ACL_ROLES_TABLE . '
 						WHERE role_id = ' . $role_id;
 					$this->db->sql_query($sql);
-					$role_name = $this->db->sql_fetchfield('role_name');
+					$role_data = $this->db->sql_fetchrow();
+					$role_name = $role_data['role_name'];
+					$role_type = $role_data['role_type'];
 
-					return $this->permission_set($role_name, $auth_option, 'role', $has_permission);
+					// Filter new auth options to match the role type: a_ | f_ | m_ | u_
+					// Set new auth options to the role only if options matching the role type were found
+					$auth_option = array_filter($auth_option,
+						function ($option) use ($role_type)
+						{
+							return strpos($option, $role_type) === 0;
+						}
+					);
+
+					if (sizeof($auth_option))
+					{
+						return $this->permission_set($role_name, $auth_option, 'role', $has_permission);
+					}
 				}
 
 				$sql = 'SELECT auth_option_id, auth_setting
--- a/tests/dbal/fixtures/migrator_permission.xml
+++ b/tests/dbal/fixtures/migrator_permission.xml
@@ -27,5 +27,139 @@
 			<value>1</value>
 			<value>0</value>
 		</row>
+		<row>
+			<value>4</value>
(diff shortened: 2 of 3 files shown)


phpBB/phpbb/db/migration/tool/permission.php
```diff
@@ -429,9 +429,23 @@ public function permission_set($name, $auth_option, $type = 'role', $has_permiss
 						FROM ' . ACL_ROLES_TABLE . '
 						WHERE role_id = ' . $role_id;
 					$this->db->sql_query($sql);
-					$role_name = $this->db->sql_fetchfield('role_name');
-
-					return $this->permission_set($role_name, $auth_option, 'role', $has_permission);
+					$role_data = $this->db->sql_fetchrow();
+					$role_name = $role_data['role_name'];
+					$role_type = $role_data['role_type'];
```
VSEphpbb: role_type is not retrieved in the query

rxu: Fixed, thanks a lot.

phpBB/phpbb/db/migration/tool/permission.php
```diff
@@ -425,13 +425,27 @@ public function permission_set($name, $auth_option, $type = 'role', $has_permiss
...
+
+					// Filter new auth options to match the role type: a_ | f_| m_ | u_
+					// Set new auth options to the role only if options matching the role type were found
+					$auth_option = array_filter($auth_option,
+						function ($option) use ($role_type)
+						{
+							return strpos($auth_option, $auth_type) === 0;
```
VSEphpbb: $auth_option should be $option? And $auth_type should be $role_type?

rxu: Hmm yeah, but ```$auth_option``` is correct as it's the method's input options array.

VSEphpbb: But you're passing the $auth_option array values it in to the closure as `$option` which is what is being tested in strpos...no?

phpBB/phpbb/db/migration/tool/permission.php
```diff
@@ -425,13 +425,27 @@ public function permission_set($name, $auth_option, $type = 'role', $has_permiss
...
 						WHERE role_id = ' . $role_id;
 					$this->db->sql_query($sql);
-					$role_name = $this->db->sql_fetchfield('role_name');
-
-					return $this->permission_set($role_name, $auth_option, 'role', $has_permission);
+					$role_data = $this->db->sql_fetchrow();
+					$role_name = $role_data['role_name'];
+					$role_type = $role_data['role_type'];
+
+					// Filter new auth options to match the role type: a_ | f_| m_ | u_
```
VSEphpbb: maybe clean up the comment ... space between `f_|` :smile:

Nicofuma:
I'd like to be able to warn the developer when it happens but for now I don't see how...

nickvergessen:
@rxu can we somehow add unit tests for this?
https://github.com/phpbb/phpbb/blob/3.1.x/tests/dbal/migrator_tool_permission_test.php

rxu:
@nickvergessen I'll look into possible tests cases.

rxu:
It looks like assigning permissions fail anyway for the group having more than 1 role assigned as the function only selects 1 role per group.
For the ```'role'``` mode, it can grant permissions regardless of the role type (f.e. ```a_``` permission for the ```u_``` role type, is it intended?).

rxu:
@nickvergessen Tests added.

Nicofuma:
:+1:

PR #3552, opened 2015-04-26
[ticket/13777] Move module management related functionailty into service

MateBartus:
[PHPBB3-13777](https://tracker.phpbb.com/browse/PHPBB3-13777)


phpBB/phpbb/module/module_manager.php
```diff
@@ -0,0 +1,547 @@
...
+	 * @var string
+	 */
+	protected $phpbb_root_path;
+
+	/**
+	 * @var string
+	 */
+	protected $php_ext;
+
+	public function __construct(\phpbb\cache\driver\driver_interface $cache,
```
Nicofuma: inline it and add a docblock please

phpBB/phpbb/module/module_manager.php
```diff
@@ -0,0 +1,552 @@
...
+	 * Get row for specified module
+	 *
+	 * @param int		$module_id		ID of the module
+	 * @param string	$module_class	Class of the module (acp, ucp, mcp etc...)
+	 *
+	 * @return array
+	 *
+	 * @throws \phpbb\module\exception\module_not_found_exception	When there is no module with $module_id
+	 */
+	public function get_module_row($module_id, $module_class)
```
MateBartus: Not sure about this, but the `$module_class` arg is there in every function, because that is how I replaced `$this->module_class`. I think it might be a good idea to get rid of it as `module_id` is a key by it self. Although if that extra check is needed for some reason, then that is that. So keep it or drop it?

phpBB/includes/acp/acp_modules.php
```diff
@@ -91,11 +93,13 @@ function main($id, $mode)
 						$db->sql_freeresult($result);
 					}
 
-					$errors = $this->delete_module($module_id);
+					$row = $module_manager->get_module_row($module_id, $this->module_class);
```
Nicofuma: this call can throw an exception right?

Nicofuma:
@nickvergessen

phpBB/phpbb/db/migration/data/v310/acp_prune_users_module.php
```diff
@@ -70,12 +72,14 @@ public function move_prune_users_module()
...
 		$this->db->sql_freeresult($result);
 
-		if (!class_exists('\acp_modules'))
-		{
-			include($this->phpbb_root_path . 'includes/acp/acp_modules.' . $this->php_ext);
-		}
-		$module_manager = new \acp_modules();
-		$module_manager->module_class = 'acp';
-		$module_manager->move_module($acp_prune_users_id, $acp_cat_users_id);
+		$module_manager = new \phpbb\module\module_manager(
```
nickvergessen: should be a service

phpBB/phpbb/db/migration/data/v310/dev.php
```diff
@@ -204,18 +206,20 @@ public function move_customise_modules()
...
-		}
 		// acp_modules calls adm_back_link, which is undefined at this point
 		if (!function_exists('adm_back_link'))
 		{
 			include($this->phpbb_root_path . 'includes/functions_acp.' . $this->php_ext);
 		}
-		$module_manager = new \acp_modules();
-		$module_manager->module_class = 'acp';
-		$module_manager->move_module($language_module_id, $language_management_module_id);
+		$module_manager = new \phpbb\module\module_manager(
```
nickvergessen: service

phpBB/phpbb/db/migration/tool/module.php
```diff
@@ -239,12 +239,14 @@ public function add($class, $parent = 0, $data = array())
 			return;
 		}
 
-		if (!class_exists('acp_modules'))
-		{
-			include($this->phpbb_root_path . 'includes/acp/acp_modules.' . $this->php_ext);
-			$this->user->add_lang('acp/modules');
-		}
-		$acp_modules = new \acp_modules();
+		$acp_modules = new \phpbb\module\module_manager(
```
nickvergessen: service

phpBB/phpbb/module/module_manager.php
```diff
@@ -0,0 +1,552 @@
...
+	{
+		$sql = 'SELECT *
+			FROM ' . $this->modules_table . "
+			WHERE module_class = '" . $this->db->sql_escape($module_class) . "'
+				AND module_id = $module_id";
```
nickvergessen: must be casted to integer or sql_escaped

phpBB/phpbb/module/module_manager.php
```diff
@@ -0,0 +1,552 @@
...
+		}
+
+		$rows = array();
+
+		$sql = 'SELECT m2.*
+			FROM ' . $this->modules_table . ' m1
+			LEFT JOIN ' . $this->modules_table . " m2 ON ($condition)
+			WHERE m1.module_class = '" . $this->db->sql_escape($module_class) . "'
+				AND m2.module_class = '" . $this->db->sql_escape($module_class) . "'
+				AND m1.module_id = $module_id
```
nickvergessen: cast or escape

phpBB/phpbb/module/module_manager.php
```diff
@@ -0,0 +1,552 @@
...
+		if (sizeof($branch))
+		{
+			return array('CANNOT_REMOVE_MODULE');
+		}
+
+		// If not move
+		$diff = 2;
+		$sql = 'DELETE FROM ' . $this->modules_table . "
+			WHERE module_class = '" . $this->db->sql_escape($module_class) . "'
+				AND module_id = $module_id";
```
nickvergessen: cast or escape

phpBB/phpbb/module/module_manager.php
```diff
@@ -0,0 +1,552 @@
...
+		$this->db->sql_query($sql);
+	}
+
+	/**
+	 * Remove module from tree
+	 *
+	 * @param int		$module_id		ID of the module to delete
+	 * @param string	$module_class	Class of the module (acp, ucp, mcp etc...)
+	 *
+	 * @return array
```
nickvergessen: please add a description

phpBB/phpbb/module/module_manager.php
```diff
@@ -0,0 +1,552 @@
...
+	 */
+	public function delete_module($module_id, $module_class)
+	{
+		$row = $this->get_module_row($module_id, $module_class);
+
+		$branch = $this->get_module_branch($module_id, $module_class, 'children', false);
+
+		if (sizeof($branch))
+		{
+			return array('CANNOT_REMOVE_MODULE');
```
nickvergessen: should we throw an exception instead?

phpBB/phpbb/module/module_manager.php
```diff
@@ -0,0 +1,552 @@
...
+		while ($row = $this->db->sql_fetchrow($result))
+		{
+			$target = $row;
+		}
+		$this->db->sql_freeresult($result);
+
+		if (!sizeof($target))
+		{
+			// The module is already on top or bottom
+			return false;
```
nickvergessen: does not match doc type

phpBB/phpbb/module/module_manager.php
MateBartus: Fixed (top of the function)

phpBB/phpbb/module/module_manager.php
MateBartus: Fixed (top of the function)

phpBB/phpbb/module/module_manager.php
MateBartus: Fixed the doc

phpBB/phpbb/db/migration/data/v310/acp_prune_users_module.php
```diff
@@ -70,12 +72,7 @@ public function move_prune_users_module()
...
 		$this->db->sql_freeresult($result);
 
-		if (!class_exists('\acp_modules'))
-		{
-			include($this->phpbb_root_path . 'includes/acp/acp_modules.' . $this->php_ext);
-		}
-		$module_manager = new \acp_modules();
-		$module_manager->module_class = 'acp';
-		$module_manager->move_module($acp_prune_users_id, $acp_cat_users_id);
+		$module_manager = $phpbb_container->get('module.manager');
```
nickvergessen: Get it from the system insted of global `$phpbb_container`

Extend `phpbb\db\migration\container_aware_migration` and then use `$this->container->`

phpBB/phpbb/db/migration/data/v310/dev.php
```diff
@@ -204,18 +206,13 @@ public function move_customise_modules()
...
-		}
 		// acp_modules calls adm_back_link, which is undefined at this point
 		if (!function_exists('adm_back_link'))
 		{
 			include($this->phpbb_root_path . 'includes/functions_acp.' . $this->php_ext);
 		}
-		$module_manager = new \acp_modules();
-		$module_manager->module_class = 'acp';
-		$module_manager->move_module($language_module_id, $language_management_module_id);
+		$module_manager = $phpbb_container->get('module.manager');
```
nickvergessen: `container_aware_migration`

phpBB/phpbb/db/migration/tool/module.php
```diff
@@ -422,12 +419,18 @@ public function remove($class, $parent = 0, $module = '')
 				include($this->phpbb_root_path . 'includes/acp/acp_modules.' . $this->php_ext);
 				$this->user->add_lang('acp/modules');
 			}
-			$acp_modules = new \acp_modules();
-			$acp_modules->module_class = $class;
+			$acp_modules = new \phpbb\module\module_manager(
```
nickvergessen: service

phpBB/phpbb/db/migration/tool/module.php
```diff
@@ -474,12 +477,16 @@ public function reverse()
...
 	{
-		if (!class_exists('acp_modules'))
-		{
-			include($this->phpbb_root_path . 'includes/acp/acp_modules.' . $this->php_ext);
-			$this->user->add_lang('acp/modules');
-		}
-		$acp_modules = new \acp_modules();
+		global $phpbb_extension_manager;
+
+		$acp_modules = new \phpbb\module\module_manager(
```
nickvergessen: service

phpBB/phpbb/module/module_manager.php
nickvergessen: I think we should throw an exception, instead of mixed returns?

MateBartus:
Fixed everything mentioned above

phpBB/phpbb/module/module_manager.php
```diff
@@ -0,0 +1,555 @@
...
+	/**
+	 * Move module around the tree
+	 *
+	 * @param int		$from_module_id	ID of the current parent module
+	 * @param int		$to_parent_id	ID of the target parent module
+	 * @param string	$module_class	Class of the module (acp, ucp, mcp etc...)
+	 */
+	public function move_module($from_module_id, $to_parent_id, $module_class)
+	{
+		$moved_modules = $this->get_module_branch($from_module_id, $module_class, 'children');
```
Nicofuma: what if there isn't any branch? get_module_branch doesn't throw any exception so the next lines will throw an error?

MateBartus: yep, I will fix it.

tests/extension/modules_test.php
```diff
@@ -89,21 +95,7 @@ public function test_get_module_infos()
 			), $acp_modules);
 
 		// Find a specific module info file (mcp_a_module)
-		$this->acp_modules->module_class = 'mcp';
-		$acp_modules = $this->acp_modules->get_module_infos('mcp_a_module');
-		$this->assertEquals(array(
```
marc1706: Does this no longer work or why did you remove this assertion?

MateBartus: Probably by accident, thanks

MateBartus: Nope, it does not work anymore, because there is no [module_manager::module_class](https://github.com/phpbb/phpbb/pull/3552/files#diff-9b05e985bd4f61bb1bf4c1893a24a624L559) member var. However, IMO this means that I have to update the get_module_infos() params in some way, or would it be enough to just throw an exception when module class is not definded? Or just default to something?

naderman: I think throwing an exception sounds fine to me?

naderman:
Tested it a bit, seems to work fine so far.

tests/extension/modules_test.php
MateBartus: Fixed

MateBartus:
I guess it should be fine for now as it is, but there is certainly a lot of room for improvements implementation wise. My aim was to extract the management logic from the controller logic as it is used in the installer as well as the ACP. Although when the ACP gets moved to actuall controllers this could get revisited as well.

naderman:
Yeah I wouldn't see this as the final goal, but certainly a step in the right direction :-)

phpBB/phpbb/module/module_manager.php
```diff
@@ -0,0 +1,570 @@
...
+	 * @param string	$module				ID of module
+	 * @param string	$module_class		Class of the module (acp, ucp, mcp etc...)
+	 * @param bool		$use_all_available	Use all available instead of just all
+	 *										enabled extensions
+	 *
+	 * @return array	Array with module information gathered from module info files.
+	 */
+	public function get_module_infos($module = '', $module_class = null, $use_all_available = false)
+	{
+		if ($module_class === null)
```
naderman: If we do this though, shouldn't we move it to the front and make it not default to null?

MateBartus: That's what I tried to ask [here](https://github.com/phpbb/phpbb/pull/3552/files#r31227530). So I would agree.

naderman: Ah sorry, I think I misunderstood you!

PR #3553, opened 2015-04-26
[ticket/12230] Remove users from new users group when post limit is 0

Elsensee:
I thought about removing that part completely, but I saw that session_begin() is used in the updater so I just changed it to an `isset()` call.

I'm not sure - do we need a special 3.1.x PR? GitHub said at first that it can't be merged without problems, but when I merged this branch into a 3.1.x branch there were no conflicts.

### DIFF
--- a/phpBB/includes/session.php
+++ b/phpBB/includes/session.php
@@ -443,7 +443,7 @@ class session
 								$db->sql_query($sql);
 							}
 
-							if ($this->data['user_id'] != ANONYMOUS && !empty($config['new_member_post_limit']) && $this->data['user_new'] && $config['new_member_post_limit'] <= $this->data['user_posts'])
+							if ($this->data['user_id'] != ANONYMOUS && isset($config['new_member_post_limit']) && $this->data['user_new'] && $config['new_member_post_limit'] <= $this->data['user_posts'])
 							{
 								$this->leave_newly_registered();
 							}


rxu:
As far as I understand, ```$config['new_member_post_limit']``` is always set, isn't it? Maybe just get rid of it here?

bantu:
@rxu This is session.php, better not assume anything. :-P

Elsensee:
@rxu This feature was added in 3.0.6 I think. Let's play it safe. ;)

rxu:
Well, I didn't mean getting read of the whole feature but of ```isset`` thing only, but yeah, I think you guys are right, this is session.php so let's leave it alone :P

bantu:
This patch seems to change behaviour/definition of "A value of 0 disables this feature."

Pre-Patch behaviour:
When 0 is entered, newly registered system group memberships do not change.
Users are not added on registration because the feature is disabled.
Users are not removed when their number of posts change, because the feature is disabled.
The newly registered group can be manually managed by administrators.

Post-Patch behaviour:
When 0 is entered, this is interpreted as the number of posts required to leave the newly registered group.
Users are not added on registration because the feature is disabled.
Users are removed because their post count is larger or equal to 0.
The newly registered group can no longer be manually managed by administrators.

On a side note: When entering 0 into the admin panel in post-patch behaviour, we could just as well completely clear the newly registered group instead of checking whether 0 <= $x holds for the non-negative post count.

In conclusion: The `!empty` was intentional, this is not a bug and this patch is not an acceptable solution.

Elsensee:
Ah! I forgot about this line. So indeed not a bug. Thanks for clarifying. :)

However, I think that disabling this feature means "disabling" the group as well, so remove them all. (Sounds like a better idea than my old patch :D)

Elsensee:
Will do a new PR.

PR #3555, opened 2015-04-26
[ticket/12101] Remove HTML redirect workaround for Microsoft/IIS webserver

akoeplinger:
The code was added as a workaround for bugs in very old versions of IIS,
dating back to 2002: see commit 849d76697444f4e3523845f8c96569ccde57d868.

Newer IIS versions handle the "Location" header just fine, so we can avoid
the unnecessary HTML page for redirects. Given that it seems to work fine
since IIS 6.0 (on WinServer 2003), I don't think it's worth adding a special
check for earlier versions as they are no longer supported by MS as well.

[PHPBB3-12101](https://tracker.phpbb.com/browse/PHPBB3-12101)

### DIFF
--- a/phpBB/includes/functions.php
+++ b/phpBB/includes/functions.php
@@ -2389,7 +2389,7 @@ function redirect($url, $return = false, $disable_cd_check = false)
 	}
 
 	// Redirect via an HTML form for PITA webservers
-	if (@preg_match('#Microsoft|WebSTAR|Xitami#', getenv('SERVER_SOFTWARE')))
+	if (@preg_match('#WebSTAR|Xitami#', getenv('SERVER_SOFTWARE')))
 	{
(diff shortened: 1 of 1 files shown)


cs278:
To be honest I wouldn't be surprised if the whole lot can be removed. Maybe a developer could check the stats data: https://github.com/phpbb/phpbb/blob/3.0.x/phpBB/includes/questionnaire/questionnaire.php#L167

bantu:
I trust/believe this patch is okay (also because of https://tracker.phpbb.com/browse/PHPBB3-12101?focusedCommentId=40795&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-40795).

akoeplinger:
Great, thanks guys!

PR #3557, opened 2015-04-27
[ticket/13778] Update wording of the reCaptcha instructions

VSEphpbb:
https://tracker.phpbb.com/browse/PHPBB3-13778

PHPBB3-13778


bantu:
Hum. It's not showing "text", it seems to always show "numbers".

bantu:
An improvement over the old version, though.

PR #3558, opened 2015-04-27
[ticket/13786] Add events to add MCP post options

javiexin:
Three new events are included:
1) Core event to perform new moderator action in includes/mcp/mcp_post.php
2) Core event to add/modify template data in includes/mcp/mcp_post.php
3) Template event to present new options in template/mcp_post.html

PHPBB3-13786

### DIFF
--- a/phpBB/includes/mcp/mcp_post.php
+++ b/phpBB/includes/mcp/mcp_post.php
@@ -280,7 +280,7 @@ function mcp_post_details($id, $mode, $action)
 	* @since 3.1.5-RC1
 	*/
 	$vars = array(
-		'poll_info',
+		'post_info',
 		'mcp_post_template_data',
 		'attachments',
 		's_additional_opts',
--- a/phpBB/includes/mcp/mcp_post.php
+++ b/phpBB/includes/mcp/mcp_post.php
@@ -26,6 +26,7 @@ function mcp_post_details($id, $mode, $action)
 {
 	global $phpEx, $phpbb_root_path, $config;
 	global $template, $db, $user, $auth, $cache;
+	global $phpbb_dispatcher;
 
 	$user->add_lang('posting');
 
@@ -104,6 +105,21 @@ function mcp_post_details($id, $mode, $action)
 					trigger_error('FORM_INVALID');
 				}
 			}
+
+		break;
+
+		default:
+
+			/**
+			* This event allows you to handle custom post moderation options
+			*
+			* @event core.mcp_post_additional_options
+			* @var	string	action		Post moderation action name
+			* @var	array	post_info	Information on the affected post
+			* @since 3.1.5-RC1
+			*/
+			$vars = array('action', 'post_info');
+			extract($phpbb_dispatcher->trigger_event('core.mcp_post_additional_options', compact($vars)));
 
 		break;
 	}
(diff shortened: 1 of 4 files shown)


imkingdavid:
Code looks fine to me.

phpBB/styles/prosilver/template/mcp_post.html
```diff
@@ -168,6 +168,10 @@
 			</form>
 		<!-- ENDIF -->
 
+		<!-- IF S_MCP_POST_ADDITIONAL_OPTS -->
```
nickvergessen: I think the event should be used without the `IF`.
Extensions can check for that themselves and not having it allows extensions which only display when no additional options are displayed.

phpBB/styles/subsilver2/template/mcp_post.html
```diff
@@ -156,6 +156,11 @@
 				<td class="row2"><form method="post" name="mcp_chgposter" action="{U_POST_ACTION}"><input class="post" type="text" name="username" value="" /> <input class="btnmain" type="submit" value="{L_CONFIRM}" name="action[chgposter]" /><br /><span class="gensmall">[ <a href="{U_FIND_USERNAME}" onclick="find_username(this.href); return false;">{L_FIND_USERNAME}</a> ]</span><!-- IF S_USER_SELECT --><br /><select name="u">{S_USER_SELECT}</select> <input type="submit" class="btnmain" name="action[chgposter_ip]" value="{L_CONFIRM}" /><!-- ENDIF -->{S_FORM_TOKEN}</form></td>
 			</tr>
 		<!-- ENDIF -->
+
+		<!-- IF S_MCP_POST_ADDITIONAL_OPTS -->
```
nickvergessen: same here

phpBB/includes/mcp/mcp_post.php
```diff
@@ -197,7 +213,7 @@ function mcp_post_details($id, $mode, $action)
 		$l_deleted_by = '';
 	}
 
-	$template->assign_vars(array(
+	$mcp_post_template_data = array(
```
nickvergessen: Not sure if this needs changing. You can simply overwrite them. but I guess it doesn't hurt either

phpBB/styles/prosilver/template/mcp_post.html
javiexin: The problem is that this is within another IF block, where the conditions are limited (see above, line 147), so the template variable is needed anyway.  Or else, the broader block should be broken, but that will create issues with the divs around it.

phpBB/styles/subsilver2/template/mcp_post.html
javiexin: same as above, the template variable is needed anyway

phpBB/includes/mcp/mcp_post.php
javiexin: Alternatively you could pick and choose what to send in the event, like for example $message, but you are right, mostly everything else is availaby by means of the $post_info array.

phpBB/styles/prosilver/template/mcp_post.html
nickvergessen: Well just imagine an extension wants to display something with:
 - S_CAN_LOCK_POST = true
- S_MCP_POST_ADDITIONAL_OPTS = false

It's not possible because you wrapped the event. So just remove that IF around the event.

javiexin: Ok, agreed.  We keep it in L147, but remove here.

nickvergessen: yes

nickvergessen:
Merged via https://github.com/phpbb/phpbb/commit/bb7f8433441343eaaf0c15aa9f27c4227256e3c7

PR #3561, opened 2015-04-28
[ticket/10922] Added support for body and subject in email BBCode

s9e:
https://tracker.phpbb.com/browse/PHPBB3-10922
```
[email subject="Hello" body="Sent from phpBB"]user@example.org[/email]
```
```html
<a href="mailto:user@example.org?subject=Hello&amp;body=Sent+from+phpBB">user@example.org</a><br>
```

Subject/body are added as optional attributes to the `[email]` tag. I think that's preferable to asking users to specify them as query parameters to an hypothetical `mailto:` URI as it was original suggested. One of the reason is that BBCode attributes are plain text and users don't have to think twice about it. If we treat the user input as a URI, then the user would have to URL-encode spaces and things like that.


tests/text_processing/tickets_data/PHPBB3-10922.txt
```diff
@@ -1 +1,5 @@
-[email]user@example.org[/email][email=user@example.org]...[/email]
\ No newline at end of file
+[email]user@example.org[/email]
+[email=user@example.org]...[/email]
```
nickvergessen: How does this case + subject work/look like?

s9e: As one would expect. `subject` and `body` are optional, they can be set in either order. `email` is required, it can be set in short form `[email=user@example.org]` or in long form `[email email="user@example.org"]` in any order, or not at all, in which case it uses the BBCode's content as email.

tests/text_processing/tickets_data/PHPBB3-10922.txt
```diff
@@ -1 +1,8 @@
-[email]user@example.org[/email][email=user@example.org]...[/email]
\ No newline at end of file
+[email]user@example.org[/email]
+[email=user@example.org ]...[/email]
```
nickvergessen: without trailing space on the email aswell please

tests/text_processing/tickets_data/PHPBB3-10922.txt
s9e: You mean at the end of the file?

tests/text_processing/tickets_data/PHPBB3-10922.txt
nickvergessen: We need to test both:
```
[email=user@example.org]...[/email]
[email=user@example.org ]...[/email]

tests/text_processing/tickets_data/PHPBB3-10922.txt
s9e: Ah ok. I didn't see it because it was a typo. I added both cases although they are not specific to `[email]`. They should probably be in tests/text_formatter/s9e/default_formatting_test.php rather than this ticket's regression test.

callumacrae:
LGTM! :tada:

PR #3563, opened 2015-04-28
[ticket/13791] Add more post buttons template events to viewtopic_body.html

rxu:
The purpose is to allow adding custom post buttons outside of the original post
buttons conditionals/list in prosilver (for subsilver2 it's not the issue).

<a href="https://tracker.phpbb.com/browse/PHPBB3-13791">PHPBB3-13791</a>.


rxu:
!set Event

PR #3566, opened 2015-04-29
[ticket/12542] Highlight textarea when files are dragged onto editor

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-12542
Video: https://www.m-a-styles.de/download/drag_n_drop.mp4

Works for both posts and private messages if the attachment tab exists (attachments are enabled).

### DIFF
--- a/phpBB/assets/javascript/core.js
+++ b/phpBB/assets/javascript/core.js
@@ -1232,6 +1232,14 @@ phpbb.applyCodeEditor = function(textarea) {
 	});
 };
 
+phpbb.applyDragNDrop = function(textarea) {
+	$('html').on('dragenter dragover', function () {
+		$(textarea).addClass('drag-n-drop');
+	}).on('dragleave dragout dragend drop', function() {
+		$(textarea).removeClass('drag-n-drop');
+	});
+};
+
 /**
 * List of classes that toggle dropdown menu,
 * list of classes that contain visible dropdown menu
--- a/phpBB/assets/javascript/core.js
+++ b/phpBB/assets/javascript/core.js
@@ -1233,10 +1233,15 @@ phpbb.applyCodeEditor = function(textarea) {
 };
 
 phpbb.applyDragNDrop = function(textarea) {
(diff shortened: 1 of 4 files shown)


nickvergessen:
@PayBas can you have a look?

VSEphpbb:
I don't think the browser prefixes are needed for all the `outline` css properties:
http://caniuse.com/#search=outline
http://caniuse.com/#search=outline-color
https://developer.mozilla.org/en-US/docs/Web/CSS/outline-offset

Nicofuma:
@prototech @callumacrae

marc1706:
As a side note, IE does not support the outline-offset. The highlight border is therefore on the outside of the textarea. It does not seem to cover anything else though and still seems to serve the intended purpose.

phpBB/assets/javascript/core.js
```diff
@@ -1233,6 +1233,31 @@ phpbb.applyCodeEditor = function(textarea) {
...
 /**
+ * Show drag and drop animation when textarea is present
+ *
+ * This function will enable the drag and drop animation for a specified
+ * textarea.
+ *
+ * @param {object} textarea Textarea DOM object to apply editor to
+ */
+phpbb.showDragNDrop = function(textarea) {
+	if (textarea === 'undefined') {
```
callumacrae: You're missing a `typeof` here, this code won't work

phpBB/styles/prosilver/theme/colours.css
```diff
@@ -977,6 +977,22 @@ fieldset.quick-login input.inputbox {
 	color: #333333;
 }
 
+#message-box textarea.drag-n-drop {
```
callumacrae: This seems overqualified. Would just `.drag-n-drop` not work?

marc1706:
Updated

callumacrae:
:+1:

Also :+1: on dropping the vendor prefixes, doesn't seem necessary at all: http://caniuse.com/#feat=outline

PR #3568, opened 2015-04-29
[ticket/10268] Added regression test

s9e:
PHPBB3-10268 got fixed by PHPBB3-11768

https://tracker.phpbb.com/browse/PHPBB3-10268


bantu:
This PR doesn't exactly fix https://tracker.phpbb.com/browse/PHPBB3-10268 which says "extend function make_clickable() to also recognize semicolons as leading URL borders" right now.

PR #3577, opened 2015-04-30
[ticket/13800] Make router's extension manager dependency optional

MateBartus:
[PHPBB3-13800](https://tracker.phpbb.com/browse/PHPBB3-13800)

### DIFF
--- a/phpBB/phpbb/routing/router.php
+++ b/phpBB/phpbb/routing/router.php
@@ -94,13 +94,14 @@ class router implements RouterInterface
 	 * Construct method
 	 *
 	 * @param \phpbb\filesystem\filesystem_interface $filesystem	Filesystem helper
-	 * @param manager	$extension_manager	Extension manager
-	 * @param string	$phpbb_root_path	phpBB root path
-	 * @param string	$php_ext			PHP file extension
-	 * @param string	$environment		Name of the current environment
-	 * @param array		$routing_files		Array of strings containing paths to YAML files holding route information
+	 * @param string		$phpbb_root_path						phpBB root path
+	 * @param string		$php_ext								PHP file extension
+	 * @param string		$environment							Name of the current environment
+	 * @param manager|null	$extension_manager						Extension manager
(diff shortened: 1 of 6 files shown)


phpBB/phpbb/routing/router.php
```diff
@@ -172,8 +173,15 @@ public function get_routes()
...
-			$this->find_routing_files($this->extension_manager->all_enabled(false))
-				->find($this->phpbb_root_path);
+			if ($this->extension_manager !== null)
+			{
+				$this->find_routing_files($this->extension_manager->all_enabled(false))
+					->find($this->phpbb_root_path);
+			}
+			else
+			{
+				$this->find($this->phpbb_root_path);
```
Nicofuma: you still have to call `find_routing_files()` with an empty array to fill the `$routing_files` array with the core one.

phpBB/phpbb/routing/router.php
```diff
@@ -94,13 +94,14 @@ class router implements RouterInterface
...
-	 * @param array		$routing_files		Array of strings containing paths to YAML files holding route information
+	 * @param string		$phpbb_root_path						phpBB root path
+	 * @param string		$php_ext								PHP file extension
+	 * @param string		$environment							Name of the current environment
+	 * @param manager|null	$extension_manager						Extension manager
+	 * @param array			$routing_files							Array of strings containing paths to YAML files
+	 * 																holding route information
 	 */
-	public function __construct(\phpbb\filesystem\filesystem_interface $filesystem, manager $extension_manager, $phpbb_root_path, $php_ext, $environment, $routing_files = array())
+	public function __construct(\phpbb\filesystem\filesystem_interface $filesystem, $phpbb_root_path, $php_ext, $environment, manager $extension_manager = null, $routing_files = array())
```
nickvergessen: Pass it into `public function get_routes()` instead?

MateBartus: `get_routes()` is called from this class, so I guess that is not really doable.

PR #3578, opened 2015-04-30
[ticket/13801] Remove user dependency from text_formatter.s9e.parser

s9e:
https://tracker.phpbb.com/browse/PHPBB3-13801

This PR depends on https://tracker.phpbb.com/browse/PHPBB3-13802

### DIFF
--- a/phpBB/phpbb/textformatter/s9e/parser.php
+++ b/phpBB/phpbb/textformatter/s9e/parser.php
@@ -56,8 +56,9 @@ class parser implements \phpbb\textformatter\parser_interface
 		* Configure the parser service
 		*
 		* Can be used to:
-		*  - register variables or custom parsers in the s9e\TextFormatter
-		*  - configure the s9e\TextFormatter parser
+		*  - toggle features or BBCodes
+		*  - register variables or custom parsers in the s9e\TextFormatter parser
+		*  - configure the s9e\TextFormatter parser's runtime settings
 		*
 		* @event core.text_formatter_s9e_parser_setup
 		* @var \phpbb\textformatter\s9e\parser parser This parser service
--- a/phpBB/phpbb/textformatter/s9e/parser.php
+++ b/phpBB/phpbb/textformatter/s9e/parser.php
@@ -186,7 +186,7 @@ class parser implements \phpbb\textformatter\parser_interface
 	/**
 	* {@inheritdoc}
 	*
-	* This will translate the log entries found in s9e\TextFormatter's logger into phpBB error
+	* This will convert the log entries found in s9e\TextFormatter's logger into phpBB error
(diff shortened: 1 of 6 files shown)


bantu:
@Nicofuma What do you think about this?

Nicofuma:
The call to lang() should always be as late as possible so :+1:

phpBB/phpbb/textformatter/s9e/parser.php
```diff
@@ -56,24 +50,20 @@ public function __construct(\phpbb\cache\driver\driver_interface $cache, $key, \
...
 		$this->parser = $parser;
-		$this->user = $user;
 		$parser = $this;
 
 		/**
 		* Configure the parser service
 		*
 		* Can be used to:
-		*  - toggle features according to the user's preferences,
-		*  - toggle BBCodes according to the user's permissions,
```
Nicofuma: After this change, how could we toggle BBCodes according to the user perms?

s9e: You couldn't. That's why I made the ticket depend on https://tracker.phpbb.com/browse/PHPBB3-13802

Nicofuma: in this case I mark this PR as WIP and we need to find another way asap

s9e: Actually I shouldn't say that one couldn't toggle BBCodes without the user object being passed in the event. It's always possible to toggle BBCodes using the global user object. That's what extensions do in 3.1 and it's still possible in 3.2. I don't know whether that's the correct way to do it though.

VSEphpbb: I don't see the user object being used in the class for anything other than loading lang keys in get_errors() below.

Why does this event need to pass the $user object in an event?

ABBC ext (3.2) already uses this event to toggle bbcodes for a user based on their group membership. The $user is handled independently within the ABBC3 extension's own methods.

I may not be seeing the bigger picture but I'm not sure how the user object here serves any purpose in this event.

s9e: The rationale for passing the user as an event argument was that if the parser used a user object, event listeners should use the same user object. But after a second look I realized that the parser service should be independent of any user and that's where we are now.

rxu: @s9e AFAIK extensions are using the same user object anyway, so this is not a problem.

s9e: Ok, then I officially declare this PR not-WIP anymore. 3.2 has the same functionality as 3.1. @Nicofuma you can merge if you want.

PR #3579, opened 2015-05-01
[ticket/13803] Implement a generic and scalable way to reparse rich text

s9e:
https://tracker.phpbb.com/browse/PHPBB3-13803
https://area51.phpbb.com/phpBB/viewtopic.php?f=126&t=47666


s9e:
!set WIP

phpBB/phpbb/textreparser/base.php
```diff
@@ -0,0 +1,210 @@
...
+			);
+		}
+
+		// Those BBCodes are disabled based on context and user permissions and that value is never
+		// stored in the database. Here we test whether they were used in the original text.
+		$bbcodes = array('flash', 'img', 'quote', 'url');
+		foreach ($bbcodes as $bbcode)
+		{
+			$field_name = 'enable_' . $bbcode . '_bbcode';
+			$record[$field_name] = $this->guess_bbcode($record, $bbcode);
```
Nicofuma: you can use the `get_records()` to get the poster id, then you can use this poster_id to get the user permissions through the \phpbb\auth\auth class. In the current state of the perm and auth system it should be possible even if you may have a instantiate a new auth class per user (it should be faster than having to reset the class between each post)

phpBB/phpbb/textreparser/base.php
```diff
@@ -0,0 +1,210 @@
...
+		{
+			$record['text'] = $parsed_text;
+			$this->save_record($record);
+		}
+	}
+
+	/**
+	* {@inheritdoc}
+	*/
+	abstract protected function save_record(array $record);
```
Nicofuma: move it just after get_record() please

phpBB/phpbb/textreparser/row_based_plugin.php
```diff
@@ -0,0 +1,119 @@
...
+	abstract public function get_table_name();
+
+	/**
+	* {@inheritdoc}
+	*/
+	public function get_max_id()
+	{
+		$columns = $this->get_columns();
+
+		$sql = 'SELECT MAX(' . $columns['id'] . ' AS max_id FROM ' . $this->get_table_name();
```
Nicofuma: the max ID should certainly be dependent of a date. This way the number of post to reparse will be static.

This date will be stored as a timestamp in a constant somewhere and set when releasing

s9e: What would be the benefits of using a timestamp rather than the value of a primary key? An auto_increment'ed primary key is monotonic, consistent across multiple servers and guaranteed to be indexed.

Nicofuma: because this value will change between each call and we don't need to reparse the posts posted after the update.

(btw, the date will be stored in the config by a migration. not set during the release)

s9e: You only get the max ID once then you work backward from it, like you would only get the timestamp once. Check out the last paragraph of the original post at https://area51.phpbb.com/phpBB/viewtopic.php?f=126&t=47666

Nicofuma: right... maybe you could add a base implementation for a cron?

Nicofuma:
did you have time to work on that recently?

s9e:
This PR adds a set of services that can be used to reparse text and update the database. To the best of my knowledge, the code is complete and functional but I've only tested the `text_reparser.post_text` service on actual data.

You can merge this PR now then we can figure out how to move forward and use those services in a cron task or as something that can be manually executed via the ACP or in CLI, or all of the above.

Nicofuma:
please rebase and then remove the WIP tag

phpBB/config/default/container/services_text_reparser.yml
```diff
@@ -0,0 +1,63 @@
+services:
+    text_reparser.contact_admin_info:
```
Nicofuma: please also add a service for the reparser plugin collection (text_reparser.lugins_collection maybe) so it can be used

Nicofuma:
how is it supposed to be used? get the reparser you need and use it? or something else involving a factory and I don't know what to automatically guess which reparser should be used?

Because in the first case, forgot what I said about the collection and remove the reparser.plugin tag everywhere

s9e:
There are 9 different services sharing the same `phpbb\textreparser\reparser_interface`. Each service can be used to reparse a different kind of text. (posts, poll options, private messages, etc...)

Get a service from the container → Call `reparse_range($start, $end)` → All records between `$start` and `$end` are updated.

Here's a working example that reparses posts: https://gist.github.com/s9e/623359cc041a82281f54

Nicofuma:
so we don't need the tags:
```
        tags:
            - { name: text_reparser.plugin }
```
because we don't need to build a list of the available reparser. Right?

s9e:
Not right now but I assume you will later for migration purposes.

phpBB/config/default/container/services_text_reparser.yml
```diff
@@ -0,0 +1,70 @@
+services:
+    text_reparser_collection:
+        class: phpbb\textreparser\reparser_collection
```
Nicofuma: don't use a custom class, it's not required

phpBB/phpbb/textreparser/base.php
```diff
@@ -0,0 +1,211 @@
...
+	*/
+	protected function reparse_record(array $record)
+	{
+		$record = $this->add_missing_fields($record);
+		$unparsed = array_merge(
+			$record,
+			generate_text_for_edit(
+				$record['text'],
+				$record['bbcode_uid'],
+				OPTION_FLAG_BBCODE | OPTION_FLAG_SMILIES | OPTION_FLAG_LINKS
```
Nicofuma: and what if bbcode was disabled?

tests/text_reparser/plugins/contact_admin_info_test.php
```diff
@@ -0,0 +1,81 @@
...
+* @copyright (c) phpBB Limited <https://www.phpbb.com>
+* @license GNU General Public License, version 2 (GPL-2.0)
+*
+* For full copyright and license information, please see
+* the docs/CREDITS.txt file.
+*
+*/
+require_once __DIR__ . '/../../../phpBB/includes/functions.php';
+require_once __DIR__ . '/../../../phpBB/includes/functions_content.php';
+require_once __DIR__ . '/../../test_framework/phpbb_database_test_case.php';
```
Nicofuma: not needed

tests/text_reparser/plugins/contact_admin_info_test.php
```diff
@@ -0,0 +1,81 @@
...
+* This file is part of the phpBB Forum Software package.
+*
+* @copyright (c) phpBB Limited <https://www.phpbb.com>
+* @license GNU General Public License, version 2 (GPL-2.0)
+*
+* For full copyright and license information, please see
+* the docs/CREDITS.txt file.
+*
+*/
+require_once __DIR__ . '/../../../phpBB/includes/functions.php';
```
Nicofuma: not needed too

Nicofuma:
About the reparser involving sql insert queries I think we should use the sql_insert_buffer class to improve the performances

phpBB/phpbb/textreparser/base.php
s9e: As far as I know, the third argument to `generate_text_for_edit()` has no effect on anything. Its value is passed to extensions and determines the return values of `allow_bbcode`, `allow_smilies` and `allow_urls` but those are not used for anything. The settings for toggling features are prefixed with `enable_`.

I could set it to `0` and it wouldn't change anything. There are 8 posts in the `post_text` test fixture that represent every combination of the 3 settings and the third argument to `generate_text_for_edit()` does not affect the test's result.

Come to think of it, I will just generate a value based on the `enable_` field so that nobody else has to wonder about it. That will be faster that explaining that the value doesn't matter.

tests/text_reparser/plugins/contact_admin_info_test.php
s9e: They're not when the test is run as part of the test suite. However, they're needed to run the test separately. If I remove them, the next time there's a bug related to the tested class the choice will be between waiting 20+ minutes for the test suite to run or spending X amount of time figuring out which files are required to run the test alone.

I'll remove them if you still think that's worth it.

s9e:
There's no insert queries, only updates.

Nicofuma:
right :/

phpBB/config/default/container/services_text_reparser.yml
Nicofuma: ^

s9e: This *should* be fixed.

s9e:
!unset WIP

s9e:
This can be merged anytime. I'd recommend waiting a bit before using it on A51 though. At least until other PRs related to text formatting are settled.

PR #3580, opened 2015-05-01
[ticket/13782] Use non-reserved word "void" for cache driver instead of null

marc1706:


marc1706:
If you don't like "void" we could also change it to something like "dummy".

rxu:
BTW, we have the same issue with ```/phpbb/log/null.php```.

s9e:
I'd be worried that `void` would become a keyword. How about `noop` for no-op. Does that make sense?

VSEphpbb:
Other options:
`null_cache`
`is_null`
`null_void`
`non`

marc1706:
I like the idea of calling it ```dummy``` more and more as that fits what it does pretty well.

phpBB/phpbb/cache/driver/void.php
```diff
@@ -16,7 +16,7 @@
 /**
 * ACM Null Caching
 */
-class null extends \phpbb\cache\driver\base
+class void extends \phpbb\cache\driver\base
```
Nicofuma: I think we should keep the old null class (extending the new one) for BC purpose

Nicofuma:
I like the idea of `dummy` too

marc1706:
Replaced by: https://github.com/phpbb/phpbb/pull/3584

PR #3581, opened 2015-05-01
[ticket/13804] Make template's user dependency optional

MateBartus:
[PHPBB3-13804](https://tracker.phpbb.com/browse/PHPBB3-13804)

### DIFF
--- a/phpBB/phpbb/template/twig/twig.php
+++ b/phpBB/phpbb/template/twig/twig.php
@@ -13,6 +13,8 @@
 
 namespace phpbb\template\twig;
 
+use phpbb\template\exception\user_object_not_available;
+
 /**
 * Twig Template class.
 */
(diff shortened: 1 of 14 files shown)


phpBB/phpbb/template/exception/user_service_not_found.php
```diff
@@ -0,0 +1,22 @@
...
+ * the docs/CREDITS.txt file.
+ *
+ */
+
+namespace phpbb\template\exception;
+
+/**
+ * This exception is thrown when the user object was not set but it is required by the called method
+ */
+class user_service_not_found extends \phpbb\exception\runtime_exception
```
Nicofuma: user_object_not_available or user_not_available

phpBB/phpbb/template/twig/twig.php
```diff
@@ -344,11 +353,20 @@ protected function get_template_vars()
...
 			array(
 				'definition'	=> new \phpbb\template\twig\definition(),
-				'user'			=> $this->user,
 				'loops'			=> $context_vars, // To get loops
 			)
 		);
 
+		if ($this->user instanceof \phpbb\user)
+		{
+			$vars = array_merge(
```
Nicofuma: `$vars['user'] = $this->user;` ?

MateBartus: My intention was to be able to keep `$context_vars['.'][0]['user']['something']` if that is something that is used. However it is not required I fix it

Nicofuma: Actually I don't see the difference....

Nicofuma: ok ^^

PR #3584, opened 2015-05-03
[ticket/13782] Rename classes that use reserved PHP7 keywords

marc1706:


phpBB/phpbb/cache/driver/dummy.php
```diff
@@ -0,0 +1,151 @@
...
+ * the docs/CREDITS.txt file.
+ *
+ */
+
+namespace phpbb\cache\driver;
+
+/**
+ * ACM Dummy Caching
+ */
+class dummy extends \phpbb\cache\driver\base
```
nickvergessen: using `git mv <old> <new>` preserves history and doesnt show the full file here

nickvergessen: Also I'd say `null_driver` is better then dummy. We use this work-around for interface already

phpBB/phpbb/cache/driver/null.php
```diff
@@ -15,137 +15,8 @@
 
 /**
 * ACM Null Caching
+* @deprecated 3.1.5 (to be removed: 3.3.0)
```
nickvergessen: 3.2, when we want to be php7 ready on release?

phpBB/phpbb/log/dummy.php
```diff
@@ -0,0 +1,81 @@
...
+* the docs/CREDITS.txt file.
+*
+*/
+
+namespace phpbb\log;
+
+/**
+* Dummy logger
+*/
+class dummy implements log_interface
```
nickvergessen: same here

phpBB/phpbb/cache/driver/dummy.php
marc1706: Which doesn't mean it's a good idea (for example it interferes with PSR-0).

phpBB/phpbb/cache/driver/null.php
marc1706: As I mentioned on IRC earlier, this doesn't break anything at all. The null driver will still work as it extends the dummy driver.

phpBB/phpbb/cache/driver/null.php
```diff
@@ -15,137 +15,8 @@
...
 */
-class null extends \phpbb\cache\driver\base
+class null extends \phpbb\cache\driver\dummy
```
nickvergessen: I think we can drop this file, since the value is coded into the config file we need to add a "workaround" there anyway, like with the namespaces: https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/phpbb/di/extension/config.php#L80-L88

phpBB/phpbb/cache/driver/dummy.php
marc1706: Also, calling it dummy is better for distinguishing between the PHP7 version and the old one.

nickvergessen:
I'd say fix this in 3.2 only, since 3.1 uses the `/e` modifier in bbcodes which breaks php7 anyway.

Then we also dont need the null implementations for backwards compatibility, since they break the language and disallows us to test-parse the files.

bantu:
Alternative name: noop

marc1706:
im-out-of-here-eccbc87e4b5ce2fe28308fd9f2a7baf3-348

Nicofuma:
So: no BC layer and `dummy` or something else?

marc1706:
I'd be ok with no BC layer and dummy for 3.2. The current null driver is only used in tests or the installation and I don't see a point where users would use it apart from that.

bantu:
Yes, no BC is fine. Dummy is fine too.

phpBB/phpbb/cache/driver/dummy.php
bantu: @nickvergessen I don't think git mv is necessary. Moves are autodetected as far as I remember. This difference here is that null.php is not actually removed.

marc1706:
Replaced by https://github.com/phpbb/phpbb/pull/3589

PR #3586, opened 2015-05-03
[ticket/13680] Updated quote notifications

s9e:
Added get_quote_authors() to text_formatter.utils service to retrieve the names used in first-level quotes.

https://tracker.phpbb.com/browse/PHPBB3-13680

### DIFF
--- a/phpBB/config/default/container/services_notification.yml
+++ b/phpBB/config/default/container/services_notification.yml
@@ -221,6 +221,8 @@ services:
             - %tables.notification_types%
             - %tables.notifications%
             - %tables.user_notifications%
+        calls:
+            - [set_utils, [@text_formatter.utils]]
         tags:
             - { name: notification.type }
 
--- a/phpBB/phpbb/notification/type/quote.php
+++ b/phpBB/phpbb/notification/type/quote.php
@@ -21,6 +21,11 @@ namespace phpbb\notification\type;
 class quote extends \phpbb\notification\type\post
 {
 	/**
+	* @var \phpbb\textformatter\utils_interface
+	*/
+	protected $utils;
+
+	/**
 	* Get notification type name
 	*
 	* @return string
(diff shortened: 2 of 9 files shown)


s9e:
I feel that the name get_quote_authors() is slightly misleading because it only returns the names used in the first level of quotes. I don't know if there's a better name for it though.

Elsensee:
> Fatal error: Call to a member function get_quote_authors() on a non-object in /home/travis/build/phpbb/phpbb/phpBB/phpbb/notification/type/quote.php on line 78

I can't think of a better name too. Maybe `get_first_quote_authors()` but I'm really not that good at naming stuff.. ^^

Nicofuma:
utils is static, I'd prefer to inject it as a service

About the name... I can't imagine a better name

s9e:
@Nicofuma It's a service, available as `text_formatter.utils` which is itself an alias to `text_formatter.s9e.utils`. The service was created as part of the s9e\TextFormatter PR, that's why it doesn't appear in the diff.

Nicofuma:
I was duped by the tests indeed

nickvergessen:
Would be nice if we can fix that on 3.1 aswell, do you have any idea?
https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/phpbb/notification/type/quote.php#L38
https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/phpbb/notification/type/quote.php#L80-L86

s9e:
I mentionned a possible algorithm in the A51 topic, but I haven't given it much thought.

> I think the way you'll have to fix it in 3.1 is you'll have to iterate over every quote tag with a regexp like `(\[/?quote(?:=&quot;.*?&quot;):\w+\])` and maintain a counter with the number of quote tags currently open. And of course, only notify the authors of the outermost quotes.

Nicofuma:
@nickvergessen what do you think about that? merge it now in 3.2 or wait for a patch against 3.1 and merge both together?

s9e:
Merge this one. 3.1 would require a completely different patch.

nickvergessen:
> Merge this one. 3.1 would require a completely different patch.

this

Nicofuma:
https://tracker.phpbb.com/browse/PHPBB3-13855 for 3.1.x

PR #3589, opened 2015-05-04
[ticket/13782] Rename classes that use reserved PHP7 keywords

marc1706:

### DIFF
--- a/phpBB/install/install_convert.php
+++ b/phpBB/install/install_convert.php
@@ -153,7 +153,7 @@ class install_convert extends module
 				unset($dbpasswd);
 
 				// We need to fill the config to let internal functions correctly work
-				$config = new \phpbb\config\db($db, new \phpbb\cache\driver\null, CONFIG_TABLE);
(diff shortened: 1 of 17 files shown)


timriker:
Will this change or a similar one get merged into 3.1.x?

marc1706:
It will not as phpBB 3.1.x would require adding feature changes like the new BBCode parser. That's what phpBB 3.2.x is for that is currently in its RC stage.

timriker:
copying phpbb/cache/driver/null.php to phpbb/cache/driver/dummy.php with class rename and editing install/install/install_update.php to reference dummy instead of null at least allowed me to get the upgrade run through on php7. Not ready to move my production instance to php 3.2.x but my development node is already running php7. It would be nice to get minimal php7 support in 3.1.x at least to support people migrating in the future.

nikitasius:
i confirm, copy&rename `null.php` in `dummy.php` (with renaming `null` to `dummy` in) and changing `/driver/null` to `/driver/dummy` in `install_convert.php` and `install_install.php` fixed for me this bug on php7 after installation.

PR #3592, opened 2015-05-04
[ticket/13388] Integrate routing and di parameters resolution

Nicofuma:
https://tracker.phpbb.com/browse/PHPBB3-13388

PHPBB3-13388

### DIFF
--- a/phpBB/phpbb/routing/router.php
+++ b/phpBB/phpbb/routing/router.php
@@ -15,6 +15,9 @@ namespace phpbb\routing;
 
 use Symfony\Component\Config\ConfigCache;
 use Symfony\Component\Filesystem\Exception\IOException;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\DependencyInjection\Exception\ParameterNotFoundException;
+use Symfony\Component\DependencyInjection\Exception\RuntimeException;
 use Symfony\Component\Routing\Matcher\Dumper\PhpMatcherDumper;
 use Symfony\Component\Routing\Generator\Dumper\PhpGeneratorDumper;
 use Symfony\Component\Routing\Matcher\UrlMatcher;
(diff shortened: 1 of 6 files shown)


phpBB/phpbb/routing/router.php
```diff
@@ -160,10 +170,118 @@ public function find($base_path = '')
...
+			foreach ($route->getDefaults() as $name => $value)
+			{
+				$route->setDefault($name, $this->resolve($value));
+			}
+
+			foreach ($route->getRequirements() as $name => $value)
+			{
+				if ($name === '_scheme' || $name === '_method')
+				{
+					continue; // ignore deprecated requirements to not trigger deprecation warnings
```
bantu: Unset using the key once instead of checking over and over again?
```php
$requirements = $route->getRequirements();
unset($requirements['_scheme']);
unset($requirements['_method']);
foreach ($requirements)
{
    $route->setRequirement($name, $this->resolve($value));
}
```

phpBB/phpbb/routing/router.php
```diff
@@ -286,8 +404,8 @@ public function get_generator()
 		{
 			return $this->generator;
 		}
-
-		$this->create_dumped_url_generator();
+		$this->create_new_url_generator();
+		//$this->create_dumped_url_generator();
```
bantu: ?

Nicofuma:
updated

phpBB/phpbb/routing/router.php
```diff
@@ -162,10 +171,117 @@ public function find($base_path = '')
...
+			foreach ($route->getMethods() as $method)
+			{
+				$methods = array_merge($methods, explode('|', $this->resolve($method)));
+			}
+
+			$route->setMethods($methods);
+			$route->setCondition($this->resolve($route->getCondition()));
+		}
+	}
+	/**
```
naderman: Missing newline?

naderman:
There isn't really any way to do this easily without making the router depend on the service container is there? Would be nice to avoid that in as many services as possible.

naderman:
Actually can't we just call getParameterBag and only pass the parameter bag to the router, rather than the entire container?

Nicofuma:
I took what was done in Symfony: https://github.com/symfony/symfony/blob/e756135f649314f810cf9370d0e0bf8d261740ed/src/Symfony/Bundle/FrameworkBundle/Routing/Router.php

About the parameter bag I don't think it's possible to inject the result of a method call...

naderman:
Yeah I don't really know how either, I guess this is fine. Just fix that newline and I'll merge it.

phpBB/phpbb/routing/router.php
Nicofuma: done

PR #3601, opened 2015-05-09
[ticket/13827] Correctly return JSON instead of HTML when valled from AJAX

nickvergessen:
https://tracker.phpbb.com/browse/PHPBB3-13827

### DIFF
--- a/phpBB/phpbb/controller/helper.php
+++ b/phpBB/phpbb/controller/helper.php
@@ -13,6 +13,7 @@
 
 namespace phpbb\controller;
 
+use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Generator\UrlGenerator;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/controller/helper.php
```diff
@@ -212,12 +212,29 @@ public function error($message, $code = 500)
 	public function message($message, array $parameters = array(), $title = 'INFORMATION', $code = 200)
 	{
 		array_unshift($parameters, $message);
+		$message_text = call_user_func_array(array($this->user, 'lang'), $parameters);
+		$message_title = $this->user->lang($title);
+
+		if ($this->request->is_ajax())
+		{
+			global $refresh_data;
```
bantu: can we avoid that?

nickvergessen: I guess we could, by creating a `redirect()` and `meta_refresh()` method in the controller helper. But that would break BC on the update from trigger_error to the controller helper way, since the old methods are no longer valid:
https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/includes/functions.php#L2478-L2487
https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/includes/functions.php#L4071-L4083

marc1706:
Typo in commit message. Should be called and not valled.

phpBB/phpbb/controller/helper.php
Nicofuma: couldn't we have a method to do both? (set a parameter and the global)

phpBB/phpbb/controller/helper.php
```diff
@@ -212,12 +212,29 @@ public function error($message, $code = 500)
...
+
+			$json_response = new \phpbb\json_response;
```
Nicofuma: `return new JsonResponse()` please

our old `\phpbb\json_response` is to be used with `trigger_error()` only (it breaks the flow by calling exit)

nickvergessen: Pushed

phpBB/phpbb/controller/helper.php
Nicofuma: so.. what do we want to do here finally?

nickvergessen: well I can add the methods and just make them call the function for now, so exts can already settle over. Then one day we dump the function. But I guess we still need the global?

Nicofuma: Fine by me, even with the global.

At least the dirty code is restricted to one specific place.

nickvergessen: So fine for now, other patch in another PR

PR #3610, opened 2015-05-15
[ticket/13311] Add core event on acp_forums for forum content deletes

Alf007:
Event added for the function delete_forum_content() in
includes/acp/acp_forums.php, so that you can insert any extension's table(s)
to be updated when forum content is deleted

PHPBB3-13311

PR #3611, opened 2015-05-15
Ticket/13841

Alf007:
Event added for the function move_topics() in
includes/functions_admin.php, so that you can insert any extension's table(s)
to be updated when topics are moved between forums

PHPBB3-13841


phpBB/includes/functions_admin.php
```diff
@@ -534,6 +534,19 @@ function move_topics($topic_ids, $forum_id, $auto_sync = true)
...
+
+	/**
+	 * Perform additional actions before topics move
+	 *
+	 * @event core.move_topics_before_query
+	 * @var	array	table_ary	Array of tables from which forum_id will be updated for all rows that hold the moved topics
+	 * @since 3.1.5-RC1
+	 */
+	$vars = array(
+			'table_ary',
```
nickvergessen: same here.
Please add all method parameters and other relevant variables to the event.

nickvergessen:
Contains commits from https://github.com/phpbb/phpbb/pull/3610 and therefor can not be merged before

Alf007:
I don't know how the commit from #3610 get there ?! It should not as these are independant changes ?! ^^

nickvergessen:
Merged via https://github.com/phpbb/phpbb/commit/7a55bcc0f3fc5a703b686affec413e891f558ff9

PR #3616, opened 2015-05-17
[ticket/13847] Move quote generation to text_formatter.utils

s9e:
https://tracker.phpbb.com/browse/PHPBB3-13847

The markup used for quotes, `[quote=".."]...[/quote]` is currently hardcoded in posting.php. This could be moved to a method in the `text_formatter.utils` service.

Ultimately, anything that manipulates markup should live in a `text_formatter.*` service.

### DIFF
--- a/phpBB/phpbb/textformatter/s9e/utils.php
+++ b/phpBB/phpbb/textformatter/s9e/utils.php
@@ -35,6 +35,43 @@ class utils implements \phpbb\textformatter\utils_interface
 	}
 
 	/**
+	* Return given string between quotes
+	*
+	* Will use either single- or double- quotes depending on whichever requires to be escaped.
+	* Quotes and backslashes are escaped with backslashes where necessary
+	*
+	* @param  string $str Original string
+	* @return string      Escaped string within quotes
+	*/
+	protected function enquote($str)
+	{
+		$quote = (strpos($str, '"') === false || strpos($str, "'") !== false) ? '"' : "'";
+
+		return $quote . addcslashes($str, '\\' . $quote) . $quote;
+	}
+
+	/**
+	* {@inheritdoc}
+	*/
+	public function generate_quote($text, array $attributes = array())
+	{
+		$quote = '[quote';
+		if (isset($attributes['author']))
+		{
+			// Add the author as the BBCode's default attribute
+			$quote .= '=' . $this->enquote($attributes['author']);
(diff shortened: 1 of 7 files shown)


phpBB/posting.php
```diff
@@ -1586,7 +1586,11 @@
 {
 	if ($config['allow_bbcode'])
 	{
-		$message_parser->message = '[quote=&quot;' . $post_data['quote_username'] . '&quot;]' . censor_text(trim($message_parser->message)) . "[/quote]\n";
+		$message_parser->message = $phpbb_container->get('text_formatter.utils')->generate_quote(
+			censor_text(trim($message_parser->message)),
+			array('author' => $post_data['quote_username'])
```
Nicofuma: couldn't we generate the quote with a link to the user profile?

s9e: Yes. You can put as much metadata as you want, and then use it in the template. I'm wary of overloading the UI though. There's [a topic about possible quote improvements](http://imgur.com/a/Ds9Y0) on A51.

Nicofuma: lets discuss that on a51 so

phpBB/phpbb/textformatter/s9e/utils.php
```diff
@@ -35,6 +35,43 @@ public function clean_formatting($xml)
...
+	*
+	* Will use either single- or double- quotes depending on whichever requires to be escaped.
+	* Quotes and backslashes are escaped with backslashes where necessary
+	*
+	* @param  string $str Original string
+	* @return string      Escaped string within quotes
+	*/
+	protected function enquote($str)
+	{
+		$quote = (strpos($str, '"') === false || strpos($str, "'") !== false) ? '"' : "'";
```
Nicofuma: and what if the string contains both quotes? you will generate `"simple-quote ' double-quote " end of string"`?

s9e: It uses singles quotes if the string contains a double quote and no single quotes, otherwise it uses double quotes. However, I'm going to change that in a second that that it uses whichever form is the shortest. Things that need to be escaped are always escaped regardless.

s9e:
@Nicofuma Can you merge this PR whenever you have time please? Other PRs will depend on it and since most of my PRs modify more or less the same files I'm getting conflicts when merging/rebasing.

PR #3617, opened 2015-05-18
[ticket/11530] Remove extra quotes when depth limit is exceeded

s9e:
https://area51.phpbb.com/phpBB/viewtopic.php?f=108&t=42779
https://tracker.phpbb.com/browse/PHPBB3-11530

Removes extra quotes before populating the posting form.
Removes extra quotes in the submitted text.
One caveat: does not remove extra quotes in the posting form if the quoted post is from the old parser. Fixes itself on preview.

### DIFF
--- a/tests/functional/posting_test.php
+++ b/tests/functional/posting_test.php
@@ -71,4 +71,85 @@ class phpbb_functional_posting_test extends phpbb_functional_test_case
 		$crawler = self::request('GET', "viewtopic.php?t={$post['topic_id']}&sid={$this->sid}");
 		$this->assertContains('&#128512;', $crawler->text());
 	}
+
+	/**
+	* @testdox max_quote_depth is applied to the text populating the posting form
+	*/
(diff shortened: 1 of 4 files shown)


tests/functional/posting_test.php
```diff
@@ -71,4 +71,85 @@ public function test_html_entities()
 		$crawler = self::request('GET', "viewtopic.php?t={$post['topic_id']}&sid={$this->sid}");
 		$this->assertContains('&#128512;', $crawler->text());
 	}
+
+	/**
+	* @testdox max_quote_depth is applied to the text populating the posting form
```
Nicofuma: `@testdox` ?

s9e: A description of the test as a declarative sentence, because it's hard to tell what is being tested just by looking at the code. It can also be used to generate nice checklists like [this one](https://github.com/s9e/TextFormatter/blob/master/docs/testdox.txt).

PR #3619, opened 2015-05-18
[ticket/13852] Consistent tab navigation when login in

javiexin:
Added tabindex attribute to input elements in index_body.html to make
tab navigation consistent with that of login_body.html.

PHPBB3-13852


Nicofuma:
merged via 1d2a127144476b29ebd37a7d239520d87bd93adc

PR #3626, opened 2015-05-23
[ticket/13733] Allow non-migration files inside migrations folder

imkingdavid:
This ignores any classes that don't extend the abstract \phpbb\db\migration\migration class.

https://tracker.phpbb.com/browse/PHPBB3-13733

### DIFF
--- a/phpBB/phpbb/extension/base.php
+++ b/phpBB/phpbb/extension/base.php
@@ -139,11 +139,21 @@ class base implements \phpbb\extension\extension_interface
 
 		foreach ($migrations as $key => $migration)
 		{
-			// If the class doesn't exist OR the class does not extend the migration class
-			// we need to skip it.
-			if (!class_exists($migration) || ($reflector = new \ReflectionClass($migration) && !$reflector->isSubclassOf('\phpbb\db\migration\migration'))) {
-				unset($migrations[$key]);
+			// If the class exists and is a subclass of the
+			// \phpbb\db\migration\migration abstract class
+			// we skip it.
+
+			// Otherwise, i.e. if it doesn't exist or it is
+			// not an extend the abstract class, we unset it
+			if (class_exists($migration)) {
+				$reflector = new \ReflectionClass($migration);
+				if ($reflector->isSubclassOf('\phpbb\db\migration\migration')) {
+					continue;
+				}
+
 			}
+
+			unset($migrations[$key]);
 		}
 
 		return $migrations;
--- a/phpBB/phpbb/extension/base.php
+++ b/phpBB/phpbb/extension/base.php
@@ -139,8 +139,9 @@ class base implements \phpbb\extension\extension_interface
 
 		foreach ($migrations as $key => $migration)
 		{
-			$reflector = new \ReflectionClass($migration);
-			if (!$reflector->isSubclassOf('\phpbb\db\migration\migration')) {
+			// If the class doesn't exist OR the class does not extend the migration class
+			// we need to skip it.
+			if (!class_exists($migration) || ($reflector = new \ReflectionClass($migration) && !$reflector->isSubclassOf('\phpbb\db\migration\migration'))) {
 				unset($migrations[$key]);
 			}
 		}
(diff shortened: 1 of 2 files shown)


phpBB/phpbb/extension/base.php
```diff
@@ -137,6 +139,30 @@ protected function get_migration_file_list()
...
+				// If the class exists and is a subclass of the
+				// \phpbb\db\migration\migration abstract class
+				// we skip it.
+
+				// Otherwise, i.e. if it doesn't exist or it is
+				// not an extend the abstract class, we unset it
+				if (class_exists($migration))
+				{
+					$reflector = new \ReflectionClass($migration);
+					if ($reflector->isSubclassOf('\phpbb\db\migration\migration'))
```
VSEphpbb: I think this needs to be `if ($reflection->IsInstantiable())`
Otherwise files that still extend the migration class but are meant to be extended themselves (like container_aware_migration) will still be run when they shouldn't.

imkingdavid: So would it be `if ($reflector->isSubclassOf('\phpbb\db\migration\migration') && $reflector->isInstantiable())`, correct?

VSEphpbb: No i do not think so. Just isInstantiable()
@Nicofuma ?

imkingdavid: The ticket specifies that devs should be able to "...keep migration helpers / abstract classes they may want to use in their migrations folder (like the way phpBB is able to do)."

If the only thing we check for here is instantiability then any "migration helpers" (i.e. classes that don't extend the migration base class) will be treated as migrations as well. Your use case is classes that "...still extend the migration class but are meant to be extended themselves"

My ticket currently covers both criteria. Or am I misunderstanding?

Nicofuma:
:+1:

phpBB/phpbb/extension/base.php
```diff
@@ -137,6 +139,30 @@ protected function get_migration_file_list()
 
 		$migrations = $this->extension_finder->get_classes_from_files($migrations);
 
+		if ($validate_classes)
+		{
+			foreach ($migrations as $key => $migration)
+			{
+				// If the class exists and is a subclass of the
+				// \phpbb\db\migration\migration abstract class
+				// we skip it.
```
VSEphpbb: This comment makes it sounds like you are going to skip classes that are valid subclasses.

How bout something like, 
> If a class does not exist or is not an extension of the \phpbb\db\migration\migration abstract class, unset it

bantu:
I am not sure this makes sense. First of all, we should check against an interface instead of inheritance (subclass). Second, I think non migration files should just be kept elsewhere such that they do not have to be scanned at all.

VSEphpbb:
@bantu it's not good that the migrations folder is straight up just for migrations only. For example, although it hasn't really been a problem, on a local environment on the Mac, hidden `.DS_store` files are considered migrations :fearful: and this change will actually help filter those out before sending them to the migrator.

But that aside, I have already come across a need in extensions to have a reusable abstract class for extending migration files, and it's much cleaner to keep such classes in the migrations folder (as phpBB does), since it is part of migrating, and not in the root or some other disconnected location.

phpBB/phpbb/extension/base.php
```diff
@@ -121,9 +121,11 @@ public function purge_step($old_state)
 	/**
 	* Get the list of migration files from this extension
 	*
+	* @var bool $validate_classes Whether or not to check that the migration
+	*		class exists and extends the base migration class.
 	* @return array
 	*/
-	protected function get_migration_file_list()
+	protected function get_migration_file_list($validate_classes = true)
```
bantu: Is there any legitimate use case for having this off, besides tests?

bantu:
@VSEphpbb .DS_store should be a non-issue as migrations have to be .php files. Another bug?

VSEphpbb:
@bantu I know its not an issue in that case, but it illustrates how anything currently in a migrations folder is attempted to be run, and will if it is a .php file.

nickvergessen:
Btw this is the reason why https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/phpbb/db/migration/profilefield_base_migration.php is not inside the folders of the migrations... its somewhat odd, so I think it would be nice to get this in.

Nicofuma:
@imkingdavid bump

marc1706:
Taking this over

PR #3630, opened 2015-05-25
Adding SCEditor to phpBB

brunoais:
This PR adds SCE to phpBB and many other WYSIWYG editor-decoupled features related to data management and translation for a WYSIWYG editor-specific.
This is definitely not complete but it is definitely good-enough for testing and for adapting.

**Styles team**: Get ready to check this on and adapt the style of SCE to work well with prosilver. You are way suitable for this than me.

PR TODO list:
- [x] Refine the UCP options and make them appear
- [x] Fix ACP options for the current idea (they are still outdated).
- [x] Fix ACP options' texts (they need a serious improvement).
- [x] Make the above selected options affect the buttons.
- [x] Make the above selected options affect the editor mode.
- [x] Implement the xsl:if tag more properly
- [ ] Remove the code used to recalculate for SCE BBCode on page load (useful when debugging, though)
- [ ] Documentation, documentation, documentation, documentation.
- [ ] Make tests to check for TextFormatter API (internal and external) changes that produce consequences to this.
- [ ] Make tests for all this code.
- [x] Make toggleable BBCode tags actually work. E.g. B, I, U. (53rd's the charm)
- [ ] Adapt to the new [quote] tag (deal with it)
- [ ] Some blocks are not easily enough killable `[code]` for example. After deleting, `<code>` comes back when typing.
- [x] Make tooltips show up correctly
- [ ] Give a GUI of override automatic settings to the admin in the ACP
- [ ] Allow overriding for more options that SCE allow.
- [ ] Add appropriate events to allow extensions to supercharge this functionality.

**NOTE:** Due to the current state and possible changes required, **rebases may happen from time-to-time**. 

https://tracker.phpbb.com/browse/PHPBB3-10418


brunoais:
!set wip

brunoais:
!set WIP

php-bb:
well done... it is great

phpBB/config/default/container/services_wysiwyg.yml
```diff
@@ -0,0 +1,35 @@
...
+    wysiwyg.converters.sce:
+        class: phpbb\bbcode\convert_editor\sce
+        arguments:
+            - @cache.driver
+            - %wysiwyg.cache.prefix%
+            - @config
+            - @dispatcher
+            - @service_container
+            - @request
+            - @template
```
nickvergessen: new line

phpBB/assets/javascript/editor.js
```diff
@@ -24,337 +6,382 @@ function helpline(help) {
...
-		theSelection = '';
-		return;
-	}
+var editor = {
 
-	//The new position for the cursor after adding the bbcode
-	var caret_pos = getCaretPosition(textarea).start;
-	var new_pos = caret_pos + bbopen.length;
+	tokenRegex: {
+		'ALPHANUM': /^[0-9A-Za-z]+$/,
```
callumacrae: Why quotes?

phpBB/assets/javascript/editor.js
```diff
@@ -24,337 +6,382 @@ function helpline(help) {
...
+	editor.insertBBCode(e.target.alt);
+	e.preventDefault();
+});
 
-	var textarea = document.forms[form_name].elements[text_name];
 
-	textarea.focus();
+var editorConstants = {
+	'NO_VALUE' 			: -1,
+	'VALUE_IN_CONTENT' 	: -2,
```
callumacrae: so this is kinda crazy

phpBB/assets/javascript/editor.js
```diff
@@ -24,337 +6,382 @@ function helpline(help) {
...
-function insert_text(text, spaces, popup) {
-	var textarea;
+	getElementDefaultDisplay: function () {
+		if(window.getComputedStyle){
+			return function (tagName){
+				var tag = document.createElement(tagName);
+				document.body.appendChild(tag);
+				var cStyle = window.getComputedStyle(tag, "").display;
+				document.body.removeChild(tag);
+				return cStyle;
```
callumacrae: Use jQuery?

phpBB/assets/javascript/editor.js
```diff
@@ -24,337 +6,382 @@ function helpline(help) {
...
+		if (standardsBrowser) {
+			var xslDoc = (new DOMParser).parseFromString(xsl, 'text/xml');
+
+			var processor = new XSLTProcessor();
+			processor.importStylesheet(xslDoc);
+
+			return {
+				'setParameter': function (name, value){
+					if(!value){
+						if(value === ''){
```
callumacrae: This will never happen as `!''` evaluates to false

phpBB/assets/javascript/editor.js
```diff
@@ -24,337 +6,382 @@ function helpline(help) {
...
+					processor.setParameter(null, name, value);
+				},
+
+				'transformToFragment' : function (xml, onDocument){
+					var xmlDoc = (new DOMParser).parseFromString(xml, 'text/xml');
+					// NOTE: importNode() is used because of https://code.google.com/p/chromium/issues/detail?id=266305
+					return onDocument.importNode(processor.transformToFragment(xmlDoc, onDocument), true);
+				}
+			};
+		}else{
```
callumacrae: I thought we don't support IE<9 anymore?

phpBB/assets/javascript/editor.js
```diff
@@ -24,337 +6,382 @@ function helpline(help) {
...
+			};
+		}
+	},
+
+	/**
+	 * Insert HTML in the editor at the selection point
+	 *
+	 */
+	insertHTML: function (editor, start, end){
+		alert("Misconfigured editor. The editor setup code does not override editor.insertHTML");
```
callumacrae: Use console.warn or a modal, not alert()

phpBB/assets/javascript/editor.js
brunoais: Why not?
It is just a standard I follow. If you don't want to, I may remove them.

phpBB/assets/javascript/editor.js
brunoais: What is?

phpBB/assets/javascript/editor.js
brunoais: jQuery does extra work to get values and, sometimes, doesn't get the value I want and gets something else instead. I want it to fail if it would be getting the value through those alternative ways.

phpBB/assets/javascript/editor.js
brunoais: `!''` is `true`, not `false`

phpBB/assets/javascript/editor.js
brunoais: It is proposed, there's a ticket for it, it hasn't been accepted yet. That code predates the time of the proposal. I'll remove it if the proposal gets accepted.

phpBB/assets/javascript/editor.js
brunoais: I want to use `alert()` because I want it to be as intrusive as possible. I want whomever works on it or the user to get a "slap in the face" that it is wrong as it currently is.
`console.warn` is too hidden, most popin are not intrusive enough.

phpBB/assets/javascript/editor.js
marc1706: ```foo = ''; console.log(!foo);``` outputs true.

phpBB/assets/javascript/editor.js
callumacrae: ??!!?!?!?!

why would you want to slap the user in the face if a dev breaks something?

phpBB/assets/javascript/editor.js
callumacrae: d'oh, misunderstood what this code was trying to do

phpBB/assets/javascript/editor.js
callumacrae: the spacing! you're not putting whitespace before your colons in other places, why are you here?

phpBB/assets/javascript/editor.js
callumacrae: There's no reason to use them, so it's better to avoid escaping properties where possible :smiley:

phpBB/assets/javascript/editor.js
brunoais: First, the dev would be slapped if he minimally tested his code.
If that fails, the user would slap the dev for not testing the code.

phpBB/assets/javascript/editor.js
brunoais: Good point. Done.

phpBB/assets/javascript/editor.js
brunoais: Escaping?

callumacrae: quoting. https://mathiasbynens.be/notes/javascript-properties

brunoais: I'm not escaping. I specifying the properties inside single quotes...

callumacrae: pls stop arguing about semantics, I already corrected myself

brunoais: Eh... Sorry.  I misread. Anyway, why should I remove them? Is that in the coding guidelines?

callumacrae: You should remove them because they're not used anywhere else and good coders write consistent code.

It's also good practice and one of the defaults in JSHint.

phpBB/assets/javascript/editor.js
VSEphpbb: spaces should go between your if's and else's and brackets or parens

phpBB/assets/javascript/editor.js
brunoais: > good coders write consistent code.
I write my code consistently.
I didn't read anything about that on any guidelines but ok...
On another note, by the guidelines, when should quotes be used?

phpBB/assets/javascript/editor.js
```diff
@@ -24,337 +6,382 @@ function helpline(help) {
...
-* Insert text at position
-*/
-function insert_text(text, spaces, popup) {
-	var textarea;
+	getElementDefaultDisplay: function () {
+		if(window.getComputedStyle){
+			return function (tagName){
+				var tag = document.createElement(tagName);
+				document.body.appendChild(tag);
+				var cStyle = window.getComputedStyle(tag, "").display;
```
marc1706: Please don't define variables in the middle of the function. That's IMHO just a bad coding practice and other real code languages would actually complain about this.

phpBB/assets/javascript/editor.js
callumacrae: You don't write your code consistently to what is already in the codebase, which is what is important!

We should get a linter, check out this topic: https://area51.phpbb.com/phpBB/viewtopic.php?f=99&t=47906

phpBB/assets/javascript/editor.js
callumacrae: It's completely fine in JavaScript. They're hoisted anyway

callumacrae: Should be using single quotes though

phpBB/assets/javascript/editor.js
marc1706: A dev would see that in his browser's console. Never use alert() for something target at the end user.

phpBB/assets/javascript/editor.js
brunoais: For that, I use the same rule as for Java:
Define the variable when first used unless scope doesn't let it.

phpBB/assets/javascript/editor.js
brunoais: It is targeted towards the dev. The user getting it is a last resort that should not happen.

callumacrae: If it is targeted towards the dev, just use the console…

brunoais: Meh... OK... suit yourself...

phpBB/assets/javascript/editor.js
brunoais: I'm OK with that. As long as the chosen limitations don't get more in the way that I nearly have to write the code twice.

phpBB/includes/acp/acp_board.php
```diff
@@ -909,6 +912,119 @@ function board_disable($value, $key)
...
+			foreach ($set_data as $name => $new_data)
+			{
+				if ($editor_data[$name]['WYSIWYG_MODES_AVAILABLE'] & $new_data['default_mode'])
+				{
+					$new_default_modes[$name]							= $new_data['default_mode'];
+					$editor_new_data[$name]['S_WYSIWYG_MODE']			= $new_data['default_mode'];
+				}
+				if ($editor_data[$name]['WYSIWYG_BUTTON_MODES_AVAILABLE'] & $new_data['default_button_mode'])
+				{
+					$new_button_modes[$name]							= $new_data['default_button_mode'];
```
brunoais: I'm unsure if this is a right way of doing it because the same data is in multiple places.
I'll think about it later.

phpBB/docs/events.md
```diff
@@ -16,6 +16,12 @@ acp_bbcodes_edit_fieldsets_after
 * Since: 3.1.0-a3
 * Purpose: Add settings to BBCode add/edit form
 
+acp_board_extra_wysiwyg_fields'
```
nickvergessen: remove trailing `'`

phpBB/docs/events.md
```diff
@@ -854,27 +860,6 @@ posting_editor_bbcode_status_after
 * Since: 3.1.4-RC1
 * Purpose: Add content after bbcode status
 
-posting_editor_buttons_after
```
nickvergessen: can we avoid deleting events? If not we need to create a list some where of events we removed.

phpBB/includes/acp/acp_board.php
```diff
@@ -909,6 +912,119 @@ function board_disable($value, $key)
 	}
 
 	/**
+	* Default settings for the WYSIWYG editors
+	*/
+	function wysiwyg_converters($submit)
```
nickvergessen: add visibility

brunoais: Oh right... Due to the others I forgot...
Fixed.

phpBB/config/default/container/services_wysiwyg.yml
brunoais: Between the arguments list and the tags? I've seen some of the others in different files and I don't see that happening.

phpBB/docs/events.md
brunoais: The events are being deleted because `posting_buttons.html` file is a goner. There's really no viable alternative.

phpBB/includes/ucp/ucp_prefs.php
```diff
@@ -460,6 +461,132 @@ function main($id, $mode)
...
 					'S_NOTIFY'	=> $data['notify'])
 				);
+
+			break;
+
+			case 'wysiwyg':
+
+				$data = array(
+					'wysiwyg_editor'	=> $request->variable('wysiwyg_editor', $user->data['user_wysiwyg_editor']),
+				);
```
brunoais: I'm unsure about the code design in here. It doesn't seem to be correct. I've been thinking on many ways of doing this but none seem to be correct.
The objective is to show the options to the user according to the options he has chosen (if they are valid).
However, this code doesn't seem to be in the best way I could do this. Help and suggestions are appreciated.

phpBB/styles/prosilver/template/ucp_prefs_wysiwyg.html
```diff
@@ -0,0 +1,60 @@
...
+		</dd>
+	</dl>
+	<!-- EVENT ucp_prefs_wysiwyg_prepend -->
+	<!-- BEGIN wysiwyg -->
+		<div id="wysiwyg_{wysiwyg.EDITOR_TITLE}_settings_container">
+			<!-- EVENT ucp_prefs_wysiwyg_options_prepend -->
+			<dl>
+				<dt><label>{L_WYSIWYG_DEFAULT_MODE}{L_COLON}</label></dt>
+				<dd>
+					<label><input type="radio" name="wysiwyg_default_mode[{{ wysiwyg.EDITOR_ID }}]" value="{{ constant('\\phpbb\\bbcode\\convert_editor\\base::DEFAULT_WYSIWYG_MODE') }}"{% if wysiwyg.S_WYSIWYG_MODE b-and constant('\\phpbb\\bbcode\\convert_editor\\base::DEFAULT_WYSIWYG_MODE') %} checked{% endif %}{% if not (wysiwyg.WYSIWYG_MODES_AVAILABLE b-and constant('\\phpbb\\bbcode\\convert_editor\\base::DEFAULT_WYSIWYG_MODE')) %} disabled{% endif %}/> {L_WYSIWYG_DEFAULT_MODE_WYSIWYG}</label>
```
brunoais: Does it make sense to have lines this long? Do I need to make them shorter? I may need to add more template logic, depending on the suggestions about the code in the server, though....

nickvergessen: Add line breaks inside the `<input />`

brunoais: 1 line break for each parameter? That seems reasonable. What do you think?

brunoais:
I just made a rebase because merging was failing.
Please confirm if I did it the right way. I tried to deal with all conflicts the best way I could think of based on the phpBB version in the master branch.

callumacrae:
Still merge conflicts though

brunoais:
Damn!
Ok... re-rebasing, then

brunoais:
Sounds like it is back to normal now.

phpBB/includes/acp/acp_board.php
```diff
@@ -925,6 +928,121 @@ function board_disable($value, $key)
...
+		* Allow manipulating the ACP data for the WYSIWYG editor defaults
+		*
+		* NOTE: The variables starting with "new_" and editor_new_data only make sense if submit is true
+		* NOTE2: The variables with new data only differ from the set_data if they contain invalid data.
+		*
+		* Below, "DI" stands for Dependency Injection(or)
+		*
+		* @event core.acp_board_wysiwyg
+		* @var	bool	submit					Is this a submission?
+		* @var	array	set_data				The data resulting from the form submission
```
brunoais: What's invalid on this line?

brunoais: It wasn't on this line O_O.

Nicofuma:
bump, what is the status here?

brunoais:
4 failed solutions and trying a new solution.
The selection the browser does in some interactions is not intuitive, unfortunately...
I'm finished making another new algorithm that should work. I'll try it later today. Yesterday my day was a mess.

brunoais:
This solution seems to be doing the trick... At least when there's no complex HTML to work on.
This should be a success of phase 1 with this. Please test it.
Only single HTML tag HTML, for now. I.e. **b**, *i*, u, img, url, email, color.
http://brunoaisserver.dtdns.net/phpBBWYSIWYG/

brunoais:
Huge question:
How should all this work when BBCode is turned off?
My current idea is either:
1. To simply not show the editor and add add an additional js file with all the necessary functions for all operations to do their job.
**OR**
2. Have all the dormant code inside the `editor.js` file like it is now.

The situation is that, with SCE doing most of the work on normalization and manipulation of data, most of the code related with textarea manipulation in the current `editor.js` is useless.  
Also, I believe loading SCE just to use the textarea with text (not even BBCode) is a waste on the browser's resources.

What do you think on this subject?

Nicofuma:
To me it doesn't make any sense to load anything bbcode related when it is globally disabled.

brunoais:
@Nicofuma For that, it is required to split `editor.js` into 2 or 3 parts. Do you want that made?

Nicofuma:
I don't know what is needed or how the js is organized.
What I know is that it doesn't make any sense to display bbcode buttons when bbcode is globally disabled

callumacrae:
please don't split it apart until we have a way to concatenate files

brunoais:
@callumacrae OK. In that case, I'll put it all in a single file

brunoais:
(From IRC)
>\<brunoais\> On another note, I may not be able to get SCE good enough for 3.2 (IMO). Work has become a much larger pressure it has been previously so I won't have enough time to dedicate to SCE + phpBB integration as I wanted
\<brunoais\> The new quote tag also helped with that as I believe it requires hardcoding which is something I really wanted to avoid
\<brunoais\> Not to say that I'm still not fully sure on the hardcoding it requires
\<brunoais\> Worst of all, though, is the browser bugs with the contentEditable. I got some too many surprises

CHItA:
I'm closing this PR for now, please feel free to reopen it anytime you want to restart working on it.

carpcharacin:
This is something I would like to see added to 3.3/4.0.

callumacrae:
@CarpCharacin: I'm looking forward to seeing your pull request.

CarpCharacin:
@brunoais I think that you are over complicating the integration.  I think that we should try to set up a basic integration that works, but is lightweight and is not more complicated than it needs to be.  I am going to try to set up a basic integration at www.fynnmood.sg and then maybe create a new pull request.

brunoais:
@CarpCharacin Don't forget to keep the minimum set of features that are required from the user's perspective.
Oh! And if you are going to re-implement the extension that already exists, don't bother with it.

CarpCharacin:
@brunoais What are the minimum set of features that are required?  Yes there is an extension that implements it, but I want to create a pull request so it has to be implemented directly, not through an extension.

brunoais:
@carpCharacin
Quite straightforward, actually. It only has 2 requirements:

1. It must display the post being edited the same way as it will show up after it is posted.
2. It must follow the expectations of a user's expectations of a WYSIWYG document. But this one is pretty much guaranteed by the editor itself without you doing anything.

As you can see. It is quite simple/short requirement.

brunoais:
@carpCharacin
Quite straightforward, actually. It only has 2 requirements:

1. It must display the post being edited the same way as it will show up after it is posted.
2. It must follow the expectations of a user's expectations of a WYSIWYG document. But this one is pretty much guaranteed by the editor itself without you doing anything.

As you can see. It is quite simple/short requirement.

CarpCharacin:
I don't think I have the coding skills to do an "advanced" integration where the user can enable/disable it, but those requirements are very basic.  I just need to know the templates to put the wysiwyg editor code into.

brunoais:
Just read this PR's core in the `*.html` files

Persian-phpBB:
is it open again?

PR #3639, opened 2015-05-28
[ticket/13875] Ignore cache, ext, and store folder in lint test

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-13875

### DIFF
--- a/tests/lint_test.php
+++ b/tests/lint_test.php
@@ -40,6 +40,8 @@ class phpbb_lint_test extends phpbb_test_case
 			dirname(__FILE__) . '/../build/new_version',
 			dirname(__FILE__) . '/../build/old_versions',
 			dirname(__FILE__) . '/../phpBB/cache',
+			dirname(__FILE__) . '/../phpBB/ext',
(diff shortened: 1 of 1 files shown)


phpunit.xml.dist
```diff
@@ -22,6 +22,9 @@
         </testsuite>
         <testsuite name="phpBB Lint Test">
             <file>./tests/lint_test.php</file>
+            <exclude>./phpBB/cache</exclude>
```
nickvergessen: https://github.com/phpbb/phpbb/blob/3.1.x/tests/lint_test.php#L82-L89

Nicofuma:
And for the vendors? What do we want to do?

marc1706:
@Nicofuma vendors are already excluded :palm_tree:

bantu:
Will extension authors complain about the ext directory no longer being covered?

nickvergessen:
I think they should use a separate one

bantu:
Aye!

PR #3640, opened 2015-05-28
[ticket/13876] Use webfontloader asyncronously.

callumacrae:
This PR adds typekit and Google's webfontloader and calls it
asyncronously, removing the SPOF in the header. Now, if google
goes down, some buttons display with a different font.

https://github.com/typekit/webfontloader

https://tracker.phpbb.com/browse/PHPBB3-13876


marc1706:
Is there a reason we shouldn't be using https://ajax.googleapis.com/ajax/libs/webfont/1/webfont.js instead of fixing it at 1.5.8?

marc1706:
Also, don't you need to set ```wf.async = 'true';```?

callumacrae:
1/ has a shorter cache time, which is why I'm not using it. I'll set async to true.

callumacrae:
Done

PR #3644, opened 2015-05-28
[ticket/13878] Display background images when printing topic in webkit

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-13878


callumacrae:
:tada: 

screen shot 2015-05-28 at 15 03 40

marc1706:
Only does it for firefox & safari. Still better than before. :)
firefox

callumacrae:
Anything else would be a hack. Better to fix something than nothing.

http://stackoverflow.com/questions/6670151/how-can-i-force-browsers-to-print-background-images-in-css

PR #3647, opened 2015-05-28
[ticket/13799] Do not rawurlencode category in gallery avatar

marc1706:
Otherwise the directory seperator will incorrectly be transformed to its
HTML replacement.

Ticket: https://tracker.phpbb.com/browse/PHPBB3-13799

PHPBB3-13799

### DIFF
--- a/phpBB/phpbb/avatar/driver/local.php
+++ b/phpBB/phpbb/avatar/driver/local.php
@@ -182,7 +182,7 @@ class local extends \phpbb\avatar\driver\driver
 					}
 					$cat = ($path == $file_path) ? $user->lang['NO_AVATAR_CATEGORY'] : str_replace("$path/", '', $file_path);
 					$avatar_list[$cat][$image] = array(
-						'file'      => ($cat != $user->lang['NO_AVATAR_CATEGORY']) ? rawurlencode($cat) . '/' . rawurlencode($image) : rawurlencode($image),
+						'file'      => ($cat != $user->lang['NO_AVATAR_CATEGORY']) ? $cat . '/' . rawurlencode($image) : rawurlencode($image),
 						'filename'  => rawurlencode($image),
 						'name'      => ucfirst(str_replace('_', ' ', preg_replace('#^(.*)\..*$#', '\1', $image))),
 						'width'     => $dims[0],
--- a/phpBB/phpbb/avatar/driver/local.php
+++ b/phpBB/phpbb/avatar/driver/local.php
@@ -84,11 +84,13 @@ class local extends \phpbb\avatar\driver\driver
 					'AVATAR_IMAGE'  => $this->phpbb_root_path . $this->config['avatar_gallery_path'] . '/' . $img['file'],
 					'AVATAR_NAME' 	=> $img['name'],
 					'AVATAR_FILE' 	=> $img['filename'],
+					'CHECKED'			=> $img['file'] === $row['avatar'],
 				));
 
 				$template->assign_block_vars('avatar_local_row.avatar_local_option', array(
 					'AVATAR_FILE' 		=> $img['filename'],
-					'S_OPTIONS_AVATAR'	=> $img['filename']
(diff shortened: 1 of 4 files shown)


bantu:
I am not sure this is a correct fix. Both $image and $cat are based on data returned from the iterator, but only one is urlencoded post-patch.

marc1706:
So yeah, the selected avatar never showed as selected either. Fixed that now, too.

phpBB/styles/prosilver/template/ucp_avatar_options_local.html
```diff
@@ -10,7 +10,7 @@
 <!-- BEGIN avatar_local_row -->
 <!-- BEGIN avatar_local_col -->
 	<label for="av-{avatar_local_row.S_ROW_COUNT}-{avatar_local_row.avatar_local_col.S_ROW_COUNT}"><img src="{avatar_local_row.avatar_local_col.AVATAR_IMAGE}" alt="" /><br />
-	<input type="radio" name="avatar_local_file" id="av-{avatar_local_row.S_ROW_COUNT}-{avatar_local_row.avatar_local_col.S_ROW_COUNT}" value="{avatar_local_row.avatar_local_col.AVATAR_FILE}" /></label>
+	<input type="radio" name="avatar_local_file" id="av-{avatar_local_row.S_ROW_COUNT}-{avatar_local_row.avatar_local_col.S_ROW_COUNT}" value="{avatar_local_row.avatar_local_col.AVATAR_FILE}"<!-- IF avatar_local_row.avatar_local_col.CHECKED -->checked="checked"<!-- ENDIF --> /></label>
```
nickvergessen: missing space before `checked`
mind breaking the line aswell?

phpBB/styles/subsilver2/template/ucp_avatar_options_local.html
```diff
@@ -19,7 +19,7 @@
 			</tr>
 			<tr>
 				<!-- BEGIN avatar_local_option -->
-					<td class="row2" align="center"><input type="radio" class="radio" name="avatar_local_file" id="av-{avatar_local_row.S_ROW_COUNT}-{avatar_local_row.avatar_local_col.S_ROW_COUNT}" value="{avatar_local_row.avatar_local_option.S_OPTIONS_AVATAR}" /></td>
+					<td class="row2" align="center"><input type="radio" class="radio" name="avatar_local_file" id="av-{avatar_local_row.S_ROW_COUNT}-{avatar_local_row.avatar_local_col.S_ROW_COUNT}" value="{avatar_local_row.avatar_local_option.S_OPTIONS_AVATAR}"<!-- IF avatar_local_row.avatar_local_option.CHECKED -->checked="checked"<!-- ENDIF --> /></td>
```
nickvergessen: same here

phpBB/phpbb/avatar/driver/local.php
```diff
@@ -182,7 +184,7 @@ protected function get_avatar_list($user)
 					}
 					$cat = ($path == $file_path) ? $user->lang['NO_AVATAR_CATEGORY'] : str_replace("$path/", '', $file_path);
 					$avatar_list[$cat][$image] = array(
-						'file'      => ($cat != $user->lang['NO_AVATAR_CATEGORY']) ? rawurlencode($cat) . '/' . rawurlencode($image) : rawurlencode($image),
+						'file'      => ($cat != $user->lang['NO_AVATAR_CATEGORY']) ? str_replace('%2F', '/', rawurlencode($cat)) . '/' . rawurlencode($image) : rawurlencode($image),
```
nickvergessen: the user can not excape the path now, by specifying `./../../` etc?

marc1706: No. The image list is retrieved from the filesystem. The user can only selected those files. Trying to select categories that do not exist in this list or images that do not exist in that list, e..g by using ```../../``` will result in an empty list for the categories and an error message for the image.

Nicofuma:
unrelated faillure

PR #3649, opened 2015-05-28
[ticket/13879] Replaced px with em where appropriate.

callumacrae:
We're using ems basically everywhere else so ¯\\\_(ツ)_/¯

This will make it much easier for people to change the font-size of their board. I can now double the size of all the text in my board by just increasing the font-size of the body.

https://tracker.phpbb.com/browse/PHPBB3-13879

### DIFF
--- a/phpBB/styles/prosilver/theme/buttons.css
+++ b/phpBB/styles/prosilver/theme/buttons.css
@@ -5,8 +5,7 @@
 	cursor: pointer;
 	display: inline-block;
 	height: 18px;
-	line-height: 18px;
-	font-size: 13px;
+	font-size: 1.2em;
 	white-space: nowrap;
 	border: 1px solid transparent;
 	border-radius: 4px;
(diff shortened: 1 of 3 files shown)


Louis7777:
I was wondering. How about using rems (root ems) instead of ems? Would there be any downsides?

callumacrae:
Yeah, not supported in IE8. If we were running the CSS through a built tool it would be fine because we could just get it to turn them into px, but we're not.

callumacrae:
Anyone going to merge this, or are we going to let it get merge conflicts and close it in six months? :smile:

phpBB/styles/prosilver/theme/buttons.css
```diff
@@ -5,8 +5,7 @@
 	cursor: pointer;
 	display: inline-block;
 	height: 18px;
-	line-height: 18px;
```
marc1706: Is line-height not needed due to the height parameter?

callumacrae: It's not needed due to the `font-size`, iirc.

RMcGirr83:
@marc1706 

http://area51.phpbb.com/phpBB/viewtopic.php?f=126&t=48186

This broke the responsive layout and needs to be fixed/reverted

callumacrae:
Fixed, not reverted, obviously.

Nicofuma:
https://github.com/phpbb/phpbb/pull/3772

PR #3650, opened 2015-05-28
Update mcp_reports.php [skip ci]

cybernet:
nothing much :)


callumacrae:
It needs to be in the commit message

http://docs.travis-ci.com/user/how-to-skip-a-build/

marc1706:
Definitely not going to happen in 3.1.x and I'm not sure whether we want to merge something that is just whitespace changes (nothing is broken here really, it's just tiny cosmetics).

bantu:
This seems to fix indentation on github, but breaks indentation in my editor (Kate). Maybe someone can work out the details on the difference between the two.

marc1706:
This actually breaks indentation as @bantu said. It was better before.

Elsensee:
@bantu Github uses a tab size of 8, while we use a tab size of 4. :/ I think there are some browser extensions which can fix that. github-forcetabsize for Chrome used to work in the past for me but somehow it has stopped working. "Github Tab Size" for Chrome should work though.

nickvergessen:
https://github.com/phpbb/phpbb/pull/3650/files?ts=4 no need for a plugin ;)

Sumanai:
> I think there are some browser extensions which can fix that.

I use stilish and behold the following code:
```
@namespace url(http://www.w3.org/1999/xhtml);

@-moz-document domain("github.com") {
.tab-size-8{-moz-tab-size:4;tab-size:4}
}
```

PR #3652, opened 2015-05-28
[ticket/12143] Make it possible to use translated group names for non-special groups

Elsensee:
Yeah, so I did this... out of the old PR #2247 

I had to do a new method in language service - I think it should be fine to do that in this PR.

I want to create another class like this, which is based on the function get_username_string, but splitted up to several functions like get_name, get_color, etc.
These methods (get_name, get_color) can be added to this class to at a later point too, I guess.

[PHPBB3-12143](https://tracker.phpbb.com/browse/PHPBB3-12143)

### DIFF
--- a/phpBB/config/default/container/services.yml
+++ b/phpBB/config/default/container/services.yml
@@ -113,6 +113,11 @@ services:
     file_downloader:
         class: phpbb\file_downloader
 
+    group_helper:
+        class: phpbb\group\helper
+        arguments:
+            - @language
+
     log:
         class: phpbb\log\log
         arguments:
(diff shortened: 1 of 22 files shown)


Nicofuma:
https://travis-ci.org/phpbb/phpbb/jobs/64566865#L1174-L1186

Nicofuma:
bump

Nicofuma:
you need to rebase again ;)

Elsensee:
*sigh* rebased.. :P

Nicofuma:
looks good

PR #3655, opened 2015-05-28
[ticket/13881] Ported changes from PHPBB3-13880

s9e:
https://tracker.phpbb.com/browse/PHPBB3-13881

Fixes an issue in 3.2 where deep quotes are not immediately removed when quoting old posts, because of the different BBCode engines.


Nicofuma:
please repush to trigger the builds

PR #3663, opened 2015-05-30
[ticket/13833] Prevent flooding if type="submit" doesn't exist

Elsensee:
[PHPBB3-13833](https://tracker.phpbb.com/browse/PHPBB3-13833)

### DIFF
--- a/phpBB/posting.php
+++ b/phpBB/posting.php
@@ -1241,6 +1241,7 @@ if ($submit || $preview || $refresh)
 	*
 	* @event core.posting_modify_submission_errors
 	* @var	array	post_data	Array with post data
+	* @var	array	poll		Array with poll data from post (must be used instead of the post_data equivalent)
 	* @var	string	mode		What action to take if the form is submitted
 	*				post|reply|quote|edit|delete|bump|smilies|popup
 	* @var	string	page_title	Title of the mode page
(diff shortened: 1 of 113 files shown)


phpBB/posting.php
```diff
@@ -43,6 +42,7 @@
 $cancel		= (isset($_POST['cancel']) && !isset($_POST['save'])) ? true : false;
 
 $refresh	= (isset($_POST['add_file']) || isset($_POST['delete_file']) || isset($_POST['cancel_unglobalise']) || $save || $load || $preview);
+$submit		= $request->is_set_post('post') && !$refresh;
```
nickvergessen: check preview here aswell?

Elsensee: Well, if `$preview` is true, `$refresh` will be true too. But I guess it wouldn't hurt to change it to check `$preview` as well here. I think it may improve the readability, so... yeah.. :D

nickvergessen: And it helps in case of "copying code" ;)

Elsensee:
Updated.

phpBB/posting.php
```diff
@@ -43,6 +42,7 @@
 $cancel		= (isset($_POST['cancel']) && !isset($_POST['save'])) ? true : false;
 
 $refresh	= (isset($_POST['add_file']) || isset($_POST['delete_file']) || isset($_POST['cancel_unglobalise']) || $save || $load || $preview);
+$submit		= $request->is_set_post('post') && !$refresh && !$preview;
```
Nicofuma: extra tab here

PR #3666, opened 2015-05-30
[ticket/9252] Do not try to disapprove a post that was just approved

marc1706:
Without this change, posts that are approved by someone right before
another moderator is trying to disapprove a post are actually deleted
instead of disapproved.

https://tracker.phpbb.com/browse/PHPBB3-9252

PHPBB3-9252


Nicofuma:
:+1: 
@nickvergessen

PR #3667, opened 2015-05-30
[ticket/9405] Only increase last_post_time when bumping topic

marc1706:
There is no need to update the last post's post time. After this
change, the topic/post is still properly reported as new (bumped).

https://tracker.phpbb.com/browse/PHPBB3-9405

PHPBB3-9405


nickvergessen:
Well the difference is that a "forum/topic sync" now resets the bump
Also the unread link is meh now?

marc1706:
In that case it might be better to just live with this "issue" and close the ticket.

marc1706:
Closing as such

PR #3672, opened 2015-05-31
[PHPBB3/13903] replacing regexp path naming for container by md5

Wardormeur:
[PHPBB3/13903] Replacement of the path creation based on a blacklist by a md5 filename to support further configurations

PHPBB3-13903

### DIFF
--- a/phpBB/phpbb/di/container_builder.php
+++ b/phpBB/phpbb/di/container_builder.php
@@ -399,7 +399,6 @@ class container_builder
 	*/
 	protected function get_container_filename()
 	{
-		$filename = str_replace(array('/', '.'), array('slash', 'dot'), $this->phpbb_root_path);
-		return $this->phpbb_root_path . 'cache/container_' . $filename . '.' . $this->php_ext;
+		return $this->phpbb_root_path . 'cache/container_' . md5($filename) . '.' . $this->php_ext;
 	}
 }
--- a/phpBB/phpbb/di/container_builder.php
+++ b/phpBB/phpbb/di/container_builder.php
@@ -399,6 +399,6 @@ class container_builder
 	*/
 	protected function get_container_filename()
 	{
-		return $this->phpbb_root_path . 'cache/container_' . md5($filename) . '.' . $this->php_ext;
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/di/container_builder.php
```diff
@@ -399,7 +399,6 @@ protected function inject_custom_parameters()
 	*/
 	protected function get_container_filename()
 	{
-		$filename = str_replace(array('/', '.'), array('slash', 'dot'), $this->phpbb_root_path);
-		return $this->phpbb_root_path . 'cache/container_' . $filename . '.' . $this->php_ext;
+		return $this->phpbb_root_path . 'cache/container_' . md5($filename) . '.' . $this->php_ext;
```
marc1706: $filename is undefined now

Nicofuma:
The commit message is still wrong ;)

It should be:
```
[ticket/13903] Replacing regexp path naming for container by md5
Replacement of the path creation based on a blacklist by a md5 filename to support further configurations

PHPBB3-13903
```

Nicofuma:
otherwise +1

Wardormeur:
I'm amazed to see he was still able to create a container without the filename (md5 of null?). Fixing it atm, sorry about that.

Nicofuma:
;)

nickvergessen:
Why is this better then the version from before?
The previous one allowed you to quickly identify which container is for which path?

Wardormeur:
Because it disallows container creation with windows full path, as the ':' isn't parsed. So either you thinks of looking for every "edge" case and parse it, either you go for a more robust and all-around solution

bantu:
I very much prefer this version over the old version.

bantu:
:+1:

bantu:
Can you please squash the two commits together and make the commit message:
```
 [ticket/13903] Replacing regexp path naming for container by md5

Replacement of the path creation based on a blacklist by a md5 filename to
support broader configurations

PHPBB3-13903
```

bantu:
Replaced by #3711

Wardormeur:
Sorry about that, haven't been on my projects lately for IRL reasons. Thanks for supporting this change :)

PR #3675, opened 2015-06-01
[ticket/13906] Fixed old signatures in post preview

s9e:
PHPBB3-13906

I couldn't create a test for it because it requires a signature parsed with 3.1. Since it cannot be created in 3.2 it has to be inserted directly in the DB, and I don't know whether it's possible to access the database of the system being tested by a functional test.


bantu:
In the functional test case class, you should have access to the database via `$this->get_db()` which returns an instance of `\phpbb\db\driver\driver_interface`. However, using a database fixture might be a better solution.

PR #3676, opened 2015-06-01
[ticket/13908] "After" clause in migration not working

javiexin:
In a migration file, if you try to add a column to a table and specify
the order, by specifying an "after" clause, it is not used at all, even
for databases that support it (like mysql).
This fixes the issue, for mysql databases.

PHPBB3-13908


javiexin:
Thanks to rxu for the fix.

javiexin:
Closing as the PR is against develop-ascareus, reopen agains 3.1.x

PR #3683, opened 2015-06-05
[ticket/13921] Fix block BBCodes inside of inline BBCodes

s9e:
https://tracker.phpbb.com/browse/PHPBB3-13921
https://area51.phpbb.com/phpBB/viewtopic.php?f=126&t=47976

### DIFF
--- a/tests/text_processing/tickets_data/PHPBB3-13921.html
+++ b/tests/text_processing/tickets_data/PHPBB3-13921.html
@@ -1,0 +1,1 @@
+<span style="font-size: 200%; line-height: normal"></span><div style="text-align:center"><span style="font-size: 200%; line-height: normal">xxx</span></div>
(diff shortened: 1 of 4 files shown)


tests/text_processing/tickets_data/PHPBB3-13921.html
```diff
@@ -0,0 +1 @@
+<span style="font-size: 200%; line-height: normal"></span><div style="text-align:center"><span style="font-size: 200%; line-height: normal">xxx</span></div>
```
Crizz0: I think, you miss a blank line after line 1 here.

s9e: Those files contain the data that corresponds to a normal post and its output. To have a LF here you'd need a LF in the original text, but then it would add a `<br>` to the output too and trailing whitespace is normally removed anyway.

PR #3685, opened 2015-06-06
[ticket/13923] Update startup.php

Nicofuma:
While suppressing the output from the 'set_magic_quotes_runtime(0)'
is sufficient with normal phpBB installs, using this file in association
with the oauth plugin will cause an error on some versions of PHP.
This fixes a potential bug where set_magic_quotes_runtime is depreciated
in newer versions of PHP.

Replaces https://github.com/phpbb/phpbb/pull/3491

https://tracker.phpbb.com/browse/PHPBB3-13923

PHPBB3-13923

### DIFF
--- a/phpBB/includes/startup.php
+++ b/phpBB/includes/startup.php
@@ -94,7 +94,7 @@ if (version_compare(PHP_VERSION, '5.4.0-dev', '>='))
 }
 else
 {
-	if(get_magic_quotes_runtime())
+	if (get_magic_quotes_runtime())
 	{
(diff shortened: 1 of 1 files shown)


phpBB/includes/startup.php
```diff
@@ -94,7 +94,11 @@ function deregister_globals()
 }
 else
 {
-	@set_magic_quotes_runtime(0);
+	if(get_magic_quotes_runtime())
```
bantu: Missing space

PR #3688, opened 2015-06-06
[ticket/13888] Fix free_result() error in author searches

CHItA:
[PHPBB3-13888](https://tracker.phpbb.com/browse/PHPBB3-13888)


rxu:
Alternatively, you can just call ```sql_freeresult()``` omitting the argument.

CHItA:
I would rather leave it this way, if there aren't any objections.

PR #3694, opened 2015-06-07
[ticket/13930] Check for correct spacing between keyword & parenthesis

marc1706:
The codesniffer should check for a correct spacing between keywords like
if or foreach and its opening parenthesis.

Ticket: https://tracker.phpbb.com/browse/PHPBB3-13930

PHPBB3-13930

### DIFF
--- a/build/code_sniffer/ruleset-php-strict.xml
+++ b/build/code_sniffer/ruleset-php-strict.xml
@@ -45,4 +45,7 @@
  <!-- There MUST NOT be unused use statements. -->
  <rule ref="./phpbb/Sniffs/Namespaces/UnusedUseSniff.php" />
 
+ <!-- There MUST be one space between control structure and opening parenthesis -->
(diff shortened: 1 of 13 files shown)


build/code_sniffer/ruleset-php-strict.xml
```diff
@@ -45,4 +45,7 @@
  <!-- There MUST NOT be unused use statements. -->
  <rule ref="./phpbb/Sniffs/Namespaces/UnusedUseSniff.php" />
 
+ <!-- There MUST be one space between control structure and opening parenthesis -->
+ <rule ref="./phpbb/Sniffs/ControlStructures/OpeningParenthesisSniff.php" />
```
bantu: Did you try putting this into legacy?

bantu: Do things explode?

marc1706: Haven't tried yet

marc1706: Testing now :)

PR #3698, opened 2015-06-09
[ticket/13935] Allow more admin-configurable schemes in post links

s9e:
https://tracker.phpbb.com/browse/PHPBB3-13936

s9e\TextFormatter uses a whitelist of allowed schemes in URLs. By default, only http and https are allowed. Schemeless URLs are always allowed.

This PR adds a config setting that holds a comma separated list of schemes, with a default value of `http,https,ftp`.

I don't know whether the migration file is correct.

### DIFF
--- a/phpBB/includes/acp/acp_board.php
+++ b/phpBB/includes/acp/acp_board.php
@@ -197,7 +197,6 @@ class acp_board
 						'allow_post_flash'		=> array('lang' => 'ALLOW_POST_FLASH',		'validate' => 'bool',	'type' => 'radio:yes_no', 'explain' => true),
 						'allow_smilies'			=> array('lang' => 'ALLOW_SMILIES',			'validate' => 'bool',	'type' => 'radio:yes_no', 'explain' => false),
 						'allow_post_links'		=> array('lang' => 'ALLOW_POST_LINKS',		'validate' => 'bool',	'type' => 'radio:yes_no', 'explain' => true),
-						'allowed_schemes_links'	=> array('lang' => 'ALLOWED_SCHEMES_LINKS',	'validate' => 'string',	'type' => 'text:0:255', 'explain' => true),
 						'allow_nocensors'		=> array('lang' => 'ALLOW_NO_CENSORS',		'validate' => 'bool',	'type' => 'radio:yes_no', 'explain' => true),
 						'allow_bookmarks'		=> array('lang' => 'ALLOW_BOOKMARKS',		'validate' => 'bool',	'type' => 'radio:yes_no', 'explain' => true),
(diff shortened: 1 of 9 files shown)


phpBB/includes/acp/acp_board.php
```diff
@@ -196,6 +196,7 @@ function main($id, $mode)
 						'allow_post_flash'		=> array('lang' => 'ALLOW_POST_FLASH',		'validate' => 'bool',	'type' => 'radio:yes_no', 'explain' => true),
 						'allow_smilies'			=> array('lang' => 'ALLOW_SMILIES',			'validate' => 'bool',	'type' => 'radio:yes_no', 'explain' => false),
 						'allow_post_links'		=> array('lang' => 'ALLOW_POST_LINKS',		'validate' => 'bool',	'type' => 'radio:yes_no', 'explain' => true),
+						'allowed_schemes_links'	=> array('lang' => 'ALLOWED_SCHEMES_LINKS',	'validate' => 'string',	'type' => 'text:0:255', 'explain' => true),
```
nickvergessen: I'm not sure whether we need a config option for this.
It has never been configurable?

I mean we can still have it as a config option, but not display it in the UI.

s9e: No option in 3.1, any scheme is allowed except for `javascript` and I think any scheme that ends in "script" too.

If the config option is not displayed in the ACP's UI, it might as well be hardcoded. An extension would still be able to configure which schemes are allowed via the `core.text_formatter_s9e_configure_after` event.

Nicofuma: what he means is that it could be configured through the config.yml file and not through the UI because there is already too much config option in the ACP

Nicofuma: @nickvergessen so? entry in the acp or not?

nickvergessen: I'd say just add the config for now and skip the UI. We can then add the UI option once we have the "advanced UI" for ACP

Nicofuma:
marked as WIP as per https://github.com/phpbb/phpbb/pull/3698#discussion_r34059211

s9e:
I removed the UI but if schemes aren't configurable I recommend not merging this PR. The only functional difference is it enables FTP by default and that can be hardcoded in 1 line of code instead.

Nicofuma:
It can be configured, but only with the CLI and not within the ACP

s9e:
Ok.

phpBB/includes/acp/acp_board.php
```diff
@@ -549,6 +549,12 @@ function main($id, $mode)
 			}
 		}
 
+		// Invalidate the text_formatter cache when posting options are changed
+		if ($mode == 'post' && $submit)
+		{
+			$phpbb_container->get('text_formatter.cache')->invalidate();
```
Nicofuma: I don't think this is needed anymore

PR #3699, opened 2015-06-09
[ticket/13823] Move the part of the query, so the update will get it …

nickvergessen:
…correctly

https://tracker.phpbb.com/browse/PHPBB3-13823

**Note:** This will create a conflict on the update from 3.1.4 when you previously updated from 3.1.3 to 3.1.4, due to the missing whitespace change on the ` AND ` which is why we need to move all this around.


Nicofuma:
https://github.com/phpbb/phpbb/pull/3700

marc1706:
I definitely prefer #3700 over this.

nickvergessen:
@marc1706 we need both, so the section is changed in 3.1.5 and triggers an update of the file from 3.1.4 #3700 just helps to avoid it in future

marc1706:
@nickvergessen I'm pretty sure this file was already included in the 3.1.4 to 3.1.5 without this change. O.o

PR #3700, opened 2015-06-09
[ticket/13823] Change diff options while packaging to not ignore all whitespaces

Nicofuma:
Removed:
    -w, --ignore-all-space       Ignore white space when comparing lines.

Added:
    -Z, --ignore-trailing-space  Ignore white space at line end.
    -b, --ignore-space-change    Ignore changes in the amount of white space.

https://tracker.phpbb.com/browse/PHPBB3-13823

PHPBB3-13823

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -105,7 +105,7 @@
 			<property name="dir" value="build/old_versions/release-${version}" />
 		</phingcall>
 
-		<exec dir="build/old_versions" command="LC_ALL=C diff -crNEBwd release-${version} release-${newversion} >
+		<exec dir="build/old_versions" command="LC_ALL=C diff -crNEBZbd release-${version} release-${newversion} >
 			../new_version/patches/phpBB-${version}_to_${newversion}.patch" escape="false" />
 	</target>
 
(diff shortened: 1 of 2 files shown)


nickvergessen:
We don't use those files I think.
You can try it by manipulating an uodate package, just remove all files but one from install/update/index.php and then try your stuff in that file. At that point the diffs generated by the commands you modify are of no use? Or was this somehow the problem with the packages that were created for 3.1.3 to 3.1.4?

Nicofuma:
actually my PR solves one part f the packaging but maybe not the automatic updater... We have to try but I can't do that for now.

But if the automatic updater is still failing with these changes I should be an issue with `install_update.php` because the lib found the difference between the files.

build/build.xml
```diff
@@ -177,13 +177,13 @@
 	<target name="package" depends="clean,prepare,prepare-new-version,old-version-diffs">
 		<exec dir="build" command="php -f package.php '${versions}' > logs/package.log" escape="false" />
 		<exec dir="build" escape="false"
-			command="diff -crNEBwd old_versions/release-${prevversion}/language new_version/phpBB3/language >
+			command="diff -crNEBZbd old_versions/release-${prevversion}/language new_version/phpBB3/language >
```
nickvergessen: Please add `LC_ALL=C ` here as well

build/build.xml
```diff
@@ -177,13 +177,13 @@
...
 				save/phpbb-${prevversion}_to_${newversion}_language.patch" />
 		<exec dir="build" escape="false"
-			command="diff -crNEBwd old_versions/release-${prevversion}/styles/prosilver new_version/phpBB3/styles/prosilver >
+			command="diff -crNEBZbd old_versions/release-${prevversion}/styles/prosilver new_version/phpBB3/styles/prosilver >
```
nickvergessen: Please add `LC_ALL=C ` here as well

build/build.xml
```diff
@@ -177,13 +177,13 @@
...
 				save/phpbb-${prevversion}_to_${newversion}_prosilver.patch" />
 		<exec dir="build" escape="false"
-			command="diff -crNEBwd old_versions/release-${prevversion}/styles/subsilver2 new_version/phpBB3/styles/subsilver2 >
+			command="diff -crNEBZbd old_versions/release-${prevversion}/styles/subsilver2 new_version/phpBB3/styles/subsilver2 >
```
nickvergessen: Please add `LC_ALL=C ` here as well

nickvergessen:
`build/build_helper.php` needs the same change at the top, otherwise the update packages are empty :dash: 

```php

	// -c - context diff
	// -r - compare recursive
	// -N - Treat missing files as empty
	// -E - Ignore tab expansions
	//		not used: -b - Ignore space changes.
	// -w - Ignore all whitespace
	// -B - Ignore blank lines
	// -d - Try to find smaller set of changes
	var $diff_options = '-crNEBwd';
```
and please also update the comment

nickvergessen:
Other then that I can see how it would have helped with the initial issue.

Nicofuma:
updated

marc1706:
From IRC:
```<Marc> just tested if the installer diff engine picks up the whitespace change and in fact it does```

The change also causes the file to be correctly packaged into the 3.1.3 to 3.1.4 update package. This didn't work before these changes.

nickvergessen:
Pr should go against the prep branch, and if there is ever a 3.0 release it should be there as welk i guess?

Nicofuma:
rebased on 3.0.x

nickvergessen:
when you merge this merge 3.0 into prep and 3.1 but not prep into 3.1!

nickvergessen:
Merged into prep-release-3.1.5 via https://github.com/phpbb/phpbb/compare/88d16b47a408...0d349a82ca64
will come to 3.1.x when 3.1.5 is being released

PR #3705, opened 2015-06-13
[ticket/13942] Check if path to language file contains language path

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-13942

PHPBB3-13942


marc1706:
Replaced by #3706

PR #3707, opened 2015-06-13
[ticket/13944] Fix the test framework compatibility with Oracle

Nicofuma:
https://tracker.phpbb.com/browse/PHPBB3-13944

PHPBB3-13944

### DIFF
--- a/tests/test_framework/phpbb_database_test_connection_manager.php
+++ b/tests/test_framework/phpbb_database_test_connection_manager.php
@@ -59,7 +59,18 @@ class phpbb_database_test_connection_manager
 			case 'sqlite2':
 			case 'sqlite':	// SQLite3 driver
 				$dsn .= $this->config['dbhost'];
-			break;
+				break;
+			case 'oci':
+				// For oracle the DSN is special: oci:dbname//host:port/db
+				$dsn .= 'dbname=//' . $this->config['dbhost'];
+
+				if ($this->config['dbport'])
+				{
+					$dsn .= ':' . $this->config['dbport'];
+				}
+
+				$dsn .= '/' . $this->config['dbname'];
+				break;
 
 			case 'sqlsrv':
 				// prefix the hostname (or DSN) with Server= so using just (local)\SQLExpress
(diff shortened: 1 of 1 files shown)


bantu:
@Nicofuma I agree with Noxwizard here. See https://tracker.phpbb.com/browse/PHPBB3-13944?focusedCommentId=56436&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-56436 It seems like you have been missing that tests on Oracle just fail because the implementations don't work on there. There is (mostly) nothing wrong with the test framework.

bantu:
Maybe don't push the Oracle status then?

bantu:
Maybe we should just drop Oracle support to make this official.

Nicofuma:
I'm not, actually there is 2 things with Oracle:
1) The tests are failing
2) It doesn't always work because only one DSN is supported and we don't handle `custom_dsn` and also because we try to delete all the tables and not only the tables related to phpBB

My fix isn't correct yet be it will be later.
And for the status, I have disable the Oracle Job and I think we can add a plan for it (like we do with MsSQL)

Noxwizard:
Can you expand on 2)? I have yet to see tests fail because of the DSN or because of errors deleting system tables.

Nicofuma:
https://bamboo.phpbb.com/browse/PHPBB3-PRDOCKER-UTP6O-44/log for the DSN
and for the system tables I don't have the logs on Bamboo because I ran the tests locally but the error was:
>   [SQLSTATE[HY000]: General error: 4098 OCIStmtExecute: ORA-04098: trigger 'SYSTEM.T_PHPBB_USERS' is invalid and failed re-validation
>  (/usr/src/php/ext/pdo_oci/oci_statement.c:148)]
> 
> /data/tests/test_framework/phpbb_database_test_case.php:115
> /data/tests/auth/provider_apache_test.php:25

Noxwizard:
None of those indicate an issue with the DSN format. If I run the entire suite, I get an output similar to Bamboo's. If I run them in filtered groups, I get normal SQL errors (e.g. invalid hex number) instead of TNS errors. This seems more related to the configuration of the database instance itself, or we need to look into connecting directly and bypass TNS.

I have not seen the error trying to drop the trigger in my runs.

Nicofuma:
I took the wrong error for the second point...
For the DNS it may be related to the install yes but the DNS I introduce is the one documented in the PHP doc. (Btw,  my install is a standard oracle XE install but running on a different host)

Nicofuma:
About the system tables, there isn't any error actually I saw that by dumping the table list before and after the clean

CHItA:
Added pending closure as there is no activity on this ticket since 2015. Also we should rewrite the test framework at some point.

PR #3708, opened 2015-06-14
[ticket/13946] Increase the CSS margin around [code] and [list]

s9e:
PHPBB3-13946

### DIFF
--- a/phpBB/styles/prosilver/theme/content.css
+++ b/phpBB/styles/prosilver/theme/content.css
@@ -383,7 +383,7 @@ dl.faq dt {
 }
 
 .content ul, .content ol {
-	margin: 1em 0 1.2em 3em;
+	margin: 0.8em 0 0.9em 3em;
 }
 
 .posthilit {
@@ -429,7 +429,7 @@ dd .signature {
 }
 
 .signature ul, .signature ol {
-	margin: 1em 0 1.2em 3em;
+	margin: 0.8em 0 0.9em 3em;
 }
 
 /* Post noticies */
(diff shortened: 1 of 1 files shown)


marc1706:
Is there a reason you closed the quote one?

marc1706:
Before:
13946_before
After:
13946_after

phpBB/styles/prosilver/theme/content.css
```diff
@@ -383,8 +383,7 @@ dl.faq dt {
 }
 
 .content ul, .content ol {
-	margin-bottom: 1em;
-	margin-left: 3em;
+	margin: 1em 0 1.2em 3em;
```
marc1706: I just checked and changing it to this creates the exact same margins as the codebox has in Chrome:
```margin: 0.8em 0 0.9em 3em;```

I'd prefer if the resulting margins would be the same.

s9e: I didn't notice they were different from each other. Then I assume that code/list have different font sizes? In that case,  would it make sense to use px or--even better--**rem**? Apparently [rem is supported on everything but IE 7](http://caniuse.com/#feat=rem).

I don't care either way, `0.8em 0 0.9em 3em` is fine by me.

marc1706: See related comment here: https://github.com/phpbb/phpbb/pull/3649#issuecomment-106742826

callumacrae:
Yes, I was wondering where that margin went :tada:

PR #3710, opened 2015-06-16
[ticket/12952] Check obvious errors before status in functional tests

marc1706:
This will change the output from just saying that the status did not match
the expected one to actually showing any errors before complaining about a
possible difference in the status code.

Ticket: https://tracker.phpbb.com/browse/PHPBB3-12952

PHPBB3-12952


Nicofuma:
:+1:

PR #3711, opened 2015-06-16
[ticket/13903] Replacing regexp path naming for container by md5.

bantu:
Replacement of the path creation based on a blacklist by a md5 filename to
support broader configurations

https://tracker.phpbb.com/browse/PHPBB3-13903

Replaces #3672


Nicofuma:
:+1:

PR #3714, opened 2015-06-18
[ticket/13951] Make sure the page number is an integer

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-13951

PHPBB3-13951


Crizz0:
From my point of view, this patch solves the problem from the ticket.

PR #3716, opened 2015-06-19
[ticket/13748] Wrong tooltip after poll vote change

javiexin:
Store the "voted" tooltip in an extra attribute in HTML, and use it
with the JS that refreshes after poll vote.

PHPBB3-13748


phpBB/styles/prosilver/template/viewtopic_body.html
```diff
@@ -80,7 +80,7 @@ <h2 class="poll-title"><!-- EVENT viewtopic_body_poll_question_prepend -->{POLL_
 			<fieldset class="polls">
 			<!-- BEGIN poll_option -->
 				<!-- EVENT viewtopic_body_poll_option_before -->
-				<dl class="<!-- IF poll_option.POLL_OPTION_VOTED -->voted<!-- ENDIF --><!-- IF poll_option.POLL_OPTION_MOST_VOTES --> most-votes<!-- ENDIF -->"<!-- IF poll_option.POLL_OPTION_VOTED --> title="{L_POLL_VOTED_OPTION}"<!-- ENDIF --> data-poll-option-id="{poll_option.POLL_OPTION_ID}">
+				<dl class="<!-- IF poll_option.POLL_OPTION_VOTED -->voted<!-- ENDIF --><!-- IF poll_option.POLL_OPTION_MOST_VOTES --> most-votes<!-- ENDIF -->"<!-- IF poll_option.POLL_OPTION_VOTED --> title="{L_POLL_VOTED_OPTION}"<!-- ENDIF --> data-alt-text="{LA_POLL_VOTED_OPTION}" data-poll-option-id="{poll_option.POLL_OPTION_ID}">
```
vinny:
```html
{L_POLL_VOTED_OPTION}
```
must be correct. right?

javiexin: I have used LA because the string is used from JS, and because the (only) example of use of the same construct in phpbb also uses LA:
https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/styles/prosilver/template/memberlist_body.html#L28

But if you tell me I must change it to L_, I will do.

vinny:
https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/styles/prosilver/template/overall_footer.html#L18

https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/styles/prosilver/template/overall_footer.html#L22

javiexin:
Ok, should we "fix" the example I provided as well?  Probably, we should, but from another ticket.  Will you open it?

javiexin:
After the change from LA_ to L_, there are two Travis failures that were not there before.  I don't understand the errors, I cannot figure out what or where are they failing (general issue with Travis...).  But it looks as if the change is creating the issues.  Should we go back to using LA_?

javiexin:
I am sorry, but I continue to think that my original patch https://github.com/javiexin/phpbb/commit/5e39a7ab72db9320ce1e38403ebb87d6a48c41f9, using the "LA_" version of the template variable, is the correct one.

Please, could someone review and comment?

PR #3720, opened 2015-06-19
[ticket/13954] Make it possible to retrieve service names from a collection

CHItA:
[PHPBB3-13954](https://tracker.phpbb.com/browse/PHPBB3-13954)


bantu:
Tests?

CHItA:
I guess it is possible, though we could only test the ArrayObject.

bantu:
Aren't there any existing tests for service_collection? There should be.

CHItA:
There are none. Though my function only relies on the fact that ArrayObject::getArrayCopy is not calling offsetGet().

CHItA:
I added tests, if you think I should add some more, please let me know.

tests/di/service_collection_test.php
```diff
@@ -0,0 +1,92 @@
...
+			$builder->without_cache();
+			$container = $builder->get_container();
+			$this->service_collection = $container->get('service_collection');
+
+			parent::setUp();
+		}
+
+		public function test_service_collection_get_service_names()
+		{
+			global $service_initialized;
```
bantu: don't use global

tests/di/service_collection_test.php
```diff
@@ -0,0 +1,92 @@
...
+			$service_initialized = false;
+
+			$service_names = $this->service_collection->get_service_names();
+			$this->assertTrue(in_array('collection_service1', $service_names));
```
bantu: First of all, there is assertContains. But then, this should also rather be assertSame with an array if the order is known. Otherwise $service_names can contain more elements.

tests/di/service_collection_test.php
CHItA: I can't really test the behavior other then this (if I would use a getter member function that would construct the class, which is exactly what I don't want to do)

tests/di/service_collection_test.php
```diff
@@ -0,0 +1,92 @@
...
+			$this->assertTrue(in_array('collection_service2', $service_names));
+
+			$this->assertFalse($service_initialized);
+		}
+
+		public function test_service_collection()
```
bantu: this tests getIterator?

Nicofuma:
?

CHItA:
I don't need this anymore

PR #3721, opened 2015-06-20
[ticket/13952] Inject event dispatcher dependency to the tidy search cron task

rxu:
<a href="https://tracker.phpbb.com/browse/PHPBB3-13952">PHPBB3-13952</a>.

### DIFF
--- a/phpBB/phpbb/cron/task/core/tidy_search.php
+++ b/phpBB/phpbb/cron/task/core/tidy_search.php
@@ -26,6 +26,7 @@ class tidy_search extends \phpbb\cron\task\base
 	protected $config;
 	protected $db;
 	protected $user;
+	protected $phpbb_dispatcher;
 
 	/**
 	* Constructor.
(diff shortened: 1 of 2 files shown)


phpBB/phpbb/cron/task/core/tidy_search.php
```diff
@@ -26,6 +26,7 @@ class tidy_search extends \phpbb\cron\task\base
 	protected $config;
 	protected $db;
 	protected $user;
+	protected $phpbb_dispatcher;
```
marc1706: Could you maybe add a doc block for this? Even though the old ones don't have that, I like it if we at least add a doc block for newly added lines of code.

rxu:
@marc1706 Updated.

bantu:
@Nicofuma Hmm, I thought the conclusion was that this wasn't a problem.

Nicofuma:
It isn't an issue if we call the constructor with to many arguments. not the contrary.

bantu:
@Nicofuma Looks like I remember it wrong, or confused the two.

PR #3723, opened 2015-06-22
fix the website issue when auto_increment_offset != 1

changyy:
When db server auto_increment_offset != 1, the default website browsing is error.

The message is: 
```
QL ERROR [ mysqli ]

You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '' at line 3 [1064]

SQL

SELECT * FROM phpbb_styles s WHERE s.style_id =

BACKTRACE

FILE: (not given by php)
LINE: (not given by php)
CALL: msg_handler()

FILE: [ROOT]/phpbb/db/driver/driver.php
LINE: 855
CALL: trigger_error()

FILE: [ROOT]/phpbb/db/driver/mysqli.php
LINE: 193
CALL: phpbb\db\driver\driver->sql_error()

FILE: [ROOT]/phpbb/db/driver/factory.php
LINE: 329
CALL: phpbb\db\driver\mysqli->sql_query()

FILE: [ROOT]/phpbb/user.php
LINE: 238
CALL: phpbb\db\driver\factory->sql_query()

FILE: [ROOT]/ucp.php
LINE: 36
CALL: phpbb\user->setup()
```

I found $this->data['user_style'] is not defined.

https://tracker.phpbb.com/browse/PHPBB3-13957

### DIFF
--- a/phpBB/phpbb/user.php
+++ b/phpBB/phpbb/user.php
@@ -229,7 +229,7 @@ class user extends \phpbb\session
 		else
 		{
 			// Set up style
-			$style_id = ($style_id) ? $style_id : ((!$config['override_user_style']) ? $this->data['user_style'] : $config['default_style']);
+			$style_id = ($style_id) ? $style_id : ((!$config['override_user_style']) && isset($this->data['user_style']) ? $this->data['user_style'] : $config['default_style']);
 		}
 
 		$sql = 'SELECT *
@@ -240,7 +240,7 @@ class user extends \phpbb\session
 		$db->sql_freeresult($result);
 
 		// Fallback to user's standard style
-		if (!$this->style && $style_id != $this->data['user_style'])
+		if (!$this->style && isset($this->data['user_style']) && $style_id != $this->data['user_style'])
 		{
 			$style_id = $this->data['user_style'];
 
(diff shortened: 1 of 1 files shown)


bantu:
Thank you for your patch. Unfortunately, adding those issets does likely not resolve the root cause of phpBB not properly supporting auto_increment_offset other than 1. I highly suspect that more/other work is necessary to address the root problem.

nickvergessen:
All the stuff we do on install requires auto_increment = 1:
https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/install/schemas/schema_data.sql

Maybe once we have the new installer from @CHItA we can get rid of that file and manually call a couple of php generated queries instead. Until then I'd say this is simply not supported, sorry.

azimut:
There is an issue to track the new installer? I am having this issue and would like to know when it gets fixed.

nickvergessen:
The new installer is already merged in 3.2.0-alpha1 I think, so maybe you can test that package

azimut:
@nickvergessen  Unfortunately that didn't work. I got the same error message and the same invalid query is being done to mysql.  Additionally the installation got stuck at the stage that sends a notification email. I will stay tuned to PHPBB3-13957.

PR #3724, opened 2015-06-23
[ticket/13960] Profile field validation breaks ACP

javiexin:
Make sure new step is calculated only if no error has been identified
during validation.  Also, cleanup redundant code in acp_profile.php.

PHPBB3-13960

### DIFF
--- a/phpBB/includes/acp/acp_profile.php
+++ b/phpBB/includes/acp/acp_profile.php
@@ -511,12 +511,13 @@ class acp_profile
 					}
 				}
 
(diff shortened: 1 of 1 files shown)


javiexin:
@nickvergessen Please, take a look. Sorry to bother.

javiexin:
Could anyone please review this, so that it can be either accepted or rejected?
Please, assign to a milestone.
Thanks in advance.

Nicofuma:
Thanks for your contribution

PR #3726, opened 2015-06-24
[ticket/13961] Add orderable service collections

CHItA:
[PHPBB3-13961](https://tracker.phpbb.com/browse/PHPBB3-13961)

### DIFF
--- a/phpBB/phpbb/di/service_collection/ordered_service_collection.php
+++ b/phpBB/phpbb/di/service_collection/ordered_service_collection.php
@@ -1,0 +1,108 @@
+<?php
+/**
+ *
+ * This file is part of the phpBB Forum Software package.
+ *
+ * @copyright (c) phpBB Limited <https://www.phpbb.com>
+ * @license GNU General Public License, version 2 (GPL-2.0)
+ *
+ * For full copyright and license information, please see
+ * the docs/CREDITS.txt file.
+ *
+ */
+
+namespace phpbb\di\service_collection;
+
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Collection of services in a specified order
+ */
+class ordered_service_collection extends service_collection
+{
+	/**
+	 * @var bool
+	 */
+	protected $is_ordered;
+
+	/**
+	 * @var array
+	 */
+	protected $service_ids;
+
+	/**
+	 * Constructor
+	 *
+	 * @param ContainerInterface $container Container object
+	 */
+	public function __construct(ContainerInterface $container)
+	{
+		$this->is_ordered = false;
+		$this->service_ids = array();
+
+		parent::__construct($container);
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function offsetExists($index)
+	{
+		if (!$this->is_ordered)
+		{
+			$this->sort_services();
+		}
+
+		return parent::offsetExists($index);
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function offsetGet($index)
+	{
+		if (!$this->is_ordered)
+		{
+			$this->sort_services();
+		}
+
+		return parent::offsetGet($index);
+	}
+
+	/**
+	 * Adds a service ID to the collection
+	 *
+	 * @param string	$service_id
+	 * @param int		$order
+	 */
+	public function add($service_id, $order)
+	{
+		if ($this->is_ordered)
+		{
+			return;
+		}
+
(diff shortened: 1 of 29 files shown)


phpBB/phpbb/di/service_collection/service_collection.php
```diff
@@ -11,7 +11,7 @@
 *
 */
 
-namespace phpbb\di;
+namespace phpbb\di\service_collection;
```
nickvergessen: Can we avoid this? It breaks all extensions unnecessarily that use a service collection. Having:
* phpBB/phpbb/di/service_collection.php
* phpBB/phpbb/di/ordered_service_collection.php

is just as fine?

phpBB/phpbb/di/service_collection/ordered_service_collection.php
```diff
@@ -0,0 +1,121 @@
...
+	 * Adds a service ID to the collection
+	 *
+	 * @param string	$service_id
+	 * @param int		$order
+	 */
+	public function add($service_id, $order = 0)
+	{
+		if ($this->is_ordered)
+		{
+			return;
```
nickvergessen: why is this blocked?

phpBB/phpbb/di/service_collection/service_collection.php
CHItA: yes, haven't thought of that

phpBB/phpbb/di/service_collection/ordered_service_collection.php
CHItA: we produce the array for the ArrayObject has to be produced only once this way

CHItA: Actually, this was not fixed, this is just GH who thinks it

CHItA: @nickvergessen: the alternative of this would be to implement ordered insertion into the array, which is way more complicated, and the `add()` method should only be called by the service container (at least IMO).

Nicofuma: Here I tend to disagree...

I think we coud just regenerate the sorted array in add() when it has already been sorted.

Explanation:
- add() // just add
- add() // just add
- get() // sort array and return entry
- get() // return entry
- add() // add and mark as not sorted
- get() // sort array and return

off course if someone try to call add() in a foreach, ex:

```
foreach ($collection as $key => $service) {
    if (...) {
        $collection->add(...);
    }
}
```
it is unsupported because it could mess up with the iterator

CHItA: Fixed, though now I disagree :)

CHItA:
@bantu Any additional thoughts?

Nicofuma:
@bantu @nickvergessen https://github.com/phpbb/phpbb/pull/3726#discussion_r33411756 thoughts?

nickvergessen:
@Nicofuma that is what is currently done? If not, it sounds like the thing I had in mind.

CHItA:
Mostly, there is no protection against adding a service in a foreach, not even sure if that is possible to implement.

phpBB/phpbb/di/service_collection/ordered_service_collection.php
bantu: I know this was changed, but anyway: Returning null might not be such a good idea. API wise, I'd rather throw an exception.
Other than that, I like @Nicofuma 's description. For iteration, I think the ideal behaviour would be to ensure the added item is only returned in the next iteration (but not the current).

CHItA: is this behaviour desired/implemented in the `service_collection_iterator`? so in other words, does the `service_collection` class work correctly and we need another iterator class for the ordered one, or should i simply rewrite `service_collection_iterator`?

bantu: I don't know. I don't think it is too important. If that behaviour can be easily achieved, implement it and document it. Otherwise, just document that the behaviour is undefined in that case.

CHItA: We could probably do it by a constructor. So the collection iterator service could be something like `__construct($collection) { parent::__construct($collection->getArrayCopy()); }`

CHItA:
Everything should be fixed.

Nicofuma:
:+1:

PR #3733, opened 2015-06-30
[ticket/13934] Add enctype clause for profile fields

javiexin:
Adds a template variable (the same) in all places where profile fields
may need an enctype clause in the corresponding forms.

PHPBB3-13934

### DIFF
--- a/phpBB/adm/style/acp_profile.html
+++ b/phpBB/adm/style/acp_profile.html
@@ -17,7 +17,7 @@
 		</div>
 	<!-- ENDIF -->
 
-	<form id="add_profile_field" method="post" action="{U_ACTION}">
+	<form id="add_profile_field" method="post" action="{U_ACTION}"{S_FORM_ENCTYPE}>
 
 	<!-- IF S_STEP_ONE -->
 
--- a/phpBB/adm/style/acp_users_profile.html
+++ b/phpBB/adm/style/acp_users_profile.html
@@ -1,4 +1,4 @@
-	<form id="user_profile" method="post" action="{U_ACTION}">
+	<form id="user_profile" method="post" action="{U_ACTION}"{S_FORM_ENCTYPE}>
 
 	<fieldset>
 		<legend>{L_USER_PROFILE}</legend>
--- a/phpBB/styles/prosilver/template/ucp_register.html
+++ b/phpBB/styles/prosilver/template/ucp_register.html
@@ -14,7 +14,7 @@
 // ]]>
 </script>
 
-<form method="post" action="{S_UCP_ACTION}" id="register">
+<form id="register" method="post" action="{S_UCP_ACTION}"{S_FORM_ENCTYPE}>
 
 <div class="panel">
 	<div class="inner">


phpBB/adm/style/acp_profile.html
```diff
@@ -17,7 +17,7 @@
 		</div>
 	<!-- ENDIF -->
 
-	<form id="add_profile_field" method="post" action="{U_ACTION}">
+	<form id="add_profile_field" method="post" action="{U_ACTION}"{S_FORM_ENCTYPE}>
```
CHItA: Please use `<!-- IF ... --> {S_FORM_ENCTYPE}<!-- ENDIF -->` all the places.

javiexin: Why? This is not the way it is done elsewhere in phpbb...

CHItA: Because it is a clearer and better way of doing this. You should be able to use `<!-- IF S_FORM_ENCTYPE is defined and S_FORM_ENCTYPE is not empty --> {S_FORM_ENCTYPE}<!-- ENDIF -->`

javiexin: Examples of usage in other areas of phpbb. And there are a lot more, and none like what you propose.
I don't understand why this would be any different. I prefer consistency across the phpbb software.
https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/styles/prosilver/template/ucp_profile_profile_info.html#L3
https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/styles/prosilver/template/posting_layout.html#L24

CHItA: Please add the `enctype` attribute as well to the template. If you are bothered by the inconsistency of the template files, please feel free to send a patch for those as well.

javiexin: I don't understand the comment "add the enctype attribute to the template".  This is not meant to be there always, if it were, we would be better off hardcoding it (as in https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/adm/style/acp_users_avatar.html#L1).  This is meant to be filled in by extensions needing it, or (as in my case) by new profile field types. But not hardcoding/always present.
And no, I am not going to change so many places all around phpbb.  I'd rather keep these as they are now.

javiexin: By the way, if you want to continue discussion, please, let's use a more appropriate way:
https://area51.phpbb.com/phpBB/viewtopic.php?f=126&t=48001

CHItA: What I meant is: `<!-- IF S_FORM_ENCTYPE is defined and S_FORM_ENCTYPE is not empty --> enctype="{S_FORM_ENCTYPE}"<!-- ENDIF -->`

Thanks

javiexin: Again, no, this would not be compatible with current uses.
S_FORM_ENCTYPE content should be ' enctype="multipart/form-data"'. Otherwise, this would be incompatible with the current use pattern in ucp_profile_profile_info, that should be identical because it is configured the same way.

CHItA: Then better yet, you can use `<!-- IF S_FORM_ENCTYPE is defined and S_FORM_ENCTYPE is not empty --> enctype="multipart/form-data"<!-- ENDIF -->` as `text/plain` would not be used ever anyway, and the other one is the defualt one.

javiexin: I want to use the following code in the extension or wherever:
$template->assign_var('S_FORM_ENCTYPE', ' enctype="multipart/form-data"');
And use this exactly the same way in all places where it is needed:
- ucp_profile_profile_info.html (already existing, and coded as I want it)
- acp_profile.html 
- acp_users_profile.html
- ucp_register.html
If I changed to what you propose, that would mean that the usage pattern would be different in each place, and the template var should have a different value depending on where it is used.  Not the way I am coding this.
Besides, if you asked me to change the already existing var to follow the "new" pattern, that would break any existing code and extensions that make use of the existing var.
So, I stand by my proposal.  I want to keep it as I originally proposed.

CHItA: nothing stops you from doing that. you can use whatever in the backend you wish, but IMO there is no difference between "The only viable option here is to modify the ACTION URL template variable to include enctype: ugly, dangerous and incorrect." and the changes you are proposing.

javiexin: I guess we need another opinion, as we are not moving with our arguments.

CHItA: that's what you wrote, though... ;)

VSEphpbb: needs a space after the quote too

phpBB/adm/style/acp_users_profile.html
```diff
@@ -1,4 +1,4 @@
-	<form id="user_profile" method="post" action="{U_ACTION}">
+	<form id="user_profile" method="post" action="{U_ACTION}"{S_FORM_ENCTYPE}>
```
VSEphpbb: space?

phpBB/styles/prosilver/template/ucp_register.html
```diff
@@ -14,7 +14,7 @@
 // ]]>
 </script>
 
-<form method="post" action="{S_UCP_ACTION}" id="register">
+<form id="register" method="post" action="{S_UCP_ACTION}"{S_FORM_ENCTYPE}>
```
VSEphpbb: space?

phpBB/adm/style/acp_profile.html
javiexin: Not if used as in other places, see links above. The space is supposed to be included in the template var. But it may be safer if included again here...

nickvergessen: I dont really understand the problem. If something is undefined, using it like above will just be fine. It will be false or whatever and not causing any output. So it has the same result.

CHItA: I just hate HTML in the backend, it should not have got there in the first place.

javiexin:
I have been looking at current phpbb 3.1.5, to identify all CURRENT uses of enctype, and these are my findings:
ACP templates: 
- 2 templates have enctype clause hardcoded (no template variable)
- 1 template have enctype clause surrounded by  ; new in 3.1, and using a new different template variable S_CAN_UPLOAD

Prosilver templates:
- 19 templates have enctype template variable S_FORM_ENCTYPE as used in this PR, without  and without space between the ACTION template variable and this

PHP code:
- 5 php files fill in the same template variable (S_FORM_ENCTYPE), and in all cases, the value includes the starting space

What I have not been able to see is the amount of extension that make use of such variables; I don't know if it is difficult for you to run such a query, at least among the validated extensions.

With all of this, my proposal and preferences are as follows:
1) Continue with my original proposal, for consistency with current uses within phpbb software
2) If that is unacceptable, I would be OK with hardcoding the ACP templates maintaining the prosilver template as proposed, although I prefer the former
3) I would not accept the use of conditional for prosilver templates unless the already existing 19 cases are updated as well.

If you all agree that introducing the initial space in the templates and removing it from the PHP files is better, someone please open a separate ticket, and I will be happy to provide a PR for that, and will update this to include the space here, as I agree this could be error prone, and even, source of potential security exploits (by appending text to the URL that is in front), and no extension should be using it anyway (at least not cleanly).

If this needs to be changed further, like updating existing uses either in ACP or in prosilve to use conditionals surrounding them, I would suggest that it goes through an RFP process in Area51 AND Extension Writers forum, as this will definitely break a number of existing or in development extensions.  It might not be a bad idea, but we need to know the impact of such a change before implementing it, as it is heavily non-backwards compatible.

Please, advise on what to do next.

nickvergessen:
The patch looks fine to me, when the variables are properly set in the corresponding php files.

javiexin:
Could anyone please review this further, so that it can be either accepted or rejected?
Thanks in advance.

Nicofuma:
thanks for your contribution

PR #3734, opened 2015-07-01
[ticket/13759] Initial commit broke timestamps when quoting a post.

RMcGirr83:
The quoting user would get the date of the quoted post instead of the server time.  `post_time` variable passed to the submit_post function within the `$data` array is never used in the submit_post function.

This commit corrects the regression.

PHPBB3-13759

### DIFF
--- a/phpBB/posting.php
+++ b/phpBB/posting.php
@@ -1322,7 +1322,6 @@ if ($submit || $preview || $refresh)
 				'enable_urls'			=> (bool) $post_data['enable_urls'],
 				'enable_indexing'		=> (bool) $post_data['enable_indexing'],
 				'message_md5'			=> (string) $message_md5,
-				'post_time'				=> (isset($post_data['post_time'])) ? (int) $post_data['post_time'] : $current_time,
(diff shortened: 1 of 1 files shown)


RMcGirr83:
https://github.com/RMcGirr83/phpbb/commit/a53b57c8a2298ea04a45680cee00c5a9fec44689

RMcGirr83:
WTF!!

Inspecting commit message of commit 1a519a7a94cd06883297ac7f7d5ecb878e37cb19
Error: heading should be a sentence beginning with a letter. 
You entered: 
This commit corrects the regression. 
Syntax error on line 1: 
This commit corrects the regression. 
Expecting: comment, header, 
Aborting commit. 

Travis is a pain in my ass.  I'm done with this crap.

Nicofuma:
That's not travis... It's because you didn't follow the rules about the commit message format.
But we can take care of it for you if you want.

RMcGirr83:
Where are these rules?  Not seeing how it is different from any other commit I made and the error that is spat out makes absolutely no sense as the heading is a sentence beginning with a letter as far as I can tell.

PR #3737, opened 2015-07-03
[ticket/13980] Uploaded avatar not removed when new is uploaded

javiexin:
Remove the previously uploaded avatar when a new one with a different
extension is uploaded

PHPBB3-13980

### DIFF
--- a/phpBB/phpbb/avatar/driver/upload.php
+++ b/phpBB/phpbb/avatar/driver/upload.php
@@ -161,14 +161,11 @@ class upload extends \phpbb\avatar\driver\driver
 			return false;
 		}
 
+		// Delete current avatar if not overwritten
 		$ext = substr(strrchr($row['avatar'], '.'), 1);
 		if ($ext && $ext !== $file->get('extension'))
 		{
-			$filename = $this->phpbb_root_path . $this->config['avatar_path'] . '/' . $this->config['avatar_salt'] . '_' . $row['id'] . '.' . $ext;
-			if (file_exists($filename))
-			{
-				@unlink($filename);
-			}
+			$this->delete($row);
 		}
 
 		return array(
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/avatar/driver/upload.php
```diff
@@ -161,6 +161,16 @@ public function process_form($request, $template, $user, $row, &$error)
 			return false;
 		}
 
+		$ext = substr(strrchr($row['avatar'], '.'), 1);
+		if ($ext && $ext !== $file->get('extension'))
+		{
+			$filename = $this->phpbb_root_path . $this->config['avatar_path'] . '/' . $this->config['avatar_salt'] . '_' . $row['id'] . '.' . $ext;
```
marc1706: Please just use ```$this->delete($row);```

javiexin: I know this is the same code, but unfortunately, it would conflict with another ticket I have submitted: 
https://github.com/phpbb/phpbb/pull/3738

If I called the delete method from here, there would be no way to differentiate a "genuine" call to delete (that would require an event being triggered) with a code reuse use (this call) that would require the event is not called, as it would have been already called in this method (around L153 here).

It could be done by passing an extra parameter to the delete function that would flag this use, or by reconsidering the way to work the events.

But please, see both tickets in conjunction, and provide your opinion/feedback.  Will be happy to implement whatever it is that we decide here.

javiexin: Ok, I have been looking at this more, and what I would like to do is to call
$this->delete($row, $error);
And have $error be passed by reference and take a default value of array().
By doing this, in case an error would happen in delete method, the result could be propagated back to the user, and the success return would only be sent in case no error happened on delete either.  I would move this block to above, L156.
If you agree, I will update this fix, and the other ticket as well to be in line (prototype for delete function).

marc1706: Yeah, sounds good to me.

CHItA:
:+1:

PR #3745, opened 2015-07-06
[ticket/13986] Add --resume option to reparser CLI

s9e:
https://tracker.phpbb.com/browse/PHPBB3-13986
~~https://tracker.phpbb.com/browse/PHPBB3-13989~~

Resumes where the last execution stopped. If the last execution stopped at 227600/583468,  ~~it resumes at 227600/583468. The caveat is that it skews the estimation in the progress bar.~~ Not anymore, it resumes at 0/355868 and the estimation is as accurate as a normal execution.

### DIFF
--- a/phpBB/config/default/container/services_console.yml
+++ b/phpBB/config/default/container/services_console.yml
@@ -154,5 +154,6 @@ services:
         arguments:
             - @user
             - @text_reparser_collection
+            - @config_text
         tags:
             - { name: console.command }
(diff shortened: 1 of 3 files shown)


Nicofuma:
:+1:

Nicofuma:
needs to be rebased

s9e:
@Nicofuma Done.

Nicofuma:
I think we should also store range-min and range-max.
This way re:re --resume is enough (but if the user has provided a new value for these vars we should use them)

Nicofuma:
set as wip because it's the name of your last commit

s9e:
Right. I need to fix my `push.default` I didn't mean to push that commit.

s9e:
I updated the implementation to handle restoring options from the last run. The logic is in `get_option()`. It uses `hasParameterOption()` to determine whether an option was passed in the command line.

It restores `range-min`, `range-max` (adjusted to match the point where last execution stopped) and `range-size`. It doesn't save the state of `dry-run`.

Nicofuma:
nice

s9e:
@Nicofuma I'm done tweaking this PR. If you're ok with it you can merge anytime.

PR #3748, opened 2015-07-07
[ticket/12291] Allow extensions to use custom topic icons and smilies

rxu:
<a href="https://tracker.phpbb.com/browse/PHPBB3-12291">PHPBB3-12291</a>.


Nicofuma:
Can you add a test case please?

rxu:
Yes, but I'm currently guessing what kind of test could it be.

rxu:
!set WIP

rxu:
This also requires changes to the textformatter @s9e.

s9e:
@rxu What changes?

rxu:
F.e. here https://github.com/phpbb/phpbb/blob/master/phpBB/phpbb/textformatter/s9e/renderer.php#L144
```$this->set_smilies_path($root_path . $config['smilies_path']);``` to ```$this->set_smilies_path($root_path);```, dunno if there're another places regarding smilies and icons in the textformatter (didn't find it so far).

**Comment by s9e** [comment]
I'm not familiar with this PR and how it relates to smilies. From what I see it removes the dynamic `smilies_path` from images. In that case you'd need to remove the following methods:

 * phpbb\textformatter\renderer_interface::set_smilies_path()
 * phpbb\textformatter\s9e\renderer::configure_smilies_path()
 * phpbb\textformatter\s9e\renderer::set_smilies_path()

Remove `{$T_SMILIES_PATH}/` from the template set in `phpbb\textformatter\s9e\factory`, and remove the `configure_smilies_path` call from `config/default/container/services_text_formatter.yml`.

@rxu I can send you a PR if you want.

**Comment by rxu** [comment]
@s9e Yeah, the idea is to dump the full relative path to icons and smilies to the corresponding DB columns (```smiley_url```/```icons_url```) to allow using icons/smilies from extensions. But ```acp_icons``` module will still be looking for core icons/smilies in ```smilies_path```/```icons_path``` to write the path to the db.

> I can send you a PR if you want.

That would be nice, thanks. Although I'm not sure if smilies path configuration should be removed fully from the renderer, as it should be pointed to the root path anyway, otherwise smilies won't be displayed as far as I see because of not prepended with the board URL.

**Comment by rxu** [comment]
Note:L this PR will be conflicting with #3799 as the latter removes ```/phpbb/install/schemas/schema_data.sql```.

**Comment by rxu** [comment]
#3748 rebased onto the latest master branch.

**Comment by rxu** [comment]
#3748 rebased onto the latest master branch (again).

**Comment by rxu** [comment]
Any pointers on what else tests to add would be much appreciated (there're already several tests involving smilies for the text formatter / text processing).

File: phpBB/includes/functions.php [hunk 1]
```diff
@@ -4402,10 +4402,10 @@ function page_header($page_title = '', $display_online_list = false, $item_id =
 		'T_TEMPLATE_PATH'		=> "{$web_path}styles/" . rawurlencode($user->style['style_path']) . '/template',
 		'T_SUPER_TEMPLATE_PATH'	=> "{$web_path}styles/" . rawurlencode($user->style['style_path']) . '/template',
 		'T_IMAGES_PATH'			=> "{$web_path}images/",
-		'T_SMILIES_PATH'		=> "{$web_path}{$config['smilies_path']}/",
+		'T_SMILIES_PATH'		=> "{$web_path}",	// {$config['icons_path']}/ is not used as for 3.2 smiley path is dumped into the DB for every smiley
```
Elsensee: So, T_SMILIES_PATH is basically `$web_path`?

phpBB/includes/functions_acp.php
```diff
@@ -80,10 +80,10 @@ function adm_page_header($page_title)
 		'U_INDEX'				=> append_sid("{$phpbb_root_path}index.$phpEx"),
 
 		'T_IMAGES_PATH'			=> "{$phpbb_root_path}images/",
-		'T_SMILIES_PATH'		=> "{$phpbb_root_path}{$config['smilies_path']}/",
+		'T_SMILIES_PATH'		=> "{$phpbb_root_path}",	// {$config['icons_path']}/ is not used as for 3.2 smiley path is dumped into the DB for every smiley
 		'T_AVATAR_PATH'			=> "{$phpbb_root_path}{$config['avatar_path']}/",
 		'T_AVATAR_GALLERY_PATH'	=> "{$phpbb_root_path}{$config['avatar_gallery_path']}/",
-		'T_ICONS_PATH'			=> "{$phpbb_root_path}{$config['icons_path']}/",
+		'T_ICONS_PATH'			=> "{$phpbb_root_path}",	// {$config['icons_path']}/ is not used as for 3.2 icon path is dumped into the DB for every icon
```
Elsensee: Really, I would just write `$phpbb_root_path`. (or `$web_path` in functions.php, see above)

phpBB/phpbb/db/migration/data/v310/update_icons_smilies_url.php
```diff
@@ -0,0 +1,80 @@
...
+*/
+
+namespace phpbb\db\migration\data\v310;
+
+class update_icons_smilies_url extends \phpbb\db\migration\migration
+{
+	static public function depends_on()
+	{
+		return array(
+			'\phpbb\db\migration\data\v310\rc5',
```
Elsensee: If this patch goes into 3.2 only, this needs to be changed and moved.

@Nicofuma in 3.2 or not?

phpBB/phpbb/db/migration/data/v310/update_icons_smilies_url.php
```diff
@@ -0,0 +1,80 @@
...
+			$smilies_url[(int) $row['smiley_id']] = $row['smiley_url'];
+		}
+		$this->db->sql_freeresult($result);
+
+		foreach ($smilies_url as $smiley_id => $smiley_url)
+		{
+			$sql = 'UPDATE ' . $this->table_prefix . "smilies
+				SET smiley_url = '" . $this->db->sql_escape("{$this->config['smilies_path']}/$smiley_url") . "'
+				WHERE smiley_id = '$smiley_id'";
+			$this->sql_query($sql);
```
Elsensee: I know some boards that have hundreds of smileys. What's the probability of a timeout there?

phpBB/includes/functions.php
rxu: Yes.

phpBB/includes/functions_acp.php
rxu: Indeed, good catch.

phpBB/phpbb/db/migration/data/v310/update_icons_smilies_url.php
rxu: No idea, but is there a way to split it in migration?

JoshyPHP: Can't you update everything at once with an `UPDATE` query? Something like that:

```php
$path   = $this->db->sql_escape($this->config['smilies_path'] . '/');
$concat = $this->db->sql_concatenate($path, 'smiley_url');

$sql = 'UPDATE ' . $this->table_prefix . "smilies
		SET smiley_url = '" . $concat . "';
```

rxu: I'll try it, thanks.

phpBB/phpbb/db/migration/data/v310/update_icons_smilies_url.php
Nicofuma: maybe 3.2.1

phpBB/phpbb/db/migration/data/v310/update_icons_smilies_url.php
Elsensee: @JoshyPHP's proposal is better, so if that works, do that.

But if not, you can have a look at this: https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/phpbb/db/migration/data/v310/passwords_convert_p1.php#L30

gooof:
Please don't remove $config['icons_path'] or $config['smilies_path'].

Simply remove phpbb_root_path and change it to $web_path, its the same without database changes.

gooof:
Extensions could use a function like `ext_image_path('icon.png')` to get `{web_path}ext/ext_vendor/ext_name/icons/icon.png`

rxu:
No config values are going to be removed. As for the rest, I don't see how you could get the oath without having it in the database.

gooof:
The same way what you are doing with "smilies_path" but also with namespaces or full path. 

With full path only `/ext/ext_vendor/ext_name/icons/icon.png` so you can replace the first `/` or `/ext/` with web_path or `$web_path/ext/` later on display.

rxu:
Icons are set by icon id, how would you guess where the icon is located? Icon names can be identical for several extensions etc.

gooof:
Extensions should be able to use the icons table? hm why? only for new topic icons on posting?

gooof:
I prefer:
- extensions should also use phpbb writeable paths
- image (avatar like) upload for icons.
- (simple phpbb) way to move files from extensions to writeable paths. That will allow extensions also to use the icons without big changes.
- function for extensions to get image paths like: `ext/ext_vendor/ext_name/icons/`
- never phpbb_root_path for images, always board_url/web_path

rxu:
I don't have in plans something like that and have no idea about such implementation, sorry. The aim of this PR is just to allow extensions bringing custom icons/smilies onboard without having to mess with phpBB core.
If you have a better solution, please feel free to send another/alternative PR. I'm done with the current one to be honest and actually i'm having hackish way to get it work for my ext without this change already.

gooof:
Which extension?

rxu:
This one https://github.com/rxu/topic_actions

gooof:
hm, one icon only?

The simple way would be: Add a option to your config page, let the user choose the icon from all in database and use it as trash icon?

You can also set a column to the topics_table and display the icon over template events to topic_titles. That also works if icons disabled or deleted.

rxu:
The ext was just a trigger for the PE idea (like for all other PRs) pretty long time ago. Since that moment I've got it workable as it is now.

gooof:
There is no good reason for that. Why should extensions intervene in these two tables. 
Smileys can be added via the [img] tag. Icons must be added over the ACP tab.

Extensions destroy the sorting system. Extension will be deleted - the icon is not found any more...

rxu:
This is not an issue as extensions have to handle adding/deleting icons data in DB via migrations, like all the rest of stuff (modules etc).

gooof:
Why than no function to handle phpbb added icons and ext icons?
replace $config['icons_path'] lines with e.g. icons_path_func()
column: smiley_url:
- phpbb added icons beginning with: icon_ (icon_rolleyes.gif)
- ext icons beginning with: ext/ (ext/ext_vendor/ext_name/icons/trash.gif)

rxu:
Actually I'm not sure anymore if this option really needed in 3.2+.

marc1706:
Why do you think that? Could you elaborate on that?

rxu:
Well, with textformatter supporting Emoji there's very small chance for extension using custom smilies will appear as using Emoji is much easier.
As for topic icons, It probably would be better to implement using Emoji as icons as well.
Just thinking.

CHItA:
I'm closing this as it seems to be abandoned, feel free to reopen the PR if you wish to continue working on it.

Leinad4Mind:
This appears on changelog of 3.2.2 but wasnt added to the release?

rxu:
This wasn't actually added, just a changelog entry (by accident).

PR #3749, opened 2015-07-08
[ticket/13995] Fixes for invalid HTML in docs and phpBB

VSEphpbb:
Projection is a deprecated media type and should not be used.
Name is an invalid attribute in anchors and ID is used instead.
Avatars are invalid by missing a space between elements.

https://tracker.phpbb.com/browse/PHPBB3-13996


Nicofuma:
You need to update some tests

VSEphpbb:
Updated

PR #3756, opened 2015-07-12
[ticket/14005] Allow extensions control post buttons displaying

rxu:
<a href="https://tracker.phpbb.com/browse/PHPBB3-14005">PHPBB3-14005</a>.


vinny:
something similar should be added to <a href="https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/styles/prosilver/template/ucp_pm_viewmessage.html#L76">ucp_pm_viewmessage.html#L76</a>

rxu:
@vinny Makes sense, thanks. Updated.

PR #3760, opened 2015-07-15
[ticket/14014] Add core.mcp_forum_view_before event

Zoddo:
This event can be used to get some data in order to execute other actions.

https://tracker.phpbb.com/browse/PHPBB3-14014


Zoddo:
!set Event

PR #3762, opened 2015-07-15
[ticket/13973] Add core.index_page_after_common event

Zoddo:
https://tracker.phpbb.com/browse/PHPBB3-13973


Zoddo:
!set Event

marc1706:
Why not just use core.user_setup right before this or core.display_forums_modify_sql right after this?

Zoddo:
Yes, It's right, we can close this PR

PR #3763, opened 2015-07-15
[ticket/13968] Add core.user_setup_after event

Zoddo:
It allows to execute code at the end of user setup.
It can replace the `phpbb_user_session_handler` hook.

https://tracker.phpbb.com/browse/PHPBB3-13968

### DIFF
--- a/phpBB/phpbb/user.php
+++ b/phpBB/phpbb/user.php
@@ -301,6 +301,14 @@ class user extends \phpbb\session
 		// After calling it we continue script execution...
 		phpbb_user_session_handler();
 
+		/**
+		* Execute code at the end of user setup
+		*
+		* @event core.user_setup_after
+		* @since 3.1.6-RC1
+		*/
+		$phpbb_dispatcher->dispatch('core.user_setup_after');
+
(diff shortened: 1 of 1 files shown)


Zoddo:
!set Event

marc1706:
This would also make #3762 useless as you could just use this and check the page you're on.

phpBB/phpbb/user.php
```diff
@@ -301,6 +301,14 @@ function setup($lang_set = false, $style_id = false)
 		// After calling it we continue script execution...
 		phpbb_user_session_handler();
 
+		/**
+		* Execute code after that sessions are initialized
```
marc1706: Execute code at the end of user setup

Zoddo: @marc1706: If the event is moved at the end of the function, it will not executed if the board is disabled, so I think it's on the good location.

marc1706: I meant that you should change the text to that :)

Zoddo: Ah, sorry :p

PR #3765, opened 2015-07-15
[ticket/13950] Hide undefined permissions

Zoddo:
This hides the permissions of the disabled extensions

https://tracker.phpbb.com/browse/PHPBB3-13950

### DIFF
--- a/phpBB/includes/acp/auth.php
+++ b/phpBB/includes/acp/auth.php
@@ -1139,6 +1139,11 @@ class auth_admin extends \phpbb\auth\auth
 			@reset($cat_array['permissions']);
 			while (list($permission, $allowed) = each($cat_array['permissions']))
 			{
+				if (!$phpbb_permissions->permission_defined($permission))
+				{
+					continue;
+				}
+
 				if ($s_view)
 				{
 					$template->assign_block_vars($tpl_cat . '.' . $tpl_mask, array(
--- a/phpBB/includes/acp/auth.php
+++ b/phpBB/includes/acp/auth.php
@@ -1113,6 +1113,11 @@ class auth_admin extends \phpbb\auth\auth
 		@reset($category_array);
 		while (list($cat, $cat_array) = each($category_array))
 		{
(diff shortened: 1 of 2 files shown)


Zoddo:
I don't know if it's the best way to do this.

phpBB/phpbb/permissions.php
```diff
@@ -160,6 +160,17 @@ public function get_permission_category($permission)
...
 
 	/**
+	* Checks if a permission has been defined
+	*
+	* @param	string	$permission	Identifier of the permission
+	* @return	bool	True if the permission is defined, false otherwise
+	*/
+	public function permission_defined($permission)
+	{
+		return isset($this->permissions[$permission]);
```
Nicofuma: if I'm not mistaken this will always return false for an extension

Zoddo: @Nicofuma : No, I've tested my modifications, and they worked as expected.

Zoddo:
It's good or there is something to fix?

phpBB/includes/acp/auth.php
```diff
@@ -1139,6 +1139,11 @@ function assign_cat_array(&$category_array, $tpl_cat, $tpl_mask, $ug_id, $forum_
 			@reset($cat_array['permissions']);
 			while (list($permission, $allowed) = each($cat_array['permissions']))
 			{
+				if (!$phpbb_permissions->permission_defined($permission))
```
Nicofuma: I think you should do the same thing a few lines upper. Because extensions can also add custom categories

phpBB/phpbb/permissions.php
VSEphpbb: Just tested and seems to work for me too.

Nicofuma: it works pretty well yes :)

phpBB/includes/acp/auth.php
Zoddo: @Nicofuma It's done :)

PR #3770, opened 2015-07-16
 [ticket/14017] Extend avatar-driver by extension in ACP not possible 

javiexin:
Create a driver method to provide the driver config name, and use it
within the manager method. Default driver config name is the same as now.
But new drivers are able to override the config name with their own.

PHPBB3-14017


phpBB/phpbb/avatar/driver/driver.php
```diff
@@ -120,6 +120,14 @@ public function get_name()
 	}
 
 	/**
+	* {@inheritdoc}
+	*/
+	public function get_config_name()
+	{
+		return preg_replace('#^phpbb\\\\avatar\\\\driver\\\\#', '', get_class($this));
```
Nicofuma: str_replace please

javiexin:
I am not sure, as I don't understand the output of Travis too well, but it looks as if the errors are coming from the fact that I included a new method in avatar driver interface, and that is not taken into consideration in the test cases.

javiexin:
No more edits needed.  All the existing drivers use the default implementation provided in the abstract base class driver.php, so no changes needed to any of these files.  That was the intention all along.  Only include an override function in your own class implementation.  Zero migration, zero conflicts with existing code, zero integration issues.
And the part of acp_users, no change needed. My patch is 100% backwards compatible.
By the way, have you cared to test these changes, including the additional method as I said?

ForumHulp:
Yes I just tested it.
The name is ok now but the acp_user problem is still there.It only looks in adm/style and not in extension adm folder.

javiexin:
Try putting your file in the styles/all folder, and see if it works.
I have the exact same construct with new profilefield types, and the templates are being accessed ok.
Maybe is an issue with the adm/ folder access.
Also, have you tested with adm/style folder in the extension?.

ForumHulp:
No it won't work, it's only looking in the adm/style folder of the board.

Off course I tested with adm/style folder in extension and that gave me the solution in acp_users..

javiexin:
@ForumHulp take a look at additional changes.  You have to override the get_acp_template_name method in your class (along with the get_config_name method), and use a template name starting by @vendor_extension, and then it will be found in the extension folder.

javiexin:
Travis errors are due to the new methods introduced in the interface, not considered in the test cases.

Nicofuma:
Hi, you need to update the tests.

javiexin:
Don't know how to change it, and anyhow I am off for some time (illnes)

rxu:
This is due to the test's ```barfoo``` driver is disabled and is not constructed and therefore doesn't inherit the parent class methods, which the ```get_config_name()``` is one of. before this patch, you didn't need to call the driver class methods to get its parameters.

javiexin:
The patch is correct.

I do not know how to fix the testing framework errors, and learning this to the point that I am confident with my changes it will take too much time for me, and I consider that this time is better used in other ways.

If this is not acceptable, reject the change and close it, but the underlying ticket will remain open.

Sorry. -javiexin

!unset WIP

javiexin:
I am sorry, but I will not update the testing framework.

@ForumHulp maybe you want to take over.  Otherwise, I think I will close this PR.

ForumHulp:
I also don't know how!

marc1706:
Replaced by #4105

PR #3772, opened 2015-07-16
[ticket/14018] Fix responsive quick link menus

VSEphpbb:
https://tracker.phpbb.com/browse/PHPBB3-14018

PHPBB3-14018

### DIFF
--- a/phpBB/styles/prosilver/theme/common.css
+++ b/phpBB/styles/prosilver/theme/common.css
@@ -327,6 +327,7 @@ ul.linklist li.responsive-menu {
 ul.linklist li.responsive-menu a.responsive-menu-link {
 	display: inline-block;
 	margin: 0 5px;
+	font-size: 16px;
 	position: relative;
 	width: 16px;
 	line-height: 1.2em;
@@ -1257,7 +1258,7 @@ ul.linklist:after,
 }
 
 #quick-links a.responsive-menu-link:before {
-	font-size: 1.6em;
+	font-size: 16px;
 	line-height: 16.5px;
 }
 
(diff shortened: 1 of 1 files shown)


callumacrae:
Why can't this be done as em?

marc1706:
Yes, please elaborate how to reproduce this issue and why em doesn't work here.

callumacrae:
Why on earth is the font-size being set to zero? That's completely the wrong approach to hiding text on mobile.

I'd suggest actually fixing the main issue instead of fixing the symptoms of the weirdness.

VSEphpbb:
Check the ticket for elaboration. Resize your browser down to mobile phone size to see the Quick Links menu vanish in the header, and get distorted in the footer.

In responsive, the font size is 0 from here: https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/styles/prosilver/theme/common.css#L1264-L1266

Changing that to a non-zero breaks the layout in the nav bars. So it's 0 for a reason. So we need to use px for the `a.responsive-menu-link`s because em relative to 0 is just 0.

VSEphpbb:
Here's another version. Using em again.
1. First, replacing omitted font size from the footer navbar
2. em is 1.455 because the base font size is 11px, and we're trying to get to 16px so that = 1.455em
3. Instead of hiding the "Quick Links" text by using font-size:0, instead use width:0 with overflow and word wrap trickery. hopefully it's browser compatible (works for me on Mac with latest Safari, Chrome and FF and on iOS).

callumacrae:
lgtm :+1:

PR #3774, opened 2015-07-16
[ticket/14027] Add mcp_forum_display_options_* template events

adri1b:
PHPBB3-14027
PHPBB3-14028

### DIFF
--- a/phpBB/styles/prosilver/template/mcp_forum.html
+++ b/phpBB/styles/prosilver/template/mcp_forum.html
@@ -96,7 +96,9 @@
 	<!-- ENDIF -->
 
 	<fieldset class="display-options">
+		<!-- EVENT mcp_forum_display_options_prepend -->
 		<label>{L_DISPLAY_TOPICS}{L_COLON} {S_SELECT_SORT_DAYS}</label>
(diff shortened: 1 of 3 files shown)


adri1b:
!set Event

marc1706:
Needs to be rebased onto current 3.1.x

phpBB/styles/prosilver/template/mcp_forum.html
```diff
@@ -96,7 +96,9 @@
 	<!-- ENDIF -->
 
 	<fieldset class="display-options">
+		<!-- EVENT mcp_forum_display_options_prepend -->
 		<label>{L_DISPLAY_TOPICS}{L_COLON} {S_SELECT_SORT_DAYS}</label>
+		<!-- EVENT mcp_forum_display_options_append -->
```
marc1706: Shouldn't the append be at the end of the sorting options instead of right after the dropdown for the timeframe?

marc1706:
Pull request will be deleted upon next release as it seems to have been abandoned.

marc1706:
Closed as previously mentioned.

PR #3773, opened 2015-07-16
[ticket/14015] Update Symfony

Nicofuma:
```
- Removing symfony/filesystem (2.8.x-dev 45cc231)
- Installing symfony/filesystem (2.8.x-dev 9f70c56)
	Cloning 9f70c5625a32b2f1e6fc37222f52b4e0eb437b0e

- Removing symfony/config (2.8.x-dev 291e3f1)
- Installing symfony/config (2.8.x-dev 358ec92)
	Cloning 358ec929e494b6f12d8508d88357cbd7383a10ca

- Removing symfony/console (2.8.x-dev 7e857a2)
- Installing symfony/console (2.8.x-dev fd85e75)
	Cloning fd85e7517e79a2bceafcee8f7e8b7bbd0919a90a

- Removing symfony/dependency-injection (2.8.x-dev 90a7527)
- Installing symfony/dependency-injection (2.8.x-dev f279142)
	Cloning f279142fb6bc21e83fcd40e0a01a62ff527268be

- Removing symfony/finder (2.8.x-dev fd26bdb)
- Installing symfony/finder (2.8.x-dev 8712d5e)
	Cloning 8712d5e8c4ad65e6b936ed9b6a581e5e6a87fddf

- Removing symfony/http-foundation (2.8.x-dev 06aa08b)
- Installing symfony/http-foundation (2.8.x-dev 5339030)
	Cloning 5339030f5af0d6fef7c897130291ac168278289f

- Removing symfony/event-dispatcher (2.8.x-dev 8766ceb)
- Installing symfony/event-dispatcher (2.8.x-dev d724688)
	Cloning d7246885b7fe4cb5a2786bda34362d2f0e40b730

- Removing symfony/debug (2.8.x-dev eda866d)
- Installing symfony/debug (2.8.x-dev d1114d8)
	Cloning d1114d892ae70f833871dc55599f17084191efaa

- Removing symfony/http-kernel (2.8.x-dev c587069)
- Installing symfony/http-kernel (2.8.x-dev f24d2cb)
	Cloning f24d2cb4778ecc5f6c8e8f7f31bc4691099a920d

- Removing symfony/routing (2.8.x-dev 76b18ed)
- Installing symfony/routing (2.8.x-dev 9e797cd)
	Cloning 9e797cd70762db18d87c59dc6bc1ab7c3eb963ed

- Removing symfony/security-core (2.8.x-dev 2c21845)
- Installing symfony/security-core (2.8.x-dev 8253e02)
[...]

PR #3776, opened 2015-07-17
[ticket/14033] Fix errors in docblocks

VSEphpbb:
Should help to pass https://github.com/phpbb/phpbb/pull/3773

PHPBB3-14033


Nicofuma:
thanks

PR #3778, opened 2015-07-17
[ticket/14036] Replace path_helper with a mock

s9e:
https://tracker.phpbb.com/browse/PHPBB3-14036

### DIFF
--- a/tests/test_framework/phpbb_test_case_helpers.php
+++ b/tests/test_framework/phpbb_test_case_helpers.php
@@ -441,7 +441,7 @@ class phpbb_test_case_helpers
 				->getMock();
 			$path_helper->expects($this->test_case->any())
 				->method('get_web_root_path')
-				->will($this->test_case->returnValue('./'));
+				->will($this->test_case->returnValue('phpBB/'));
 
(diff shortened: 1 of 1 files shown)


tests/test_framework/phpbb_test_case_helpers.php
```diff
@@ -435,20 +435,15 @@ public function set_s9e_services(ContainerInterface $container = null, $fixture
...
-				'path_helper',
-				$path_helper
-			);
+			$path_helper = $this->test_case->getMockBuilder('phpbb\\path_helper')
+				->disableOriginalConstructor()
+				->setMethods(array('get_web_root_path'))
+				->getMock();
+			$path_helper->expects($this->test_case->any())
+				->method('get_web_root_path')
+				->will($this->test_case->returnValue('./'));
```
VSEphpbb: Should this be something like $phpbb_root_path instead of ./

s9e: It can be anything, it's only used to set the path to smilies images. I've just checked the other tests and they expect it to be `phpBB/`.

PR #3782, opened 2015-07-20
[ticket/13711] Notifications are sent to inactive users

RMcGirr83:
Notifications are sent to inactive users for emails and jabber.  There is no check if a user is set to inactive.

[PHPBB3-13711] (https://tracker.phpbb.com/browse/PHPBB3-13711)

### DIFF
--- a/phpBB/phpbb/notification/method/messenger_base.php
+++ b/phpBB/phpbb/notification/method/messenger_base.php
@@ -69,7 +69,7 @@ abstract class messenger_base extends \phpbb\notification\method\base
 
 			$user = $this->user_loader->get_user($notification->user_id);
 
-			if ($user['user_type'] == USER_IGNORE || in_array($notification->user_id, $banned_users))
+			if ($user['user_type'] == USER_IGNORE || ($user['user_type'] == USER_INACTIVE && $user['user_inactive_reason'] == INACTIVE_MANUAL) || in_array($notification->user_id, $banned_users))
 			{
 				continue;
 			}


Nicofuma:
Why wouldn't you send the notifications to the inactive users?

RMcGirr83:
Please read the ticket that I linked to.

VSEphpbb:
But when the user becomes activated, their notifications will be waiting for them.

bantu:
Without looking into it, I doubt the patch is a good idea as is. There are many reasons why a user account may be considered inactive. I'd argue that it makes sense to receive notifications, for example when an account is only temporarily inactive. Also see https://wiki.phpbb.com/Inactive_Users

everflux:
Currently there is no way to differentiate between a deactivated user that should receive notifications and one that should receive notifications. With the older phpbb versions they didn't get notifications, which was fine this changed with phpbb 3.1

If there is a good reason to keep to this change either
- let the admin decide (configure) how to handle it
- or introduce a different state "deactivated without notifications"

The reason is that there are use cases when you don't want to delete a user right away, but disable him (f.e. for review purposes). There is no reason keep sending the user notifications. According to german (and probably other countries) laws it may be even forbidden to keep sending emails.

bantu:
@everflux I am not sure that is actually true. We track how an user account became inactive in the `user_inactive_reason` of the `phpbb_users` table. See https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/includes/constants.php#L53

everflux:
I don't get your point. If an admin disables a user instead of deleting him so he can be reactivated later, why should the system keep on sending emails?

DavidIQ:
Because an admin deactivating an account is not the only reason an account would be inactive, which is bantu's point. Sending emails to an inactive account could also serve as a reminder that they need to reactivate their account.

everflux:
That is a good point, but the only kind of mail that I can think of that would make a valid use case is if a) the account has been deactivated due to a profile change (not by the admin) and b) the email to be sent is an email address validation mail.
All other cases, users sending contact mails, mass emails, topic subscriptions, notifications, ... should not be sent to an inactive account.

RMcGirr83:
Further an admin can always remind inactive users via the acp

VSEphpbb:
An inactive user is still a user. If their account is receiving notifications they should be notified of the activity related to their account, regardless if they are currently active or not.

RMcGirr83:
By that logic a banned user is still a user too.  Seems there is a
consensus that this shouldn't be committed so set this to closed and set
the ticket to will not fix with an appropriate reason as to why.

RMcGirr83:
Just to be clear this only affects emails not notifications on the forum
itself

everflux:
You are not serious about closing it and won't fix?!

bantu:
I believe the condition you're adding simply needs to be `$user['user_type'] == USER_INACTIVE && $user['user_inactive_reason'] == INACTIVE_MANUAL`. What do you guys think about that?

bantu:
https://wiki.phpbb.com/Inactive_Users actually already defines desired behaviour ...

everflux:
Thanks for pointing that out - I believe this behaviour is a regression in phpbb 3.1, afaik it was correct in phpbb 3.0

RMcGirr83:
@bantu that would seem to satisfy the request.

PR #3784, opened 2015-07-21
[ticket/14043] Add core.get_avatar_after php event

Senky:
[PHPBB3-14043](https://tracker.phpbb.com/browse/PHPBB3-14043)

### DIFF
--- a/phpBB/includes/functions.php
+++ b/phpBB/includes/functions.php
@@ -4766,7 +4766,7 @@ function phpbb_get_avatar($row, $alt, $ignore_config = false)
 {
 	global $user, $config, $cache, $phpbb_root_path, $phpEx;
 	global $request;
-	global $phpbb_container;
+	global $phpbb_container, $phpbb_dispatcher;
 
 	if (!$config['allow_avatar'] && !$ignore_config)
 	{
@@ -4811,7 +4811,7 @@ function phpbb_get_avatar($row, $alt, $ignore_config = false)
 	*
 	* @event core.get_avatar_after
 	* @var	string	html		The HTML <img> tag of generated avatar
-	* @since 3.1.6
+	* @since 3.1.6-RC1
 	*/
 	$vars = array('html');
 	extract($phpbb_dispatcher->trigger_event('core.get_avatar_after', compact($vars)));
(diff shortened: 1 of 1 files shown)


phpBB/includes/functions.php
```diff
@@ -4806,6 +4806,16 @@ function phpbb_get_avatar($row, $alt, $ignore_config = false)
...
 
+	/**
+	* Event to modify HTML <img> tag of avatar
+	*
+	* @event core.get_avatar_after
+	* @var	string	html		The HTML <img> tag of generated avatar
+	* @since 3.1.6
+	*/
+	$vars = array('html');
+	extract($phpbb_dispatcher->trigger_event('core.get_avatar_after', compact($vars)));
```
rxu: ```$phpbb_dispatcher``` should be listed in ```global```.

phpBB/includes/functions.php
```diff
@@ -4806,6 +4806,16 @@ function phpbb_get_avatar($row, $alt, $ignore_config = false)
 			'alt="' . ((!empty($user->lang[$alt])) ? $user->lang[$alt] : $alt) . '" />';
 	}
 
+	/**
+	* Event to modify HTML <img> tag of avatar
+	*
+	* @event core.get_avatar_after
+	* @var	string	html		The HTML <img> tag of generated avatar
+	* @since 3.1.6
```
rxu: 3.1.6-RC1

phpBB/includes/functions.php
```diff
@@ -4806,6 +4806,16 @@ function phpbb_get_avatar($row, $alt, $ignore_config = false)
...
 	}
 
+	/**
+	* Event to modify HTML <img> tag of avatar
+	*
+	* @event core.get_avatar_after
+	* @var	string	html		The HTML <img> tag of generated avatar
+	* @since 3.1.6-RC1
+	*/
+	$vars = array('html');
```
javiexin: These event should have more (context) variables to allow for extensions to choose what to do.  For example, information about the user and the unprocessed avatar data, at a minimum, should be provided, just in case an extension might need them.  Creating this event just with html as the only exposed variable seems too limiting.
Even data that is available through html, if needed, would require parsing of the provided html code.
I would recommend adding extra variables as parameters to the event.

PR #3785, opened 2015-07-21
[ticket/9687] Refactoring banning

Elsensee:
Our current banning system is not very reliable, especially if we're talking about IPs. (And even there only IPv4 does work)
That's why I thought that we just start at 0 with this. (Not really at 0 because the current user and email banning is almost 1:1 from the core code).
I wanna split it into several classes so some extension developers could easily integrate new banning types.
- [x] Ban by username
- [x] Ban by email
- [ ] Ban by IP
  - [ ] Single IPs
  - [ ] CIDR Range
  - [ ] Current used ban range
  - [ ] Hostname
  - [ ] Nested exclude
- [ ] Cron job
- [ ] Migration
- [ ] Output to ACP/MCP
- [ ] Replace old parts with new service

[PHPBB3-9687](https://tracker.phpbb.com/browse/PHPBB3-9687)


Nicofuma:
please, for every function handling both IPv6 and IPv4 split them to make it easier to read (don't change the current interface but move the IPv6 and IPv4 logic in dedicated methods)

phpBB/phpbb/ban/type/ip.php
```diff
@@ -0,0 +1,550 @@
...
+
+		// We will overwrite all blocks anyway, we just need to keep the order
+		$diff_blocks = $bytes_ip1;
+		// Set carry to -1 to add 1 to the overall result
+		$carry = -1;
+
+		for ($i = 8; $i >= 1; $i--)
+		{
+			$diff = $bytes_ip2[$i] - $bytes_ip1{$i} - $carry;
+			$carry = ($diff < 0) ? 1 : 0;
```
CHItA: When `$bytes_ip2[8] = 0xffff; $bytes_ip1[8] = 0;`, this won't work. 

I know, it is my fault. ;)

Elsensee: I knew it was too easy... :P
Also nice `{$i}` there.. xD Me idiotus :D

Elsensee:
@Nicofuma mhm.. That's probably not necessary. The function `normalise_cidr()` should be okay - pretty small, I think.
More problematic is `range_to_cidr()` which is damn big. But the issue here is that most code is duplicated which is already pretty bad but I didn't know any way around it. It's: is it valid? is it the largest range? is it the smallest range? unpack address and swap if necessary. and so on. It's just the same.
I'll probably go with getting rid of the big if and make the ipv6 specific functions work with ipv4 as well. (but in a nice way - you'll see :D )

Oyabun1:
Something to consider in refactoring:

There have been several support topics in 3.0, presumably it can still occur with 3.1, where lock-up problems occur in the ACP when there are a lot of records in the banlist table. The tipping point probably depends on server resources but, commonly, the problems start to appear when there are somewhere between 8K and 20K rows. The issue seems to occur more often with IP bans, but that may just be because they can be entered as a range.

The problem is only reported maybe a couple of times a year. However, it may not be more common simply because not many boards have thousands of bans.

The problems can be resolved by deleting bans of a particular type or truncating the table, neither of which is ideal.

Elsensee:
@Oyabun1 Thank you very much, I'll try to reduce the output time.
The fact, that this occurs more often with IP bans is not very surprising, though, since a range of IPs is converted to every single IP within that range **before** giving it to the database, which is, of course, very bad for the performance. (Although sometimes wildcard are added at the end)

phpBB/phpbb/ban/manager.php
```diff
@@ -0,0 +1,270 @@
...
+	{
+		$this->type_map = array();
+
+		foreach ($types as $type)
+		{
+			$this->type_map[$type->get_type()] = $type;
+		}
+	}
+
+	public function ban($mode, $ban_list, $ban_length, $ban_length_other, $ban_exclude, $ban_reason, $ban_reason_display = '')
```
Nicofuma: doc block ?

phpBB/phpbb/ban/manager.php
```diff
@@ -0,0 +1,270 @@
...
+			else
+			{
+				$this->log->add($log_mode, $this->user->data['user_id'], $this->user->ip, $log_entry . strtoupper($ban_type->get_type()), false, $data);
+			}
+		}
+
+		return true;
+	}
+
+	public function check($mode = 'all', $data = false)
```
Nicofuma: doc block?

phpBB/phpbb/ban/manager.php
```diff
@@ -0,0 +1,270 @@
...
+			if ($ban_result == 'exclude')
+			{
+				return false;
+			}
+		}
+
+		return $ban_result;
+	}
+
+	public function unban($mode, $ban_ids)
```
Nicofuma: doc block?

phpBB/phpbb/ban/manager.php
```diff
@@ -0,0 +1,270 @@
...
+			else
+			{
+				$this->log->add($log_mode, $this->user->data['user_id'], $this->user->ip, 'LOG_UNBAN_' . strtoupper($ban_type->get_type()), false, $data);
+			}
+		}
+
+		return true;
+	}
+
+	public function tidy()
```
Nicofuma: doc block?

phpBB/phpbb/ban/manager.php
```diff
@@ -0,0 +1,270 @@
...
+	 *
+	 * @param \phpbb\di\service_collection
+	 */
+	protected function fill_type_map($types)
+	{
+		$this->type_map = array();
+
+		foreach ($types as $type)
+		{
+			$this->type_map[$type->get_type()] = $type;
```
Nicofuma: avoid loading the services it's better to keep the collection and use the service name for $mode

phpBB/phpbb/ban/type/ip.php
```diff
@@ -0,0 +1,524 @@
...
+				//$wildcards = substr_count($ban_entry, '*');
+				//$banned_ips[] = str_replace('*', '0', $ban_entry) . '/' . (32 - $wildcards * 4);
+				$banned_ip = implode('/', $ban_item);
+				if (stripos($banned_ip, '/') !== false)
+				{
+					$banned_ip = $this->normalise_cidr($banned_ip);
+				}
+				$banned_ips[] = $banned_ip;
+			}
+			// Okay, we are really lazy here...
```
Nicofuma: doesn't add anything, please explain what you are doing

Elsensee: I am sorry. :D But the ban function is not in its final state - I will probably change everything again. This applies to everything I already wrote because I'm almost never satisfied with the result, I have.
However, I will look through everything and change the comments if necessary.

phpBB/phpbb/ban/manager.php
Elsensee: (and for the following comments regarding doc blocks)
I am not finished with the manager yet - at least I'm not even close satisfied. I just took some code from the old add_ban, check_ban and remove_ban functions in functions_user.php and user.php and put them into this file. Thanks for the reminder, though. ;)

Nicofuma:
I don't think it will make it for 3.2. Sorry. Maybe 3.3?

Elsensee:
Pretty sure it won't make it for 3.2, yes. 3.3 sounds good.

CHItA:
Would that be possible that in `type` classes you don't use queries but create a `ban_data_loader` class for example, + an interface for it, and do the db queries there?

Elsensee:
:+1: 

Also, regarding IPs, I'll probably convert IPs to decimals and store them in multiple columns in the database. I asked in ##php for their opinions about this:
```
<Bittarman> Elsensee, normalize them to their integer values on the way in, and then you can query based on that...
```

Elsensee:
I'm closing this. There's a new one now, with a rewrite of my rewrite of my rewrite of the old code. And it's going to be the final one. #5393

PR #3789, opened 2015-07-23
[ticket/13982] Remove extra new line

marc1706:
PHPBB3-13982


Nicofuma:
there is one other line 1905

PR #3791, opened 2015-07-26
[ticket/14049] Use headers from config when sending Plupload delete request.

prototech:
[PHPBB3-14049](https://tracker.phpbb.com/browse/PHPBB3-14049)


bantu:
What problems does this cause?

prototech:
None in the core, but it makes it impossible for extensions to use additional headers in these requests.

PR #3792, opened 2015-07-26
[ticket/14051] Replaces the SQL query of the Viewforum topics sort function

adri1b:
https://tracker.phpbb.com/browse/PHPBB3-14051

### DIFF
--- a/phpBB/viewforum.php
+++ b/phpBB/viewforum.php
@@ -274,14 +274,29 @@ if ($sort_days)
 {
 	$min_post_time = time() - ($sort_days * 86400);
 
-	$sql = 'SELECT COUNT(topic_id) AS num_topics
-		FROM ' . TOPICS_TABLE . "
-		WHERE forum_id = $forum_id
-			AND (topic_last_post_time >= $min_post_time
-				OR topic_type = " . POST_ANNOUNCE . '
-				OR topic_type = ' . POST_GLOBAL . ')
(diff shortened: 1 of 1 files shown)


adri1b:
!set Event

phpBB/viewforum.php
```diff
@@ -274,14 +274,29 @@
...
+	);
+
+	/**
+	* Modify the sort data SQL query for getting additional fields if needed
+	*
+	* @event core.viewforum_modify_sort_data_sql
+	* @var	array	sql_ary			The SQL array
+	* @since 3.1.6-RC1
+	*/
+	$vars = array('sql_ary');
```
javiexin: I would recommend to include more vars in the event, so that extensions might do whatever they need, and not limited to modifying the sql_ary array (without any context information).

marc1706: I agree. You should include more variables that might be of use here.

marc1706:
Pull request will be deleted upon next release as it seems to have been abandoned.

phpBB/viewforum.php
RMcGirr83: @marc1706 I will take over but what other variables would you want to have included into the event?

RMcGirr83: @marc1706 bump

marc1706: Sort settings, start, forum ID, sort days, sort key and sort direction are some values that might be of use.

marc1706:
Replaced by #4238

PR #3806, opened 2015-08-04
[ticket/14067] Add template events to overall_header.html around feeds

rxu:
<a href="https://tracker.phpbb.com/browse/PHPBB3-14067">PHPBB3-14067</a>.


rxu:
!set Event

Nicofuma:
I don't think we need before/after.

`overall_header_feeds` should be enough

rxu:
> I don't think we need before/after.

Agreed, updated #3806.

PR #3807, opened 2015-08-05
[ticket/14068] Add template events before and after poll panel 

javiexin:
In viewtopic.

PHPBB3-14068


Nicofuma:
it's to display poll related information or just for the location?

javiexin:
Both. To fully override poll panel, by setting S_HAS_POLL to false (for that we would only need one event), but also to include a panel either before or after the poll. For example, an ad that goes before the first post would be between poll and post, might not be desirable.

Nicofuma:
ok

marc1706:
Replaced by #3852

PR #3808, opened 2015-08-05
[ticket/14070] Disabled avatar types is still displayed on the forum

Sumanai:
If you turn off of any type of avatar (or all at once), then they will still be displayed at the forum and in the profiles of users using a disconnected type of avatar.
[PHPBB3-14070](https://tracker.phpbb.com/browse/PHPBB3-14070)

### DIFF
--- a/phpBB/includes/functions.php
+++ b/phpBB/includes/functions.php
@@ -4839,6 +4839,11 @@ function phpbb_get_avatar($row, $alt, $ignore_config = false, $lazy = false)
 	$phpbb_avatar_manager = $phpbb_container->get('avatar.manager');
 	$driver = $phpbb_avatar_manager->get_driver($row['avatar_type'], $ignore_config);
 	$html = '';
+	
+	if (!($phpbb_avatar_manager->is_enabled($driver) || $ignore_config))
+	{
+		return '';
+	}
 
 	if ($driver)
 	{


phpBB/includes/functions.php
```diff
@@ -4839,6 +4839,11 @@ function phpbb_get_avatar($row, $alt, $ignore_config = false, $lazy = false)
 	$phpbb_avatar_manager = $phpbb_container->get('avatar.manager');
 	$driver = $phpbb_avatar_manager->get_driver($row['avatar_type'], $ignore_config);
 	$html = '';
+	
+	if (!($phpbb_avatar_manager->is_enabled($driver) || $ignore_config))
```
marc1706: This is wrong. ```$ignore_config``` signals that the avatar should be displayed no matter if it's been disabled.

marc1706:
The phpBB avatar manager's get_driver() method already checks if the selected driver has been disabled. This shouldn't be necessary.

phpBB/includes/functions.php
Sumanai: In phpBB 3.0 was as follows:
if (!$config['allow_avatar_upload'] && !$ignore_config)

Sumanai:
Apparently parameter $ ignore_config cancels this action. Check it out, avatars downloaded even if you disable all avatar types.

marc1706:
I can confirm the bug, Your solution does however not undo the obviously incorrect way get_driver is called but instead adds another piece of code to prevent his from happening.

marc1706:
Closing as such.

Sumanai:
An alternative way would be to add a check in each method get_data avatar drivers.

marc1706:
Replaced by #3809

Sumanai:
Thanks.

PR #3811, opened 2015-08-05
[ticket/14069] Corrected sql_fetchfield() in style_update_p1 migration

marcovo:
Changed the sql_fetchfield() call to comply to the function definition.

PHPBB3-14069


bantu:
Patch looks good. Good catch. However, depending on what this migration does, we have to ensure it is correctly applied in later updates. @nickvergessen @Nicofuma

Nicofuma:
I don't think we should do anything if this migration has already been run.

bantu:
@Nicofuma In this case, this should be good to go.

PR #3812, opened 2015-08-05
[ticket/13700] Add dependency to migration profilefield_interests

marcovo:
\phpbb\db\migration\data\v310\profilefield_interests should
be dependent on \phpbb\db\migration\data\v310\dev , otherwise
the database column field_show_on_pm is not created in time. This
dependency is added in this commit.

PHPBB3-13700


Elsensee:
It already depends on `\phpbb\db\migration\data\v310\dev` by depending on:
- `\phpbb\db\migration\data\v310\profilefield_types`
  - `\phpbb\db\migration\data\v310\alpha2`
    - `\phpbb\db\migration\data\v310\alpha1`
      - `\phpbb\db\migration\data\v310\acp_style_components_module`
        - `\phpbb\db\migration\data\v310\dev`

VSEphpbb:
This is not needed

PR #3813, opened 2015-08-05
[ticket/14072] Add core event to the function format_display()

rxu:
<a href="https://tracker.phpbb.com/browse/PHPBB3-14072">PHPBB3-14072</a>.


rxu:
!set Event

PR #3816, opened 2015-08-07
[ticket/14074] Fix mark notifications as read in the ucp

Nicofuma:
https://tracker.phpbb.com/browse/PHPBB3-14074

PHPBB3-14074


Nicofuma:
Note: hot fix already applied on area51 (and reverted but still present in opcache)

PR #3818, opened 2015-08-08
[ticket/14077] Fall back to 1 as focusOffset if length is unsupported

marc1706:
The setBaseAndExtent() in Microsoft's Edge browser is incompatible with
the one Webkit browsers use. As a result of that, we have to fall back to
setting the focusOffset to 1 instead of the text length.

Ticket: https://tracker.phpbb.com/browse/PHPBB3-14077

### DIFF
--- a/phpBB/styles/prosilver/template/forum_fn.js
+++ b/phpBB/styles/prosilver/template/forum_fn.js
@@ -156,7 +156,11 @@ function selectCode(a) {
 		// Safari and Chrome
 		if (s.setBaseAndExtent) {
 			var l = (e.innerText.length > 1) ? e.innerText.length - 1 : 1;
-			s.setBaseAndExtent(e, 0, e, l);
+			try {
+				s.setBaseAndExtent(e, 0, e, l);
+			} catch (error) {
+				s.setBaseAndExtent(e, 0, e, 1);
+			}
 		}
 		// Firefox and Opera
 		else {


bantu:
> The code block is not a proper node in Microsoft's Edge browser

Now why is that?

marc1706:
Turns out it's a proper node but Microsoft's Edge browser just fails miserably at setBaseAndExtent().

callumacrae:
lol can we just remove the IE8 specific code?

marc1706:
@callumacrae We could but that wouldn't fix the issue. It's trying to use the safari and chrome approach with Microsoft's Edge.

callumacrae:
If you're thinking of an approach that only works in certain browsers, the approach is probably wrong. Edge is closer to Blink browsers than it is to IE8.

marc1706:
@callumacrae Apparently Edge pretends to be able to use the setBaseAndExtend Method but fails at using it. I'd add a try {} catch () {} block and revert to the legacy part in the catch part.

callumacrae:
why not see how TinyMCE did it? http://www.tinymce.com/develop/changelog/?ctrl=version&act=view&pr_id=1&vr_id=884

marc1706:
Well their change doesn't seem to be the best approach for us:
https://github.com/tinymce/tinymce/commit/375d2e0c4225aa27a3936bd7dc95418dca4acee2

marc1706:
Updated

callumacrae:
This will require an update when future versions of Edge add support for this. Just use a try-catch block. Browser detection is never the way!

marc1706:
Fair enough. Will go for that then.

marc1706:
@callumacrae Updated with a try/catch block and a way of using setBaseAndExtent on Edge, too.

phpBB/styles/prosilver/template/forum_fn.js
```diff
@@ -156,7 +156,11 @@ function selectCode(a) {
 		// Safari and Chrome
 		if (s.setBaseAndExtent) {
 			var l = (e.innerText.length > 1) ? e.innerText.length - 1 : 1;
-			s.setBaseAndExtent(e, 0, e, l);
+			try {
+				s.setBaseAndExtent(e, 0, e, l);
+			} catch (nil) {
```
VSEphpbb: `nil` ?

marc1706: It's not needed but there must be one variable there. If you have a better idea, feel free to share.

VSEphpbb: How about `e`

VSEphpbb: Oo wait, your using an 'e' already... 'error' ?

marc1706: Do you really dislike nil that much? :D

Elsensee: What about `ex`? :P

callumacrae: Yeah I wasn't a huge fan of `nil` either, but didn't want to nitpick. It's misleading.

marc1706:
updated

callumacrae:
lgtm :+1:

PR #3826, opened 2015-08-12
[ticket/14094] Fix segmentation fault error on PHP 7 tests

rxu:
<a href="https://tracker.phpbb.com/browse/PHPBB3-14094">PHPBB3-14094</a>.


Nicofuma:
:+1: should be fixed upstream in PHP, but I think we can merge it for now in order to work on the php7 compat

PR #3831, opened 2015-08-15
[ticket/14101] Add core event to the download/file.php

rxu:
<a href="https://tracker.phpbb.com/browse/PHPBB3-14101">PHPBB3-14101</a>.


rxu:
!set Event

PR #3832, opened 2015-08-15
[ticket/14103] Reset text-decoration in abbr element

cyberalien:
https://tracker.phpbb.com/browse/PHPBB3-14103

### DIFF
--- a/phpBB/adm/style/admin.css
+++ b/phpBB/adm/style/admin.css
@@ -17,6 +17,10 @@
 	font-size: 100%;
 }
 
+abbr {
+	text-decoration: none;
+}
+
 body, div, p, th, td, li, dd {
 	font-size: x-small;
 	voice-family: "\"}\"";
--- a/phpBB/styles/prosilver/theme/common.css
+++ b/phpBB/styles/prosilver/theme/common.css
@@ -43,6 +43,9 @@ table {
 	border-collapse: collapse;
 	border-spacing: 0;
 }
+abbr {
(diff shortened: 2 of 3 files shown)


lavigor:
This PR was intended to fix an issue in prosilver.
But it also introduces inconsistency across browsers in subsilver2 and ACP style.
Was this solution tested on a board with subsilver2?
There're no other style rules for `<abbr>` in that style. It means that those elements were previously underlined in most browsers on boards where that style is enabled.

Moreover, [useful debug information](https://github.com/phpbb/phpbb/blob/668defcf5705699be3a15548c08a3ea6c79086af/phpBB/includes/functions.php#L5318) about SQL time and PHP time is also affected by current solution.

@marc1706, could you confirm the reliability of the fixes for subsilver2 and ACP style (that are not related to the ticket of this PR) please?

lavigor:
A week is almost over, but no replies from the team.
Can I assume that the edits were merged without any testing?

marc1706:
```<abbr>``` is not used anywhere else apart from the debug info in the footer. The patch fixes the inconsistency introduced with the latest update to Firefox and makes sure that this looks the same in all browsers.
This solution was properly reviewed and tested. Most browsers until now didn't underline ```<abbr>```. We do not expect this to be underlined in the footer so it should stay this way.

Please stop implying that these fixes were improper or that they we're not reviewed correctly.
Feel free to create a ticket that suggests this for our next feature release 3.2.

PR #3833, opened 2015-08-15
[ticket/14102] Add core event to the mcp_topic.php

rxu:
<a href="https://tracker.phpbb.com/browse/PHPBB3-14102">PHPBB3-14102</a>.


rxu:
!set Event

PR #3837, opened 2015-08-16
[ticket/14107] Fix dropdown rtl states

hanakin:
[ticket/14107] Fix dropdown rtl states

https://tracker.phpbb.com/browse/PHPBB3-14107
PHPBB3-14107

### DIFF
--- a/phpBB/styles/prosilver/theme/bidi.css
+++ b/phpBB/styles/prosilver/theme/bidi.css
@@ -116,7 +116,11 @@
 
 .rtl .dropdown-extended .header .header_settings, .rtl .dropdown-container-right {
 	float: left;
-}rtl .notification_list ul li img {
+}
+
+/* Notifications
+-----------------------------------------*/
+rtl .notification_list ul li img {
 	float: right;
 	margin-left: 5px;
 	margin-right: 0;
--- a/phpBB/styles/prosilver/theme/bidi.css
+++ b/phpBB/styles/prosilver/theme/bidi.css
@@ -88,7 +88,7 @@
 
 /* Dropdown menu
 ---------------------------------------- */
-.rtl .dropdown-container.topic-tools {
+.rtl .dropdown-container.topic-tools, .rtl .dropdown-container-left {
 	float: right;
 }
 
@@ -114,13 +114,9 @@
 	text-align: right;
 }
 
-.rtl .dropdown-extended .header .header_settings {
+.rtl .dropdown-extended .header .header_settings, .rtl .dropdown-container-right {
 	float: left;
-}
-
-/* Notifications
-----------------------------------------*/
-.rtl .notification_list ul li img {
+}rtl .notification_list ul li img {
 	float: right;
 	margin-left: 5px;
 	margin-right: 0;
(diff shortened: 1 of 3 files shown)


hanakin:
!set Styles

phpBB/styles/prosilver/theme/bidi.css
```diff
@@ -114,13 +114,9 @@
...
 
-.rtl .dropdown-extended .header .header_settings {
+.rtl .dropdown-extended .header .header_settings, .rtl .dropdown-container-right {
 	float: left;
-}
-
-/* Notifications
-----------------------------------------*/
-.rtl .notification_list ul li img {
+}rtl .notification_list ul li img {
```
Nicofuma: You removed the comment?

phpBB/styles/prosilver/theme/bidi.css
```diff
@@ -114,13 +114,13 @@
...
-.rtl .dropdown-extended .header .header_settings {
+.rtl .dropdown-extended .header .header_settings, .rtl .dropdown-container-right {
 	float: left;
 }
 
 /* Notifications
-----------------------------------------*/
-.rtl .notification_list ul li img {
+-----------------------------------------*/
+rtl .notification_list ul li img {
```
marc1706: Removing the dot in front of rtl doesn't seem to make any sense to me? Typo?

hanakin:
This is ready for review @cyberalien

cyberalien:
In topic at bottom of page tools dropdown is on left side with RTL language. It should be on right side.

cyberalien:
Jumpbox is also on wrong side

hanakin:
this is what topic looks like for me with rtl
screenshot 2015-08-27 05 56 12

is this not what you are seeing?

cyberalien:
No, this is what I'm seeing:

rtl

Its a fresh installation directly from your branch. 

In LTR language jumpbox is wrapped in dropdown-container-right that has float:right.
In RTL language jumpbox is wrapped in dropdown-container-left that normally has float:left, but in bidi.css it is overwritten with float:right

So in both directions it floats to right. Same with quickmod dropdown.

hanakin:
have to look into this more when i get home tonight as that makes no sense.

dropdown-container-right = float:right
dropdown-container-left = float: left

in bidi they are switched so that they float to the opposite side I think you may have cache issues

cyberalien:
No, there are no cache issues. It was fresh installation directly from your repo in new directory, there is nothing to cache.

cyberalien:
From jumpbox.html:

    <div class="dropdown-container dropdown-container-{S_CONTENT_FLOW_END}

From bidi.css:

    .rtl .dropdown-container.topic-tools, .rtl .dropdown-container-left {
    	float: right;
    }


First it changes direction in HTML code by changing class name, so float:right becomes float:left. Then it again changes direction in bidi.css, which reverts it back to float:right.

hanakin:
ah thats why I was not aware the it was changing it in the html in this way. this seems counter intuitive

hanakin:
@cyberalien can please have another look at this

marc1706:
Looks good now.

cyberalien:
.jumpbox .dropdown-contents a

has margin on wrong side. fix:

    .rtl .jumpbox .dropdown-contents a { margin-right: 0; margin-left: 20px; }

everything else looks good

PR #3838, opened 2015-08-16
[ticket/14109] InnoDB requires multiple queries when setting FULLTEXT

brunoais:
> InnoDB presently supports one FULLTEXT index creation at a time [1795]

https://tracker.phpbb.com/browse/PHPBB3-14109

### DIFF
--- a/phpBB/phpbb/search/fulltext_mysql.php
+++ b/phpBB/phpbb/search/fulltext_mysql.php
@@ -889,7 +889,10 @@ class fulltext_mysql extends \phpbb\search\base
 				$alter[] = 'MODIFY post_subject text NOT NULL';
 			}
 			$alter[] = 'ADD FULLTEXT (post_subject)';
+			$this->db->sql_query('ALTER TABLE ' . POSTS_TABLE . ' ' . implode(', ', $alter));
 		}
+
+		$alter = array();
 
 		if (!isset($this->stats['post_content']))
 		{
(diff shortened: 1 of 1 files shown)


bantu:
There is quite some repetition already and even more post-patch.

brunoais:
What would be preferable, then?

dhruvgoel92:
Closing this PR as its a duplicate of https://github.com/phpbb/phpbb/pull/3839

Nicofuma:
that's not a duplicate actually, the other PR is just based on this one (there is 2 different tickets)

marc1706:
@Nicofuma I guess this should be reopened then?

Nicofuma:
I think yes (forgot to reopen)

marc1706:
Isn't this obsolete after the bug was fixed in MySQL itself?

brunoais:
@marc1706  if you want my opinion, now I'd just bump the minimum MySQL version on phpBB and call it a day.

bantu:
The error in https://bugs.mysql.com/bug.php?id=67004 is triggered on the ALTER statement. This patch thus only partially addresses the issue. It is my understanding that any table changes would have to drop the FULLTEXT indexes before adding e.g. columns. This is rather ridiculous and would require unnecessary additional changes. I am very much in favour of just increasing the required version.

bantu:
Replaced with #4426

PR #3839, opened 2015-08-16
[ticket/14108] MySQL search backend add post_text index

brunoais:
When creating the indexes for MySQL, the index for just post_text is not done.
Such index is required for topic searches when searching on the message only (without the title)

https://tracker.phpbb.com/browse/PHPBB3-14108

### DIFF
--- a/phpBB/phpbb/search/fulltext_mysql.php
+++ b/phpBB/phpbb/search/fulltext_mysql.php
@@ -889,7 +889,10 @@ class fulltext_mysql extends \phpbb\search\base
 				$alter[] = 'MODIFY post_subject text NOT NULL';
 			}
 			$alter[] = 'ADD FULLTEXT (post_subject)';
+			$this->db->sql_query('ALTER TABLE ' . POSTS_TABLE . ' ' . implode(', ', $alter));
 		}
+
+		$alter = array();
 
 		if (!isset($this->stats['post_content']))
 		{
@@ -903,10 +906,6 @@ class fulltext_mysql extends \phpbb\search\base
 			}
 
 			$alter[] = 'ADD FULLTEXT post_content (post_text, post_subject)';
-		}
-
-		if (sizeof($alter))
-		{
 			$this->db->sql_query('ALTER TABLE ' . POSTS_TABLE . ' ' . implode(', ', $alter));
 		}
 
--- a/phpBB/phpbb/search/fulltext_mysql.php
+++ b/phpBB/phpbb/search/fulltext_mysql.php
@@ -944,6 +944,7 @@ class fulltext_mysql extends \phpbb\search\base
 		if (isset($this->stats['post_content']))
 		{
 			$alter[] = 'DROP INDEX post_content';
+			$alter[] = 'DROP INDEX post_text';
 		}
 
 		if (sizeof($alter))
--- a/phpBB/phpbb/search/fulltext_mysql.php
+++ b/phpBB/phpbb/search/fulltext_mysql.php
@@ -907,6 +907,8 @@ class fulltext_mysql extends \phpbb\search\base
 
 			$alter[] = 'ADD FULLTEXT post_content (post_text, post_subject)';
 			$this->db->sql_query('ALTER TABLE ' . POSTS_TABLE . ' ' . implode(', ', $alter));
+
+			$this->db->sql_query('ALTER TABLE ' . POSTS_TABLE . ' ADD FULLTEXT post_text (post_text)');
 		}
 
 		$this->db->sql_query('TRUNCATE TABLE ' . SEARCH_RESULTS_TABLE);


brunoais:
Depends on #3838

brunoais:
Git really screwed up hard this time...

bantu:
The post_content (post_text, post_subject) index can be used for searches on post_text only.

bantu:
See https://tracker.phpbb.com/browse/PHPBB3-11621

brunoais:
@bantu 
> The post_content (post_text, post_subject) index can be used for searches on post_text only.

Tell that to MySQL

> Can't find FULLTEXT index matching the column list [1191]

	SELECT SQL_CALC_FOUND_ROWS p.post_id FROM phpbb_topics t, phpbb_posts p WHERE MATCH (p.post_text) AGAINST ('+limited* ' IN BOOLEAN MODE)

Generated by url, for example:

	http://localhost/phpBB/search.php?keywords=limited*&t=4&sf=msgonly

bantu:
What the fuck.

bantu:
@brunoais Please start by adding a test case for reproducing this. There are other search tests already.

brunoais:
@bantu I need your help, then.
I need to know, using phpBB's test API, how to **make sure the DB is created with MySQL tables using the InnoDB engine.**
I've looked for it and there's no tests for index creation. I've read all the code for the tests and I can't find any DB connections being made in the whole search backend tests.

VSEphpbb:
Unit tests are done using SQLite. Functional tests are done using MySQL with InnoDB. Except the PHP 5.3 environment, which is MyISAM https://github.com/phpbb/phpbb/blob/3.1.x/.travis.yml#L8

bantu:
@brunoais Is this an InnoDB-only issue? I am pretty sure @dhruvgoel92 tested this, at least on MyISAM.

brunoais:
@VSEphpbb Where is it set to MyISAM/InnoDB?
@bantu It is an InnoDB issue.

bantu:
@brunoais I believe MySQL on Travis CI is InnoDB by default. Here we change to MyISAM specifically. https://github.com/phpbb/phpbb/blob/3.1.x/travis/setup-database.sh#L25

It indeed seems that we do not test MySQL Fulltext Index using InnoDB on Travis CI, which would explain why we missed this.

bantu:
If this is an InnoDB-only issue, I would like to avoid re-adding the additional post_text index, because it is rather expensive storage-wise (as per https://tracker.phpbb.com/browse/PHPBB3-11621). This means this patch can not be accepted as is and a better solution needs to be found.

bantu:
As for testing MySQL Fulltext on InnoDB: We have a mariadb environment, which should already be using InnoDB. Question is whether we test for it and MariaDB passes, which means this would be a MySQL only issue or whether the MariaDB version is too old to support Fulltext Indexes.

bantu:
The MariaDB environment says
```
17) phpbb_functional_search_mysql_test::test_search_backend
Search backend is not supported/running
```
https://travis-ci.org/phpbb/phpbb/jobs/75981209#L1153

bantu:
@brunoais Feel free to confirm that phpbb_functional_search_mysql_test already contains a test case for testing for the issue you're having (message only search).

bantu:
Maybe we should have "MySQL MyISAM Fulltext Search Index" and "MySQL InnoDB Fulltext Search Index".

brunoais:
@bantu 
> They can work even without a FULLTEXT index, although a search executed in this fashion would be quite slow.

source: https://dev.mysql.com/doc/refman/5.5/en/fulltext-boolean.html

Question:
Is the current search with that system in MyISAM tables using the index?
(time to find out)

brunoais:
My current answer:
No:
id 	select_type 	table 	type 	possible_keys 	key 	key_len 	ref 	rows 	Extra
1 	SIMPLE 	p 	ALL 	NULL	NULL	NULL	NULL	13 	Using where

The query made is as follows

	EXPLAIN SELECT  p.post_id FROM phpbb_topics t, phpbb_posts p WHERE MATCH (p.post_text) AGAINST ('+limited* ' IN BOOLEAN MODE)

I've confirmed the indexes available and everything else on MySQL 5.6.25. They were the same as for InnoDB.

**In other words**, It equally does not work with the set indexes in MyISAM. The difference is that it can work without indexes in MyISAM, whereas in InnoDB it can't do the search process without using the indexes

brunoais:
@bantu Do you still need those tests?

marc1706:
Bump @bantu 
FWIW, I think we should have tests for this.

dhruvgoel92:
We have test for this, the reason why they dont run on travis is travis is using older mysql version. Fulltext for innodb is present since mysql 5.6.4 which is why mysql full text search shows as not supported in functional tests. BTW I ran the tests on 5.6.22 and manually tested this use case, was not able to reproduce the issue.

brunoais:
This has been fixed in MySQL. I thought I had made a comment in the ticket about it.
I think there's no use spending more time in this. I couldn't find any of web servers with the version intervals that contain this bug.
Close this and the bug?

PR #2601, opened 2014-06-15
[ticket/12719] Convert to Normalize & SUITCSS - base for reset

hanakin:
[ticket/12719] Convert to Normalize & SUITCSS - base for reset

Todo list:

- [x] Chrome [Mac]
- [x] Firefox [Mac]
- [x] Safari [Mac]
- [x] IE9 [Win7]
- [x] IE10 [Win7]
- [x] IE11 [Win7]
- [ ] EDGE [Win10]

https://tracker.phpbb.com/browse/PHPBB3-12719

PHPBB3-12719

PR #3840, opened 2015-08-16
[ticket/12719] Convert to Normalize & SUITCSS

hanakin:
[ticket/12719] Convert to Normalize & SUITCSS - base for reset

Todo list:

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] IE9 [Win7]
- [x] IE10 [Win7]
- [x] IE11 [Win7]
- [x] Edge [Win10]

Old PR: https://github.com/phpbb/phpbb/pull/2601
Tracker: https://tracker.phpbb.com/browse/PHPBB3-12719

PHPBB3-12719


hanakin:
!set Styles

Nicofuma:
Could it be tested Microsoft edge too please?

phpBB/styles/prosilver/theme/normalize.css
```diff
@@ -0,0 +1,331 @@
...
+ */
+a:active,
+a:hover { outline: 0 }
+
+/* Text-level semantics
+   ========================================================================== */
+/**
+ * Address styling not present in IE 8/9/10/11, Safari, and Chrome.
+ */
+abbr[title] { border-bottom: 1px dotted }
```
marc1706: Also see #3832 for that

hanakin:
I am still trying to work out some issues in my ability to test on windows at the moment i do not have a clean solution. I managed to hack a test on 9-11 but I do not have access to edge

phpBB/styles/prosilver/theme/normalize.css
hanakin: marc igor was actually correct and the abbr underline in the spec is not stylistic it is for usability/accessibility. The fact that we never used it before is the error not the fact that firefox changed it. That is why it is part of normalize and has been since the beginning.

cyberalien: While it is recommended for usability/accessibility, I think it doesn't belong in prosilver.

Currently abbr tag is used only in footer (also one line in doc and debug output, but those abbreviations aren't something that users ever see, so they are irrelevant). Underline looks out of place there.

There are many other abbreviations in styles, but none of them use abbr tag so they are not affected: MCP, FAQ, PM, BBCode - those are used all over the forum. Unlike time zone, users see those abbreviations all the time. If abbreviations are changed, it should be done properly - all those abbreviations should also be changed to abbr tag.

Changing it only for one abbreviation that also appears to be the most irrelevant one is not the way to go.

As for reverting patch. Patch fixes an issue. Issue is abbreviations used to be displayed in prosilver without underline for ages and now suddenly it behaves differently in latest version of one browser. Patch fixes browser inconsistency, forcing Firefox to display abbreviations in same way as other browser do and old Firefix did. If you want to change layout of abbreviations, that's a different issue. That's what Igor fails to understand - that patch not about having or not having underline, its about fixing inconsistent layout caused by change in one browser. Adding underline is a different issue that needs to be discussed separately from that ticket (this ticket seems like a good place for that).

hanakin: patch should never have been applied to 3.2 I can see your point for 3.1 based on it not being displayed that way before. With 3.2 we are trying to bring prosilver up to industry standards which is utilizing normalize and fix several inconsistencies with it based on those standards such as the fact that it has not been displaying abbr in this way.

cyberalien: That's a good point. Maybe find all abbreviations in language, add abbr tag where necessary and change style for 3.2?

hanakin: sounds good but obviously outside the scope of this PR

lavigor: @hanakin, I absolutely agree with you. I don't know what's wrong with @cyberalien today, but he clutched to his solution with both hands.
I don't think that we need to specify titles for every abbreviation in the code, but I think that the right way to save the 'historical look' of the time zone in the footer is to use the `<span>` element there.
@hanakin, what do you think about the suggestion not to affect every abbreviation just because of one historic occurrence in the footer?

As for that PR, it introduces inconsistency in styles other than prosilver (and those 'fixes' are not related to PR's ticket).

marc1706: @lavigor Please stay on topic. I have different things to do than trying to keep everyone happy.

lavigor: @mark1706, is there any topic where I can complain on the situation?
It's a pity that the comment
> @hanakin, I absolutely agree with you.

is considered 'an offensive part' while the insult of being an idiot from a 13 years long phpBB member is OK.

Nicofuma: First, the PR is definitely not the place to discuss about all of that, so all please STOP, create a dedicated ticket and discuss about it on A51.
Second, your last is offensive! you can't say publicly that someone you don't know is an idiot as you can't say he is 13 years old! And for your other message, it was less offensive but you can't say it wasn't either. And the removed was off topic anyway.

lavigor: @Nicofuma, 13 years long member means a member registered on the board for 13 years (since 2002), not a human of 13 years old.
Compare `long` to `old`.

It's very frustrating that those words were misunderstood. I'm not the person that could say anything like that.
I meant that Arty is experienced enough if he is with phpBB for 13 years.

Nicofuma: well, that's right.

hanakin:
all moving forward as @Nicofuma has stated the rest should be discussed elsewhere. I find none it offensive but he is right its not best discussed here this PR is about using normalize...

phpBB/styles/prosilver/theme/normalize.css
marc1706: @lavigor feel free to send me a PM on .com. This is definitely not the place to discuss these things.

hanakin:
Finally managed to fix my router settings to get my testing suite backup for IE. Tested and working on all IE 9 - Edge, Safari, Firefox, Chrome & IOS

Nicofuma:
nice, but don't forget to rebase please ^^

Nicofuma:
Finally, @hanakin it is ready?
@cyberalien thoughts?
@phpbb/development-team any css guy here? :/

cyberalien:
@Nicofuma patch looks good. I think it should be applied to ACP style as well, probably in separate PR.

hanakin:
@Nicofuma yes should probably be a tag for that like we use for website but its ready. @cyberalien agree admin needs some love but its secondary and separate.

hanakin:
nevermind spoke too soon looks like the old reset got add back in when i rebased somehow

Nicofuma:
Just leave a comment when it is ready for merge please

hanakin:
ready to merge

Nicofuma:
now I have two underlines for abbr elements, it is intended?

hanakin:
@Nicofuma abbr should have ........ under them as per normalize. Weather or not our usaged of the abbr elemement in the html is correct is another story and should be its own PR

cyberalien:
I've tested latest version with Firefox, it does show double border for abbr tag.

t12719

hanakin:
on pc cause on mac there is only 1

cyberalien:
No, that screenshot is made on Mac. Firefox 40.0.2

lavigor:
@hanakin, `text-decoration` should be removed for `<abbr>` for compatibility with modern browsers (currently only Firefox 40).
See my PR #3841.

hanakin:
yeah there is a newer version of normalize i need to update it

hanakin:
seems firefox 40 is newer normalize has not caught up yet

cyberalien:
That fixed it. Looks good in FF now.

hanakin:
should be good to merge now

Nicofuma:
thank you

cyberalien:
I've noticed one possible side effect of normalize patch: line-height:normal for body element makes layout different in IE9. I don't think it looked like that before, though I'm not sure.

If that rule is removed, layout looks the same in IE9 as in other browsers.

PR #3842, opened 2015-08-17
[ticket/14097] Improve error and failure handling in the installer

CHItA:
[PHPBB3-14097](https://tracker.phpbb.com/browse/PHPBB3-14097)

### DIFF
--- a/phpBB/phpbb/install/installer.php
+++ b/phpBB/phpbb/install/installer.php
@@ -92,7 +92,8 @@ class installer
 		$module_found = false;
 
 		// Variable used to check if the install process have been finished
-		$install_finished = false;
+		$install_finished	= false;
+		$fail_cleanup		= false;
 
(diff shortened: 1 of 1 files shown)


Nicofuma:
That's a first good step :)
A second step would be to send the full error, including the stacktrace. The latter being displayed only in the logs/console. (We could use https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/Debug/Exception/FlattenException.php to serialize the exception)

phpBB/phpbb/install/installer.php
```diff
@@ -209,13 +210,20 @@ public function run()
 		{
 			// Do nothing
 		}
+		catch (\Exception $e)
+		{
+			// Most likely there were a PHP failure, so let's die like a gentleman
```
VSEphpbb: this comment probably isn't needed, but it also has many English problems too.

PR #3843, opened 2015-08-18
add core event on index for board statistics

svennd:
[PHPBB3-14112] add event for board statistics

Event added so we can access the board
statistics from an extension;

PHPBB3-14112

### DIFF
--- a/phpBB/index.php
+++ b/phpBB/index.php
@@ -161,8 +161,7 @@ if ($config['load_birthdays'] && $config['allow_birthdays'] && $auth->acl_gets('
 	$db->sql_freeresult($result);
 }
 
-// Assign index specific vars
-$template->assign_vars(array(
+$index_vars = array(
 	'TOTAL_POSTS'	=> $user->lang('TOTAL_POSTS_COUNT', (int) $config['num_posts']),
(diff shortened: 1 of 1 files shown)


rxu:
You don't need additional event to modify template variables from ```index.php```. You can just use existing event ```'core.index_modify_page_title'``` to simply reassign them via ```assign_vars()``` method.

svennd:
The name of that event is rather restricting and I think renaming events is not allowed.

rxu:
Well, that is just a name. It doesn't affect the way you can use it or restrict anything. So renaming is also would be senseless.

svennd:
Yeah thank, I know, but it would be crazy to add statistics variables to an event called index_modify_page_title. Either way this build failed and I am clueless, I'm gone leave the pro's to fix or to forget about it :)

rxu:
> it would be crazy to add statistics variables to an event called index_modify_page_title

Not at all :)

Nicofuma:
The failure is about your commit message: https://wiki.phpbb.com/Git#Commit_Messages

About the events they are extension points and their name is just indicative, to have a way to find them easily. (they could be name event-1, event-2 etc... but then it will be very to talk about them)

svennd:
Oke Thx @rxu and @Nicofuma I will cancel this and retry later :)

PR #3845, opened 2015-08-19
[ticket/14112] Add variable to index event for board statistics

svennd:
Variable added so we can access the board
statistics from an extension;

PHPBB3-14112

### DIFF
--- a/phpBB/index.php
+++ b/phpBB/index.php
@@ -161,8 +161,7 @@ if ($config['load_birthdays'] && $config['allow_birthdays'] && $auth->acl_gets('
 	$db->sql_freeresult($result);
 }
 
-// Assign index specific vars
-$template->assign_vars(array(
+$board_statistics = array(
 	'TOTAL_POSTS'	=> $user->lang('TOTAL_POSTS_COUNT', (int) $config['num_posts']),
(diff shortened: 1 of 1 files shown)


phpBB/index.php
```diff
@@ -192,11 +191,15 @@
 *
 * @event core.index_modify_page_title
 * @var	string	page_title		Title of the index page
+* @var	array	board_statistics		Array with the board statistics, legend and birthday list
 * @since 3.1.0-a1
```
marc1706: Missing ```@changed```

phpBB/index.php
```diff
@@ -161,8 +161,7 @@
 	$db->sql_freeresult($result);
 }
 
-// Assign index specific vars
-$template->assign_vars(array(
+$board_statistics = array(
```
marc1706: This is not just the board statistics but rather most of the index template data apart from the forums. Maybe just use a variable name like $template_data

marc1706:
@svennd Are you going to take care of the mentioned issues? If not, please give us a heads up so we can take care of it.

svennd:
@marc1706 I'm sorry I thought you could edit it as a patch, I am unfamilar with git, I'm willing to look into it tomorrow, but if its only a click on some button for you feel free.

svennd:
I don't understand why it failed ?

marc1706:
I'll take care of it ;)

PR #3847, opened 2015-08-19
[ticket/14105] Remove sid= from local URLs

JoshyPHP:
PHPBB3-14105

### DIFF
--- a/phpBB/phpbb/textformatter/s9e/parser.php
+++ b/phpBB/phpbb/textformatter/s9e/parser.php
@@ -393,4 +393,24 @@ class parser implements \phpbb\textformatter\parser_interface
 
 		return $url;
 	}
+
+	/**
+	* Filter a valid URL
+	*
+	* @see message_parser::validate_url()
+	*
+	* @param  string $url Original URL
+	* @return string      Same URL minus any sid= parameter
+	*/
+	static public function filter_url($url)
+	{
+		if (strpos($url, generate_board_url()) !== false && strpos($url, 'sid=') !== false)
+		{
+			$url = preg_replace('/([&?])sid=[0-9a-f]{32}&/', '$1', $url);
+			$url = preg_replace('/[&?]sid=[0-9a-f]{32}$/', '', $url);
+			$url = append_sid($url);
+		}
+
+		return $url;
+	}
 }
--- a/phpBB/phpbb/textformatter/s9e/parser.php
+++ b/phpBB/phpbb/textformatter/s9e/parser.php
@@ -404,11 +404,10 @@ class parser implements \phpbb\textformatter\parser_interface
 	*/
 	static public function filter_url($url)
 	{
-		if (strpos($url, generate_board_url()) !== false && strpos($url, 'sid=') !== false)
+		if (strpos($url, 'sid=') !== false)
 		{
 			$url = preg_replace('/([&?])sid=[0-9a-f]{32}&/', '$1', $url);
 			$url = preg_replace('/[&?]sid=[0-9a-f]{32}$/', '', $url);
-			$url = append_sid($url);
 		}
 
 		return $url;
--- a/tests/text_processing/tickets_data/PHPBB3-14105.html
+++ b/tests/text_processing/tickets_data/PHPBB3-14105.html
@@ -1,0 +1,2 @@
+<a href="http://localhost/index.php" class="postlink">...</a><br>
+<a href="http://localhost/index.php?foo=1" class="postlink">...</a>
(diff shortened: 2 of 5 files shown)


phpBB/phpbb/textformatter/s9e/parser.php
```diff
@@ -393,4 +393,24 @@ static public function filter_img_url($url, array $url_config, Logger $logger, $
...
+	* Filter a valid URL
+	*
+	* @see message_parser::validate_url()
+	*
+	* @param  string $url Original URL
+	* @return string      Same URL minus any sid= parameter
+	*/
+	static public function filter_url($url)
+	{
+		if (strpos($url, generate_board_url()) !== false && strpos($url, 'sid=') !== false)
```
s9e: I think this block sucks and should be replaced with this:

```php
if (strpos($url, 'sid=') !== false)
{
	$url = preg_replace('/([&?])sid=[0-9a-f]{32}&/', '$1', $url);
	$url = preg_replace('/[&?]sid=[0-9a-f]{32}$/', '', $url);
}
```

rxu: Then this will affect all URLs hgaving ```sid``` but not just local ones?

tests/text_processing/tickets_data/PHPBB3-14105.html
```diff
@@ -0,0 +1,2 @@
+<a href="http://localhost/index.php" class="postlink">...</a><br>
+<a href="http://localhost/index.php?foo=1" class="postlink">...</a>
```
rxu: Newline :P

s9e: Doing that means adding a `<br>`. Is there a reason for adding one?

bantu: No, just adding a newline to the HTML file.

marc1706:
Do I see this correctly that this currently filters out any given URL with a sid parameter in it?
I think it should only filter the sid parameter for a URL of the given board.

Nicofuma:
Actually, I said the exact opposite on A51. I'm failing to see any use case where it is a good idea to keep the sid.

marc1706:
The thing is that the sid parameter doesn't have to stand for session ID and one might actually want to include a session ID for other sites (for whatever reason that is). My point is that we shouldn't mess with URLs just because it's nice to have for phpBB URLs.

bantu:
> Actually, I said the exact opposite on A51. I'm failing to see any use case where it is a good idea to keep the sid.

Your assumption of `sid` always standing for session id is simply wrong and arbitrary. `sid` might simply be an integer for example standing for `store id` or whatever `s... id`. Removing arbitrary parameters from posted URLs is an absolute no-go.

tests/text_processing/tickets_data/PHPBB3-14105.html
s9e: Which is impossible, since adding a new line in a post results in a newline in the HTML.

bantu: I agree. This is a fixture, so it's okay.

Nicofuma:
To sum-up:
We want to remove the sid only from the local urls, that's right? And do we want to readd the sid on display? (I don't think so)

s9e:
As far as I'm concerned, the sid shouldn't be removed from any URLs. If people want to remove them for SEO purposes, they can do it in an extension.

Checking whether a URL is a local URL adds a hidden dependency to the parser because `generate_board_url()` uses the `request` service. I don't know what the implications are, especially in CLI.

marc1706:
In that case I'd suggest to not remove any SID from URLs.

Nicofuma:
I kinda agree on this point. Anyway it always looked weird to me to remove the sid in the url.

If someone have a different opinion, please talk about that again on area51 http://area51.phpbb.com/phpBB/viewtopic.php?p=282611

Nicofuma:
I kinda agree on this point. Anyway it always looked weird to me to remove the sid in the url.

If someone have a different opinion, please talk about that again on area51 http://area51.phpbb.com/phpBB/viewtopic.php?p=282611

dragosvr92:
I only just realised that this was reconsidered... And i am still wondering... Why? 
We wasnt going to remove them for SEO purposes. We were going to remove them because they are useless and it makes no sense having it there? 

All of you agreed with this from the beginning.

Elsensee:
You can read the comments (especially from @bantu and @JoshyPHP) to see why. I think there is no further discussion necessary.

PR #3850, opened 2015-08-21
[ticket/14117] Add core events to index.php to allow modifying birthdays list

rxu:
<a href="https://tracker.phpbb.com/browse/PHPBB3-14117">PHPBB3-14117</a>.

### DIFF
--- a/phpBB/index.php
+++ b/phpBB/index.php
@@ -119,7 +119,7 @@ $db->sql_freeresult($result);
 $legend = implode($user->lang['COMMA_SEPARATOR'], $legend);
 
 // Generate birthday list if required ...
-$birthday_list = array();
+$birthdays = $birthday_list = array();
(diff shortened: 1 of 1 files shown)


rxu:
!set Event

phpBB/index.php
```diff
@@ -132,33 +132,67 @@
...
+		$birthday_list[] = $birthday_username . (($birthday_age) ? " ({$birthday_age})" : '');
 	}
-	$db->sql_freeresult($result);
+
+	/**
+	* Event to modify the birthdays list
+	*
+	* @event core.index_modify_birthdays_list
+	* @var	array	birthdays		Array with the users birhtdays data
+	* @var	array	birthday_list	Array with the birthdays list items
```
Nicofuma: It is only used for 3.0 compatibility. I don't think we should add it to the event (or at least repeat the deprecated comment here)

rxu:
@Nicofuma Updated #3850.

Nicofuma:
:+1:

PR #2653, opened 2014-06-25
[3.2][WIP][ticket/12769] Add font-awesome

hanakin:
#### TODO:

- [ ] Add CDN version of Font-Awesome 4.1.0
- [ ] Merge search changes
- [ ] Replace all icons with fa best guess !placeholders
- [ ] Remove all icon images
- [ ] Remove all css rules related to applying the various icons
- [ ] Remove all the unnecessary HTML bits
- [ ] Cleanup the semantics in the HTML
- [ ] Cross browser checks
- [ ] Determine final choices for icons
- [ ] Fix any gap icons maybe we use [Fontastic](http://app.fontastic.me/#) implementation of fa + whatever else we need?

PR: https://tracker.phpbb.com/browse/PHPBB3-12769

PHPBB3-12769

PR #3851, opened 2015-08-21
[ticket/12769] Font-Awesome Integration & Icon overhaul

hanakin:
#### TODO:

- [x] Add CDN version of Font-Awesome
- [x] Search Icons
- [x] Navbar icons
- [x] Re-fator Drop-Downs
- [x] Re-factor Buttons
- [x] Form Icons
- [x] Post Icons

- [x] Misc Icons

- [x] Cleanup the semantics in the HTML
- [x] Cleanup /Remove unnecessary CSS
- [x] Remove replaced images
- [x] Refactor any javascript breaks
- [x] Cross browser checks
- [x] Finalize icon choices
- [x] Fix rtl
- [x] Fix responsive

Old PR: https://github.com/phpbb/phpbb/pull/2653
Ticket: https://tracker.phpbb.com/browse/PHPBB3-12769

Requirments: https://github.com/phpbb/phpbb/pull/3837

PHPBB3-12769


hanakin:
!set WIP

hanakin:
!set Styles

hanakin:
This is pretty much ready for review other than one small js fix for switching subscribe icons. I am going to ask for as much help with the review and cross-browser checks on this as possible as there is alot of page changes here. Not to mention alot of icons and I might have missed one

phpBB/styles/prosilver/template/search_results.html
```diff
@@ -64,17 +76,36 @@ <h2 class="searchresults-title"><!-- IF SEARCH_TITLE -->{SEARCH_TITLE}<!-- ELSE
...
-							<!-- IF searchresults.S_TOPIC_REPORTED --><a href="{searchresults.U_MCP_REPORT}">{REPORTED_IMG}</a><!-- ENDIF --><br />
+
+
+							<!-- IF searchresults.S_TOPIC_UNAPPROVED or searchresults.S_POSTS_UNAPPROVED -->
+							<a href="{searchresults.U_MCP_QUEUE}" title="{TOPIC_UNAPPROVED}">
+								<i class="icon fa-question fa-fw icon-blue"></i><span class="sr-only">{TOPIC_UNAPPROVED}</span>
+							</a> 
+							<!-- ENDIF -->
+							<!-- IF searchresults.S_TOPIC_DELETED -->
+							<a href="{searchresults.U_MCP_QUEUE}" title="{TOPIC_DELETED">
```
VSEphpbb: missing closing bracket in TOPIC_DELETED

VSEphpbb:
Nice work! My main concern is the backwards compatibility issues this will introduce for extensions. Mainly in the form of where you are deleting classes used by extensions. For example, icons in the navbar.

Is it possible we can have a small compatibility layer in the CSS where some of the (deprecated) classes you eliminated are still retained so some extensions don't get too clobbered between 3.1 and 3.2? At least for `.small-icon`, and if anybody can think of any others that have been important to extensions.

phpBB/styles/prosilver/template/navbar_header.html
```diff
@@ -40,50 +78,101 @@
...
 		<!-- ENDIF -->
 		<!-- IF S_NOTIFICATIONS_DISPLAY -->
-			<li class="small-icon icon-notification dropdown-container dropdown-{S_CONTENT_FLOW_END} rightside" data-skip-responsive="true">
-				<a href="{U_VIEW_ALL_NOTIFICATIONS}" id="notification_list_button" class="dropdown-trigger"><span>{L_NOTIFICATIONS} [</span><strong>{NOTIFICATIONS_COUNT}</strong><span>]</span></a>
-				<!-- INCLUDE notification_dropdown.html -->
-			</li>
+		<li class="dropdown-container dropdown-{S_CONTENT_FLOW_END} rightside" data-skip-responsive="true">
+			<a href="{U_VIEW_ALL_NOTIFICATIONS}" id="notification_list_button" class="dropdown-trigger">
+				<i class="icon fa-bell fa-fw"></i><span><span>{L_NOTIFICATIONS} [</span><strong>{NOTIFICATIONS_COUNT}</strong><span>]</span></span>
+			</a></span>
```
VSEphpbb: This closing span tag is a stray that has no opening

phpBB/styles/prosilver/template/mcp_topic.html
```diff
@@ -112,14 +112,17 @@ <h3 id="review">
...
+					<i class="icon fa-file fa-fw icon-lightgray icon-sm"></i><span class="sr-only">{postrow.MINI_POST}</span>
+				</a> {L_POSTED} {postrow.POST_DATE} {L_POST_BY_AUTHOR} <strong>{postrow.POST_AUTHOR_FULL}</strong><!-- IF postrow.U_MCP_DETAILS --> [ <a href="{postrow.U_MCP_DETAILS}">{L_POST_DETAILS}</a> ]<!-- ENDIF --></p>
 
 				<!-- IF postrow.S_POST_UNAPPROVED -->
 				<p class="post-notice unapproved">
 					<a href="{postrow.U_MCP_APPROVE}"><strong>{L_POST_UNAPPROVED}</strong></a>
 				</p>
 				<!-- ENDIF -->
-
+
```
VSEphpbb: Whitepsace

phpBB/styles/prosilver/template/posting_review.html
```diff
@@ -14,7 +14,15 @@
 
 	<div class="postbody" id="ppr{post_review_row.POST_ID}">
 		<h3><a href="#ppr{post_review_row.POST_ID}">{post_review_row.POST_SUBJECT}</a></h3>
-		<p class="author"><!-- IF S_IS_BOT -->{post_review_row.MINI_POST_IMG}<!-- ELSE --><a href="{post_review_row.U_MINI_POST}">{post_review_row.MINI_POST_IMG}</a><!-- ENDIF --> {L_POST_BY_AUTHOR}<strong>  {post_review_row.POST_AUTHOR_FULL}</strong> &raquo; {post_review_row.POST_DATE}</p>
+		<p class="author">
+		<!-- IF S_IS_BOT -->
```
VSEphpbb: Indentation please

phpBB/styles/prosilver/template/posting_review.html
```diff
@@ -14,7 +14,15 @@
...
+		<span><i class="icon fa-file fa-fw icon-lightgray icon-sm"></i><span class="sr-only">{post_review_row.MINI_POST}</span></span>
+		<!-- ELSE -->
+		<a href="{post_review_row.U_MINI_POST}" title="{post_review_row.MINI_POST}">
+			<i class="icon fa-file fa-fw icon-lightgray icon-sm"></i><span class="sr-only">{post_review_row.MINI_POST}</span>
+		</a>
+		<!-- ENDIF -->
+		{L_POST_BY_AUTHOR}<strong>  {post_review_row.POST_AUTHOR_FULL}</strong> &raquo; {post_review_row.POST_DATE}</p>
```
VSEphpbb: Closing p tag should be on its own line

phpBB/styles/prosilver/template/posting_topic_review.html
```diff
@@ -28,22 +28,30 @@ <h3 id="review">
...
+						<i class="icon fa-quote-left fa-fw"></i><span>{L_QUOTE} {topic_review_row.POST_AUTHOR}</span>
 					</a>
 				</li>
 			<!-- ENDIF -->
 			</ul>
 			<!-- ENDIF -->
 
-			<p class="author"><!-- IF S_IS_BOT -->{topic_review_row.MINI_POST_IMG}<!-- ELSE --><a href="{topic_review_row.U_MINI_POST}">{topic_review_row.MINI_POST_IMG}</a><!-- ENDIF --> {L_POST_BY_AUTHOR} <strong>{topic_review_row.POST_AUTHOR_FULL}</strong> &raquo; {topic_review_row.POST_DATE}  </p>
+			<p class="author">
+			<!-- IF S_IS_BOT -->
```
VSEphpbb: Indentation please and closing p tag again, same as above.

phpBB/styles/prosilver/template/ucp_pm_viewmessage.html
```diff
@@ -8,148 +8,177 @@
 
 <!-- IF S_DISPLAY_HISTORY and (U_VIEW_PREVIOUS_HISTORY or U_VIEW_NEXT_HISTORY) -->
 	<fieldset class="display-options clearfix">
-		<!-- IF U_VIEW_PREVIOUS_HISTORY --><a href="{U_VIEW_PREVIOUS_HISTORY}" class="left-box arrow-{S_CONTENT_FLOW_BEGIN}">{L_VIEW_PREVIOUS_HISTORY}</a><!-- ENDIF -->
-		<!-- IF U_VIEW_NEXT_HISTORY --><a href="{U_VIEW_NEXT_HISTORY}" class="right-box arrow-{S_CONTENT_FLOW_END}">{L_VIEW_NEXT_HISTORY}</a><!-- ENDIF -->
+		<!-- IF U_VIEW_PREVIOUS_HISTORY -->
+		<a href="{U_VIEW_PREVIOUS_HISTORY}" class="left-box arrow-{S_CONTENT_FLOW_BEGIN}">
```
VSEphpbb: Please indent content inside `<!-- IF -->` blocks here and elsewhere.

phpBB/styles/prosilver/template/viewtopic_body.html
```diff
@@ -354,25 +384,30 @@ <h3 <!-- IF postrow.S_FIRST_ROW -->class="first"<!-- ENDIF -->><!-- IF postrow.P
 <!-- ENDIF -->
 
 <!-- EVENT viewtopic_body_topic_actions_before -->
-<div class="action-bar bottom">
-	<div class="buttons">
+<div class="action-bar bar-bottom">
 		<!-- EVENT viewtopic_buttons_bottom_before -->
```
VSEphpbb: Can you fix indentation levels from here down please

hanakin:
@VSEphpbb there is going to be a lot of backwards compatibility issues I am sure I hade to completely rewrite several things ie buttons, action-bars, serach, drop-downs to mention a few to get this to work properly so not sure small-icon is really as big an issue by comparisons.

VSEphpbb:
@hanakin I know .small-icon will be an issue. As for phpBB's core, we have always been trying to include as much b.c. as possible so the transition from 3.1. to 3.2 will be as smooth as possible for users, mostly with respect to extensions. Obviously there are some major changes here that may break a couple extensions in a way we can't help or forsee, but for a simple class as .small-icon and as many extensions out there that may be using (including two flagship official extension, Pages and Board Rules) I see no reason why we can't keep the old class' CSS around during the 3.2.x line.

hanakin:
@VSEphpbb we should probably update the official extension icons to be svg to avoid pixel icons mixing but I will look into the BC stuff, but this brings up an old discussion I remember having with a few of the others about including a deprecated.css file to handle such issues like this

VSEphpbb:
@hanakin We will of course update the extensions. The issue is that it currently takes several months for an extension to get validated. The reality is it will be a long time after 3.2 is released before validated extensions can catch up. We want to mitigate issues where users who choose to upgrade and find extensions are broken, and there's no update to fix it anytime soon. We really want the 3.1 to 3.2 transition to be painless. 

So yeah, a deprecated css file would be great. As I said, so far I think it only needs .small-icon, though others might think of a few other rules that would help too.

hanakin:
As for now I just added the deprecated class to the end of the original files with a comment as this way you do not have to worry as much about specification with in its own file.

hanakin:
Short of any issues anyone has with choices for the icons this is good to go as far as Browser compatability. 

It has one error that has to do with the method of inclusion of Font-Awesome being from a CDN. As for some reason there is a test to ensure all styles are coming from the theme folder which obviously fails? Need to determine how we want to handle this.

@VSEphpbb if there are any more BC stuff its probably easier to fix those in alpha/beta when users can put it through the ringers. We can add BC back in at that point.

@cyberalien when you get a chance this is a large one!

hanakin:
!unset WIP

VSEphpbb:
@hanakin I like all the icons I've seen so far except for New Post/PM Buttons. One of the fa plus symbols makes more sense than a star to me. I know we use star currently, but we don't need to keep with it IMO. Not sure how others feel.

VSEphpbb:
I just noticed the close button icon in Ajax dialogs needs help with positioning now. It's flush up against the sides in the corner, could use some padding/margin

cyberalien:
In navigation icon color doesn't match links. Its too saturated. Maybe something like this?

    .navbar a .icon {
        color: inherit;
        opacity: 0.9;
    }

Code above breaks .icon.icon-gray color, so that rule will have to be duplicated with .navbar as well.

Orange color for last post link in topics list doesn't match layout. All icons are red, orange link near red icon looks rather bad. It is also way too small, but making it larger looks equally bad. Maybe use fa-file-o instead of fa-file for it?

Similar issue with post icon inside posts (one below post subject on left of post time). It is very small.

Everything else looks good.

VSEphpbb:
Didn't dig into find the issue, so here's a screenshot of a button I see that is messed up now by these changes: 
screen shot 2015-08-30 at 2 06 43 pm

hanakin:
@VSEphpbb 
- I prefer plus unless anyone else has issues. 
- The close button is there for a reason I forget how but there was a weird bg issue with the icon when floating off the container. Ill look at it again. 
- I will look into the button issue, what browser is screenshot from?

@cyberalien 
- The icon color is slightly different because that was the color of the original icons. I can make them the same as the link if prefered. 
- icon sizes were chosen to match original icon sizes can adjust based on input I actually created three modifier classes for this reason. 
- Can you be more specific about orange icon screen shot as everything used is original colors.

VSEphpbb:
@hanakin That button issue is not browser related. It's just using the button2 class which you have removed, so it lost its formatting.

hanakin:
its fixed now

cyberalien:
@hanakin 

Orange icon:
icon

Small icon in post:
icon2

I've tested all possible combinations of colors and FA glyphs for those icons, don't like any of them. Maybe keep these small icons as images instead of glyphs? Update images to have HD versions. Images look much better. FontAwesome file icons are too ugly at small font size.

hanakin:
i do not really see the problem the icon is not that different than the
originals.

cyberalien:
But it is very different. This is current image:

old

It is sharp, its color matches the rest of layout and has nice width/height ratio.

FA glyph is bland, blurred and looks fat. The more I look at it, the more I dislike it. It is very ugly. It might look nicer if its big, but when font size is tiny it looks like crap.

hanakin:
ok small fix to the notification drop-down. Swapped out the * for +. added slight alteration to the target post icon.
@cyberalien how about the icon now?
@VSEphpbb thoughts on the close button based on the attached screenshot of the issue?

anything else anyone sees?

VSEphpbb:
@hanakin :+1: close button looks good on the outer corner like that

hanakin:
k will update later tonight with those changes

hanakin:
@cyberalien when you get a chance can you look at this once more I think the target icon is the only thing left. unless anyone else has any objections. This should be ready to go.

cyberalien:
I don't see any difference.

cyberalien:
Never mind, its a bit bigger. But its still very blurred and orange.

hanakin:
It also narrower. I tweaked the scale slightly and fixed the color. should look pretty good now

hanakin:
@VSEphpbb @cyberalien can you check the target icon once more. 

I also replaced the contact menu icon and made some more changes to remove alot of the bloated icon/img based stuff from the theme

cyberalien:
Bug: icons in contact dropdown in messages are missing

Target icon: might be a bit better, but still very ugly. Now that I see you have SVG PR, maybe use something SVG instead of font?

Also its still orange for new posts. There is no orange color anywhere in style, so it doesn't belong there. It looked fine when it was just few pixels inside old icon, but if whole icon is orange it doesn't fit layout.

phpBB/styles/prosilver/theme/stylesheet.css
```diff
@@ -7,6 +7,7 @@
     --------------------------------------------------------------
 */
 
+@import url("https://maxcdn.bootstrapcdn.com/font-awesome/4.4.0/css/font-awesome.min.css");
```
nickvergessen: Should be added to our assets instead, so it also works if your forum is not connected to the internet or e.g. CDN is disabled?

tests/functional/paging_test.php
```diff
@@ -35,7 +35,7 @@ public function test_pagination()
 		$this->assertNotContains('post no4', $crawler->text());
 		$this->assertContains('post no16', $crawler->text());
 
-		$prev_link = $crawler->filter('.pagination > ul > li.previous > a')->attr('href');
+		$prev_link = $crawler->filter('.pagination > ul > li.prev > a')->attr('href');
```
nickvergessen: unnecessary, or why can't we keep the old name?

phpBB/styles/prosilver/template/viewtopic_body.html
```diff
@@ -342,43 +373,47 @@ <h3 <!-- IF postrow.S_FIRST_ROW -->class="first"<!-- ENDIF -->><!-- IF postrow.P
...
+				<span class="caret"><i class="icon fa-sort-down fa-fw"></i></span>
+			</span>
+			<div class="dropdown">
 				<div class="pointer"><div class="pointer-inner"></div></div>
 				<ul class="dropdown-contents">
-				<!-- BEGIN quickmod -->
-					<!-- DEFINE $QUICKMOD_AJAX = (quickmod.VALUE in ['lock', 'unlock', 'delete_topic', 'restore_topic', 'make_normal', 'make_sticky', 'make_announce', 'make_global']) -->
-					<li><a href="{quickmod.LINK}"<!-- IF $QUICKMOD_AJAX --> data-ajax="true" data-refresh="true"<!-- ENDIF -->>{quickmod.TITLE}</a></li>
-				<!-- END quickmod -->
+					<!-- BEGIN quickmod -->
```
nickvergessen: please avoid retabbing to reduce the diff size of the style changes on release

hanakin:
sorry do not see the problem with the icon choice looks fine too me, as for the color orange is used in pms for marked messages. If not orange what? unread color?

tests/functional/paging_test.php
hanakin: sorry just common practice naming convention to use nxt and prev. can revert if need be

phpBB/styles/prosilver/template/viewtopic_body.html
hanakin: This was due to tabbing be introduced early on that I had to fix and did not know what it look like before so I just fixed it all.

phpBB/styles/prosilver/theme/stylesheet.css
hanakin: agree, it will not even cache this way. We need a way to utilize the cdn when we can but also as you stated include it when no connection is present. I just did not know how the assets worked

tests/functional/paging_test.php
Nicofuma: yes please, it will make the life of style author easier

marc1706:
The text inside post buttons do not seem to be centered. This looks especially weird since the icons in the post buttons are properly centered.

hanakin:
@marc1706 what buttons are you referring too? and in what browser? the post reply buttons or the post action buttons?

marc1706:
I'm talking about these. The text is not properly vertically centered while the icons are.
centered
Browser is Chrome 45 on Windows.

hanakin:
the text is a pixel smaller and has always been i can increase it to the same as the icon, but the text has also always been aligned to the base of the icon not centered. 
image

marc1706:
Hm, probably didn't notice it because the older icon was a little bit blurred

hanakin:
Ok update so everyone is aware of where this is at.

@cyberalien  I reverted all the previous changes to the small post icons and simplified them as this is a base theme we should stick to font-awsome for all icons if it does not work then thats probably a justification for removal of that icon or reworking the UI for said feature in the future.

@VSEphpbb I believe I have addressed all the previous issues

@nickvergessen just so we are on the same page as far the assets inclusion of font-awesome you are just wanting this included as a template variable similar to how jquery is included? This is the only thing I have left to work on for this pending any one else's issues I should be able to get to this by the end of the week.

hanakin:
@nickvergessen @Nicofuma I have fixed the inclusion of font-awesome for connection fallback but I do not think we should require this failing functional test. It makes absolutely no sense to test that all links are pointing at the styles directory to me. This just leads to hack js workarounds. Such as the one we are using to include the google font.

@cyberalien @VSEphpbb Can you please give this another look when you get time.

VSEphpbb:
@hanakin it's trying to load fonts locally from assets/fonts, which doesn't exist when CDN is off.

Also, for New Post/PM/Reply buttons, I think the `fa-pencil` icon will look a hell of a lot better, instead of plus sign? 
screen shot 2015-09-13 at 9 29 09 pm

Should the PM icon in the header be a `fa-envelope` instead of an inbox?
screen shot 2015-09-13 at 9 30 25 pm

VSEphpbb:
This jump boxes triangle needs to be colored correctly:
screen shot 2015-09-13 at 9 32 51 pm

hanakin:
@VSEphpbb 

- pm icon choice is to avoid confusion as the envelope is used elsewhere for the contact us icon. 
- the pencil i do not think so as again to avoid confusion we are using the pencil for edit post.
- jump-box triangle is a seperate pr as thats not apart of this the issue with that is it needs its own class applied when used in that location as its an mcp jump which is rendered down instead of up like the normal jump.
- as for the CDN issue i see whats wrong and will fix that later tonight

VSEphpbb:
@hanakin I feel strongly about pencil though. It's not confusing as editing a post is the same result as writing a new post/pm... It leads you to the same thing, which is composing text.

VSEphpbb:
..adding to above, the New Topic (PM) buttons are the largest buttons we show. We should put our best foot forward with them by making them look nice. The plus sign is just crap. :-1:

VSEphpbb:
..and more. The edit post icon and the New Topic/PM buttons are never on the same page, so I see no confusion.

hanakin:
thats true they are not shown at the same time

VSEphpbb:
I don't think it'd be confusing to also use mail icon for PM notifications as we have currently been using the mail icon for PM notifications and the Contact link in 3.1, with no problems.

rxu:
Also, what if we'll have different icon colors for PMs/Notifications 0 and >0? F.e. for 0 it can be regular color and if you have PMs/Notifications it can be different (yellow/white/whatever suitable for the style).

VSEphpbb:
@hanakinfor the failing test, try adding `->eq(1)`

So in forum_style_test.php, replace all:
`$crawler->filter('head > link[rel=stylesheet]')->attr('href'));`
with:
`$crawler->filter('head > link[rel=stylesheet]')->eq(1)->attr('href'));`

This should skip over the first stylesheet link, which is now for fontawesome, and test the stylesheet for prosilver as before.

By the way, it would be good to add a similar test that does the same but checks to make sure fontawesome css is there.

```php
$this->assertContains('assets/css/font-awesome', $crawler->filter('head > link[rel=stylesheet]')->eq(0)->attr('href'));
```

hanakin:
@rxu outside the scope of this pr, but i would rather just not show it unless we have them. I never really seen the point of the 0 in the first place but thats maybe just me.

VSEphpbb:
Topic icons next to Private messages are missing in pm viewfolder.

hanakin:
Ok so I completely rebased for a cleaner diff to help avoid merge conflicts. So its going to need another round of reviews. Pending any other issues with the icon choices everything is good to go afterwards

cyberalien:
I still don't like mini post glyph, but if all images are supposed to be replaced, I don't see a better alternative.

Contact dropdown has gif icons instead of FA glyphs.

hanakin:
@cyberalien contact icons themselves are problematic for numerous reasons so I left it alone it will need to be its own pr as it will require a lot of rewriting in the backend as everything is include based on individual DB values it may be better handleded as svgs rather than font as well

cyberalien:
Sounds good.

Then everything seems fine. I don't see any errors.

marc1706:
I get three 404s:
```
GET http://localhost/phpbb3/assets/fonts/fontawesome-webfont.woff2?v=4.4.0 
GET http://localhost/phpbb3/assets/fonts/fontawesome-webfont.woff?v=4.4.0 
GET http://localhost/phpbb3/assets/fonts/fontawesome-webfont.ttf?v=4.4.0
```

Nicofuma:
I don't have these 404. @marc1706 can you check please?

Otherwise looks good

Nicofuma:
@cyberalien By the way, do you know/have an idea how much work it will require to have prosilver se compatible?

hanakin:
@Nicofuma the 404 was fixed by https://github.com/hanakin/phpbb/commit/ff6d590b3693c5a20f1626664bb92bb18f384d3a

PR #3855, opened 2015-08-23
[ticket/14123] Add descriptive help to the CLI help output

VSEphpbb:
https://tracker.phpbb.com/browse/PHPBB3-14123

Appends the Help section at the bottom of the image:
screen shot 2015-08-28 at 7 52 16 am

PHPBB3-14123


Nicofuma:
:+1: thanks you

VSEphpbb:
Will update for 3.2 after this is merged into 3.1 and master 😉

phpBB/language/en/cli.php
```diff
@@ -54,7 +54,7 @@
 	'CLI_DESCRIPTION_DELETE_CONFIG'				=> 'Deletes a configuration option',
 	'CLI_DESCRIPTION_DISABLE_EXTENSION'			=> 'Disables the specified extension.',
 	'CLI_DESCRIPTION_ENABLE_EXTENSION'			=> 'Enables the specified extension.',
-	'CLI_DESCRIPTION_FIND_MIGRATIONS'			=> 'Finds migrations that are not depended on.',
+	'CLI_DESCRIPTION_FIND_MIGRATIONS'			=> 'Finds migrations that are not depended upon.',
```
bantu: This is a language fix. Consider extracting from this PR.

bantu:
I certainly agree that some commands need to be described better and possibly need to use setHelp() to set detailed command documentation. This PR seems to introduce quite some duplication however, which indates to me that there might be more generic problems that should be solved instead.

For example:

Commands already know information like "The key name and increment amount are required.". Thus, maybe, such text should be automatically generated instead of having it manually typed out in the language file.

Then, the CLI_HELP_DELETE_CONFIG definition is basically a repetition of CLI_DESCRIPTION_DELETE_CONFIG. Maybe the description should be shown when no specific help section is set?

Nicofuma:
related to https://github.com/symfony/symfony/issues/15498

Nicofuma:
https://github.com/symfony/symfony/pull/15601

VSEphpbb:
The help contains example usages while the descriptions do not.

VSEphpbb:
It's also worth mentioning that if some commands just use the description, then not all will use setHelp, so when new commands are added, the setHelp may be forgotten or overlooked, and that will lead to inconsistencies as well - where some commands have great help and some have none.

phpBB/language/en/cli.php
VSEphpbb: It is a lang fix, but one that was made for consistency with the new help line.

VSEphpbb:
After looking over these changes again, I do agree with @bantu and think this does appear to be a lot of duplication that may be more trouble than it's worth.

I have just updated the basic CLI docs here with more info about understanding how the CLI help shows us when args/opts are optional or required. 
https://github.com/phpbb/documentation/pull/49

Nicofuma:
Well, some commands still deserve a better description (like config:set-atomic) and some may deserve a few warnings in the description (no idea in mind). But it's fine if some don't have a long description (we will fallback to the short one in this case)

VSEphpbb:
We could re-use the description in the help for most in language files, like so:
`'CLI_HELP_CRON_LIST' => $lang['CLI_DESCRIPTION_CRON_LIST'],`

And if more help explanation is needed:
`'CLI_HELP_SET_ATOMIC_CONFIG' => $lang['CLI_DESCRIPTION_SET_ATOMIC_CONFIG'] . ' Say more stuff here',`

Nicofuma:
The short description will be used automatically if a long one isn't provided

VSEphpbb:
Isn't that not available yet?

Nicofuma:
I'm waiting to see if it's merged in Symfony. If not I'll make the patch in phpBB

VSEphpbb:
OK, removed redundant help messages. Now there's just help on the commands requiring a little more explanation, and re-using the description lang key within the new help key.

phpBB/language/en/cli.php
```diff
@@ -54,8 +54,8 @@
 	'CLI_DESCRIPTION_DELETE_CONFIG'				=> 'Deletes a configuration option',
 	'CLI_DESCRIPTION_DISABLE_EXTENSION'			=> 'Disables the specified extension.',
 	'CLI_DESCRIPTION_ENABLE_EXTENSION'			=> 'Enables the specified extension.',
-	'CLI_DESCRIPTION_FIND_MIGRATIONS'			=> 'Finds migrations that are not depended on.',
-	'CLI_DESCRIPTION_GET_CONFIG'				=> 'Gets a configuration option’s value',
+	'CLI_DESCRIPTION_FIND_MIGRATIONS'			=> 'Finds migrations that are not depended upon.',
+	'CLI_DESCRIPTION_GET_CONFIG'				=> 'Gets a configuration option’s integer value',
```
Nicofuma: that's not always an integer

VSEphpbb: Oops, wrong line

Nicofuma:
related to https://github.com/phpbb/phpbb/pull/3877

PR #2540, opened 2014-06-03
[ticket/12610] Add command to check if the board is up to date.

Stephannos:
[ticket/12610] Add command to check if the board is up to date.
https://tracker.phpbb.com/browse/PHPBB3-12610

PR #3856, opened 2015-08-24
[ticket/12610] Add command to check if the board is up to date.

Nicofuma:
PHPBB3-12610


Nicofuma:
replaces https://github.com/phpbb/phpbb/pull/2540

Nicofuma:
to be done: long help

@VSEphpbb ping? =)

marc1706:
The tests fail because version check is now ran by default on all extensions in the list. This causes the two actual extensions to output errors because they don't have any version check info inside them. I'm not sure if this is behavior we should want by default?
@paul999 @VSEphpbb @Derky

paul999:
Can't we skip extension which doesn't have any version check information? Having the command is usefull, especially for extensions as they are updated more often.

Derky:
The version check info in composer is optional so it shouldn't error when an extension doesn't come with a version check URL.

marc1706:
I'd recommend to let this fail silently on the extensions page.

paul999:
Or maybe just display something like N/A?

marc1706:
This already happens on the overview page. There is no update info displayed on that page. It's rather just green/red or black for no info.

paul999:
I would leave it black in that case.

marc1706:
It seems like right now this does not properly check for the latest version on the extensions overview page. One needs to click on details to have phpBB check if extensions are up to date.

Before that, error messages like this one are displayed:
```
The “forumhulp/realpostcount” extension is not valid.
Failed to obtain latest version information.
```
This is obviously incorrect as this message disappears once one clicks on the extension details.

marc1706:
@Nicofuma what's the status of this?

Nicofuma:
kinda forgot about it. Will update the branch soon

marc1706:
Status?

Nicofuma:
well... I just need time to finish it (I need to not forget about it too...)

Nicofuma:
rebased and fixed

phpBB/bin/phpbbcli.php
```diff
@@ -42,6 +42,7 @@
 require($phpbb_root_path . 'includes/functions.' . $phpEx);
 require($phpbb_root_path . 'includes/functions_admin.' . $phpEx);
 require($phpbb_root_path . 'includes/utf/utf_tools.' . $phpEx);
+include($phpbb_root_path . 'includes/functions_compatibility.' . $phpEx);
```
marc1706: Only conflict right now is that this was already added by another pull request.

phpBB/phpbb/console/command/update/check.php
```diff
@@ -0,0 +1,311 @@
...
+			{
+				$this->display_versions($output, $updates_available);
+			}
+
+			return 1;
+		}
+		else
+		{
+			$output->writeln('');
+			$output->writeln('<question>' . $this->user->lang('NOT_UP_TO_DATE', $metadata['name']) . '</question>');
```
marc1706: Looks like a copy paste typo. Should display that it's up to date.

phpBB/phpbb/console/command/update/check.php
```diff
@@ -0,0 +1,311 @@
...
+			}
+
+			return 1;
+		}
+		else
+		{
+			if ($input->getOption('verbose'))
+			{
+				$output->writeln('');
+				$output->writeln('<question>' . $this->user->lang('UPDATE_NOT_NEEDED') . '</question>');
```
marc1706: I think this should always be displayed, even if verbose is disabled. Right now this just returns empty.

marc1706:
Don't you think it might be good to use the SymfonyStyle I/O styling that symfony introduced in 2.8?
http://symfony.com/blog/new-in-symfony-2-8-console-style-guide

VSEphpbb:
^ What @marc1706 said :)

marc1706:
Also, it looks like you're still using `$user->lang()` instead of the new language class. I guess that's still from when you initially based this for 3.1.x?

marc1706:
I think you should also catch the exception and do a "nice" output when someone tries to check a non-existent extension for updates:
```
[RuntimeException]                                                                                       
  The requested file could not be found: phpBB/bin/../ext/foobar/composer.json                                                                                                         
```

Nicofuma:
right, I'll do that later

Nicofuma:
@marc1706 updated

marc1706:
Looks like you forgot to update the tests.

PR #3857, opened 2015-08-24
[ticket/14124] Automatically translate exceptions in CLI 

Nicofuma:
https://tracker.phpbb.com/browse/PHPBB3-14124

PHPBB3-14124

### DIFF
--- a/phpBB/phpbb/console/application.php
+++ b/phpBB/phpbb/console/application.php
@@ -13,6 +13,7 @@
 
 namespace phpbb\console;
 
+use phpbb\exception\exception_interface;
 use Symfony\Component\Console\Shell;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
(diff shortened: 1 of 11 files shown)


phpBB/phpbb/console/application.php
```diff
@@ -115,6 +116,16 @@ public function doRun(InputInterface $input, OutputInterface $output)
...
 
-		return parent::doRun($input, $output);
+		try
+		{
+			$result = parent::doRun($input, $output);
+		}
+		catch (exception_interface $e)
+		{
+			$message = call_user_func_array(array($this->language, 'lang'), array_merge(array($e->getMessage()), $e->get_parameters()));
+			throw new \RuntimeException($message, $e->getCode(), $e);
```
Nicofuma: one question: do we want to link to the previous exception? I think yes because otherwise it is hard to debug.
But this way the whole exception chain will be displayed... 

EDIT: one solution could be to embed the previous exception only if we have debug enabled.

Nicofuma:
Actually it should be done with a listener, listening to `ConsoleEvents::EXCEPTION` (like we do for the controllers)

I let it like for now, because of my upper question

Nicofuma:
updated

marc1706:
giphy

PR #3863, opened 2015-08-26
[ticket/14128] Fix img bbcode regression, lost postimage class

VSEphpbb:
https://tracker.phpbb.com/browse/PHPBB3-14128

PHPBB3-14128


VSEphpbb:
@s9e

s9e:
Looks fine.

By the way, beside the hardcoded default being different from 3.1 this issue highlights the fact the img template in prosilver's bbcode.html is not being used because its bit in `phpbb_styles.template_bitfield` isn't set.

VSEphpbb:
@s9e Can you make a ticket for that issue please?

VSEphpbb:
@s9e Nevermind the ticket, I guess. According to our knowledgebase, Prosilver does not, by default, use that HTML replacement for IMG tags (as well as some others).

https://www.phpbb.com/support/docs/en/3.1/kb/article/how-to-template-bitfield-and-bbcodes/ See bettom of "Generate the bitlist"	

So this appears to be the intended behavior.

PR #3864, opened 2015-08-30
[ticket/8920] Add a MCP PM reports view permission

Zoddo:
Anybody who has the permission to read reports in one forum can read PM reports.

Solving this problem by adding a new permission to read PM reports.

https://tracker.phpbb.com/browse/PHPBB3-8920

### DIFF
--- a/phpBB/phpbb/db/migration/data/v31x/m_pm_report.php
+++ b/phpBB/phpbb/db/migration/data/v31x/m_pm_report.php
@@ -1,0 +1,43 @@
+<?php
+/**
+*
+* This file is part of the phpBB Forum Software package.
+*
+* @copyright (c) phpBB Limited <https://www.phpbb.com>
+* @license GNU General Public License, version 2 (GPL-2.0)
+*
+* For full copyright and license information, please see
+* the docs/CREDITS.txt file.
+*
+*/
+
+namespace phpbb\db\migration\data\v31x;
+
+class m_pm_report extends \phpbb\db\migration\migration
+{
+	static public function depends_on()
+	{
+		return array('\phpbb\db\migration\data\v31x\v316rc1');
+	}
+
+	public function update_data()
+	{
+		return array(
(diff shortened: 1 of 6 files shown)


phpBB/phpbb/db/migration/data/v31x/m_pm_report.php
```diff
@@ -0,0 +1,43 @@
...
+{
+	static public function depends_on()
+	{
+		return array('\phpbb\db\migration\data\v31x\v316rc1');
+	}
+
+	public function update_data()
+	{
+		return array(
+			array('permission.add', array('m_pm_report', true, 'm_report')),
```
Nicofuma: you should add the permission to the MODERATOR rule too

Zoddo: The permission is copied from `m_report`.

Nicofuma: didn't see... that's good so

Nicofuma:
after running the migration, the permission was set to no for the full moderator on my test board...

rxu:
It probably needs the ```permission.permission_set``` as well to apply new permission for groups/roles.

Zoddo:
@rxu : So, a bug with the "permission copy" function?

rxu:
Tested locally like that
```php
$permission = new \phpbb\db\migration\tool\permission($db, $cache, $auth, $phpbb_root_path, $phpEx);
$permission->add('m_pm_report', true, 'm_report');
```

Checked the permission for the Full moderator and it's been set to ```Yes```.

@Nicofuma any chance you have set the ```m_report``` permission to ```No``` for the full moderator so it's just copied to the new permission?

Zoddo:
Hi,

Any news?

marc1706:
Looks good. Can't reproduce the issue mentioned by @Nicofuma

PR #3865, opened 2015-08-30
[ticket/13992] Fix HTML5 error from w3.org validator

Zoddo:
https://tracker.phpbb.com/browse/PHPBB3-13992

### DIFF
--- a/phpBB/styles/prosilver/template/overall_header.html
+++ b/phpBB/styles/prosilver/template/overall_header.html
@@ -46,7 +46,7 @@
 <!-- ENDIF -->
 <link href="{T_STYLESHEET_LINK}" rel="stylesheet">
 <link href="{T_STYLESHEET_LANG_LINK}" rel="stylesheet">
-<link href="{T_THEME_PATH}/responsive.css?assets_version={T_ASSETS_VERSION}" rel="stylesheet" media="all and (max-width: 700px), all and (max-device-width: 700px)">
+<link href="{T_THEME_PATH}/responsive.css?assets_version={T_ASSETS_VERSION}" rel="stylesheet" media="all and (max-width: 700px)">
 
 <!-- IF S_CONTENT_DIRECTION eq 'rtl' -->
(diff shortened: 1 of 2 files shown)


VSEphpbb:
Why are you getting rid of media=screen? The only invalid thing we had there was "projection" which we removed. screen is still very valid.

Zoddo:
I removed these changes

phpBB/styles/prosilver/template/overall_header.html
```diff
@@ -46,7 +46,7 @@
 <!-- ENDIF -->
 <link href="{T_STYLESHEET_LINK}" rel="stylesheet">
 <link href="{T_STYLESHEET_LANG_LINK}" rel="stylesheet">
-<link href="{T_THEME_PATH}/responsive.css?assets_version={T_ASSETS_VERSION}" rel="stylesheet" media="all and (max-width: 700px), all and (max-device-width: 700px)">
+<link href="{T_THEME_PATH}/responsive.css?assets_version={T_ASSETS_VERSION}" rel="stylesheet" media="(max-width: 700px)">
```
VSEphpbb: I think it should be all and (max-width: 700px) ?

phpBB/styles/prosilver/template/simple_header.html
```diff
@@ -24,7 +24,7 @@
 <!-- ENDIF -->
 <link href="{T_STYLESHEET_LINK}" rel="stylesheet">
 <link href="{T_STYLESHEET_LANG_LINK}" rel="stylesheet">
-<link href="{T_THEME_PATH}/responsive.css?assets_version={T_ASSETS_VERSION}" rel="stylesheet" media="only screen and (max-width: 700px), only screen and (max-device-width: 700px)">
+<link href="{T_THEME_PATH}/responsive.css?assets_version={T_ASSETS_VERSION}" rel="stylesheet" media="(max-width: 700px)">
```
VSEphpbb: I think it should be only screen and (max-width: 700px) ?

PR #3866, opened 2015-08-30
[Ticket/14038] Jumpbox improvements

hanakin:
Provide some visualization back to the jumpbox

Tracker: https://tracker.phpbb.com/browse/PHPBB3-14038

PHPBB3-14038

PR #3867, opened 2015-08-30
[Ticket/14131] Completely overhaul all colors in Colours.css

hanakin:
Completely overhaul all colors in Colours.css to be consistent and standardized. Match all colors to a color styleguide using set color variables.

Tracker: https://tracker.phpbb.com/browse/PHPBB3-14131

PHPBB3-14131

### DIFF
--- a/phpBB/styles/prosilver/theme/colours.css
+++ b/phpBB/styles/prosilver/theme/colours.css
@@ -341,7 +341,7 @@ a:hover .icon-black {
 	color: #FFFFFF;
 }
 
-.jumpbox-cat-link:hover { 
+.jumpbox-cat-link:hover {
 	background-color: #13A4EC;
 	border-top-color: #13A4EC;
 	color: #FFFFFF;
@@ -511,7 +511,7 @@ blockquote {
 	background-image: url("./images/quote_rtl.gif");
 }
 
-blockquote blockquote {	
+blockquote blockquote {
 	background-color: #F0F0DB;
 }
 
@@ -730,7 +730,7 @@ Colours and backgrounds for buttons.css
 }
 
 .caret {
-	border-color: #CCCCCC; 
+	border-color: #CCCCCC;
 }
 
 .contact-icons a:hover {
@@ -950,7 +950,7 @@ ul.cplist {
 /* Link styles for the sub-section links */
 .navigation a {
 	color: #333333;
-	backgroun-color: #AFC5CF;
+	background-color: #AFC5CF;
 	background-image: -webkit-linear-gradient(left, #AFC5CF 50%, #C8DEEA 100%);
 	background-image: linear-gradient(to right, #AFC5CF 50%, #C8DEEA 100%);
 	filter: progid:DXImageTransform.Microsoft.gradient(startColorstr='#AFC5CF', endColorstr='#C8DEEA', GradientType=1);
@@ -1159,8 +1159,6 @@ a.button1,
 input.button1 {
 	border-color: #666666;
 }
-
-[Define New Theme] 8. Utilize Font Awesome icons where icons are needed to simplify server requests. 
 
 input.button3 {
 	background-image: none;
--- a/phpBB/styles/prosilver/theme/colours.css
+++ b/phpBB/styles/prosilver/theme/colours.css
@@ -8,14 +8,16 @@ Colours
 
 // Grayscale Colors
 @color-black:					#000000;
+@color-black-90:				#1A1A1A;
 @color-black-80:				#333333;
 @color-black-70:				#4D4D4D;
 @color-black-60:				#666666;
 @color-black-50:				#7F7F7F;
 @color-black-40:				#999999;
+@color-black-30:				#B3B3B3;
 @color-black-25:				#BFBFBF;
 @color-black-20:				#CCCCCC;
-@color-black-23:				#DDDDDD;
+@color-black-13:				#DDDDDD;
 @color-black-10:				#E6E6E6;
 @color-black-5:					#F2F2F2;
 @color-black-2:					#FAFAFA;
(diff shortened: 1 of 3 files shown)


hanakin:
!set Styles

Zoddo:
See https://github.com/phpbb/phpbb/pull/3866#issuecomment-136151656 :)

hanakin:
Ready for review!

Nicofuma:
I guess that not every browser supports this feature, right? What about the one we officially support?

Nicofuma:
Well actually, I don't really understand what you are doing... just cleaning up the formatting and some weird color definitions?

hanakin:
Nothing is supported by a browser those are just the variables I used when I fixed all the incorrect colors in the theme which is the main purpose of this. I removed all the slight variances and replaced all the incorrect colors with the proper ones based on the standardized guide in the comments. I left the comments in for documentation purposes on what colors to use in what situations. Overall reducing the total number of colors in half

Nicofuma:
ok, thanks.

By the way, you need to rebase (it conflicts with another of your PRs)

hanakin:
yeah working that now there is alot of manual review lol

Nicofuma:
@cyberalien Again, could you have a look please? (Yes, there is a lot of styling PRs these days :-))

phpBB/styles/prosilver/theme/colours.css
```diff
@@ -1,24 +1,106 @@
...
+
+html,
+body {
+	color: #5C758C;
 	background-color: #F5F7FA;
 }
 
-h1 {
-	color: #FFFFFF;
+h1 {
```
marc1706: Trailing whitespace should be removed

marc1706:
Your PR seems to contain a lot of trailing whitespace. They're nicely highlighted by github, too.

hanakin:
@marc1706 yeah have not really done much work on as every merge affects it drastically it going to have to go as the last merge anyway. So its best to fix it all during the long dynamic rebase

hanakin:
!set WIP

cyberalien:
Sorry for delay. Color changes look good.

Zoddo:
@hanakin : You need to add `PHPBB3-14131` at the end of each commit message for the tracker.

hanakin:
ready for review @cyberalien

hanakin:
!unset WIP

hanakin:
ready for review @prototech should not be any BC issues just helping to standardize colours.

VSEphpbb:
Umm,forum background color is awful. Please revert that. There's just a weird wash of blue-i-ness to this that makes phpBB uglier.

hanakin:
@VSEphpbb used bg instead of 2 how about now?

VSEphpbb:
Actually, I was talking about the body/html background.

hanakin:
@VSEphpbb

marc1706:
The post text in the editor is also highlighted in the same color as the outside of the text box border. This looks kind of weird.

hanakin:
@marc1706 ill have a look when i get home later tonight

hanakin:
@marc1706 try now switched it back to the darker blue

VSEphpbb:
Text in post editor should be black. If that's what's being discussed here.

hanakin:
hmm can make it default color but its always been dark blue

VSEphpbb:
It's always been black. technically, `#333333`
https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/styles/prosilver/theme/colours.css#L973-L987

hanakin:
hmm guess the specificity was coded backwords originally not sure why any of the inputs was blue to begin with possibly easier to leave it as all black and remove/deprecate that line.

VSEphpbb:
I have to say I'm still not a fan of some of the color changes, or why they are even needed. To my eyes, this PR is just hurting things with the color changes/choices.

That said, still many inconsistencies with font colors, leading to them harder to read. Font colors are too washed out, and sometimes a weird off blue/green when they should be black.

THIS PR:
screen shot 2015-11-23 at 7 35 48 pm
screen shot 2015-11-23 at 7 35 36 pm

Original (better):
screen shot 2015-11-23 at 7 36 41 pm
screen shot 2015-11-23 at 7 36 53 pm

hanakin:
the problem was the overuse of colors. there were an extreme amount of unique colors being used in the theme making it harder to manage and properly fix things. This is a perfect example of poor design/color usage from the start as you stated should probably be black and I agree but the original was not black. it was a randomly chose tint of blue which had no ties to any of the primary or what should have been the primary colors. A primary color scheme was never chosen, instead the individual colors were chosen on each mockup and thus we ended up with hundreds of unique colors which is a nightmare for web/design. It makes it harder for theme authors to modify or extension authors to add too if they can not easily see what color to use.

There was probably 5-6 different colors for basic text/headers alone. which should not be the case you should only have 1 and possibly a lighter shade of it for subs.

The methods that I used for replace was not optimal for fixing every individual case and but it was the easiest way to standardize the majority.

here is with the secondary dark color to attempt keeping some of the original blue
screenshot 2015-11-24 07 48 30

VSEphpbb:
Still, I don't see any value to messing with Prosilver, when these efforts could be better spent on work for an all new design/theme. This is probably going to make the 3.1 to 3.2 transition for style authors another nightmare (and the 3.0 to 3.2 even more so).

hanakin:
it should have little if any impact on authors current themes as no BC is broken all thats changes is the default colors for more consistancy. As for the why yes a new theme is needed and probably better spent working on it but it will be some time until a new theme is finished can be included with the core and completely replaces prosilver to the point where we can not support it anymore given the extent of extensions and styles already based of it.

VSEphpbb:
Maybe the bug fixes should be separated from this conceptual change into two PRs (bugs such as message editor text currently being blue)

hanakin:
I agree that fixes should be separate and for the most part this should adhere to it. the only deviations is either formating to make it easier to read which can be undone or implemented in 3.2 ie the icon colors.

PR #3868, opened 2015-08-30
[ticket/7362] Add alt attribute on topic icons

Zoddo:
https://tracker.phpbb.com/browse/PHPBB3-7362

### DIFF
--- a/phpBB/includes/functions_posting.php
+++ b/phpBB/includes/functions_posting.php
@@ -310,6 +310,7 @@ function posting_gen_topic_icons($mode, $icon_id)
 			{
 				$template->assign_block_vars('topic_icon', array(
 					'ICON_ID'		=> $id,
+					'ICON_NAME'		=> $data['img'],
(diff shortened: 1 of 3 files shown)


Nicofuma:
While it is already an improvement, it looks 'too small'. With this patch we display the image name but it's often meaningless. Instead we could provide an optional description and if the description is not set, fallback to the image name.

By the way, what about the alt/title thing?

Zoddo:
@Nicofuma : So, I think it would be better to do this in 3.2 in this case, no?

Nicofuma:
We could do that indeed: keep this patch in 3.1 and do more in 3.2

Zoddo:
@Nicofuma : Yes, I think we can merge this patch for 3.1 and open an another ticket to do more in 3.2

Nicofuma:
please, create a ticket for 3.2 and send a PR if you want :)

Nicofuma:
Could you also add the alt to the topic icon list in posting please?

PR #3869, opened 2015-08-30
[ticket/8800] Add "mark topics read" link to "View unread posts"

Zoddo:
https://tracker.phpbb.com/browse/PHPBB3-8800

### DIFF
--- a/phpBB/search.php
+++ b/phpBB/search.php
@@ -427,6 +427,8 @@ if ($keywords || $author || $author_id || $search_id || $submit)
 
 				gen_sort_selects($limit_days, $sort_by_text, $sort_days, $sort_key, $sort_dir, $s_limit_days, $s_sort_key, $s_sort_dir, $u_sort_param);
 				$s_sort_key = $s_sort_dir = $u_sort_param = $s_limit_days = '';
+
+				$template->assign_var('U_MARK_ALL_READ', ($user->data['is_registered'] || $config['load_anon_lastread']) ? append_sid("{$phpbb_root_path}index.$phpEx", 'hash=' . generate_link_hash('global') . '&amp;mark=forums&amp;mark_time=' . time()) : '');
 			break;
 
 			case 'newposts':
--- a/phpBB/styles/prosilver/template/search_results.html
+++ b/phpBB/styles/prosilver/template/search_results.html
@@ -33,6 +33,7 @@
 		<!-- EVENT search_results_searchbox_after -->
 
 		<div class="pagination">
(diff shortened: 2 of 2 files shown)


phpBB/search.php
```diff
@@ -427,6 +427,8 @@
 
 				gen_sort_selects($limit_days, $sort_by_text, $sort_days, $sort_key, $sort_dir, $s_limit_days, $s_sort_key, $s_sort_dir, $u_sort_param);
 				$s_sort_key = $s_sort_dir = $u_sort_param = $s_limit_days = '';
+
+				$template->assign_var('U_MARK_FORUMS', ($user->data['is_registered'] || $config['load_anon_lastread']) ? append_sid("{$phpbb_root_path}index.$phpEx", 'hash=' . generate_link_hash('global') . '&amp;mark=forums&amp;mark_time=' . time()) : '');
```
Nicofuma: maybe use U_MARK_TOPICS or U_MARK_ALL_READ instead of U_MARK_FORUMS?

phpBB/styles/prosilver/template/search_results.html
```diff
@@ -33,6 +33,7 @@ <h2 class="searchresults-title"><!-- IF SEARCH_TITLE -->{SEARCH_TITLE}<!-- ELSE
 		<!-- EVENT search_results_searchbox_after -->
 
 		<div class="pagination">
+			<!-- IF U_MARK_FORUMS --><a href="{U_MARK_FORUMS}" class="mark-read" accesskey="m">{L_MARK_ALL_READ}</a> &bull;<!-- ENDIF -->
```
Nicofuma: should be added to subsilver2 too

phpBB/search.php
Zoddo: Done

Nicofuma:
and what about subsilver2?

Zoddo:
@Nicofuma : I didn't really see where add the link (because of the structure of subsilver2)

Nicofuma:
@hanakin @cyberalien any idea?

Nicofuma:
@nickvergessen or just don't add it to subsilver2?

hanakin:
I do not see why it could not be there but i do not know if it should be part of 3.1 as it may have some unforseen side effects on currently released themes. Ill let @cyberalien chime in on that

marc1706:
@cyberalien any comments on this?

cyberalien:
Tested it, works correctly.

As for subsilver2, I think screw it. This is not an essential feature, subsilver2 is already missing some stuff and subsilver2 is going to be removed anyway.

Nicofuma:
Ok, thanks for your contibution @Zoddo and for your comments the others

PR #3874, opened 2015-08-31
[ticket/14134] Send warning notification PM in user's language.

lavigor:
PHPBB3-14134

### DIFF
--- a/phpBB/includes/mcp/mcp_warn.php
+++ b/phpBB/includes/mcp/mcp_warn.php
@@ -530,7 +530,6 @@ function add_warning($user_row, $warning, $send_pm = true, $post_id = 0)
 		include_once($phpbb_root_path . 'includes/functions_privmsgs.' . $phpEx);
 		include_once($phpbb_root_path . 'includes/message_parser.' . $phpEx);
 
-		// Okay we need to get the reason from the warned users language
 		// Attempt to translate warning to language of user being warned if user's language differs from issuer's language
 		if ($user_row['user_lang'] != $user->lang_name)
 		{
--- a/phpBB/includes/mcp/mcp_warn.php
+++ b/phpBB/includes/mcp/mcp_warn.php
@@ -530,12 +530,30 @@ function add_warning($user_row, $warning, $send_pm = true, $post_id = 0)
 		include_once($phpbb_root_path . 'includes/functions_privmsgs.' . $phpEx);
 		include_once($phpbb_root_path . 'includes/message_parser.' . $phpEx);
 
-		$user_row['user_lang'] = (file_exists($phpbb_root_path . 'language/' . $user_row['user_lang'] . "/mcp.$phpEx")) ? $user_row['user_lang'] : $config['default_lang'];
-		include($phpbb_root_path . 'language/' . basename($user_row['user_lang']) . "/mcp.$phpEx");
+		// Okay we need to get the reason from the warned users language
+		// Only load up the language pack if the language is different to the current one
+		if ($user_row['user_lang'] != $user->lang_name)
+		{
+			// Load up the language pack
+			$lang = array();
+
+			$user_row['user_lang'] = (file_exists($phpbb_root_path . 'language/' . basename($user_row['user_lang']) . "/mcp." . $phpEx)) ? $user_row['user_lang'] : $config['default_lang'];
+			include($phpbb_root_path . 'language/' . basename($user_row['user_lang']) . "/mcp." . $phpEx);
+
+			$warn_pm_subject = $lang['WARNING_PM_SUBJECT'];
+			$warn_pm_body = sprintf($lang['WARNING_PM_BODY'], $warning);
+
+			unset($lang); // Free memory
+		}
+		else
+		{
+			$warn_pm_subject = $user->lang('WARNING_PM_SUBJECT');
+			$warn_pm_body = $user->lang('WARNING_PM_BODY', $warning);
+		}
 
 		$message_parser = new parse_message();
 
-		$message_parser->message = $user->lang('WARNING_PM_BODY', $warning);
+		$message_parser->message = $warn_pm_body;
 		$message_parser->parse(true, true, true, false, false, true, true);
 
 		$pm_data = array(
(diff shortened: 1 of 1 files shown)


Nicofuma:
I wonder if this shouldn't use the notification system instead of the PM... anyway that wouldn't be for 3.1

phpBB/includes/mcp/mcp_warn.php
```diff
@@ -530,12 +530,30 @@ function add_warning($user_row, $warning, $send_pm = true, $post_id = 0)
...
-		$user_row['user_lang'] = (file_exists($phpbb_root_path . 'language/' . $user_row['user_lang'] . "/mcp.$phpEx")) ? $user_row['user_lang'] : $config['default_lang'];
-		include($phpbb_root_path . 'language/' . basename($user_row['user_lang']) . "/mcp.$phpEx");
+		// Okay we need to get the reason from the warned users language
+		// Only load up the language pack if the language is different to the current one
+		if ($user_row['user_lang'] != $user->lang_name)
+		{
+			// Load up the language pack
+			$lang = array();
+
+			$user_row['user_lang'] = (file_exists($phpbb_root_path . 'language/' . $user_row['user_lang'] . "/mcp.$phpEx")) ? $user_row['user_lang'] : $config['default_lang'];
```
Nicofuma: don't mix the notations please (avoid $var . 'text' . "$vartext$var" and choose one notation)

you can also move this check to the if.

phpBB/includes/mcp/mcp_warn.php
```diff
@@ -530,12 +530,30 @@ function add_warning($user_row, $warning, $send_pm = true, $post_id = 0)
...
+			include($phpbb_root_path . 'language/' . basename($user_row['user_lang']) . "/mcp.$phpEx");
```
Nicofuma: use exactly the same construction as in the check, use  `$user_row['user_lang']` in both cases

phpBB/includes/mcp/mcp_warn.php
lavigor: I haven't changed the previous code for this line. Do I need to make this adjustment?

phpBB/includes/mcp/mcp_warn.php
lavigor: > use  $user_row['user_lang']  in both cases

The same code is [here](https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/includes/mcp/mcp_queue.php#L1253).
I treated it as an example.

phpBB/includes/mcp/mcp_warn.php
```diff
@@ -530,12 +530,30 @@ function add_warning($user_row, $warning, $send_pm = true, $post_id = 0)
...
-		$user_row['user_lang'] = (file_exists($phpbb_root_path . 'language/' . $user_row['user_lang'] . "/mcp.$phpEx")) ? $user_row['user_lang'] : $config['default_lang'];
-		include($phpbb_root_path . 'language/' . basename($user_row['user_lang']) . "/mcp.$phpEx");
+		// Okay we need to get the reason from the warned users language
+		// Only load up the language pack if the language is different to the current one
+		if ($user_row['user_lang'] != $user->lang_name)
+		{
+			// Load up the language pack
+			$lang = array();
+
+			$user_row['user_lang'] = (file_exists($phpbb_root_path . 'language/' . basename($user_row['user_lang']) . "/mcp." . $phpEx)) ? $user_row['user_lang'] : $config['default_lang'];
```
Nicofuma: I think you should move the file_exists check to the upper if

lavigor: I thought about that, but the current approach is not the same.
No file in user's language => get the file in default language.
Default language may be different from the language of moderator who issued the warning.
That's why this line is not inside the if.

phpBB/includes/mcp/mcp_warn.php
```diff
@@ -530,12 +530,30 @@ function add_warning($user_row, $warning, $send_pm = true, $post_id = 0)
 		include_once($phpbb_root_path . 'includes/functions_privmsgs.' . $phpEx);
 		include_once($phpbb_root_path . 'includes/message_parser.' . $phpEx);
 
-		$user_row['user_lang'] = (file_exists($phpbb_root_path . 'language/' . $user_row['user_lang'] . "/mcp.$phpEx")) ? $user_row['user_lang'] : $config['default_lang'];
-		include($phpbb_root_path . 'language/' . basename($user_row['user_lang']) . "/mcp.$phpEx");
+		// Okay we need to get the reason from the warned users language
+		// Only load up the language pack if the language is different to the current one
```
marc1706: Please use a more objective comment. I'd recommend this:
```
Attempt to translate warning to language of user being warned if user's language differs from issuer's language
```

phpBB/includes/mcp/mcp_warn.php
```diff
@@ -530,12 +530,30 @@ function add_warning($user_row, $warning, $send_pm = true, $post_id = 0)
...
+		if ($user_row['user_lang'] != $user->lang_name)
+		{
+			// Load up the language pack
```
marc1706: Unneeded comment and actually wrong as you're only instantiating the array here.

phpBB/includes/mcp/mcp_warn.php
```diff
@@ -530,12 +530,30 @@ function add_warning($user_row, $warning, $send_pm = true, $post_id = 0)
...
+			$lang = array();
+
+			$user_row['user_lang'] = (file_exists($phpbb_root_path . 'language/' . basename($user_row['user_lang']) . "/mcp." . $phpEx)) ? $user_row['user_lang'] : $config['default_lang'];
+			include($phpbb_root_path . 'language/' . basename($user_row['user_lang']) . "/mcp." . $phpEx);
+
+			$warn_pm_subject = $lang['WARNING_PM_SUBJECT'];
+			$warn_pm_body = sprintf($lang['WARNING_PM_BODY'], $warning);
+
+			unset($lang); // Free memory
```
marc1706: Please also remove this comment. What this line of code does is very clear and there is no need in stating it with a comment.

lavigor:
@marc1706 mentioned comments are not mine.
It's just a copypaste from [existing core code](https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/includes/mcp/mcp_queue.php#L1265-L1279).

I guess comment adjustments require a separate issue and PR - `mcp_warn.php` is not the only place where fixes should be made.

marc1706:
You're adding those comments here though. I don't care about changing old comments but I do care about adding new ones. Even if it's just a copy/paste from somewhere else, we shouldn't continue using improper comments just because we have used them before.

lavigor:
@marc1706 Done.

marc1706:
The "Okay ..." part is still in there

Nicofuma:
@lavigor bump. Please remove the line starting with "Okay"

lavigor:
@Nicofuma, done.

PR #3876, opened 2015-09-01
[ticket/14138] Use span tags instead of abbr

VSEphpbb:
https://tracker.phpbb.com/browse/PHPBB3-14138

PHPBB3-14138


bantu:
This removes the underlining which probably was the reason why abbr was used in the first place.

marc1706:
We never used underlining for these elements. Most browsers didn't even show the underline until recently. The fix should be fine like this.

Nicofuma:
it looks good to me and that's what was discussed in the other PRs

PR #3879, opened 2015-09-02
[Ticket/14139] Stop styling with IDs

hanakin:
Its considered bad practice and adds to specificity nightmare when combining styling of IDs and classes. As such we should add duplicate classes to all elements that are currently being styled with IDs in order to move the style to a class to improve specificity and transparency of the theme.

Ticket: https://tracker.phpbb.com/browse/PHPBB3-14139

PHPBB3-14139

### DIFF
--- a/phpBB/styles/prosilver/template/mcp_header.html
+++ b/phpBB/styles/prosilver/template/mcp_header.html
@@ -21,14 +21,14 @@
 
 	<div style="width: 100%;">
 
-	<div id="cp-menu">
-		<div id="navigation" role="navigation">
+	<div id="cp-menu" class="cp-menu">
+		<div id="navigation" class="navigation" role="navigation">
 			<ul>
 			<!-- BEGIN l_block1 -->
 			<!-- IF l_block1.S_SELECTED -->
 				<!-- BEGIN l_block2 -->
 				<!-- IF l_block1.l_block2.S_SELECTED -->
-				<li id="active-subsection"><a href="{l_block1.l_block2.U_TITLE}"><span>{l_block1.l_block2.L_TITLE}<!-- IF l_block1.l_block2.ADD_ITEM --> ({l_block1.l_block2.ADD_ITEM})<!-- ENDIF --></span></a></li>
+				<li id="active-subsection" class="active-subsection"><a href="{l_block1.l_block2.U_TITLE}"><span>{l_block1.l_block2.L_TITLE}<!-- IF l_block1.l_block2.ADD_ITEM --> ({l_block1.l_block2.ADD_ITEM})<!-- ENDIF --></span></a></li>
 				<!-- ELSE -->
 				<li><a href="{l_block1.l_block2.U_TITLE}"><span>{l_block1.l_block2.L_TITLE}<!-- IF l_block1.l_block2.ADD_ITEM --> ({l_block1.l_block2.ADD_ITEM})<!-- ENDIF --></span></a></li>
 				<!-- ENDIF -->
@@ -39,7 +39,7 @@
 		</div>
 	</div>
 
-	<div id="cp-main" class="mcp-main panel-container">
+	<div id="cp-main" class="cp-main mcp-main panel-container">
 		<!-- IF MESSAGE -->
 		<div class="content">
 			<h2 class="message-title">{L_MESSAGE}</h2>
--- a/phpBB/styles/prosilver/template/mcp_post.html
+++ b/phpBB/styles/prosilver/template/mcp_post.html
@@ -106,7 +106,7 @@
 			</p>
 		<!-- ENDIF -->
 
-		<div class="content" id="post_details">
+		<div id="post_details" class="content post_details">
 			{POST_PREVIEW}
 		</div>
 
--- a/phpBB/styles/prosilver/template/mcp_topic.html
+++ b/phpBB/styles/prosilver/template/mcp_topic.html
@@ -11,7 +11,7 @@
 	<!-- DEFINE $SHOW_PANEL = 'display-panel' -->
 <!-- ENDIF -->
 
-<div id="minitabs" class="sub-panels" data-show-panel="{$SHOW_PANEL}" role="tablist">
+<div id="minitabs" class="minitabs sub-panels" data-show-panel="{$SHOW_PANEL}" role="tablist">
 	<ul>
 		<li id="display-panel-tab" class="tab<!-- IF not S_MERGE_VIEW --> activetab<!-- ENDIF -->">
 			<a href="#minitabs" data-subpanel="display-panel" role="tab" aria-controls="display-panel">{L_DISPLAY_OPTIONS}</a>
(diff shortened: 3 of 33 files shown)


hanakin:
!set Styles

phpBB/styles/prosilver/template/mcp_post.html
```diff
@@ -106,7 +106,7 @@
 			</p>
 		<!-- ENDIF -->
 
-		<div class="content" id="post_details">
+		<div id="post_details" class="content post_details" >
```
vinny: extra space

Crizz0:
Aren't we removing a tool for specifity and put additional not needed code to the template with an ID and class. 

I don't get the point of "bad practice", too. 

What will change for the developer and style authors with that?

hanakin:
@Crizz0

the use of styling on an ID is bad practice. It should never be used. Its overly specific and uses more resources to render than classes. Its also limited in scope. That is why I added classes for all cases where IDs where used in the CSS. I left the IDs in for BC & JS. As some of these are still incorrectly being used as JS hooks.

What this will change is the need for over specificity in our css. This helps prevent us from overly chaining selectors together (exp: .rtl ul.topiclist dd.info > span). Instead we can use a simple class to target pretty much anything. It also means we no longer need CSS 1.0 spec for ul.class or a.class which is also extremely bad practice as its also overly specific.

This is a small step forward there are still a lot of specificity issues.

here is a screen shot of the results for the current area51 board index scan i performed for reference:

screenshot 2015-09-07 13 51 08

you can see there are a total of 170 selectors used to render the page and all the IDs are at the top when organized be specificity.

you can check this for yourself using the http://cssdig.com/ browser extension.

phpBB/styles/prosilver/template/mcp_header.html
```diff
@@ -19,31 +19,31 @@
...
-				<!-- ELSE -->
-				<li><a href="{l_block1.l_block2.U_TITLE}"><span>{l_block1.l_block2.L_TITLE}<!-- IF l_block1.l_block2.ADD_ITEM --> ({l_block1.l_block2.ADD_ITEM})<!-- ENDIF --></span></a></li>
-				<!-- ENDIF -->
-				<!-- END l_block2 -->
-			<!-- ENDIF -->
-			<!-- END l_block1 -->
-			</ul>
-		</div>
-	</div>
+			<div id="cp-menu" class="cp-menu">
```
marc1706: Same as previously mentioned by nickvergessen. Please do not indent code unless it's really necessary.

phpBB/styles/prosilver/template/ucp_header.html
```diff
@@ -17,85 +17,85 @@ <h2 class="ucp-title">{L_UCP}</h2>
...
-
-		<!-- IF S_PRIVMSGS -->
-			<!-- BEGIN t_block2 -->
-				<!-- IF S_PRIVMSGS and not t_block2.S_LAST_ROW -->
-				<ul>
-					<!-- IF t_block2.S_SELECTED -->
-						<li id="active-subsection"><a href="{t_block2.U_TITLE}"><span>{t_block2.L_TITLE}</span></a></li>
-					<!-- ELSE -->
-						<li><a href="{t_block2.U_TITLE}"><span>{t_block2.L_TITLE}</span></a></li>
+		<div style="width: 100%;">
```
marc1706: See above comment for indentation.

phpBB/styles/prosilver/template/mcp_topic.html
```diff
@@ -87,12 +87,12 @@
 <div class="panel">
 	<div class="inner">
 
-	<h3 id="review">
+	<h3 id="review" class="review">
```
marc1706: The margin-top is overriden by the rules for .panel h3 in colours.css

phpBB/styles/prosilver/template/posting_attach_body.html
```diff
@@ -32,8 +32,8 @@
 						<th class="attach-status">{L_PLUPLOAD_STATUS}</th>
 					</tr>
 				</thead>
-				<tbody class="responsive-skip-empty" id="file-list">
-					<tr class="attach-row" id="attach-row-tpl">
+				<tbody class="responsive-skip-empty file-list" id="file-list">
+					<tr class="attach-row attach-row-tpl" id="attach-row-tpl">
```
marc1706: ```display: none``` for this is not overriden by plupload right now. Attachments do not show after upload in the filelist. They properly show after clicking on preview.

marc1706:
The loading_indicator is broken on normal pages that do not use simple_footer.html. You'll have to also add the class here:
https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/assets/javascript/core.js#L30

phpBB/styles/prosilver/theme/colours.css
```diff
@@ -942,12 +942,12 @@ dl.mini dt {
 }
 
 /* Avatar gallery */
-#gallery label {
+,gallery label {
```
marc1706: Typo, needs to be a dot not a comma.

marc1706:
The rest looks fine to me.

hanakin:
not sure what is issue with loading indicator? there is no loading indicator when include anywhere but simple by default?

marc1706:
@hanakin this will fix the issue with loading indicator: https://gist.github.com/marc1706/0384b9b7040ad5b21e55

It's created using jQuery in core.js which is why it's only in simple_footer.html.

marc1706:
Could you fix your last two commit messages? This one has the incorrect ticket ID https://github.com/hanakin/phpbb/commit/945bc3f7c06171032f4fee7b6d11be76c67f8930
and the second one has two leading square brackets instead of one:
https://github.com/hanakin/phpbb/commit/db1d292617344fbef6d925c76cf73da7e1b305bf

hanakin:
hmm sware i fixed that will do

marc1706:
I'd merge this once you've reverted the indentation that is unrelated to this PR. :)

phpBB/styles/prosilver/template/mcp_header.html
hanakin: @marc1706 @nickvergessen this is not really a vaild argument anymore as recent changes to github compound highlight on changes. You can see the spacing added as well as any other changes to the code. All indentation made is to fix inconsistent spacing that exists for cleaner codebase.

marc1706: At least split this into two separate commits so one will be able to determine that the indentation change is unrelated to the change of IDs to classes.

nickvergessen: > @marc1706 @nickvergessen this is not really a vaild argument anymore as recent changes to github compound highlight on changes.

@hanakin this is not about github, but about the .diff file we hand out to users when they try to update their custom styles.

hanakin: hmm not sure i understand, are you meaning merge conflicts for patches? either way the fact that the indentation is all over the place and inconsistent is a constant problem when making changes and needs to be fixed I guess we need a scrub PR to fix all this. Thats sort of a daunting task though with template inclusion

hanakin:
All indentation should be undone...

PR #3880, opened 2015-09-04
[ticket/12911] Add SVG Icons

hanakin:
convert all the forum/topic imageset icons into svgs

for now this requires the Font-Awesome PR: https://github.com/phpbb/phpbb/pull/3851

Old PR: https://github.com/phpbb/phpbb/pull/2794
Ticket: https://tracker.phpbb.com/browse/PHPBB3-12911
PHPBB3-12911


Louis7777:
Having just an imageset.png sprite was so fine. Imo I'd rather have to edit just one imageset.psd. Meh...

Louis7777:
However, the clients won't download just one file right? Whereas with the sprite it would just be one HTTP request for an imageset.png file weighing 5,02 KB.

hanakin:
it can still be served as a single file, svg is just html so it can combined in a single file for one request. Then you choose how to reference either through css just like a normal sprite or svg use tags. Eventually even through a simple image tag but its not fully supported yet

hanakin:
!set WIP

hanakin:
!set Styles

phpBB/styles/prosilver/template/forumlist_body.html
```diff
@@ -11,98 +11,125 @@
...
-						<dt><div class="list-inner"><!-- IF forumrow.S_IS_CAT --><a href="{forumrow.U_VIEWFORUM}">{forumrow.FORUM_NAME}</a><!-- ELSE -->{L_FORUM}<!-- ENDIF --></div></dt>
-						<dd class="topics">{L_TOPICS}</dd>
-						<dd class="posts">{L_POSTS}</dd>
-						<dd class="lastpost"><span>{L_LAST_POST}</span></dd>
-					</dl>
-					<!-- EVENT forumlist_body_category_header_row_append -->
-				</li>
-			</ul>
-			<ul class="topiclist forums">
+				<ul class="topiclist">
```
nickvergessen: please avoid retabbing like this. It will make any diff unusable

gooof: https://github.com/phpbb/phpbb/pull/3880/files?w=0 gives a better diff

nickvergessen: I know the ?w=* thing,
but that is not what style authors will get when trying to update their style.

phpBB/styles/prosilver/theme/stylesheet.css
```diff
@@ -7,6 +7,7 @@
     --------------------------------------------------------------
 */
 
+@import url("https://maxcdn.bootstrapcdn.com/font-awesome/4.4.0/css/font-awesome.min.css");
```
nickvergessen: Should be added to our assets instead, so it also works if your forum is not connected to the internet or e.g. CDN is disabled?

phpBB/styles/prosilver/template/forumlist_body.html
gooof: Thats true! ;)

Nicofuma:
these changes are awesome but.... I'm not sure we want that in 3.2 because it will break a lot of things....
- Every style will be broken because the authors can't just drag and drop their new icons
- The extensions displayed in the navbar and using the icons defined in `colours.css` will be broken

In order to merge these PRs in 3.2, I really think we need some kind of BC layer because we don't want to break every extensions and we don't want to force every style author to start over again :/

hanakin:
@Nicofuma its one of the most requested changes which is why I am attempting it. But this is still experimental I do not think that there is an easy way to replace these easily with a sprite no matter how we approach it. At least not one a novice would be able to work with. It would require some sort of build process unless we created something in the core to handle it.

hanakin:
@Nicofuma tested the BC and this should not break any custom icons in current themes as inherency will trump these. I am still working on labeling all the icons properly in the svg that way for those who want to modify the color can do so easily in the code.

hanakin:
!set WIP

Nicofuma:
ok thanks

hanakin:
ok ignoring the font-awesome changes everything seems to be good as far as BC goes and the changes to the svg should make editing it via code a lot easier. As far as extensions go this should be an improvement as the imageset is no a physical element to work with as well.

hanakin:
!unset WIP

hanakin:
@callumacrae @cyberalien thoughts/review please?

cyberalien:
This is really cool.

Ideally I'd love this to be applied to all icons, including FontAwesome glyphs. FA looks very blurry and inconsistent on different browsers and operating systems, but SVG icons are sharp and consistent.

cyberalien:
Icons look great on white background, however background in forum row isn't white. I think icons should have white circle background similar to old images.

Another issue is icons are not vertically centered. I suggest to absolutely position it in left center and add padding-left for content.

svg

cyberalien:
Another issue is when forum description is long, it goes below icon:

svg2

It can also be fixed by absolutely positioning icon and adding padding-left for content.

hanakin:
Hmm bg on icon did not look good. I actually prefer the open look I think we might play with the colors a bit first.

As for the center issue the icon has always been top aligned I think with the layout of pro silver it fits better as top aligned if the forums were more grid style then rows it might not look better with center but I will play around with it

cyberalien:
No, it wasn't top aligned. Icon has always been a background image aligned to center:

    dl.icon {
        background-position: 10px 50%;
        background-repeat: no-repeat;
    }

hanakin:
hmm guess i never noticed I will look into it

VSEphpbb:
> Ideally I'd love this to be applied to all icons, including FontAwesome glyphs.

Especially now that @callumacrae popped up to say that font icons are passé. http://area51.phpbb.com/phpBB/viewtopic.php?p=283781#p283771

callumacrae:
sorryyyyy

hanakin:
@VSEphpbb @callumacrae to avoid confusion I replied to that on area51 https://area51.phpbb.com/phpBB/viewtopic.php?f=81&t=45671&start=50

hanakin:
!set WIP

hanakin:
!unset WIP

hanakin:
This should just need color overhaul and then testing against https://github.com/phpbb/phpbb/pull/4014

hanakin:
!set WIP

PR #3881, opened 2015-09-04
[ticket/14142] Remove unused ignore_configs from avatar drivers

VSEphpbb:
The `$ignore_config` appears to never be used in any avatar driver. It is only passed to the upload driver, but not used in it. None of the other drivers even have an argument for it.

PHPBB3-14142


nickvergessen:
@marc1706

marc1706:
I think this was left from a previous change to the avatar drivers, maybe even before the merge of the avatar PR.

PR #3882, opened 2015-09-05
[ticket/10824] Make styles/languages use composer.json files

nickvergessen:
https://tracker.phpbb.com/browse/PHPBB3-10824


phpBB/includes/acp/acp_styles.php
```diff
@@ -756,22 +756,26 @@ protected function find_available($all)
...
-			if ($cfg === false)
+
+			try
+			{
+				$style_data = $this->read_style_composer_file($dir);
+			}
+			catch (\DomainException $e)
 			{
-				// Invalid style.cfg
+				// Invalid composer.json
```
Nicofuma: why do we skip the exception? if the composer.json is invalid the following checks may error

phpBB/styles/prosilver/style.cfg
```diff
@@ -1,32 +0,0 @@
...
-#
-
-# General Information about this style
-name = prosilver
-copyright = © phpBB Limited, 2007
-style_version = 3.2.0-a1-dev
-phpbb_version = 3.2.0-a1-dev
-
-# Defining a different template bitfield
-# template_bitfield = lNg=
```
Nicofuma: please add it to the prosilver composer.json file (commented)

phpBB/phpbb/db/migration/data/v310/style_update_p1.php
```diff
@@ -69,10 +69,11 @@ public function styles_update()
 		$skip_dirs = array('.', '..', 'prosilver');
 		foreach ($iterator as $fileinfo)
 		{
-			if ($fileinfo->isDir() && !in_array($fileinfo->getFilename(), $skip_dirs) && file_exists($fileinfo->getPathname() . '/style.cfg'))
+			if ($fileinfo->isDir() && !in_array($fileinfo->getFilename(), $skip_dirs) && file_exists($fileinfo->getPathname() . '/composer.json'))
```
Nicofuma: what will happen in case of update from 3.0.13 if their the installed styles are not 3.2 compatible?

phpBB/styles/prosilver/style.cfg
nickvergessen: there are no comments in json?

phpBB/includes/acp/acp_styles.php
nickvergessen: continue means go to the next entry in the array ;)

phpBB/phpbb/db/migration/data/v310/style_update_p1.php
nickvergessen: the style will not be handled, just like in 3.1?

phpBB/styles/prosilver/style.cfg
nickvergessen: Also I don't get what this value is for?

Nicofuma: right... :/ IIRC it is used to determine what bbcodes should be overridden by the style bbcode.html file

phpBB/phpbb/db/migration/data/v310/style_update_p1.php
Nicofuma: ok

phpBB/includes/acp/acp_styles.php
Nicofuma: I'm tired... but we don't display any error in this case? should be fixed one day...

phpBB/language/en/composer.json
```diff
@@ -0,0 +1,27 @@
+{
+	"name": "en",
```
nickvergessen: @naderman what would be a good naming pattern for languages and styles?

phpBB/language/en/composer.json
```diff
@@ -0,0 +1,27 @@
...
+	"description": "phpBB Forum Software default style",
```
nickvergessen: needs fixing

phpBB/includes/acp/acp_styles.php
```diff
@@ -756,22 +756,26 @@ protected function find_available($all)
...
 
 			// Style should be available for installation
-			$parent = $cfg['parent'];
+			$parent = $style_data['extra']['parent-style'];
 			$style = array(
 				'style_id'			=> 0,
-				'style_name'		=> $cfg['name'],
-				'style_copyright'	=> $cfg['copyright'],
+				'style_name'		=> $style_data['name'],
+				'style_copyright'	=> $style_data['license'],
```
michaelcullum: The old copyright is not equivalent to the licence field in composer. The old copyright field detailed who had the copyright, this details the license (e.g. MIT)

phpBB/language/en/composer.json
```diff
@@ -0,0 +1,27 @@
...
+	"type": "phpbb-language",
+	"version": "3.2.0-a1-dev",
+	"homepage": "https://www.phpbb.com",
+	"license": "GPL-2.0",
+	"authors": [
+		{
+			"name": "phpBB Limited",
+			"email": "operations@phpbb.com",
+			"homepage": "https://www.phpbb.com/go/authors"
```
michaelcullum: phpbb.com homepage

phpBB/language/en/composer.json
```diff
@@ -0,0 +1,27 @@
+{
+	"name": "en",
+	"description": "phpBB Forum Software default style",
+	"type": "phpbb-language",
+	"version": "3.2.0-a1-dev",
```
michaelcullum: Needs adding to the release todo list

phpBB/language/en/composer.json
michaelcullum: Always need a vendor name. `phpbb/en-language-pack` Naming here doesn't really matter (and I don't think that requiring specific naming conventions would be a good idea to be honest).

phpBB/styles/prosilver/composer.json
```diff
@@ -0,0 +1,29 @@
+{
+	"name": "prosilver",
```
michaelcullum: `phpbb/prosilver`

phpBB/styles/prosilver/composer.json
```diff
@@ -0,0 +1,29 @@
...
+	"description": "phpBB Forum Software default style",
+	"type": "phpbb-style",
+	"version": "3.2.0-a1-dev",
```
michaelcullum: Release todo list

phpBB/styles/prosilver/composer.json
```diff
@@ -0,0 +1,29 @@
...
+	"homepage": "https://www.phpbb.com",
+	"license": "GPL-2.0",
+	"authors": [
+		{
+			"name": "phpBB Limited",
+			"email": "operations@phpbb.com",
+			"homepage": "https://www.phpbb.com/go/authors"
```
michaelcullum: .com homepage again?

michaelcullum:
Main composer.json needs to provide `phpbb/prosilver` and `phpbb/en-language-pack`'s `self-version`?

phpBB/language/en/composer.json
nickvergessen: Well we want to be able to know the name, so you don't have to look it up first when you install a forum + language via composer, so yes  a naming convention makes sense

phpBB/includes/acp/acp_styles.php
nickvergessen: Yeah not sure why we  had that thou, we don't do that for exts either and it's only displayed in the acp, so i think removing should be fine?

phpBB/language/en/composer.json
nickvergessen: This is copied from phpBB/composer.json

phpBB/styles/prosilver/composer.json
nickvergessen: Already is, because  the style.cfg needs modifying as well

phpBB/includes/acp/acp_styles.php
michaelcullum: +1

phpBB/language/en/composer.json
michaelcullum: True. But you'll never know the vendor without looking it up.

Also, we need to change phpBB

Nicofuma: or just set the name to `<vendor>/<iso>`?

Nicofuma:
I guess it will impact titania? @prototech ? (just keep it in mind)

Nicofuma:
the composer.json file is still executable by the way

marc1706:
Bump :)

nickvergessen:
Feel free to take over, won't have time in the near future

marc1706:
Will take this over

PR #3885, opened 2015-09-06
[ticket/14129] Cache custom extensions autoloaders

rxu:
<a href="https://tracker.phpbb.com/browse/PHPBB3-14129">PHPBB3-14129</a>.

### DIFF
--- a/phpBB/includes/functions.php
+++ b/phpBB/includes/functions.php
@@ -27,26 +27,51 @@ if (!defined('IN_PHPBB'))
 */
 function phpbb_load_extensions_autoloaders($phpbb_root_path)
 {
-	$iterator = new \RecursiveIteratorIterator(
-		new \phpbb\recursive_dot_prefix_filter_iterator(
-			new \RecursiveDirectoryIterator(
-				$phpbb_root_path . 'ext/',
-				\FilesystemIterator::SKIP_DOTS | \FilesystemIterator::FOLLOW_SYMLINKS
-			)
-		),
-		\RecursiveIteratorIterator::SELF_FIRST
-	);
-	$iterator->setMaxDepth(2);
+	global $phpbb_config_php_file;
 
-	foreach ($iterator as $file_info)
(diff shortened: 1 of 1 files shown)


rxu:
!set WIP

rxu:
!unset WIP

Elsensee:
Is this really necessary? Does this really increase the performance if many extensions are installed?

rxu:
Yes, it is. Without caching, it takes about 10% of page generation time as per my measurements.

Nicofuma:
I don't like hacking the cache like that.... instead I think I'd prefer to write a cache/<env>/autoload.php file which will load all these autoloaders:

```
<?php

require '<root>/ext/my/ext1/vendor/autoload.php'
....
```

This file will be written (and then loaded) here instead of loading the individual files https://github.com/phpbb/phpbb/blob/master/phpBB/phpbb/di/container_builder.php#L380-L385 and also loaded here https://github.com/phpbb/phpbb/blob/master/phpBB/phpbb/di/container_builder.php#L132 if `$this->use_extensions` is `true`. I think I'd prefer something like that.

rxu:
Anyway, I guess this PR should be closed as it's made against 3.1.x.

Nicofuma:
didn't notice that :p

for 3.1... I don't know. @nickvergessen what do you think? And for 3.2, you know what I think ^^

rxu:
!set WIP

marc1706:
Performance "hack"/improvement == new feature --> 3.2

rxu:
I have <a href="https://gist.github.com/rxu/af652c7ffbb79c05e9dc">another approach</a> for this, probably hackish again (for 3.1 only though, as in 3.2 the relevant code has been changed fully).

marc1706:
@nickvergessen your call I guess

Nicofuma:
replaced by https://github.com/phpbb/phpbb/pull/4142

rxu:
Actually #4142 doesn't replace this PR as the current one was for 3.1 :wink:

PR #3888, opened 2015-09-06
[ticket/14146] Add viewtopic_body_post_subject_before event

Senky:
[PHPBB3-14146](https://tracker.phpbb.com/browse/PHPBB3-14146)


Nicofuma:
please add this event to subsilver2 too

Senky:
I think it's not needed there. In subsilver2, `viewtopic_body_postrow_post_details_before` event takes it's place.

Nicofuma:
I think it's preferable to mirror them. This way an ext author can use the same event (and why not the same code) for bot subsilver and prosilver

PR #3895, opened 2015-09-11
[ticket/14154] Add extension composer name as tooltip

Zoddo:
[PHPBB3-14154](https://tracker.phpbb.com/browse/PHPBB3-14154)

### DIFF
--- a/phpBB/adm/style/acp_ext_list.html
+++ b/phpBB/adm/style/acp_ext_list.html
@@ -48,7 +48,7 @@
 		</tr>
 		<!-- BEGIN enabled -->
 		<tr class="ext_enabled">
-			<td><strong>{enabled.META_DISPLAY_NAME}</strong></td>
+			<td><strong title="{enabled.NAME}">{enabled.META_DISPLAY_NAME}</strong></td>
 			<td style="text-align: center;">
 				<!-- IF enabled.S_VERSIONCHECK -->
 				<strong <!-- IF enabled.S_UP_TO_DATE -->style="color: #228822;"<!-- ELSE -->style="color: #BC2A4D;"<!-- ENDIF -->>{enabled.META_VERSION}</strong>
@@ -73,7 +73,7 @@
 		</tr>
 		<!-- BEGIN disabled -->
 		<tr class="ext_disabled">
-			<td><strong>{disabled.META_DISPLAY_NAME}</strong></td>
+			<td><strong title="{disabled.NAME}">{disabled.META_DISPLAY_NAME}</strong></td>
 			<td style="text-align: center;">
 				<!-- IF disabled.S_VERSIONCHECK -->
 				<strong <!-- IF disabled.S_UP_TO_DATE -->style="color: #228822;"<!-- ELSE -->style="color: #BC2A4D;"<!-- ENDIF -->>{disabled.META_VERSION}</strong>
--- a/phpBB/includes/acp/acp_extensions.php
+++ b/phpBB/includes/acp/acp_extensions.php
@@ -369,6 +369,7 @@ class acp_extensions
 
 		foreach ($enabled_extension_meta_data as $name => $block_vars)
 		{
+			$block_vars['NAME'] = $name;
 			$block_vars['U_DETAILS'] = $this->u_action . '&amp;action=details&amp;ext_name=' . urlencode($name);
 
 			$this->template->assign_block_vars('enabled', $block_vars);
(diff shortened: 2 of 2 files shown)


Nicofuma:
I'd rather have it with a float:right or with brackets

Zoddo:
@Nicofuma : Oh, yes, I've forgot the brackets!

VSEphpbb:
I don't like this. This is what the **Details** link is for if a user really must know. This just adds clutter and inconsistency. It seems to be addressing a problem that doesn't really exist.

dragosvr92:
VSE, i have fully explained why this is needed. After you uninstall an extension, you need to know where to browse to delete the extension files. Why should we have to check the details to find out where it is installed? But, if it is indeed considered disturbing, maybe it would be best to have the clean name on all extensions while we hover on their Extension name, as i suggested in the ticket. There is a topic and a ticket where we can discuss this.

VSEphpbb:
Having the clean name everywhere is a disaster. Nobody has ever complained of this, it's just not been an issue. The details link, 1 click, has all the info you need.

dragosvr92:
Its not everywhere. It is somewhere where people need it. People also never complained publicly about the difficulty to follow an extension, without a hover highlighter on the extensions list. Besides me. And you wrote the patch for it yesterday. 

When you are set to remove an extension, you have two steps in mind; disable it, then you need to know where it is stored and go delete its files. Why add the Check details page to see where it is stored, in between? Checking the Details page for the path can be frustrating. 

I have already suggested moving the clean name to the title twice. What you saying ? :P

VSEphpbb:
We are currently working on an extension management interface that should eliminate the need to go to FTP, so this whole topic is going to be a moot point soon. ;)

dragosvr92:
Nice. Can i have a peek in its current design? :grinning: I am guessing that the extension management would go out in 3.2 or 4.0. So, we could add this as a title hover until then. It is easy enough to merge.

VSEphpbb:
I have always complained about the row thing. Just last week in #phpbb-dev: 
> VSE: Anyone else think it'd be nice to have hover row highlighting in the ACP extensions manager?

Also, bear in mind. People don't normally remove an extension one day out of the blue. They install one. they update one. Both of which involve immediate handling of the package, so they know what it's called. Normally when somebody uninstalls an extension, it's in close proximity time-wise to when they installed (as in they realize they don't like it or have problems because of it). It's just very rare that users decide to uninstall one so far away from the time when they know what the package naming was.

VSEphpbb:
> Nice. Can i have a peek in its current design? :grinning: I am guessing that the extension management would go out in 3.2 or 4.0. So, we could add this as a title hover until then. It is easy enough to merge.

A hover/title attribute would be OK, even into the future. Yes, you will be able to see the code once PRs start popping up.

dragosvr92:
I have installed and updated many extensions i have installed. i still have no clue where many of them are stored. When you find out there is a new version, you check the details, see the path, and the download link for the update. Also, there are those people that dont update their software. 

I have about 32 extensions. I only remember some of the paths of developers im acquainted to. 

The other day i realised i had one active extension that i never used and i need to delete it. 
There were also disabled extensions that i was considering to use, but never did. 

 @Zoddo Could you please move the clean name to a title? 
Not sure how to implement it. Maybe change the <strong> with a <span> and add the title tag and set the bold in css ?

VSEphpbb:
> Not sure how to implement it. Maybe change the < strong > with a < span > and add the title tag and set the bold in css ?

No, just add the title attribute to the strong tag. Simple.

dragosvr92:
We can do that? I mean, i know we can, and it works. But, im not sure if it is html valid...?

VSEphpbb:
It is. title is a global attribute, available in any html tag.
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes

dragosvr92:
hum, ok. Thanks for letting me know. I had the impression that its not recommended to use style and id/class attributes in simple elements like < b > or < i > and we should use spans or divs for complexish things.

dragosvr92:
Btw, i love how git sends messages instantly and updates the page also instantly. It works so very smooth. I hope phpbb will inspire from gits behaviour. Its a bit annoying to have it load and refresh so much.

VSEphpbb:
> Thanks for letting me know. I had the impression that its not recommended to use style and id/class attributes

It's not styling. It's an HTML tag attribute.

Zoddo:
> @Zoddo Could you please move the clean name to a title? 

Done

dragosvr92:
Thank you @Zoddo :smile: 
@Nicofuma, would it be possible to rush this into the 3.1.6 Branch?

VSEphpbb:
> would it be possible to rush this into the 3.1.6 Branch?
No it won't.

dragosvr92:
I recently realised it has been released. I thought there was still time.

PR #3898, opened 2015-09-12
[ticket/14149] Serve concatenated/minified/compressed css

hanakin:
This is still WIP and open for discussion and feedback to be discussed here:  https://area51.phpbb.com/phpBB/viewtopic.php?f=81&t=45669

There are countless reasons to stop serving a static css file using the @import statement. But because It drastically improves performance & allows us to run diagnostics on the theme is enough in my book.

The easiest way to accomplish this based on the latest discussion is to include a build process and some sort of setting to the themes that allow for the choice of weather to use minified version or not.

With included documentation both on the wiki and .com on how to use the minification process.

The actual minification should be accomplished via gulp. This also allows us more power in the development cycle of the theme as we can start leveraging preprocessing/build tools.

Down the road we can possibly include some sort of automated php equivalent for simple magnification but for now an unobtrusive method to allow the switch to a minified workflow should work well and speed up development and easy maintainability.

Ticket : https://tracker.phpbb.com/browse/PHPBB3-14149

PHPBB3-14149


hanakin:
!set WIP

hanakin:
!set Styles

Nicofuma:
I don't think we will do that for 3.2 but more for the next version and using a more complex and complete system. But it isn't designed yet and so I think it's to early to start working on that.

VSEphpbb:
Also when we do this, we should probably have gulp be run from the pre-commit hook and/or the build script.

hanakin:
@Nicofuma this is not meant to be rushed only here for discussion/testing however from a development standpoint it should still be looked at with the highest importance. The main issue is that we should not be including an @import css file but instead a minified file upon release. So we would have a minified file and the unminified version that we have always used. Simple inclusion of a custom.css file for all modifications will accommodate most users and for those who are making custom themes for release. Documentation on how to use the development files should work quite well.

VSEphpbb:
For development, having minified CSS in use sucks. Luckily in 3.2 we have environments. So we can deliver minified in the production and testing environments, and use the unminified in the devlopment environment. That way devs can make CSS and JS tweaks and see the results immediately, in the unminified files.

All that is needed, really, is simple logic like:
`stylesheet<!-- IF PHPBB_ENVIRONMENT != 'development' -->.min<!-- ENDIF -->.css`

hanakin:
Not really as long as you are using sourcemaps, but you would only minify as needed. The big issue here is not really the minification, but more importantly the @import as its not supported by browsers. It prevents cahcing and source analysis.

As for the env where/how is this set as this is news to me.

VSEphpbb:
Environment is a constant defined in the config.php file.

hanakin:
closing this for now will reopen with new approach later

PR #3901, opened 2015-09-12
[ticket/13716] Check phpBB version constant against config version

Zoddo:
On the ACP display a warning if the version constant (files) does not
match the `phpbb_config` version (often indicating that the database updater
hasn't been run)

[PHPBB3-13716](https://tracker.phpbb.com/browse/PHPBB3-13716)

### DIFF
--- a/phpBB/adm/style/acp_update.html
+++ b/phpBB/adm/style/acp_update.html
@@ -6,11 +6,16 @@
 
 <p>{L_VERSION_CHECK_EXPLAIN}</p>
 
+<!-- IF S_UPDATE_INCOMPLETE -->
+	<div class="errorbox">
+		<p>{L_UPDATE_INCOMPLETE} {L_UPDATE_INCOMPLETE_MORE}</p>
+	</div>
+<!-- ENDIF -->
 <!-- IF S_UP_TO_DATE -->
 	<div class="successbox">
 		<p>{L_VERSION_UP_TO_DATE_ACP} - <a href="{U_VERSIONCHECK_FORCE}">{L_VERSIONCHECK_FORCE_UPDATE}</a></p>
 	</div>
-<!-- ELSE -->
+<!-- ELSEIF not S_UPDATE_INCOMPLETE -->
 	<div class="errorbox">
 		<p>{L_VERSION_NOT_UP_TO_DATE_ACP} - <a href="{U_VERSIONCHECK_FORCE}">{L_VERSIONCHECK_FORCE_UPDATE}</a></p>
 	</div>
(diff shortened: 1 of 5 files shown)


Zoddo:
It's possible that the text contains some formulation and typo errors (the english isn't that in which I'm better).

phpBB/language/en/install.php
```diff
@@ -439,6 +440,7 @@
 	'FILES_NOT_MODIFIED_EXPLAIN'	=> 'The following files are not modified and represent the original phpBB files from the version you want to update from.',
 	'FILES_UP_TO_DATE'				=> 'Already updated files',
 	'FILES_UP_TO_DATE_EXPLAIN'		=> 'The following files are already up to date and do not need to be updated.',
+	'FILES_VERSION'					=> 'Files version',
```
VSEphpbb: Files version doesn't sound right. Either "Package version" or "Package version installed"

I would also use `PACKAGE_VERSION` instead of `FILES_VERSION` here and elsewhere.

phpBB/language/en/install.php
```diff
@@ -530,6 +532,27 @@
 	'UPDATE_DATABASE_SCHEMA'		=> 'Updating database schema',
 	'UPDATE_FILES'					=> 'Update files',
 	'UPDATE_FILES_NOTICE'			=> 'Please make sure you have updated your board files too, this file is only updating your database.',
+	'UPDATE_INCOMPLETE'				=> 'Your phpBB installation has not been correctly updated.',
```
VSEphpbb: All of this here seems like a big confusing mess of information (to me, and so probably to anybody else). It's also not necessary to have installation instructions in the middle of the installation process.

All you are checking is if the db version is < the package version. If that's true, all that means is the user needs to run the database update script. So just provide a link to the update script at root/install/database_update.php.

VSEphpbb: Like instead of all that, just explain that it appears the database has not yet been updated...here's the link to do so...

Zoddo: > All of this here seems like a verbose mess of information not really needed.

Why don't write an explicative text? This can be benefit for administrators that don't have much knowledge of phpBB. Or there is some reason to not include this text?

> So just provide a link to the update script at root/install/database_update.php

Yes, but the `install` folder may have been deleted and some administrators can be novice, so I think that we need to explain to the user, the complete way to retrieve the archive and finish the update.

VSEphpbb: This will only appear to somebody who has already downloaded and installed the latest package. So re-explaining to them where and how to get the latest package is redundant.

If the user deleted their install folder, they will get an error that it couldn't be found. So all you have to explain here is that users should visit the database_update script, and if it is missing, they should download the package version from phpbb.com again, upload the install folder, and then visit the script.

Nicofuma: bump, whats up here?

phpBB/adm/style/acp_update.html
```diff
@@ -18,10 +23,21 @@
...
 <fieldset>
 	<legend></legend>
+	<!-- IF not S_UPDATE_INCOMPLETE -->
 	<dl>
 		<dt><label>{L_CURRENT_VERSION}</label></dt>
 		<dd><strong>{CURRENT_VERSION}</strong></dd>
 	</dl>
+	<!-- ELSE -->
+	<dl>
+		<dt><label>{L_FILES_VERSION}</label></dt>
```
Nicofuma: This language string is missing

phpBB/language/en/install.php
Zoddo: So delete extras informations and just keep a short text that say to run database_update?

Nicofuma: to run database_update or the cli command yes

phpBB/adm/style/acp_update.html
Nicofuma: @Zoddo bump

Nicofuma: @Zoddo bump again

marc1706:
Closing as there is now #4281

PR #3904, opened 2015-09-13
[ticket/14162] Add CLI commands to manage migrations

Zoddo:
| Command | Description |
| ------------- | -------------- |
| db:list | List all installed and not installed migrations. |
| db:revert | Revert a migration. |

Note: For `db:list`, the class is named `list_command`, because `list` is a reserved word and can't be used as class name in PHP.

[PHPBB3-14162](https://tracker.phpbb.com/browse/PHPBB3-14162)

### DIFF
--- a/phpBB/language/en/cli.php
+++ b/phpBB/language/en/cli.php
@@ -51,6 +51,7 @@ $lang = array_merge($lang, array(
 	'CLI_DESCRIPTION_CRON_RUN'					=> 'Runs all ready cron tasks.',
 	'CLI_DESCRIPTION_CRON_RUN_ARGUMENT_1'		=> 'Name of the task to be run',
 	'CLI_DESCRIPTION_DB_MIGRATE'				=> 'Updates the database by applying migrations.',
+	'CLI_DESCRIPTION_DB_REVERT'					=> 'Revert a migration.',
 	'CLI_DESCRIPTION_DELETE_CONFIG'				=> 'Deletes a configuration option',
 	'CLI_DESCRIPTION_DISABLE_EXTENSION'			=> 'Disables the specified extension.',
 	'CLI_DESCRIPTION_ENABLE_EXTENSION'			=> 'Enables the specified extension.',
@@ -93,6 +94,8 @@ $lang = array_merge($lang, array(
 	'CLI_EXTENSIONS_ENABLED'			=> 'Enabled',
 
 	'CLI_FIXUP_RECALCULATE_EMAIL_HASH_SUCCESS'	=> 'Successfully recalculated all email hashes.',
+
+	'CLI_MIGRATION_NAME'					=> 'Migration name, including the namespace (use forward slashes instead of backslashes to avoid problems).',
 
 	'CLI_REPARSER_REPARSE_REPARSING'		=> 'Reparsing %1$s (range %2$d..%3$d)',
 	'CLI_REPARSER_REPARSE_REPARSING_START'	=> 'Reparsing %s...',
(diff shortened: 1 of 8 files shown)


phpBB/phpbb/console/command/db/list_command.php
```diff
@@ -0,0 +1,69 @@
...
+			{
+				$uninstalled[] = $name;
+			}
+		}
+
+		if (!empty($installed) && $show_installed)
+		{
+			$output->writeln('<info>' . $this->user->lang('CLI_MIGRATION_INSTALLED') . $this->user->lang('COLON') . '</info>');
+			$output->writeln($installed);
+		}
```
Nicofuma: You should display a message if there isn't any installed migration

phpBB/phpbb/console/command/db/list_command.php
```diff
@@ -0,0 +1,69 @@
...
+
+		if (!empty($installed) && $show_installed && !empty($uninstalled))
+		{
+			$output->writeln('');
+		}
+
+		if (!empty($uninstalled))
```
Nicofuma: You should display a message if there isn't any uninstalled migration

Nicofuma: by the way, this list will also be displayed without the option `-u`

phpBB/phpbb/console/command/db/list_command.php
```diff
@@ -0,0 +1,69 @@
...
+		{
+			$output->writeln('<info>' . $this->user->lang('CLI_MIGRATION_UNINSTALLED') . $this->user->lang('COLON') . '</info>');
```
Nicofuma: I don't think uninstall is a good name here: thes migrations can be both uninstalled and not installed. I'd prefer the more generic `not installed`or `available` here

Nicofuma:
beside my comments, :+1:

phpBB/phpbb/console/command/db/list_command.php
imkingdavid: +1 for `available` rather than `not installed` or `uninstalled`

Zoddo:
@Nicofuma : Done

phpBB/language/en/cli.php
```diff
@@ -50,7 +50,9 @@
 	'CLI_DESCRIPTION_CRON_LIST'					=> 'Prints a list of ready and unready cron jobs.',
 	'CLI_DESCRIPTION_CRON_RUN'					=> 'Runs all ready cron tasks.',
 	'CLI_DESCRIPTION_CRON_RUN_ARGUMENT_1'		=> 'Name of the task to be run',
+	'CLI_DESCRIPTION_DB_LIST'					=> 'List all installed and not installed migrations.',
```
VSEphpbb: List all installed, and uninstalled, migrations.

phpBB/language/en/cli.php
```diff
@@ -94,6 +96,12 @@
 
 	'CLI_FIXUP_RECALCULATE_EMAIL_HASH_SUCCESS'	=> 'Successfully recalculated all email hashes.',
 
+	'CLI_MIGRATION_NAME'					=> 'Full migration name (you can replace the antislashs by slashs to avoid problems)',
```
VSEphpbb: Is it name or path and name?
Rewrite: (use forward slashes instead of backslashes to avoid problems)

Zoddo: It's the class name (including namespace)

VSEphpbb: Then please say that: 
> Migration name, including the namespace (use forward slashes instead of backslashes to avoid problems).

Zoddo:
@VSEphpbb : Done :)

marc1706:
:+1:

PR #3905, opened 2015-09-13
[ticket/14078] Fix SQL error when editing a post with a quote

Zoddo:
[PHPBB3-14078](https://tracker.phpbb.com/browse/PHPBB3-14078)


Zoddo:
!set Blocker

PR #3906, opened 2015-09-13
[ticket/14164] Add instructions to use the CLI for database updates

Zoddo:
[PHPBB3-14164](https://tracker.phpbb.com/browse/PHPBB3-14164)

### DIFF
--- a/phpBB/docs/INSTALL.html
+++ b/phpBB/docs/INSTALL.html
@@ -265,7 +265,7 @@
 
 <p>If you are currently using a stable release of phpBB, updating to this version is straightforward. You would have downloaded one of four packages and your choice determines what you need to do. <strong>Note</strong>: Before updating, we heavily recommend you do a <em>full backup of your database and existing phpBB files</em>! If you are unsure how to achieve this please ask your hosting provider for advice.</p>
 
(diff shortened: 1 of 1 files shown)


VSEphpbb:
Should probably say something more like: 
> If you have ssh access to your server, you may wish to update via the command line interface. From your board's root, execute the following command

Nicofuma:
:+1: for both of you ;)

Oyabun1:
It probably needs to be clear that the CLI command is only for updating the database, so maybe:
> If you have shell access to your server, you may wish to perform the database update via the command line interface. From your board's root, execute the following command: <code>php bin/phpbbcli.php --safe-mode db:migrate</code>.

VSEphpbb:
> have a shell

is wrong.. Remove the "a"

Zoddo:
@VSEphpbb : Done

bantu:
This goes into the right direction, but it unfortunately looks kind of half-baked right now. `4.i. Full package` to `4.iii. Patch file` still make direct references to (only) `install/database_update.php`.

Zoddo:
@bantu : Done :)

PR #3908, opened 2015-09-15
[ticket/14163] Properly fix select all for code tag in Edge

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-14163

PHPBB3-14163

### DIFF
--- a/phpBB/styles/prosilver/template/forum_fn.js
+++ b/phpBB/styles/prosilver/template/forum_fn.js
@@ -159,7 +159,10 @@ function selectCode(a) {
 			try {
 				s.setBaseAndExtent(e, 0, e, l);
 			} catch (error) {
-				s.setBaseAndExtent(e, 0, e, 1);
+				r = document.createRange();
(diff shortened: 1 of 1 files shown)


bantu:
@callumacrae Here we are again. :-D

marc1706:
On a side note: it turns out Edge is using the number of lines of the textarea for the select Index instead of the actual character count. Hence why it fails with the old method.

bantu:
@callumacrae Any comments on this?

callumacrae:
No comments, have no idea what is going on!

marc1706:
If you have questions, feel free to ask.

bantu:
This piece of code was previously discussed in #3818

marc1706:
The code we added in #3818 does work, but only for one line. Based on Edge's behavior the API seems to work as outlined here: https://github.com/phpbb/phpbb/pull/3908#issuecomment-140651715

This approach does work perfectly fine for Edge and is probably what we should've used to begin with.

PR #3910, opened 2015-09-17
[ticket/14170] Fix mcp_change_poster_after event

kinerity:
Move the mcp_change_poster_after event to before renewing $post_info. I
changed the "since" version as this would be fixed in 3.1.7-RC1.

PHPBB3-14170

### DIFF
--- a/phpBB/includes/mcp/mcp_post.php
+++ b/phpBB/includes/mcp/mcp_post.php
@@ -549,6 +549,17 @@ function change_poster(&$post_info, $userdata)
 	$from_username = $post_info['username'];
 	$to_username = $userdata['username'];
 
+	/**
+	* This event allows you to perform additional tasks after changing a post's poster
+	*
+	* @event core.mcp_change_poster_after
+	* @var	array	userdata	Information on a post's new poster
+	* @var	array	post_info	Information on the affected post
+	* @since 3.1.7-RC1
+	*/
+	$vars = array('userdata', 'post_info');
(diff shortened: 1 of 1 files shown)


phpBB/includes/mcp/mcp_post.php
```diff
@@ -549,27 +549,27 @@ function change_poster(&$post_info, $userdata)
...
-	$post_info = $post_info[$post_id];
-
 	/**
 	* This event allows you to perform additional tasks after changing a post's poster
 	*
 	* @event core.mcp_change_poster_after
 	* @var	array	userdata	Information on a post's new poster
 	* @var	array	post_info	Information on the affected post
-	* @since 3.1.6-RC1
+	* @since 3.1.7-RC1
```
Nicofuma: You shouldn't change @since but add a @change tag, example: https://github.com/phpbb/phpbb/pull/3946/files#diff-a661e9c889d836377e0341f8e90befacR4288

kinerity: Fixed in the commit below. Out of curiosity, why wasn't this mentioned sooner?

Nicofuma:
The event name is `core.mcp_change_poster_after` therefore it seems logical to me to dispatch it after changing the poster name

kinerity:
It still is... https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/includes/mcp/mcp_post.php#L473-L498

Nicofuma:
ho ok I see.

Seems fine

kinerity:
Um, bumpage? This is a month old. If it's good, what's holding up the merge?

PR #3913, opened 2015-09-19
[ticket/14168] Refactor attachment management functions into classes

marc1706:
Todo:

- [x] Put delete_attachments() into own class
- [x] Improve test coverage
    - [x] delete
    - [x] resync
    - [x] upload
- [x] Split up code into methods

Ticket: https://tracker.phpbb.com/browse/PHPBB3-14168

PHPBB3-14168


phpBB/config/default/container/services_files.yml
```diff
@@ -1,4 +1,35 @@
 services:
+    attachment.delete:
```
CHItA: these services should be moved to somewhere else (as they are crashing the install env).

Nicofuma: you are using services_files in the installer?

CHItA: yep, that's where filesystem is. We came up with the same filename in the past.

Nicofuma: actually we have 3 things here:
- filesystem
- upload
- attachments

I think we should split them

CHItA: i moved the previous attachment management tasks (so `files.*`) services under `services_attachment.yml` for quick fix, though that's might not the best solution, so do whatever you fancy, and i'll rebase.

marc1706: I guess moving the attachment classes to services_attachments.yml sounds like a reasonable way of handling this?

marc1706:
Removed WIP label

phpBB/phpbb/attachment/delete.php
```diff
@@ -0,0 +1,323 @@
...
+		// Now update the topics. This is a bit trickier, because there could be posts still having attachments within the topic
+		$this->resync->resync('topic', $this->topic_ids);
+
+		return $this->num_deleted;
+	}
+
+	/**
+	 * Set attachment IDs
+	 *
+	 * @param array $ids
```
CHItA: If it can be anything else then an array, it would be nice if you could add a proper docblock here.

phpBB/phpbb/attachment/resync.php
```diff
@@ -0,0 +1,125 @@
...
+ */
+
+namespace phpbb\attachment;
+
+use \phpbb\db\driver\driver_interface;
+
+/**
+ * Attachment delete class
+ */
+
```
CHItA: unnecessary new line

CHItA:
looks good to me :+1:

CHItA:
`services_attachment.yml` is not included in `services.yml`

marc1706:
done

Nicofuma:
you have to rebase

phpBB/phpbb/attachment/resync.php
```diff
@@ -0,0 +1,124 @@
...
+ * the docs/CREDITS.txt file.
+ *
+ */
+
+namespace phpbb\attachment;
+
+use \phpbb\db\driver\driver_interface;
+
+/**
+ * Attachment delete class
```
Nicofuma: wrong docblock

phpBB/config/default/container/services_attachment.yml
```diff
@@ -0,0 +1,31 @@
...
+            - @attachment.resync
+            - %core.root_path%
+
+    attachment.resync:
+        class: phpbb\attachment\resync
+        scope: prototype
+        arguments:
+            - @dbal.conn
+
+    attachment.upload:
```
Nicofuma: these services seems to be 3 different part of an attachments manager.

I think i'd prefer to have and expose a global manager (using these services).

marc1706: Ok, should be a rather easy change.

marc1706:
Will add tests for the manager and post an update once I finished that.

marc1706:
@Nicofuma done

phpBB/phpbb/attachment/manager.php
```diff
@@ -0,0 +1,99 @@
...
+	 * Wrapper method for uploading attachment
+	 *
+	 * @param string			$form_name		The form name of the file upload input
+	 * @param int			$forum_id		The id of the forum
+	 * @param bool			$local			Whether the file is local or not
+	 * @param string			$local_storage	The path to the local file
+	 * @param bool			$is_message		Whether it is a PM or not
+	 * @param array		$local_filedata	An file data object created for the local file
+	 *
+	 * @return object filespec
```
Nicofuma: `@return \phpbb\files\filespec`

phpBB/phpbb/attachment/upload.php
```diff
@@ -0,0 +1,334 @@
...
+	 * Uses upload class
+	 *
+	 * @param string			$form_name		The form name of the file upload input
+	 * @param int			$forum_id		The id of the forum
+	 * @param bool			$local			Whether the file is local or not
+	 * @param string			$local_storage	The path to the local file
+	 * @param bool			$is_message		Whether it is a PM or not
+	 * @param array		$local_filedata	An file data object created for the local file
+	 *
+	 * @return object filespec
```
Nicofuma: `@return \phpbb\files\filespec`

Nicofuma:
You should probably use the new manager everywhere instead of using the dedicated services. What do you think?

marc1706:
Yeah, I knew I forgot to change something :)

PR #3912, opened 2015-09-17
[ticket/14175] Refactor responsive implementation for easier manipulation

hanakin:
The current way we include and use the responsive stylesheet is very dated and improperly handled. All media queries should be handled in the file itself and all code should be consolidated by breakpoint. This makes for setting of a style at a specific breakpoint very simplistic.

ticket: https://tracker.phpbb.com/browse/PHPBB3-14175
PHPBB3-14175

PR #3918, opened 2015-09-21
Remove deprecated max-device-width

MikkCZ:
See https://drafts.csswg.org/mediaqueries/#mf-deprecated.

From HTML removed in https://github.com/phpbb/phpbb/commit/96521767f8db215e36e7a00264dd491fc5816caa.

https://tracker.phpbb.com/browse/PHPBB3-14183

### DIFF
--- a/phpBB/adm/style/admin.css
+++ b/phpBB/adm/style/admin.css
@@ -129,14 +129,14 @@ hr {
 	display: none;
 }
 
-@media only screen and (max-width: 800px), only screen and (max-device-width: 800px)
+@media only screen and (max-width: 800px)
 {
 	body {
 		margin: 5px 5px 0;
 	}
 }
 
-@media only screen and (max-width: 700px), only screen and (max-device-width: 700px)
+@media only screen and (max-width: 700px)
 {
 	html, body {
 		height: auto;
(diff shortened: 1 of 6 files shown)


Louis7777:
Man this was so useful when you wanted - for example - a website to transform only for smartphones and not for PCs... I wonder why they are deprecating it...

bantu:
@hanakin What do you think about this? Should this be part of #3912 then?

hanakin:
@bantu yes its actually something I was going to refactor as well I can easily address it in my pr

bantu:
I also don't think we want this in 3.1.x. @nickvergessen

bantu:
This follows #3865 though, which was merged into 3.1.x.

bantu:
@marc1706 You merged #3865. Want to review this / decide whether we want it?

marc1706:
@bantu the difference is that the HTML part caused the 3.1.x prosilver to be invalid HTML5. This however does not. It's also not invalid CSS level 3. Therefore I don't think we should merge this into 3.1.x. A patch against master should be the correct way to go.

marc1706:
@hanakin I guess easiest solution would be if you'd address that in your PR?

nickvergessen:
> the difference is that the HTML part caused the 3.1.x prosilver to be invalid HTML5. This however does not. It's also not invalid CSS level 3. Therefore I don't think we should merge this into 3.1.x. A patch against master should be the correct way to go.

agreed

PR #3920, opened 2015-09-22
[ticket/14186] Do correct string concatenation in phpbb_mcp_sorting()

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-14186


bantu:
:+1:

PR #3925, opened 2015-09-23
[ticket/14191] Add core.get_gravatar_url_after event

Senky:
[PHPBB3-14191](https://tracker.phpbb.com/browse/PHPBB3-14191)

### DIFF
--- a/phpBB/phpbb/avatar/driver/gravatar.php
+++ b/phpBB/phpbb/avatar/driver/gravatar.php
@@ -172,6 +172,8 @@ class gravatar extends \phpbb\avatar\driver\driver
 	*/
 	protected function get_gravatar_url($row)
 	{
+		global $phpbb_dispatcher;
+
 		$url = self::GRAVATAR_URL;
 		$url .=  md5(strtolower(trim($row['avatar'])));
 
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/avatar/driver/gravatar.php
```diff
@@ -180,6 +182,17 @@ protected function get_gravatar_url($row)
...
+		/**
+		* Modify gravatar url
+		*
+		* @event core.get_gravatar_url_after
+		* @var	string	row	User data or group data
+		* @var	string	url	Gravatar URL
+		* @since 3.1.7-RC1
+		*/
+		$vars = array('row', 'url');
+		extract($phpbb_dispatcher->trigger_event('core.get_gravatar_url_after', compact($vars)));
```
Nicofuma: I'm failing to see why we would need to have an event here, could you give me a use case please?

Senky: I want to append `&d=identicon` to the URL.

Nicofuma: I see. Thanks

PR #3926, opened 2015-09-23
[ticket/14192] Add core.memberlist_memberrow_before event

Senky:
[PHPBB3-14192](https://tracker.phpbb.com/browse/PHPBB3-14192)

### DIFF
--- a/phpBB/memberlist.php
+++ b/phpBB/memberlist.php
@@ -1448,6 +1448,20 @@ switch ($mode)
 				usort($user_list,  'phpbb_sort_last_active');
 			}
 
+			// do we need to display contact fields as such
+			$use_contact_fields = false;
+
+			/**
+			 * Modify list of users before member row is created
+			 *
+			 * @event core.memberlist_memberrow_before
+			 * @var array	user_list			Array containing list of users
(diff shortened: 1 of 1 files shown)


phpBB/memberlist.php
```diff
@@ -1448,6 +1448,20 @@
 				usort($user_list,  'phpbb_sort_last_active');
 			}
 
+			// do we need to display contact fields as such
+			$use_contact_fields = false;
+
+			/**
+			 * Modify the query used to get the users for the team page
```
Nicofuma: this event doesn't allow you to edit any SQL query

Senky: In fact, it allows. Just see what `$use_contact_fields` variable modifies - parameter of `$cp->generate_profile_fields_template_data` call, which modifies SQL query inside that method.

Nicofuma: yes but it doesn't allow you to edit the sql query itself but if yes or no the contact fields should be used

Senky: True. So the problem is just the description of the event? I can change that easily.

Nicofuma: yep

PR #3928, opened 2015-09-25
[Ticket/14197]Modularize/refactor all buttons

hanakin:
To allow for a more consistent UI as well as easier maintainability and customization we should refactor all button variations to work off of the primary button module with modifer classes for the different varions. Then consolidate all of it into the buttons.css file and reorganize it.

Tickets: 
https://tracker.phpbb.com/browse/PHPBB3-14197
https://tracker.phpbb.com/browse/PHPBB3-14193

PHPBB3-14197

### DIFF
--- a/phpBB/styles/prosilver/template/captcha_default.html
+++ b/phpBB/styles/prosilver/template/captcha_default.html
@@ -12,7 +12,7 @@
 		<dt><label for="confirm_code">{L_CONFIRM_CODE}{L_COLON}</label></dt>
 		<dd class="captcha captcha-image"><img src="{CONFIRM_IMAGE_LINK}" alt="{L_CONFIRM_CODE}" /></dd>
 		<dd><input type="text" name="confirm_code" id="confirm_code" size="8" maxlength="8" tabindex="{$CAPTCHA_TAB_INDEX}" class="inputbox narrow" title="{L_CONFIRM_CODE}" />
-		<!-- IF S_CONFIRM_REFRESH --><input type="submit" name="refresh_vc" id="refresh_vc" class="button2" value="{L_VC_REFRESH}" /><!-- ENDIF -->
+		<!-- IF S_CONFIRM_REFRESH --><input type="submit" name="refresh_vc" id="refresh_vc" class="button button-form-bold" value="{L_VC_REFRESH}" /><!-- ENDIF -->
 		<input type="hidden" name="confirm_id" id="confirm_id" value="{CONFIRM_ID}" /></dd>
 		<dd>{L_CONFIRM_CODE_EXPLAIN}</dd>
 	</dl>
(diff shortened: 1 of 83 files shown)


VSEphpbb:
IMO, within Prosilver, the "old" buttons (Submit, Reset, LogIn) should be changed to look like the New Topic or Custom BBCode buttons, so all buttons have a consistent appearance in Prosilver. Likewise, the "old" pull downs should be made to look like the newer Jump Box pull downs too.

hanakin:
I agree with you but there was push back when previously discussing such changes which is why I opted to just take it step by step. use modifiers that we can easily remove. 

As for the old pull downs are you referring to making the dropdown menu it self look like the jumpbox menu? I somewhat agree with that. The jumpbox button itself should be the same as the other buttons. the dropdowns menus of all the others should be updated to be similar to the way the highlighting happens in the jumpbox menu basically color changes without the underlines

Nicofuma:
I agree with both of you: we should do that but in small and specialized PRs :)

VSEphpbb:
I'm just reffering to the fact that prosilver now has new skool and old skool form elements. Which is inconsistent.

screen shot 2015-09-25 at 10 13 45 am

screen shot 2015-09-25 at 10 14 24 am
(By the way, that drop down menu is still broken!)

hanakin:
@vse its not broken its intended to be that color as thats the default row color. If I change it to match the cat then when you scroll its broken again. Its a catch 22 not sure how to really handle it.

VSEphpbb:
@hanakin The triangle is the wrong color. It needs to be dark blue.

hanakin:
@VSEphpbb look at it on area51 and you will see what I mean

Used in footer
screenshot 2015-09-25 19 31 04

Used in MCP
screenshot 2015-09-25 19 32 03

VSEphpbb:
Then can you add some space above the first elem in the list, just 5px or so. Something like:
screen-shot-2015-09-25-at-10 42 02-am

hanakin:
hmm that would only fix it before you scroll and it would make the first row larger if the first element was not a category

VSEphpbb:
It would be a separate element, maybe a CSS:before psuedo thing on the first list element.

VSEphpbb:
Or move the triangle behind the list instead of on top of it.?

hanakin:
!set WIP

hanakin:
Have broken this up into smaller parts to address specific issues
https://github.com/phpbb/phpbb/pull/3932
https://github.com/phpbb/phpbb/pull/3937
https://github.com/phpbb/phpbb/pull/3940
https://github.com/phpbb/phpbb/pull/3938

marc1706:
@hanakin is this still WIP?

hanakin:
Its been broken up into smaller piers so should be closed

PR #3929, opened 2015-09-25
[ticket/12618] Allow extension author to use SSL for version-check.

kilianr:
Fixes Ticket 12618 (https://tracker.phpbb.com/browse/PHPBB3-12618)

In order to be able to use SSL with fsockopen, one has to append 'ssl://' to the host. Thus, the class file_downloader needed to be changed.
'ssl://' will always be appended now, if the port is 443.
I have checked, that this does not influence any other uses of that class (there are none).

### DIFF
--- a/phpBB/phpbb/file_downloader.php
+++ b/phpBB/phpbb/file_downloader.php
@@ -42,7 +42,7 @@ class file_downloader
 		$this->error_number = 0;
 		$this->error_string = '';
 
-		if ($socket = @fsockopen(($port == 443 ? 'ssl://' : '') . $host, $port, $this->error_number, $this->error_string, $timeout))
+		if ($socket = @fsockopen(($port == 443 ? 'tls://' : '') . $host, $port, $this->error_number, $this->error_string, $timeout))
 		{
 			@fputs($socket, "GET $directory/$filename HTTP/1.0\r\n");
 			@fputs($socket, "HOST: $host\r\n");
--- a/phpBB/phpbb/file_downloader.php
+++ b/phpBB/phpbb/file_downloader.php
@@ -42,7 +42,7 @@ class file_downloader
 		$this->error_number = 0;
 		$this->error_string = '';
 
-		if ($socket = @fsockopen($host, $port, $this->error_number, $this->error_string, $timeout))
+		if ($socket = @fsockopen(($port == 443 ? 'ssl://' : '') . $host, $port, $this->error_number, $this->error_string, $timeout))
 		{
 			@fputs($socket, "GET $directory/$filename HTTP/1.0\r\n");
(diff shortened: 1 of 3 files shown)


bantu:
Unfortunately the way in which SSL/TLS is used here lacks security in the sense that certificates are not validated at all.

kilianr:
@bantu Ok, maybe I'm missing the point here. So far, it has been impossible to connect via https to get the information for the version check. This PR would solve that. True, for older versions of php, fsockopen doesn't care about certificates (it does starting with PHP 5.6 by the way).
So in order to make https connections at least possible it would still be nice to merge this PR, but if you have other requirements, its ok for me to close it. However, in this case, it would be nice to state those requirements in the ticket (e.g. what are trusted root-certs, to you want to pin some certs, do you want to consider DNSSec, maybe even DANE, ....) because right now, they are unclear.
My personal opinion would be, that if you just validate that the certificate is for the domain, you gain nothing. So you need to establish a chain of trust, which brings us back to root-certs or DANE.

Nicofuma:
imo, the best solution would be to use `Guzzle` in the `file_downloader` instead of using the sockets manually.

ButI don't think it can be done in 3.1.x

phpBB/phpbb/file_downloader.php
```diff
@@ -42,7 +42,7 @@ public function get($host, $directory, $filename, $port = 80, $timeout = 6)
 		$this->error_number = 0;
 		$this->error_string = '';
 
-		if ($socket = @fsockopen($host, $port, $this->error_number, $this->error_string, $timeout))
+		if ($socket = @fsockopen(($port == 443 ? 'ssl://' : '') . $host, $port, $this->error_number, $this->error_string, $timeout))
```
bantu: This should certainly be tls://

bantu:
@kilianr Maybe this patch isn't all too bad. My problem is mostly the lack of certificate checking and lack of authentication as a result. Maybe we should just decide that this is a not an issue for version information.

Elsensee:
> Maybe we should just decide that this is a not an issue for version information.

Yes. Also, as @kilianr already said, this is fixed with PHP 5.6.

Nicofuma:
I think we can say it's not an issue for version check.

:+1: 

@bantu you confirm it's good for you?

phpBB/phpbb/file_downloader.php
marc1706: Yeah, I'd go for ```tls://``` here

kilianr: Changed it to tls:// in last recent commit.

Elsensee:
Thank you. :) :+1:

PR #3932, opened 2015-09-27
[ticket/14193]Fix secondary buttons

hanakin:
ticket: https://tracker.phpbb.com/browse/PHPBB3-14193
PHPBB3-14193

### DIFF
--- a/phpBB/styles/prosilver/template/mcp_topic.html
+++ b/phpBB/styles/prosilver/template/mcp_topic.html
@@ -100,7 +100,7 @@
 			<div class="postbody" id="pr{postrow.POST_ID}">
 				<ul class="post-buttons">
 					<li>
-						<a href="{postrow.U_POST_DETAILS}" title="{L_POST_DETAILS}" class="button">
+						<a href="{postrow.U_POST_DETAILS}" title="{L_POST_DETAILS}" class="button button-icon-only">
 							<i class="icon fa-info fa-fw"></i><span class="sr-only">{L_POST_DETAILS}</span>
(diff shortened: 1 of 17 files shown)


hanakin:
!set Styles

hanakin:
@VSEphpbb please review

VSEphpbb:
Unfortunately, this has changed a lot of button icons red. Advanced search, New Topic for example. That looks wrong/bad.

Also, text inside of the message editor has been turned blue!?! It should be black.

The custom BBCode buttons look good though. Except for one issue. They should have some top/bottom  spacing. If you have a lot of bbcodes, and they wrap down a line, the buttons touch.

screen shot 2015-09-28 at 2 05 40 pm

hanakin:
yeah these are all kind of wipish as I am trying to muddle through splitting up 14197

hanakin:
hmm @VSEphpbb  can not reproduce the red icon issue on my end maybe a cache thing? but try everything now

VSEphpbb:
Issues are cleared up, but this seems like a lot of "other" changes just to fix the custom bbcode buttons.

Can't you just make a fixer class for the bbcode buttons only, rather than having made a change that seems to be affecting other buttons and resulting in this awkward "button-icon-only" class being added in so many places? Just a thought.

hanakin:
well the fix for the bbcode button is simply the button-secondary class I could break it up into separate prs if absolutely necessary one for the button-secondary and one for the extra spacing between rows

hanakin:
test

PR #3935, opened 2015-09-27
[ticket/10343] Allows a_user to show inactive accounts in memberlist

Zoddo:
https://tracker.phpbb.com/browse/PHPBB3-10343

### DIFF
--- a/phpBB/memberlist.php
+++ b/phpBB/memberlist.php
@@ -151,7 +151,7 @@ switch ($mode)
 		$db->sql_freeresult($result);
 
 		$sql_ary = array(
-			'SELECT'	=> 'u.user_id, u.group_id as default_group, u.username, u.username_clean, u.user_colour, u.user_rank, u.user_posts, u.user_allow_pm, g.group_id',
+			'SELECT'	=> 'u.user_id, u.group_id as default_group, u.username, u.username_clean, u.user_colour, u.user_type, u.user_rank, u.user_posts, u.user_allow_pm, g.group_id',
 
 			'FROM'		=> array(
 				USER_GROUP_TABLE => 'ug',
@@ -313,6 +313,8 @@ switch ($mode)
 
 							'RANK_IMG'		=> $user_rank_data['img'],
 							'RANK_IMG_SRC'	=> $user_rank_data['img_src'],
+
+							'S_INACTIVE'	=> $row['user_type'] == USER_INACTIVE,
 
 							'U_PM'			=> ($config['allow_privmsg'] && $auth->acl_get('u_sendpm') && ($row['user_allow_pm'] || $auth->acl_gets('a_', 'm_') || $auth->acl_getf_global('m_'))) ? append_sid("{$phpbb_root_path}ucp.$phpEx", 'i=pm&amp;mode=compose&amp;u=' . $row['user_id']) : '',
 
@@ -1346,7 +1348,7 @@ switch ($mode)
 		}
 
 		$user_types = array(USER_NORMAL, USER_FOUNDER);
-		if ($auth->acl_get('a_user') || $user->data['user_type'] == USER_FOUNDER)
+		if ($auth->acl_get('a_user'))
 		{
 			$user_types[] = USER_INACTIVE;
 		}
@@ -1357,7 +1359,7 @@ switch ($mode)
 		$sql = "SELECT u.user_id
 			FROM " . USERS_TABLE . " u
 				$sql_from
-			WHERE u.user_type IN (" . implode(', ', $user_types) . ")
+			WHERE " . $db->sql_in_set('u.user_type', $user_types) . "
 				$sql_where
 			ORDER BY $order_by";
 		$result = $db->sql_query_limit($sql, $config['topics_per_page'], $start);
@@ -1472,6 +1474,7 @@ switch ($mode)
 
 					'S_CUSTOM_PROFILE'	=> (isset($cp_row['row']) && sizeof($cp_row['row'])) ? true : false,
 					'S_GROUP_LEADER'	=> $is_leader,
+					'S_INACTIVE'		=> $row['user_type'] == USER_INACTIVE,
 
 					'U_VIEW_PROFILE'	=> get_username_string('profile', $user_id, $row['username']),
 				));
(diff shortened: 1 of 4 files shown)


bantu:
Does this only affect the ACP?

bantu:
https://tracker.phpbb.com/browse/PHPBB3-10343?focusedCommentId=56817&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-56817 mentions clearly marking inactive users. Is that the case?

Zoddo:
@bantu: No, it affect all member listing

I will add a clear mark of inactive users in the week :)

!set WIP

Zoddo:
@bantu Done :)

!unset WIP

Nicofuma:
could you share a screenshot please?

phpBB/memberlist.php
```diff
@@ -1345,13 +1347,19 @@
 			);
 		}
 
+		$user_types = array(USER_NORMAL, USER_FOUNDER);
+		if ($auth->acl_get('a_user') || $user->data['user_type'] == USER_FOUNDER)
```
Nicofuma: the check on the user type shouldn't be necessary. It is already enforced in the auth class: https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/phpbb/auth/auth.php#L424

phpBB/memberlist.php
```diff
@@ -1345,13 +1347,19 @@
...
+		}
+
 		$start = $pagination->validate_start($start, $config['topics_per_page'], $config['num_users']);
 
 		// Get us some users :D
 		$sql = "SELECT u.user_id
 			FROM " . USERS_TABLE . " u
 				$sql_from
-			WHERE u.user_type IN (" . USER_NORMAL . ', ' . USER_FOUNDER . ")
+			WHERE u.user_type IN (" . implode(', ', $user_types) . ")
```
Nicofuma: you should use `$db->sql_in_set()` here

phpBB/styles/prosilver/theme/content.css
```diff
@@ -842,3 +842,7 @@ table.fixed-width-table {
 	background: rgba(0, 0, 0, .3);
 	border-radius: 3px;
 }
+
+#memberlist tr.inactive, #team tr.inactive {
+	font-style: italic;
+}
```
Nicofuma: missing new line a the end

Zoddo:
@Nicofuma : Fixed.

Here a screenshot (desactivated users are in italic):
image  
image

Nicofuma:
thanks

:+1:

DavidIQ:
The "clear" mark is that they're in italics?  I'm not sure that's very clear at all.  How does an administrator looking at this list know that users in italics are inactive?

Nicofuma:
thinking back about it... you're right... maybe we could display them in gray too?

DavidIQ:
I think the italicizing of the text is fine but maybe add some text too:
_test (Inactive)_

It would require adding a language variable (INACTIVE) so this change can be "enhanced" like so on line 105 of memberlist_body.html:
```
		<td><span class="rank-img"><!-- EVENT memberlist_body_rank_prepend --><!-- IF memberrow.RANK_IMG -->{memberrow.RANK_IMG}<!-- ELSE -->{memberrow.RANK_TITLE}<!-- ENDIF --><!-- EVENT memberlist_body_rank_append --></span><!-- IF S_IN_SEARCH_POPUP and not S_SELECT_SINGLE --><input type="checkbox" name="user" value="{memberrow.USERNAME}" /> <!-- ENDIF --><!-- EVENT memberlist_body_username_prepend -->{memberrow.USERNAME_FULL}<!-- IF memberrow.S_INACTIVE --> ({L_INACTIVE})<!-- ENDIF --><!-- EVENT memberlist_body_username_append --><!-- IF S_IN_SEARCH_POPUP --><br />[&nbsp;<a href="#" onclick="insert_single_user('#results', '{memberrow.A_USERNAME}'); return false;">{L_SELECT}</a>&nbsp;]<!-- ENDIF --></td>
```

And line 22 of memberlist_team.html:
```
		<td><!-- IF group.user.RANK_IMG --><span class="rank-img">{group.user.RANK_IMG}</span><!-- ELSE --><span class="rank-img">{group.user.RANK_TITLE}</span><!-- ENDIF -->{group.user.USERNAME_FULL}<!-- IF group.user.S_INACTIVE --> ({L_INACTIVE})<!-- ENDIF --></td> 
```

Zoddo:
So, a new ticket/branch with add this?

Nicofuma:
David is taking care of it at the moment I think

ntonksie:
Hi guys, i don't know what prompted this addition, but you do realize thay on huge forums, that list is going to be enormous.
could you for the next version make it an option and revert back to only showing active members. Thanks

Zoddo:
@ntonksie : A checkbox in `Find a member` for example?

Lordsamus:
Hi ! Same case for me too. Have a whole list of inactive users is not really usefull for me (and a lot of friends).
See inactive users is interesting only in <i>Find a member</i> in certain case. Actually, to hide inactives, i must write a theme which extends from Prosilver to  only add : 
.inactive {
 display : none;
}
(because adding it in prosilver will be erased at any update)
It's not really a good solution and disturb row colors.

ntonksie:
@Zoddo Like the checkbox for invited users you mean ? I'm down with that ! I understand there was a bug in previous versions in ACP and that is a bug fixing but in doing so, you made using the memberlist on the live forum difficult. I know i can go around it by sorting them by latest connection, but i didn't find if / how i can put that as default in ACP. For me default is alphabetical. I'll be happy with anything that resolve this.

Lordsamus:
Using CSS to hide inactive users was not a good solution too  : 
badidea

Rows colors are broken with this trick.

Elsensee:
Also, this isn't a very good solution, since you can easily disable the CSS with browser tools and then see a list of all inactive users (even with the status "inactive") without having the permission.

DavidIQ:
This change has already been merged. You guys should just open a new ticket in our tracker to deal with this or continue the conversation in the original ticket.  Commenting here won't really do anything as a new merge request would be needed.

PR #3936, opened 2015-09-27
[ticket/14200] Update FAQ to reflect that hidden users can't see himself

Zoddo:
https://tracker.phpbb.com/browse/PHPBB3-14200

### DIFF
--- a/phpBB/language/en/help_faq.php
+++ b/phpBB/language/en/help_faq.php
@@ -81,7 +81,7 @@ $help = array(
 	),
 	array(
 		0 => 'How do I prevent my username appearing in the online user listings?',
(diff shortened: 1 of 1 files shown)


bantu:
Can you dig up the ticket or pull request that changed the behaviour?

Zoddo:
@bantu : Ah no, I remembered that it functioned like that, but after searching in commit history, it appears that the behaviour has never been changed.

I will update my comment on tracker.

CHItA:
I think we might be better fixing the behavior rather then the language file. Though that's kind of another issue.

Nicofuma:
If it is not hard  to do I'd prefer to fix the behaviour too.

@Zoddo can you do that?

Zoddo:
@Nicofuma : Yes, I will open a new PR for that :)

Zoddo:
Replaced by #3957

PR #3937, opened 2015-09-27
[ticket/14204]Fix form buttons to use consistent button codebase

hanakin:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-14204
PHPBB3-14204

### DIFF
--- a/phpBB/styles/prosilver/template/captcha_default.html
+++ b/phpBB/styles/prosilver/template/captcha_default.html
@@ -12,7 +12,7 @@
 		<dt><label for="confirm_code">{L_CONFIRM_CODE}{L_COLON}</label></dt>
 		<dd class="captcha captcha-image"><img src="{CONFIRM_IMAGE_LINK}" alt="{L_CONFIRM_CODE}" /></dd>
 		<dd><input type="text" name="confirm_code" id="confirm_code" size="8" maxlength="8" tabindex="{$CAPTCHA_TAB_INDEX}" class="inputbox narrow" title="{L_CONFIRM_CODE}" />
-		<!-- IF S_CONFIRM_REFRESH --><input type="submit" name="refresh_vc" id="refresh_vc" class="button2" value="{L_VC_REFRESH}" /><!-- ENDIF -->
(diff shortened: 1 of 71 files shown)


hanakin:
!set Styles

hanakin:
!set WIP

hanakin:
!unset WIP

hanakin:
ready for review @VSEphpbb @cyberalien

prototech:
@hanakin Please see IRC note.

hanakin:
@prototech this is more of a standardization as button class will most likely be similar in future theme/s but can be pushed til 3.3

hanakin:
!set WIP

VSEphpbb:
Buttons look good. I think we should give them a little border radius like all our other buttons have (3px)?

VSEphpbb:
Why in some place (like UCP prefs) is the Reset button Bold, and submit plain? Should be the opposite IMO.

VSEphpbb:
..Also, `Go` buttons, next to drop downs are often bold. I don't think they need to be bold. I think bold buttons are really only effective when you have 2 (or more buttons) in a row, and you want the bold button to stand out as the primary choice.

VSEphpbb:
On Posting, `Save Draft` is bold?? Should be `Submit` that is bold ;)

VSEphpbb:
Also Add Attachments is bold, since that button is a singleton, it' should not be bold. In general all lone buttons should not be bold IMO.

VSEphpbb:
In general, comparing this to the existing phpBB, it looks to me like the bold buttons are all reversed... What should be bold is not bold, and vice versa. So I think in general you need to reverse all your bold and unbold button assignments here.

hanakin:
hmm looks like I replace the styles in reverse easy fix

hanakin:
Refactored

marc1706:
There seems to be an issue with the alignment of text inside the buttons. The marging between the text and the top border is 5px while it's 8px underneath the text. Not sure if this is only limited to my linux mint & chrome installation or if it also happens on other machines but it certainly looks weird.

marc1706:
image

hanakin:
Chrome on OSX
screenshot 2017-03-18 12 23 10

PR #3941, opened 2015-09-28
[ticket/14195]Fix pupload & polls hiding

hanakin:
ticket: https://tracker.phpbb.com/browse/PHPBB3-14195
PHPBB3-14195

### DIFF
--- a/phpBB/styles/prosilver/theme/utilities.css
+++ b/phpBB/styles/prosilver/theme/utilities.css
@@ -60,8 +60,7 @@
 }
 
 .hidden {
-	display: none !important;
-	visibility: hidden !important;
+	display: none ;
 }
 
 .affix { position: fixed }


hanakin:
!set Styles

VSEphpbb:
Inline style? :smirk_cat:

VSEphpbb:
This isn't really fixing the problems introduced by the original change. This does nothing to help prevent breaking all the extensions that are using and depending on phpBB's current `hidden` tags combined with phpBB's `toggleDisplay` JS.

VSEphpbb:
The rules added to the original `hidden` tag are an issue for parts of phpBB and extensions (especially if they use some for of JS to unhide).

The issue is the Font Awesome PR changing `hidden` to be waaaay more strict...
from:
```css
.hidden { display: none; }
```
to
```css
.hidden {
   display: none !important;
   visibility: hidden !important;
}
```

The important flags and the added visibility rule make ours (and extensions) JS unhide commands fail to work... `$(obj).show()` will fail, for example.

This simply isn't backwards compatible. I see two solution options:
1. Since I don't know why the Font Awesome PR needed to make this change, simply revert `hidden` back to the original rules if there is no specific reason why they were made so strict.
2. Revert `hidden` back to the original rules and add another new rule, `invisible` and give that the new stricter rules. Then use the `invisible` class wherever it was required per Font Awesome PR changes, and leaving the hidden tag to function as expected elsewhere in the core and among extensions.

hanakin:
the class was modified in an attempt to fix issues with the responsive menus at the time. I have since fixed the issue I was originally having it seems requiring the more stricter class. It seems it was being caused by some conflict with the parser functionality in the FA pr already. so simply reverting the hidden class should work. The changes I made to inline here was based on searching through the codebase as I noticed that there is a lot of inconsistency with how we are handling the hiding of content through use of obscured functionality(parser function), hidden classes, and the display:none inline which is used countlessly other places. More so then any other method. All of these should be addressed.

VSEphpbb:
Cool, glad it worked out in the end then :+1:

hanakin:
ready for merge

PR #3945, opened 2015-10-04
[ticket/14216] Do not run thumbnail test if gd is disabled

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-14216

PHPBB3-14216


tests/console/thumbnail_test.php
```diff
@@ -87,6 +87,11 @@ protected function tearDown()
 
 	public function test_thumbnails()
 	{
+		if (!@extension_loaded('gd'))
+		{
+			$this->markTestSkipped('Thumbnail tests require gd extension.');
```
Nicofuma: you should do that in the setUp() I think

Nicofuma:
thanks

PR #3946, opened 2015-10-06
[ticket/13591] Change SQL query into array to allow

RMcGirr83:
extension authors to modify SQL query

PHPBB-13591

https://tracker.phpbb.com/browse/PHPBB3-14217

### DIFF
--- a/phpBB/includes/functions.php
+++ b/phpBB/includes/functions.php
@@ -4264,10 +4264,14 @@ function obtain_users_online_string($online_users, $item_id = 0, $item = 'forum'
 
 	if (sizeof($online_users['online_users']))
 	{
-		$sql = 'SELECT username, username_clean, user_id, user_type, user_allow_viewonline, user_colour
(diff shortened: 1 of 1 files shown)


RMcGirr83:
!set Event

phpBB/includes/functions.php
```diff
@@ -4279,13 +4283,14 @@ function obtain_users_online_string($online_users, $item_id = 0, $item = 'forum'
 		* @var	string	item			Restrict online users to a certain
 		*								session item, e.g. forum for
 		*								session_forum_id
-		* @var	string	sql				SQL query to obtain users online data
+		* @var	string	sql_ary				SQL query to obtain users online data
```
Nicofuma: you have an extra tab here => `		* @var	string	sql_ary			SQL query to obtain users online data`

RMcGirr83:
Okay will fix and update pull request

PR #3952, opened 2015-10-08
[task/phpunit5] Fix php7 tests

Nicofuma:
Several tests to try to make php7 happy


marc1706:
Current issues on PHP7 seem to be caused by the use of extract() and this bug in PHP7 itself:
https://bugs.php.net/bug.php?id=70682

PR #3954, opened 2015-10-08
fix invalid redirect from http://... to https://...:80 when host is u…

scootergrisen:
…sing 80 for HTTPS SERVER_PORT

See:
https://www.phpbb.com/community/viewtopic.php?f=466&t=2293861
https://tracker.phpbb.com/browse/PHPBB3-13256

### DIFF
--- a/phpBB/includes/functions.php
+++ b/phpBB/includes/functions.php
@@ -2255,15 +2255,6 @@ function generate_board_url($without_script_path = false)
 		$script_path = $user->page['root_script_path'];
 	}
 
-	if ($server_port && (($cookie_secure && $server_port <> 443) || (!$cookie_secure && $server_port <> 80)))
-	{
-		// HTTP HOST can carry a port number (we fetch $user->host, but for old versions this may be true)
-		if (strpos($server_name, ':') === false)
-		{
-			$url .= ':' . $server_port;
-		}
-	}
-
 	if (!$without_script_path)
 	{
 		$url .= $script_path;


phpBB/includes/functions.php
```diff
@@ -2255,15 +2255,6 @@ function generate_board_url($without_script_path = false)
 		$script_path = $user->page['root_script_path'];
 	}
 
-	if ($server_port && (($cookie_secure && $server_port <> 443) || (!$cookie_secure && $server_port <> 80)))
```
Nicofuma: this is necessary in case you run phpBB under a not conventional port

Nicofuma: does it fixes your issue if in request.php you change `is_secure()` to 
```php
	public function is_secure()
	{
		$https = $this->server('HTTPS');
		
		return !empty($https) && strtolower($https) !== 'off';
	}
```

scootergrisen: Changing is_secure() to the above does not fix the problem.

marc1706:
Have you checked if setting "Force server URL settings:" in the ACP server settings (General -> Server configuration -> Server settings) solves the described issue?

scootergrisen:
Forcing the setting does seem to "work" but it says its for forcing either http or https.
I don't want to force the user to use one or the other. I want them both to work.

marc1706:
You have both http and https running on the same server port? That just sounds like an improper setup on the server side. As far as I'm aware, apache does not support this.

scootergrisen:
Yeah i tried asking my webhotel about changing SERVER_PORT but they don't want to.
SERVER_PORT is in PHP always 80 both from http and https.

I think it is suppose to be 443 when visiting from https.

Strange thing now.... i can't seem to reproduce the error.
I just forced that setting, uploaded my functions.php with the out commented code, then reversed the setting and uploaded again with the commented code (like before).... but now i cant seem to get it to _not_ work.

Strange.

marc1706:
Closing as such.

PR #3959, opened 2015-10-11
[ticket/13684] Prevent horizontal resize of attach comment textarea

marc1706:
Replaces #3482 

https://tracker.phpbb.com/browse/PHPBB3-13684


prototech:
Doesn't seem like autowidth is of much use in reality. :P Let's go with a 100% width on that instead?

marc1706:
The autowidth seemed to work fine for me?

marc1706:
autowidth looked fine on full HD but not on other resolutions. This seems to look fine though.

PR #3960, opened 2015-10-11
[ticket/14229]Responsive topic specific paging

hanakin:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-14229
image

PHPBB3-14229


hanakin:
!set Style

Nicofuma:
It used to be displayed like that but I like how it looks with this PR

marc1706:
Are the report and topic approval icons supposed to split lines?
bildschirmfoto vom 2015-10-17 12 23 16
This happens on latest Chrome right before the <dd> tags are hidden while reducing the viewport size.

hanakin:
its inline-block need to add inline to .icons possibly

PR #3970, opened 2015-10-14
[ticket/13789] Upgrade reCaptcha plugin to 2.0 API.

prototech:
[PHPBB3-13789](https://tracker.phpbb.com/browse/PHPBB3-13789)

### DIFF
--- a/phpBB/composer.lock
+++ b/phpBB/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at http://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "3a1fb55478d50b5b7b5888f3a07e0b43",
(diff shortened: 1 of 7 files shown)


marc1706:
The lock file is not up-to-date. Please run this command to update the lock file for the recaptcha entry:
```php ../composer.phar update google/recaptcha```

**Comment by Nicofuma** [comment]
@prototech you forgot to commit the .lock file ;)

**Comment by prototech** [comment]
Updated.

File: phpBB/composer.lock [hunk 1]
```diff
@@ -1977,12 +2022,12 @@
             "version": "v1.0.2",
             "source": {
                 "type": "git",
-                "url": "https://github.com/silexphp/Pimple.git",
```
marc1706: @Nicofuma why are these changes still happening?

Nicofuma: I really don't know... especially that this change shouldn't happen in this direction...

PR #3971, opened 2015-10-14
[ticket/14237] Use $language class for notifications

VSEphpbb:
Includes other minor fixes to the notifications classes I found while working on this.

PHPBB3-14237

### DIFF
--- a/phpBB/phpbb/notification/manager.php
+++ b/phpBB/phpbb/notification/manager.php
@@ -848,7 +848,7 @@ class manager
 	 *
 	 * @param string $notification_type_name
 	 * @param array  $data
-	 * @return \phpbb\notification\type\type_interface
+	 * @return type\type_interface
 	 */
 	public function get_item_type_class($notification_type_name, $data = array())
 	{
@@ -863,7 +863,7 @@ class manager
 	 * Helper to get the notifications method class and set it up
 	 *
 	 * @param string $method_name
-	 * @return \phpbb\notification\method\method_interface
+	 * @return method\method_interface
 	 */
 	public function get_method_class($method_name)
 	{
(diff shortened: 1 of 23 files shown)


Nicofuma:
Couldn't it be possible to get ride of the language dependency here, use language strings everywhere and translate them in the template using the `lang()` function and/or the maybe incoming `lang` filter?

VSEphpbb:
Doesn't seem like it. It's using the sprintf thing in some places, as well as lang's is_set() method for some lang keys (see report_pm for example). Not to mention add_lang()

Nicofuma:
humf ok :/

phpBB/phpbb/notification/manager.php
```diff
@@ -858,7 +873,8 @@ public function get_method_class($method_name)
 	/**
 	 * Helper to load objects (notification types/methods)
 	 *
-	 * @return method\method_interface|type\type_interface
+	 * @param string $object_name
+	 * @return \phpbb\notification\method\method_interface|\phpbb\notification\type\type_interface
```
Nicofuma: `method\method_interface|type\type_interface`

phpBB/phpbb/notification/manager.php
```diff
@@ -848,7 +862,8 @@ public function get_item_type_class($notification_type_name, $data = array())
 	/**
 	 * Helper to get the notifications method class and set it up
 	 *
-	 * @return method\method_interface
+	 * @param string $method_name
+	 * @return \phpbb\notification\method\method_interface
```
Nicofuma: `method\method_interface`

phpBB/phpbb/notification/manager.php
```diff
@@ -834,7 +846,9 @@ public function get_default_methods()
 	/**
 	 * Helper to get the notifications item type class and set it up
 	 *
-	 * @return type\type_interface
+	 * @param string $notification_type_name
+	 * @param array  $data
+	 * @return \phpbb\notification\type\type_interface
```
Nicofuma: `method\method_interface`

You need to be consistent actually, pick one form and use it everywhere

VSEphpbb: I think my IDE caused this

PR #3978, opened 2015-10-14
[ticket/13851] Fix ignore flood limit check

RMcGirr83:
[PHPBB3-13851](https://tracker.phpbb.com/browse/PHPBB3-13851)

### DIFF
--- a/phpBB/posting.php
+++ b/phpBB/posting.php
@@ -1027,7 +1027,8 @@ if ($submit || $preview || $refresh)
 		$message_parser->bbcode_bitfield = $post_data['bbcode_bitfield'];
 	}
 
-	if ($mode != 'edit' && !$preview && !$refresh && $config['flood_interval'] && !$auth->acl_get('f_ignoreflood', $forum_id))
+	$ignore_flood = $auth->acl_get('u_ignoreflood') ? true : $auth->acl_get('f_ignoreflood', $forum_id);
+	if ($mode != 'edit' && !$preview && !$refresh && $config['flood_interval'] && !$ignore_flood)
 	{
 		// Flood check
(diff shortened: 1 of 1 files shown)


RMcGirr83:
https://tracker.phpbb.com/browse/PHPBB3-13851

Nicofuma:
This is not an easy question.

The problem is that we have 2 permissions (one user based and one forum based) and that we have only 2 status for the permissions: yes/no and the default value for f_ignoreflood is no...
Therefore there is some case to decide:

u_ignoreflood | f_ignoreflood | result
------------------|--------------------|--------
yes|yes|yes
yes|no| ?
no|no|no
no|yes| ?

With this PR the result is the following (and as you see, f_ignoreflood can be removed):
u_ignoreflood | f_ignore_flood | result
------------------|--------------------|--------
yes|yes|yes
yes|no| yes
no|no|no
no|yes| no

@DavidIQ

RMcGirr83:
@Nicofuma I don't agree with this logic

u_ignoreflood | f_ignoreflood | result
----------------- | ----------------- | -------
no | yes | no

believe it would be instead?

u_ignoreflood | f_ignoreflood | result
----------------- | ----------------- | -------
no | yes | yes

In fact I did test it by setting groups can ignore flood permission to no and forum group permission to yes and it seems to work as I state above.

Nicofuma:
right

Nicofuma:
fine with me this way

marc1706:
Fine with me, too.

DavidIQ:
Since I was tagged, although a little late in responding: yes the user permission should take precedence over the forum permission, as everyone else has stated.

Nicofuma:
thanks

PR #3979, opened 2015-10-14
[ticket/9153] Allow users without u_sendpm permission to delete existing PM's

prototech:
The u_sendpm check in ucp_pm.php is merely for the purpose of displaying the
error message within the UCP template. Otherwise it could be removed as
compose_pm() will check the appropriate permission pertaining to the attempted
action.

[PHPBB3-9153](https://tracker.phpbb.com/browse/PHPBB3-9153)


Nicofuma:
:+1:

PR #3981, opened 2015-10-14
[ticket/14235] Font Awesome in simple_header.html

ErnadoO:
PHPBB3-14235


Nicofuma:
This is the best way indeed :) Thanks for your contribution

PR #3984, opened 2015-10-15
[ticket/12616] Remove report notifications when post is deleted/disapproved.

prototech:
[PHPBB3-12616](https://tracker.phpbb.com/browse/PHPBB3-12616)

### DIFF
--- a/phpBB/includes/functions_admin.php
+++ b/phpBB/includes/functions_admin.php
@@ -829,6 +829,7 @@ function delete_posts($where_type, $where_ids, $auto_sync = true, $posted_sync =
 		'notification.type.quote',
 		'notification.type.approve_post',
 		'notification.type.post_in_queue',
+		'notification.type.report_post',
 	);
 
 	/**


bantu:
There appear to be quite some weirdnesses when trying this using a single user (admin) account:
* Change permissions to require approval
* Post something
* The user itself gets a notification for <<Post approval request by admin: "Test Topic 1">> but this notification is already read and the top shows <<Notifications [0]>> (new issue?)
* The notification redirects to the approval screen where the message can be approved.
* Checking "Notify poster about approval?" creates a new notification about the approval which is unread (this works properly)
* Regardless of whether "Notify poster about approval?" is checked or not, the original <<Post approval request by admin: "Test Topic 1">> notification is removed (i.e. I can not reproduce this issue).

prototech:
The ticket is oddly worded and rather vague I suppose. The issue here is with reported posts. The approved state doesn't make any difference upon further inspection. If the reported post is disapproved or deleted, the notification stating "Post reported by X" will be left behind. Clicking the link for that results in a "No report found" error.

prototech:
I believe the "Post approval" notification is intentionally marked read for the author seeing as he would be aware of the post's existence and need for approval in the posting confirmation screen.

bantu:
Oh oops. I think i misread.

PR #3989, opened 2015-10-17
Fix on memberlist the sort method.

Leinad4Mind:
This way, after clicking on posts or other, it will sort desc. so the high numbers will apear.
In asc mode, we need to click twice to be able to see who posted more, etc.
Just the username makes sense to be as asc mode.

https://tracker.phpbb.com/browse/PHPBB3-14242

### DIFF
--- a/phpBB/memberlist.php
+++ b/phpBB/memberlist.php
@@ -88,7 +88,7 @@ $submit = (isset($_POST['submit'])) ? true : false;
 
 $default_key = 'c';
 $sort_key = request_var('sk', $default_key);
-$sort_dir = request_var('sd', 'a');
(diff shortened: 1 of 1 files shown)


CHItA:
Your commit message isn't in the [required format](https://wiki.phpbb.com/Git#Commit_Messages).
Also your branch should have the name `ticket/14242`. If you don't know how to do these things, feel free to ask for further help.

Thanks for your contribution!

Leinad4Mind:
Sorry don't know how to do it. :/

CHItA:
All right, we can do that when merging this PR.

prototech:
Replaced by #3996

PR #3995, opened 2015-10-18
[ticket/14044] Deduplicate the installers

CHItA:
[PHPBB3-14044](https://tracker.phpbb.com/browse/PHPBB3-14044?filter=-1)


Nicofuma:
By the way, could you base this PR on your updater PR?

Nicofuma:
A few things to fix the tests:
https://gist.github.com/Nicofuma/0913d5aab83e6854a7b6
and then add at the end of install_board: (to avoid a giant memory leak)
```

		global $phpbb_container, $cache, $phpbb_dispatcher, $request, $user, $auth, $db, $config, $phpbb_log, $symfony_request, $phpbb_filesystem, $phpbb_path_helper, $phpbb_extension_manager, $template;
		$phpbb_container->reset();
		unset($phpbb_container, $cache, $phpbb_dispatcher, $request, $user, $auth, $db, $config, $phpbb_log, $symfony_request, $phpbb_filesystem, $phpbb_path_helper, $phpbb_extension_manager, $template);
```

Nicofuma:
You kept the wrong language file for `language/en/install.php` ~100 keys are missing (the file with this PR: https://github.com/CHItA/phpbb/blob/ticket/14044/phpBB/language/en/install.php install_new.php: https://github.com/phpbb/phpbb/blob/cc1a96a8045b6abe5b1234afb045912d53669fee/phpBB/language/en/install_new.php)

CHItA:
Fixed

Nicofuma:
Still an issue with the database updater due to the constants used in the migrations: in `constants.php` you need to make `$table_prefix` available by adding `global $table_prefix;`. I hate globals :/

CHItA:
No idea why it is not working... it is a global where `constants.php` is included

CHItA:
fixed?

Nicofuma:
I still have the 2 issues when installing:
- The progress bar isn't displayed, I have a spinner instead
- `The "../language//email" directory does not exist.`

But it's not a blocker for the alpha1

CHItA:
both should be fixed

Nicofuma:
```
<Nicofuma> Found the issue :/
<Nicofuma> #language_fallback: array:3 [▼ 0 => "" 1 => "" 2 => "en" ]
<Nicofuma> don't why but that's what I have
<Nicofuma> I also found out why I don't have the progress bar, that's just because no data is sent through the connection (no log, nothing)
<Nicofuma> the use class is using the global config to fetch the defaut language :/
<Nicofuma> the global config and the config returned by the container are wrong :/
<Nicofuma> but the config object created manually in notify_user is good
 ```
I can see 2 solutions:
1) Try to found out why the config object is wrong it only contains the rand_seed)
2) Set the default language manually in notify_user using $this->config (which is constructed manually)

By the way please apply this small patch to language file:
```
-if ($this->user_language !== false)
+if ($this->user_language)
{
	$fallback_array[] = $this->user_language;
}

-if ($this->default_language !== false)
+if ($this->default_language)
{
	$fallback_array[] = $this->default_language;
}
```

**Comment by Nicofuma** [comment]
about the travis error, adding `global $table_prefix;` to `tests/bootstrap.php` line 19 is enough

**Comment by CHItA** [comment]
"The progress bar isn't displayed, I have a spinner instead" issue is probably caused by OB settings on your server, i don't think we can do too much improvement on that from PHP but that can wait till a2.

All the other things seems to be fixed now.

**Comment by Elsensee** [comment]
In your startup.php:
```
$other_config_path = $phpbb_root_path . 'install/update/new/config';
$config_path = (file_exists($other_config_path . '/installer/config.yml')) ? $other_config_path : $phpbb_root_path . 'config';
```
I don't think this is enough. Will the complete folder be included if there's even the slightest change in the config folder? If not, then we have a problem.

CHItA:
That is solved for a while now, see `build/package.php`

Nicofuma:
Their is a last issue: the messenger class is using the global $config object to determine the language fallback. But.... in this case the global object is wrong (that's why you are creating one manually in the finish task) and it leads to an error (dir doesn't exists)

CHItA:
@Nicofuma

PR #3996, opened 2015-10-19
[ticket/14242] Fix memberlist sort method

RMcGirr83:
This way, after clicking on posts or other, it will sort desc. so the high numbers will apear.
In asc mode, we need to click twice to be able to see who posted more, etc.
Just the username makes sense to be as asc mode.

[PHPBB3-14242](https://tracker.phpbb.com/browse/PHPBB3-14242)

### DIFF
--- a/phpBB/memberlist.php
+++ b/phpBB/memberlist.php
@@ -88,7 +88,7 @@ $submit = (isset($_POST['submit'])) ? true : false;
 
 $default_key = 'c';
 $sort_key = request_var('sk', $default_key);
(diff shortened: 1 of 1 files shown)


RMcGirr83:
https://github.com/phpbb/phpbb/pull/3989

RMcGirr83:
El bumpage :confused:

RMcGirr83:
Crap I closed the wrong damn ticket.  El nubbage strikes again and I blame Paul

PR #4000, opened 2015-10-22
[ticket/14250] Review/bump dependencies

Elsensee:
Last night I looked through all dependencies and checked if they are really required by phpBB or if there is an update.

* Some Symfony components are never referenced in our code so I removed them. (One other was listed as dev-requirement even though we reference it in our core code so I moved it)
* I upgraded lusitanian/oauth so some extension authors might be able to integrate new OAuth services as they were added in later versions.
* patchwork/utf8 added support for php7 in 1.2 (kind of.. they just removed it from the allowed failures, so I think that might be worth it)
* I upgraded s9e/text-formatter to php5.4 for obvious reasons.
* Regarding dev requirements:
```
<Nicofuma> Elsensee: by the way, in your PR add the @dev suffix to all dev dependencies
```

[PHPBB3-14250](https://tracker.phpbb.com/browse/PHPBB3-14250)


marc1706:
Tests fail due to oauth changes.

Elsensee:
Aw man. Why don't they use semver? :(
So keep oauth at 0.8.1 but add these methods to the class or go back to 0.2.1?
I'd rather go with 0.8.1 tbh. ^^

Nicofuma:
What changed in 0.8.1?

Elsensee:
Looks like OAuth\Common\Storage\TokenStorageInterface has added some methods that we (obviously) don't implement in phpbb\auth\provider\oauth\token_storage

Nicofuma:
ok

Elsensee:
Alright. The failure is obviously that we're missing Symfony/CssSelector 2.8 (we only have CssSelector 2.7.5).
This is required by Symfony/DomCrawler 2.8 which we have listed as an requirement in our composer file.
So why doesn't composer download CssSelector 2.8 even though this is listed as an requirement by DomCrawler 2.8? This makes no sense to me.

Elsensee:
```
<alcohol> also, composer does not install dev requirements of dependencies
```
The more you know... (please note that this is a real username in irc.. in case you didn't know)

Also (for the record):
```
<Nicofuma> About @dev I meant the one we require in dev stability :p
<Nicofuma> Master - dev per example
<Nicofuma> Only symfony and text formatter actually
```

Elsensee:
I feel like having a conversation with myself.

@JoshyPHP told me that we could target the version as well.
What would you prefer? Targeting the branch or the version? I don't really care about that, tbh.

JoshyPHP:
Basically you have a choice between `^0.4.2` and `dev-release/php5.4`. If you use the former, you get to track a version number which is a bit more meaningful than a commit id. If you use the latter, you get a slightly smaller codebase and fewer workarounds that are related to PHP 5.3.

I'd softly recommend the former.

marc1706:
@Nicofuma we should probably try to incorporate these changes with the update to symfony 2.8.

Nicofuma:
Sorry but you have to rebase :)

Elsensee:
... Yep... Will do tomorrow. ^^

Elsensee:
@Nicofuma I.. erm.. rebased it now. I think I messed up a bit because I accidentally removed two commits but no changes are lost.

Nicofuma:
You still have a conflict :/

marc1706:
I think again, otherwise the tests wouldn't have run.

Nicofuma:
The conflict is about the service definition you changed

Nicofuma:
bump

Elsensee:
Okay, I rebased it. Finally. Sorry for the delay.
Will add @michaelcullum's patch in a minute if he's still awake. Otherwise it might be 2 minutes.

marc1706:
@Elsensee I hope you tested if oauth still works. :)

Elsensee:
I'll test it again, just to be sure.

Regarding symfony components. I know we want to use Symfony 2.8 so we'll have to require all Symfony components we don't require directly (but indirectly via our requirements) to be 2.8 at least.
Is there a better way?

Nicofuma:
Require symfony/symfony?

Elsensee:
We don't need all symfony components, do we?

Nicofuma:
We don't but it's a way to avoid listing all of them :p

Elsensee:
Hah. I consider that as a joke, then? :P
Although it might be a good idea. We get a bigger package but extension developers don't have to provide any symfony components that might conflict. :D :P

Nicofuma:
Well it was partially a joke...

Elsensee:
Removed symfony/debug and symfony/process since we only listed them because they were required by one of our requirements and we wanted them in 2.8. Obviously composer is smart and downloads them now in 2.8 anyway (probably because it's stable now.. whatever)
Also removed the `@dev` after the symfony dev-requirements - I forgot that there.

And, @marc1706, I just tested the OAuth stuff and it still works. Yay. :)

Nicofuma:
debug is required: https://github.com/phpbb/phpbb/pull/4076

Elsensee:
Thanks, updated.

marc1706:
I can confirm that oauth still works, even after updating while previously being logged in using oauth.

PR #3973, opened 2015-10-14
[ticket/14236] Disable Opcache on Travis when running the tests

Nicofuma:
https://tracker.phpbb.com/browse/PHPBB3-14236

PHPBB3-14236

PR #3965, opened 2015-10-12
[ticket/14234] Use replacement variables instead of references in events

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-14234

PR #4003, opened 2015-10-23
[ticket/14256] Remove PHP7 from the allowed failures

Elsensee:
We're currently doing a pretty good job at letting our test suite not fail on PHP7 and I really want us to be compatible with PHP7 so I think we should be safe now to remove this from the allowed failures.

[PHPBB3-14256](https://tracker.phpbb.com/browse/PHPBB3-14256)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,6 @@ matrix:
     - php: hhvm
       env: DB=mysqli
   allow_failures:
-    - php: 7.0
     - php: hhvm
   fast_finish: true
 


bantu:
How does this interact with #3952?

marc1706:
The PHP7 issues were fixed in other PRs. #3952 is just a leftover of testing what the real issues for the PHP7 issues in the test suite are.

marc1706:
Related PRs are https://github.com/phpbb/phpbb/pull/3973 and https://github.com/phpbb/phpbb/pull/3965

PR #4004, opened 2015-10-24
[ticket/14258] Add event into auth::login before session creation

paul999:
Add a new event in the login method just before the session create so extensions can do some additional validation during login.

PHPBB3-14258

### DIFF
--- a/phpBB/phpbb/auth/auth.php
+++ b/phpBB/phpbb/auth/auth.php
@@ -984,7 +984,7 @@ class auth
 			}
 
 			/**
-			 * TEvent is triggered after checking for valid username and password, and before the actual session creation.
+			 * Event is triggered after checking for valid username and password, and before the actual session creation.
 			 *
 			 * @event core.auth_login_session_create_before
 			 * @var	array	login				Variable containing login array
--- a/phpBB/phpbb/auth/auth.php
+++ b/phpBB/phpbb/auth/auth.php
@@ -984,14 +984,13 @@ class auth
 			}
 
 			/**
-			 * This event is triggered just after the user has been checked if he has a valid username/password,
-			 * but before the actual session creation.
+			 * TEvent is triggered after checking for valid username and password, and before the actual session creation.
 			 *
-			 * @event core.auth_login_before_session_create
-			 * @var	array	login				Variable containing login array value
-			 * @var	bool	admin				Variable containing if is logging into the acp
-			 * @var	string	username			Variable containing the username
-			 * @var	bool	autologin			Variable containing if this is by auto login
+			 * @event core.auth_login_session_create_before
+			 * @var	array	login				Variable containing login array
+			 * @var	bool	admin				Boolean variable whether user is logging into the ACP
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/auth/auth.php
```diff
@@ -982,6 +983,25 @@ function login($username, $password, $autologin = false, $viewonline = 1, $admin
 				redirect($url);
 			}
 
+			/**
+			 * This event is triggered just after the user has been checked if he has a valid username/password,
```
marc1706: I'd prefer ```Event is triggered after checking for valid username and password, and before the actual session creation```

phpBB/phpbb/auth/auth.php
```diff
@@ -982,6 +983,25 @@ function login($username, $password, $autologin = false, $viewonline = 1, $admin
...
+			 * but before the actual session creation.
+			 *
+			 * @event core.auth_login_before_session_create
```
marc1706: core.auth_login_session_create_before

phpBB/phpbb/auth/auth.php
```diff
@@ -982,6 +983,25 @@ function login($username, $password, $autologin = false, $viewonline = 1, $admin
...
+			 * @var	array	login				Variable containing login array value
```
marc1706: Variable containing login array

phpBB/phpbb/auth/auth.php
```diff
@@ -982,6 +983,25 @@ function login($username, $password, $autologin = false, $viewonline = 1, $admin
...
+			 * @var	bool	admin				Variable containing if is logging into the acp
```
marc1706: Boolean variable whether user is admin

phpBB/phpbb/auth/auth.php
```diff
@@ -982,6 +983,25 @@ function login($username, $password, $autologin = false, $viewonline = 1, $admin
...
+			 * @var	string	username			Variable containing the username
```
marc1706: Username of user to log in

phpBB/phpbb/auth/auth.php
paul999: It is not if user is admin, but if he is going to login into the ACP

phpBB/phpbb/auth/auth.php
```diff
@@ -982,6 +983,25 @@ function login($username, $password, $autologin = false, $viewonline = 1, $admin
...
+			 * @var	bool	autologin			Variable containing if this is by auto login
```
marc1706: Boolean variable signaling whether login is triggered via auto login

phpBB/phpbb/auth/auth.php
```diff
@@ -982,6 +983,24 @@ function login($username, $password, $autologin = false, $viewonline = 1, $admin
 				redirect($url);
 			}
 
+			/**
+			 * TEvent is triggered after checking for valid username and password, and before the actual session creation.
```
marc1706: TEvent? :)

PR #4005, opened 2015-10-24
[ticket/14257] Reparse after update

Elsensee:
One new cron task for each reparser.
Advantage of this approach is that the decision which reparser will run in the cron job is made before running it.
Disadvantage is that extension authors have to create their own cron task but since they only have to define that in the services.yml and they have to create their own reparser class anyway, I don't think this is a big issue.
So basically no disadvantages. :grinning: 

Cron job and console command can't interfere with eachother because the reparser acquires a lock, of course.
However, cron job and console command will be synced. So the console command can resume where the cron job stopped and vice versa.

This also means unfortunately that an administator can set weird values by using the console command (weird start, end, size IDs) which might lead to an unfinished reparsing. I don't know if this is something we have to care about?

And since this is your baby, @JoshyPHP, please have a look as well. :) ;) Thank you! :)

[PHPBB3-14257](https://tracker.phpbb.com/browse/PHPBB3-14257)

### DIFF
--- a/phpBB/language/en/acp/common.php
+++ b/phpBB/language/en/acp/common.php
@@ -289,6 +289,7 @@ $lang = array_merge($lang, array(
 	'PROCEED_TO_ACP'					=> '%sProceed to the ACP%s',
 
 	'REMIND'							=> 'Remind',
+	'REPARSE_LOCK_ERROR'				=> 'Could not obtain reparse lock.',
 	'RESYNC'							=> 'Resynchronise',
 
 	'RUNNING_TASK'			=> 'Running task: %s.',
--- a/phpBB/language/en/acp/common.php
+++ b/phpBB/language/en/acp/common.php
@@ -289,7 +289,7 @@ $lang = array_merge($lang, array(
 	'PROCEED_TO_ACP'					=> '%sProceed to the ACP%s',
 
 	'REMIND'							=> 'Remind',
-	'REPARSE_LOCK_ERROR'				=> 'Could not obtain reparse lock.',
+	'REPARSE_LOCK_ERROR'				=> 'Reparsing is already in progress by another process.',
 	'RESYNC'							=> 'Resynchronise',
 
 	'RUNNING_TASK'			=> 'Running task: %s.',
--- a/phpBB/phpbb/console/command/reparser/reparse.php
+++ b/phpBB/phpbb/console/command/reparser/reparse.php
@@ -61,9 +61,9 @@ class reparse extends \phpbb\console\command\command
 	* Constructor
 	*
 	* @param \phpbb\user $user
-	* @param \phpbb\di\service_collection $reparsers
-	* @param \phpbb\config\db_text $config_text
 	* @param \phpbb\lock\db $reparse_lock
+	* @param \phpbb\textreparser\manager $reparser_manager
+	* @param \phpbb\di\service_collection $reparsers
 	*/
 	public function __construct(\phpbb\user $user, \phpbb\lock\db $reparse_lock, \phpbb\textreparser\manager $reparser_manager, \phpbb\di\service_collection $reparsers)
 	{
--- a/phpBB/phpbb/console/command/reparser/reparse.php
+++ b/phpBB/phpbb/console/command/reparser/reparse.php
@@ -13,6 +13,7 @@
 
 namespace phpbb\console\command\reparser;
 
+use phpbb\exception\runtime_exception;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputOption;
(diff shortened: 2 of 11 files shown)


prototech:
I don't think we really need cron tasks for the forum data, contact admin info, and group descriptions. We should be able to run these from a single migration. The contact admin info is just a single config entry for example.

Elsensee:
Indeed, I didn't thought about that. ^^

phpBB/phpbb/console/command/reparser/reparse.php
```diff
@@ -163,29 +171,39 @@ protected function execute(InputInterface $input, OutputInterface $output)
 		$this->input = $input;
 		$this->output = $output;
 		$this->io = new SymfonyStyle($input, $output);
-		$this->load_resume_data();
 
-		$name = $input->getArgument('reparser-name');
-		if (isset($name))
+		if (!$this->reparse_lock->acquire())
```
JoshyPHP: Is that conditional correct? Looks like it should not be negated.

Either way, I think it would be better to simply have a conditional with only one branch that throws an exception and no else, e.g..

```php
if (!$this->reparse_lock->acquire())
{
    throw new Exception;
}

// ...the rest executes as usual
```

Elsensee: Argh, thanks. That's the way I wanted to write it first, but then I changed it (mainly to be consistent with everything else that acquires a lock) and forgot to change the conditional too. :/

JoshyPHP:
What do you think of moving the loading/saving of resume data into its own class? Perhaps even the locking mechanism? Not sure about that one. Maybe it could have a simplified API to schedule reparsing via the cron. It would be useful for extension authors I guess. You could have something like that:
```php
namespace phpbb\textreparser;
class manager
{
    public function get_resume_data($name);
    public function update_resume_data($name, $min, $current, $size);
    public function schedule($name);
}
```

Elsensee:
Well, that would reduce code duplication, so I guess we can do that.

The locking mechanism already reuses the phpbb\lock\db class. Also why simplifying it further? I think the `$reparse_lock->acquire();` and `$reparse_lock->release();` should be simple enough?

Extension authors actually don't have to write any class for this. For the cronjob they can just reuse this one and change the call in the service definition, while for the CLI command adding the tag to their reparser class is enough.

More concerning should be that an administrator can do weird stuff in the console command which can have unexpected results for the cron job, right? What do you think about that?

JoshyPHP:
What kind of weird stuff can they do? Reparse a single post and interrupt the cron maybe?

I don't know. I think that people who use CLI commands should know better, and if they interrupt the cron (by overwriting the resume data) they'll be able to restart reparsing. I don't have a strong opinion.

Elsensee:
> What kind of weird stuff can they do? Reparse a single post and interrupt the cron maybe?

Exactly.
Well I think, that we maybe could just remove the possibility to define a start and end ID in the CLI command?

Elsensee:
@JoshyPHP Writing the manager resume_data_manager right now. What's the schedule() supposed to do actually?

JoshyPHP:
`schedule('post_text')` would create/enable/update a cron task that reparses all `post_text`. Like the reparser CLI, instead of post_text it could be the name of a reparser service such as `text_reparser.poll_option`. `schedule_all()` would cause every available reparser service to be scheduled to be run via cron.

Elsensee:
Added the text_reparser manager now. (I also fixed stuff so console command and cron task now can really work together ^^)
I guess I'll remove the start, end, size options from the CLI command and then.. what's left then? I think that's it. :grin:

Nicofuma:
Why would you remove these options from the CLI command? This command isn't meant only for the update, it shouldn't be a command you can use only once in the life of your forum but anytime you need (or think you need) to reparse the posts (or a sub set) of your forum.

The console is meant to be used by advanced user, please keep it flexible

phpBB/language/en/acp/common.php
```diff
@@ -289,6 +289,7 @@
 	'PROCEED_TO_ACP'					=> '%sProceed to the ACP%s',
 
 	'REMIND'							=> 'Remind',
+	'REPARSE_LOCK_ERROR'				=> 'Could not obtain reparse lock.',
```
Nicofuma: This is a very technical message, you should explain what the issue is (ie. an other process might already ben using the reparser)

Nicofuma:
@phpbb/native ^

phpBB/language/en/acp/common.php
```diff
@@ -289,6 +289,7 @@
 	'PROCEED_TO_ACP'					=> '%sProceed to the ACP%s',
 
 	'REMIND'							=> 'Remind',
+	'REPARSE_LOCK_ERROR'				=> 'Reparsing is already in progress by another process.',
```
marc1706: How about ```Another process is already running reparsing.```?

marc1706:
:+1:

phpBB/language/en/acp/common.php
VSEphpbb: `Another reparsing process is running` ?

marc1706: ```Another reparsing process is already running```?

PR #4007, opened 2015-10-27
[ticket/13202] Remove old and unsed code from session_begin()

Zoddo:
This code was to be removed in 3.0.2. So, I think that we can safely remove it in 3.2.x.

Also, the `$result` variable doesn't longer defined by update_session(), so the variable come from a `SELECT` query few lines above and this condition is always false.

[PHPBB3-13202](https://tracker.phpbb.com/browse/PHPBB3-13202)

### DIFF
--- a/phpBB/phpbb/session.php
+++ b/phpBB/phpbb/session.php
@@ -465,15 +465,6 @@ class session
 
 							$db->sql_return_on_error(false);
 
-							// If the database is not yet updated, there will be an error due to the session_forum_id
-							// @todo REMOVE for 3.0.2
-							if ($result === false)
-							{
-								unset($sql_ary['session_forum_id']);
-
-								$this->update_session($sql_ary);
-							}
-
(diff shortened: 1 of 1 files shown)


marc1706:
I think we still need this piece of code for updates from versions prior to 3.0.2. However, we should apply the change that was suggested in the ticket as the update session code was incorrectly refactored while adding the update_session() method.

Nicofuma:
@Zoddo you need to rebase your PR.
About marc comment.... I think we can drop phpBB 3.0.1 and 3.0.2 support here

Nicofuma:
@marc1706 ^

marc1706:
@Nicofuma this was already removed by #4008 
Also, I'm pretty sure this hasn't been working since the refactor to use the update_session() method.
```$result``` was set using the previous query instead of the session update query and this resulted in it never being false. Unless there would have been an error in the previous query that is not inside sql_return_on_error() and therefore would've resulted in a general error.

marc1706:
Closing as such (see above comment).

Nicofuma:
didn't notice it was removed in the other PR. Not a big loss :)

PR #4008, opened 2015-10-27
[ticket/14261] Move the update of session informations to page_footer()

Zoddo:
Currently, the unique way to disable the update of session_page is to pass `false` to the parameter of `session_begin()`. This method is directly called in `app.php`, so pages served from the routing system can't disable the update of session informations.

By moving the update to `page_footer`, we can allow controllers to tell to the session manager that we don't want to update the session infos.

[PHPBB3-14261](https://tracker.phpbb.com/browse/PHPBB3-14261)

### DIFF
--- a/phpBB/phpbb/session.php
+++ b/phpBB/phpbb/session.php
@@ -446,39 +446,6 @@ class session
 
 					if (!$session_expired)
 					{
-						// Only update session DB a minute or so after last update or if page changes
-						if ($this->time_now - $this->data['session_time'] > 60 || ($this->update_session_page && $this->data['session_page'] != $this->page['page']))
-						{
-							$sql_ary = array('session_time' => $this->time_now);
-
-							// Do not update the session page for ajax requests, so the view online still works as intended
-							if ($this->update_session_page && !$request->is_ajax())
-							{
-								$sql_ary['session_page'] = substr($this->page['page'], 0, 199);
-								$sql_ary['session_forum_id'] = $this->page['forum'];
-							}
-
-							$db->sql_return_on_error(true);
-
-							$this->update_session($sql_ary);
-
-							$db->sql_return_on_error(false);
-
-							// If the database is not yet updated, there will be an error due to the session_forum_id
-							// @todo REMOVE for 3.0.2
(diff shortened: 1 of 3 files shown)


Nicofuma:
Could you explain me why we wouldn't update the session infos in some case?

Zoddo:
@Nicofuma: For example, if a route is used to load images or attachments of an extension.

With the current behaviour, viewonline page always display wrong informations because images are loaded after the page, so session_page contain the URL to the download route.

Currently, the unique way to fix that is to set the parameter of session_begin() to `false` ... but this method is called in `app.php` so we can't use that.

Nicofuma:
I see ok

phpBB/phpbb/session.php
```diff
@@ -1599,4 +1554,48 @@ public function update_session($session_data, $session_id = null)
...
+			$db->sql_return_on_error(false);
+
+			// If the database is not yet updated, there will be an error due to the session_forum_id
+			// @todo REMOVE for 3.0.2
+			if ($result === false)
+			{
+				unset($sql_ary['session_forum_id']);
+
+				$this->update_session($sql_ary);
+			}
```
Zoddo: `$result` doesn't exists and fails tests. I will remove this code. So, if this PR is merged, #4007 will be obselete.

Elsensee:
Why are you moving code around? Can't you just create an event?

Zoddo:
Create an event for this is hacking. 

If we create an event for this, we need to parse the URL (the routing system isn't initialized at this moment) to know if it's an URL for which we don't want to update the sessions infos.

Additionally, if for some reason we want to disable this update only in certain conditions, we need to check all these conditions ... conditions that are probably also checked in controller for the generation of the page.

For me, it's simpler to just update a variable/call a method from the controller to say that we don't want update the session page.

Zoddo:
Hi,

Any news?

rxu:
I guess this PR should be reverted as it brakes a lot of things. See 
https://tracker.phpbb.com/browse/PHPBB3-14409
https://www.phpbb.com/community/viewtopic.php?f=466&t=2353601

PR #4016, opened 2015-11-04
[ticket/14270] Purge cache when the installer is finished

CHItA:
[PHPBB3-14270](https://tracker.phpbb.com/browse/PHPBB3-14270)

### DIFF
--- a/phpBB/install/startup.php
+++ b/phpBB/install/startup.php
@@ -81,4 +81,5 @@ $config_path = (file_exists($other_config_path . '/installer/config.yml')) ? $ot
 
 $phpbb_installer_container = $phpbb_installer_container_builder
 	->with_config_path($config_path)
+	->with_custom_parameters(array('cache.driver.class' => 'phpbb\cache\driver\file'))
 	->get_container();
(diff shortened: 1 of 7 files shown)


Nicofuma:
You forgot to update the service definition ;)

CHItA:
Yeah, but it is wip still, i'm not sure if without a valid config.php `%cache.driver.class%` exists or not.

phpBB/install/startup.php
```diff
@@ -81,4 +81,5 @@ function phpbb_include_updated($path, $phpbb_root_path, $optional = false)
 
 $phpbb_installer_container = $phpbb_installer_container_builder
 	->with_config_path($config_path)
+	->with_custom_parameters(array('cache.driver.class' => 'phpbb\cache\driver\file'))
```
CHItA: @Nicofuma: Do we need anything more advanced than this?

Nicofuma: I don't like this patch because it's a hack.... but it should work and I don't see any other "small" solution :/

marc1706: This is just the installer so there are things that are really more important than manually specifying the cache type.

Nicofuma: The real issue is that the cache and php file storage are separated but we rely on the cache to purge the stored php files. We need to build a proper API for that but that will not happen before 3.3

PR #4017, opened 2015-11-05
[ticket/14262] Move convertor to controller

CHItA:
[PHPBB3-14262](https://tracker.phpbb.com/browse/PHPBB3-14262)

TODO:
- [x] Add missing language vars
- [x] Connect the controller and converter logic
- [x] Investigate the missing language selector
- [x] Implement navigation stage tracking
- [x] Keep the main Convert menu as active (broken by the use of multiple routes)
- [x] Add missing template structure


marc1706:
@CHItA bump

CHItA:
basically the convert backend itself is what's missing, i hope it can be copy pasted and then it should be ready for the weekend

Nicofuma:
Bump: status?

CHItA:
https://github.com/phpbb/phpbb/pull/4017/files#diff-5faa55522410c6deccdf20d8942d2c5fR337 and the related legacy file should be connected but i won't have time until wednesday to do anything.

Nicofuma:
ok

CHItA:
@phpbb/development-team ready for testing. it is ugly here and there but works so i don't tell anyone about it if you guys don't

CHItA:
@prototech if you have some time and could have a look, it would be much appreciated

Nicofuma:
Actually it hasn't been converted to the new installer. Just converted to controllers. Rigth?

CHItA:
more or less, some parts been moved but mostly because the corresponding parts of the old installer have been removed

Nicofuma:
In `phpBB/install/convert/convertor.php` you need to implement error and db_error (at least display them in the template please)

Something like 
```
$this->template->assign_block_vars('errors', array(...));
```

phpBB/install/convert/convert.php
```diff
@@ -0,0 +1,60 @@
...
+	// Number of rows to be inserted at once (extended insert) if supported
+	// For installations having enough memory a value of 60 may be good.
+	var $num_wait_rows = 20;
+
+	// Mysqls internal recoding engine messing up with our (better) functions? We at least support more encodings than mysql so should use it in favor.
+	var $mysql_convert = false;
+
+	var $p_master;
+
+	function convert(&$p_master)
```
Nicofuma: should be

```

	public function __construct($p_master)
	{
		$this->p_master = $p_master;
	}
```

Nicofuma:
besides my 2 comments, :+1:

CHItA:
Done

Nicofuma:
:+1:

phpBB/install/convert/controller/convertor.php
```diff
@@ -0,0 +1,857 @@
...
+		));
+
+		$convertors = $sort = array();
+		$get_info = true; // Global flag
+
+		$handle = @opendir($this->phpbb_root_path . 'install/convertors/');
+
+		if (!$handle)
+		{
+			die('Unable to access the convertors directory');
```
Elsensee: Is it intentional?

CHItA: more or less, that is not the ugliest thing with this anyway

Elsensee: Thought it was a leftover and you wanted to change it into an exception or something like that.

PR #4021, opened 2015-11-06
[ticket/14272] Use html5 valid input elements

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-14272

### DIFF
--- a/phpBB/includes/functions_acp.php
+++ b/phpBB/includes/functions_acp.php
@@ -269,7 +269,7 @@ function build_cfg_template($tpl_type, $key, &$new, $config_key, $vars)
 		case 'datetime-local':
 		case 'month':
 		case 'week':
-			$tpl = '<input id="' . $key . '" type="' . $tpl_type[0] . '" name="' . $name . '" value="' . $new[$config_key] . '"' . (($tpl_type[0] === 'password') ?  ' autocomplete="off"' : '') . ' />';
+			$tpl = '<input id="' . $key . '" type="' . $tpl_type[0] . '" name="' . $name . '" value="' . $new[$config_key] . '" />';
 		break;
 
 		case 'date':
--- a/phpBB/includes/functions_acp.php
+++ b/phpBB/includes/functions_acp.php
@@ -254,6 +254,16 @@ function build_cfg_template($tpl_type, $key, &$new, $config_key, $vars)
 		case 'text':
 		case 'url':
 		case 'email':
+		case 'tel':
+		case 'search':
+			// maxlength and size are only valid for these types and will be
+			// ignored for other input types.
+			$size = (int) $tpl_type[1];
+			$maxlength = (int) $tpl_type[2];
+
+			$tpl = '<input id="' . $key . '" type="' . $tpl_type[0] . '"' . (($size) ? ' size="' . $size . '"' : '') . ' maxlength="' . (($maxlength) ? $maxlength : 255) . '" name="' . $name . '" value="' . $new[$config_key] . '"' . (($tpl_type[0] === 'password') ?  ' autocomplete="off"' : '') . ' />';
+		break;
+
 		case 'color':
 		case 'date':
 		case 'time':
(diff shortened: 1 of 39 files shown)


marc1706:
@cyberalien  @hanakin

Elsensee:
Can't we use the min/max values from php code?

cyberalien:
As Elsensee wrote, using values from php makes more sense than random numbers.

Those maxlength attributes weren't specific before, so using them as base for new max value doesn't make sense.

phpBB/includes/functions_acp.php
```diff
@@ -254,46 +254,49 @@ function build_cfg_template($tpl_type, $key, &$new, $config_key, $vars)
...
 
 			$tpl = '<input id="' . $key . '" type="' . $tpl_type[0] . '"' . (($size) ? ' size="' . $size . '"' : '') . ' maxlength="' . (($maxlength) ? $maxlength : 255) . '" name="' . $name . '" value="' . $new[$config_key] . '"' . (($tpl_type[0] === 'password') ?  ' autocomplete="off"' : '') . ' />';
 		break;
 
+		case 'color':
+		case 'datetime':
+		case 'datetime-local':
+		case 'month':
+		case 'week':
+			$tpl = '<input id="' . $key . '" type="' . $tpl_type[0] . '" name="' . $name . '" value="' . $new[$config_key] . '"' . (($tpl_type[0] === 'password') ?  ' autocomplete="off"' : '') . ' />';
```
Elsensee: I owe you a beer if these will ever be passwords.

marc1706: What about buffer overflows?

Elsensee: Very, very unlikely.

marc1706:
Please explain how you'd use min/max values from php. For some of these settings, this is utterly pointless. The current change implements the max length / size settings using min/max settings instead of incorrectly using max length and/or size.

hanakin:
@marc1706 i believe he meant what the max field length in the DB or the setting should match what the max input is. ie for avatar sizes the user should not be able to enter 999 if the max for it is 350

marc1706:
@hanakin the issue with avatar width and height is that part of the form is hidden by default and setting proper min and max values will prevent you from submitting the form at all.

hanakin:
@marc1706 not sure I follow will have to test it out to see what you mean maybe. If the hidden issues is cousing a propblem then that is something that needs fix in whatever js is used to control that. Im at work now but I look at it later if i get time

marc1706:
I guess we'd have to disable all form elements that are not visible.

hanakin:
that should be the case anyway. If its not visible its because that content is not necessary so it should be disabled to prevent sending a null value forward especially if the field is marked required when it is nessecary

marc1706:
I'll have to look into this further. There is some JS function taking care of this and I'll have to take a further look on how we actually hide those elements.

marc1706:
@hanakin added min and max values for avatar settings now

hanakin:
@marc1706 lookin good for me but ill let @cyberalien chime in since it was his point to begin with

marc1706:
bump @cyberalien

cyberalien:
@marc1706 sorry, I haven't been working with phpbb for few months. Will take a look asap.

cyberalien:
Avatar settings: missing values for uploaded avatars. Maxlength used to be set to 3, which is a reasonable number, so max should be set to 999.

I don't understand logic behind setting max width/height values for remote avatars based on config value for uploaded avatars. It makes configuring avatars more complex if admin only wants to allow remote avatars.

Max file size for attachments is set to 999999999999999, but it is limited by php settings. Wouldn't it be more helpful for admins to use php setting and add php attachments limit below it in _explain string?

marc1706:
Uploaded avatars do not have a setting for width or height.

The remote avatars are already checked against the config value set in the acp for all avatars (min/max width/height).

We can't just limit the input of the max filesize via the input element as admins are also able to choose to input Bytes, KiB, or MiB.

marc1706:
Added a minor change that will allow setting the maximum filesize to something like ```1.5 MiB```

Nicofuma:
@cyberalien bump
@marc1706 bump

marc1706:
Is IMHO ready to merge. Not sure if cyberalien has anymore input?

cyberalien:
Yes, I think its ready

Nicofuma:
Please send another PR against 3.2 and solving the merge conflict. Be careful to to functions_acp (at least) because of the `new_ary` stuff and all new changes

SiteSplat:
This is creating some issues. https://tracker.phpbb.com/browse/PHPBB3-14615 New standards should be properly tested and not rushed. Especially when we want to see HTML5 implementation. Why worrying about min and max when you sill use unneeded slash /> to close your inputs or redundant markup like type="text/javascript" for example. The theme needs some major overhaul if you want to truly implement HTML5 and the (style) team has know what that implementation implies and how it works to avoid redundancy or a painful UX...

DavidIQ:
Thanks for the feedback but this is not the place to give it as this merge request has been merged and closed already and so very few people will see your comment.

SiteSplat:
David, you read the comment and so did the other participants involved on this.  + I created a ticket. 
That's good enough for the right people involved. ;) No point on posting this on area51 where you would get a bunch of people who just spend their day arguing for things they do not comprehend or rather then being productive...

DavidIQ:
The correct place for your comments should have been the ticket that you created, not a closed pull request where you're reaching a bare minimum of people that may or may not take action on your comments and where there's absolutely nothing that can be done.

PR #4023, opened 2015-11-08
[ticket/14277] Fix undefined index error in migrations

CHItA:
[PHPBB3-14277](https://tracker.phpbb.com/browse/PHPBB3-14277)

### DIFF
--- a/phpBB/assets/javascript/installer.js
+++ b/phpBB/assets/javascript/installer.js
@@ -247,8 +247,7 @@
 
 		try {
 			responseObject = JSON.parse(messageJSON);
-		}
-		catch (err) {
+		} catch (err) {
 			if (window.console) {
 				console.log('Failed to parse JSON object\n\nMessage: ' + err.message + '\n\nServer Response: ' + messageJSON);
 			} else {
--- a/phpBB/assets/javascript/installer.js
+++ b/phpBB/assets/javascript/installer.js
@@ -243,8 +243,21 @@
 	 */
 	function parseMessage(messageJSON) {
 		$('#loading_indicator').css('display', 'none');
+		var responseObject;
 
-		var responseObject = JSON.parse(messageJSON);
+		try {
+			responseObject = JSON.parse(messageJSON);
+		}
+		catch (err) {
+			if (window.console) {
+				console.log('Failed to parse JSON object\n\nMessage: ' + err.message + '\n\nServer Response: ' + messageJSON);
+			} else {
+				alert('Failed to parse JSON object\n\nMessage: ' + err.message + '\n\nServer Response: ' + messageJSON);
+			}
+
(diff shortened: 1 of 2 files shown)


phpBB/phpbb/db/migration/tool/module.php
```diff
@@ -262,7 +262,7 @@ public function add($class, $parent = 0, $data = array())
 
 			// Success
 			$module_log_name = ((isset($this->user->lang[$data['module_langname']])) ? $this->user->lang[$data['module_langname']] : $data['module_langname']);
-			$phpbb_log->add('admin', $user->data['user_id'], $user->ip, 'LOG_MODULE_ADD', false, array($module_log_name));
+			$phpbb_log->add('admin', ANONYMOUS, $user->ip, 'LOG_MODULE_ADD', false, array($module_log_name));
```
marc1706: I'd prefer checking if user id exists and use it if it is. Or will this always be undefined?

CHItA: As far as I'm aware yes. But that might change in the future.

phpBB/assets/javascript/installer.js
```diff
@@ -243,8 +243,21 @@
...
 	function parseMessage(messageJSON) {
 		$('#loading_indicator').css('display', 'none');
+		var responseObject;
 
-		var responseObject = JSON.parse(messageJSON);
+		try {
+			responseObject = JSON.parse(messageJSON);
+		}
+		catch (err) {
+			if (window.console) {
```
marc1706: I'd rather not output the error via console or alert. Just catch it and reset the polling by default.

phpBB/assets/javascript/installer.js
```diff
@@ -243,8 +243,21 @@
 	 */
 	function parseMessage(messageJSON) {
 		$('#loading_indicator').css('display', 'none');
+		var responseObject;
 
-		var responseObject = JSON.parse(messageJSON);
+		try {
+			responseObject = JSON.parse(messageJSON);
+		}
+		catch (err) {
```
marc1706: Needs to be ```} catch (err) {```

phpBB/assets/javascript/installer.js
CHItA: I disagree, as in that case, you get no notification about the installer failing.

marc1706: I guess we can leave this in for now but I'd rather not have alert messages popping up in a stable release ...

CHItA: only in IE8

marc1706: Anyway, pls poke me once tests finished so I can try again ;)

PR #4024, opened 2015-11-09
[ticket/14278] Check if user_id is set and fall back to ANONYMOUS

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-14278

### DIFF
--- a/phpBB/phpbb/install/module/update_database/task/update.php
+++ b/phpBB/phpbb/install/module/update_database/task/update.php
@@ -172,7 +172,7 @@ class update extends task_base
 		{
 			$this->log->add(
 				'admin',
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/install/module/install_finish/task/notify_user.php
```diff
@@ -141,7 +141,7 @@ public function run()
 		// Create log
 		$this->log->add(
 			'admin',
-			$this->user->data['user_id'],
+			(isset($this->user->data['user_id'])) ? $this->user->data['user_id'] : ANONYMOUS,
```
CHItA: This change is unnecessary as a session is initialized at this point.

PR #4026, opened 2015-11-10
[ticket/14281] Fix installer CLI after recent changes

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-14281

### DIFF
--- a/phpBB/phpbb/install/console/command/install/config/show.php
+++ b/phpBB/phpbb/install/console/command/install/config/show.php
@@ -96,7 +96,7 @@ class show extends \phpbb\console\command\command
 
 		if (!is_file($config_file))
 		{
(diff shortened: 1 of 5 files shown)


Nicofuma:
:+1:

phpBB/phpbb/install/console/command/install/config/show.php
```diff
@@ -96,7 +96,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
 		if (!is_file($config_file))
 		{
-			$iohandler->add_error_message(array('MISSING_FILE', array($config_file)));
```
CHItA: This doesn't seem to be right

marc1706: Should be correct now.

PR #4030, opened 2015-11-14
[ticket/14283] Add a Manage Group link on a group page

RMcGirr83:
https://tracker.phpbb.com/browse/PHPBB3-14283

PHPBB3-14283

### DIFF
--- a/phpBB/memberlist.php
+++ b/phpBB/memberlist.php
@@ -1088,7 +1088,7 @@ switch ($mode)
 		if ($mode == 'group')
 		{
 			// We JOIN here to save a query for determining membership for hidden groups. ;)
-			$sql = 'SELECT g.*, ug.user_id
+			$sql = 'SELECT g.*, ug.user_id, ug.group_leader
 				FROM ' . GROUPS_TABLE . ' g
 				LEFT JOIN ' . USER_GROUP_TABLE . ' ug ON (ug.user_pending = 0 AND ug.user_id = ' . $user->data['user_id'] . " AND ug.group_id = $group_id)
 				WHERE g.group_id = $group_id";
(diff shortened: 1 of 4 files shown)


phpBB/memberlist.php
```diff
@@ -1147,6 +1147,20 @@
 					$user_rank_data['img'] .= '<br />';
 				}
 			}
+			// include modules for manage groups link display or not
+			// need to ensure the modulr is active
```
Mauron: _module_

phpBB/memberlist.php
```diff
@@ -1147,6 +1147,20 @@
 					$user_rank_data['img'] .= '<br />';
 				}
 			}
+			// include modules for manage groups link display or not
+			// need to ensure the module is active
```
marc1706: While I can see the intention, I'm not sure if I approve of the general implementation. Including the module functions and instantiating a new instance of p_master() adds overhead that I'm not sure is acceptable.

RMcGirr83: It is, pretty much, no different than viewing a profile (p_master is called then too).  As far as overhead without the edits I get `Queries: 13 | Peak Memory Usage: 10.02 MiB` with the edits I get on first run `Queries: 14 | Peak Memory Usage: 10.21 MiB` one additional query to build the cache and .19mb of memory.  Could add a `if user_logged in` check to only run for registered users but the "overhead" seems a bit minimal to me.  It would seem that visiting a user profile creates more overhead than this does?

Up to devs really, can't really see another way to accomplish the task and handle the ticket request.

RMcGirr83:
El bumpage por favor @marc1706

phpBB/memberlist.php
```diff
@@ -1147,6 +1147,24 @@
...
+			$can_manage_group = false;
+			if ($user->data['is_registered'])
```
marc1706: Check if the user is a group leader here and not after instantiating p_master, listing the modules, and checking if the module is even active. This should get rid of this overhead for most users. Otherwise we'll add this check every time any user accesses this page without any use to it.

phpBB/memberlist.php
```diff
@@ -1159,7 +1177,8 @@
 				'RANK_IMG'		=> $user_rank_data['img'],
 				'RANK_IMG_SRC'	=> $user_rank_data['img_src'],
 
-				'U_PM'			=> ($auth->acl_get('u_sendpm') && $auth->acl_get('u_masspm_group') && $group_row['group_receive_pm'] && $config['allow_privmsg'] && $config['allow_mass_pm']) ? append_sid("{$phpbb_root_path}ucp.$phpEx", 'i=pm&amp;mode=compose&amp;g=' . $group_id) : '',)
+				'U_PM'			=> ($auth->acl_get('u_sendpm') && $auth->acl_get('u_masspm_group') && $group_row['group_receive_pm'] && $config['allow_privmsg'] && $config['allow_mass_pm']) ? append_sid("{$phpbb_root_path}ucp.$phpEx", 'i=pm&amp;mode=compose&amp;g=' . $group_id) : '',
+				'U_MANAGE'		=> ($can_manage_group) ? append_sid("{$phpbb_root_path}ucp.$phpEx", 'i=ucp_groups&amp;mode=manage') : false,)
```
marc1706: I guess the parenthesis at the end of this line shouldn't be there?

dragosvr92:
Thanks for the patch @RMcGirr83 :smile: Now i cant wait to upgrade to 3.1.7.

PR #4035, opened 2015-11-18
[ticket/14112] Add variable to index event for board statistics

marc1706:
Replaces #3845 

Ticket: https://tracker.phpbb.com/browse/PHPBB3-14112


rxu:
I guess this is not needed as you can just reassign any of those vars in extension over the old values.

marc1706:
Closing as such.

PR #4036, opened 2015-11-20
[ticket/14301] Fix missing subject line in subsilver2

RMcGirr83:
Within contact page if using as a registered user.

https://tracker.phpbb.com/browse/PHPBB3-14301

PHPBB3-14301


RMcGirr83:
Crap, already done.  Why isn't it in my damn repo!!

PR #4044, opened 2015-11-25
[ticket/14313] Don't display quote button on unapproved posts

RMcGirr83:
https://tracker.phpbb.com/browse/PHPBB3-14313

PHPBB3-14313


prototech:
post_visibility is not a boolean value so this check is incorrect. ITEM_DELETED and ITEM_REAPPROVE will pass this check while ending with the same behavior described in the ticket. Check the value for ITEM_APPROVED explicitly

RMcGirr83:
https://github.com/phpbb/phpbb/pull/4045

PR #4046, opened 2015-11-27
[ticket/14269] Use http_exceptions in the installer instead of die()

CHItA:
[PHPBB3-14269](https://tracker.phpbb.com/browse/PHPBB3-14269)

### DIFF
--- a/phpBB/language/en/install.php
+++ b/phpBB/language/en/install.php
@@ -83,8 +83,8 @@ $lang = array_merge($lang, array(
 
 	'ACP_LINK'	=> 'Take me to <a href="%1$s">the ACP</a>',
 
-	'INSTALL_PHPBB_IS_ALREADY_INSTALLED'	=> 'phpBB is already installed.',
-	'INSTALL_PHPBB_IS_NOT_INSTALLED'		=> 'phpBB is not installed yet.'
+	'INSTALL_PHPBB_INSTALLED'		=> 'phpBB is already installed.',
+	'INSTALL_PHPBB_NOT_INSTALLED'	=> 'phpBB is not installed yet.'
 ));
 
 // Requirements translation
--- a/phpBB/language/en/install.php
+++ b/phpBB/language/en/install.php
@@ -82,6 +82,9 @@ $lang = array_merge($lang, array(
 	<p>Only those databases supported on your server will be displayed.',
 
 	'ACP_LINK'	=> 'Take me to <a href="%1$s">the ACP</a>',
+
+	'INSTALL_PHPBB_IS_ALREADY_INSTALLED'	=> 'phpBB is already installed.',
+	'INSTALL_PHPBB_IS_NOT_INSTALLED'		=> 'phpBB is not installed yet.'
 ));
 
 // Requirements translation
--- a/phpBB/phpbb/install/controller/install.php
+++ b/phpBB/phpbb/install/controller/install.php
@@ -105,7 +105,7 @@ class install
 	{
 		if ($this->install_helper->is_phpbb_installed())
 		{
-			throw new http_exception(404, 'INSTALL_PHPBB_INSTALLED');
+			throw new http_exception(403, 'INSTALL_PHPBB_INSTALLED');
 		}
 
 		$this->template->assign_vars(array(
--- a/phpBB/phpbb/install/controller/install.php
+++ b/phpBB/phpbb/install/controller/install.php
@@ -105,7 +105,7 @@ class install
 	{
 		if ($this->install_helper->is_phpbb_installed())
 		{
-			throw new http_exception(404, 'INSTALL_PHPBB_IS_ALREADY_INSTALLED');
+			throw new http_exception(404, 'INSTALL_PHPBB_INSTALLED');
 		}
 
 		$this->template->assign_vars(array(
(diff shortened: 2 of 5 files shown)


phpBB/language/en/install.php
```diff
@@ -82,6 +82,9 @@
 	<p>Only those databases supported on your server will be displayed.',
 
 	'ACP_LINK'	=> 'Take me to <a href="%1$s">the ACP</a>',
+
+	'INSTALL_PHPBB_IS_ALREADY_INSTALLED'	=> 'phpBB is already installed.',
```
marc1706: How about just using INSTALL_PHPBB_ALREADY_INSTALLED and INSTALL_PHPBB_NOT_INSTALLED?
The "is" is kind of useless.

phpBB/phpbb/install/event/kernel_exception_subscriber.php
```diff
@@ -0,0 +1,125 @@
...
+		$this->language				= $language;
+		$this->template				= $template;
+	}
+
+	/**
+	 * This listener is run when the KernelEvents::EXCEPTION event is triggered
+	 *
+	 * @param GetResponseForExceptionEvent	$event
+	 */
+	public function on_kernel_exception(GetResponseForExceptionEvent $event)
```
marc1706: Similar to the above use of "is", the "on" is IMHO useless and doesn't add any info.

phpBB/phpbb/install/event/kernel_exception_subscriber.php
```diff
@@ -0,0 +1,125 @@
...
+		{
+			$response->setStatusCode($exception->getStatusCode());
+			$response->headers->add($exception->getHeaders());
+		}
+
+		$event->setResponse($response);
+	}
+
+	/**
+	 * Returns an array of the events the object is subscribed to
```
marc1706: an array of events the object

marc1706:
Rest looks good :+1:

phpBB/language/en/install.php
VSEphpbb: How 'bout just INSTALL_PHPBB_INSTALLED

phpBB/phpbb/install/event/kernel_exception_subscriber.php
CHItA: that's the naming convention and AFAIK not just in phpBB.

CHItA:
@phpbb/development-team anyone has something to add about HTTP response code?

phpBB/phpbb/install/controller/install.php
```diff
@@ -97,12 +98,14 @@ public function __construct(helper $helper, factory $factory, navigation_provide
...
 	 * @return Response|StreamedResponse
+	 *
+	 * @throws http_exception When phpBB is already installed
 	 */
 	public function handle()
 	{
 		if ($this->install_helper->is_phpbb_installed())
 		{
-			die ('phpBB is already installed');
+			throw new http_exception(404, 'INSTALL_PHPBB_INSTALLED');
```
marc1706: I'd suggest a 418 but apart from that a 403 might be better.

phpBB/phpbb/install/controller/update.php
```diff
@@ -93,12 +94,16 @@ public function __construct(helper $controller_helper, installer $installer, ins
...
+	 * @return Response|StreamedResponse
+	 *
+	 * @throws http_exception When phpBB is not installed
 	 */
 	public function handle()
 	{
 		if (!$this->install_helper->is_phpbb_installed())
 		{
-			die ('phpBB is not installed');
+			throw new http_exception(404, 'INSTALL_PHPBB_NOT_INSTALLED');
```
marc1706: See above

phpBB/phpbb/install/controller/install.php
Nicofuma: I like 418 ;) but yeah +1 for 403

marc1706:
Are you planning on extending the subscriber? Otherwise we might just change this to a listener and get rid of the call to getSubscribedEvents() and then rename on_kernel_exception() to onKernelException() and it will automatically be run if you define the kernel.exception event in the tag.

CHItA:
I think this is fine as is for now

marc1706:
Well this will certainly reduce the complexity of the file. If we're not having more than the kernel exception in this file we should really go for a listener. It gets rid of the getSubscribedEvents() method in the file completely and will autoload the onKernelException() method.

marc1706:
@michaelcullum for the php part

Nicofuma:
Moving the event registration to the service definition doesn't really reduce the complexity. It just moves it to the yaml file.

Nicofuma:
I like using subscribers because this way your class is self explained. You don't need to check another file to see if and how (which event and priority?) the event listener is registered.

marc1706:
Ok, I'm fine with that, too. Was just a suggestion.

michaelcullum:
I'd agree with Marc as onKernelException() is self-explanatory also I think.

Nicofuma:
It explains the name yes, but not the priority. And even if the name is clear you can forget to update the definition but to know that you have to check 2 different files.
Actually I don't see any good reason to do that in the service definition (

CHItA:
This is how it is everywhere in the core, this has basically no disadvantages, therefore i think consistency is the best option for now. if this really bothers someone this can be modified in an other pr.

Nicofuma:
I completely agree with the latest message of CHItA

PR #4051, opened 2015-11-28
[ticket/14311] Show errors in installer invoked by trigger_error

CHItA:
[PHPBB3-14311](https://tracker.phpbb.com/browse/PHPBB3-14311)

### DIFF
--- a/phpBB/install/startup.php
+++ b/phpBB/install/startup.php
@@ -47,6 +47,36 @@ function phpbb_include_updated($path, $phpbb_root_path, $optional = false)
 	}
 }
 
+function installer_msg_handler($errno, $msg_text, $errfile, $errline)
+{
+	switch ($errno)
+	{
+		case E_NOTICE:
+		case E_WARNING:
+		case E_USER_WARNING:
+		case E_USER_NOTICE:
+			$msg = '[phpBB debug] "' . $msg_text . '" in file ' . $errfile . ' on line ' . $errline;
+			throw new \phpbb\exception\runtime_exception($msg);
+		break;
+		case E_USER_ERROR:
+			$msg = '<b>General Error:</b><br />' . $msg_text . '<br /> in file ' . $errfile . ' on line ' . $errline;
+
+			$backtrace = get_backtrace();
+			if ($backtrace)
+			{
+				$msg .= '<br /><br />BACKTRACE<br />' . $backtrace;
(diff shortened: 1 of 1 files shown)


Nicofuma:
:+1: I don't see any other solution

CHItA:
maybe using globals and iohandler

Nicofuma:
When there is an error I think we want to stop the installer and display it

phpBB/install/startup.php
```diff
@@ -47,6 +47,36 @@ function phpbb_include_updated($path, $phpbb_root_path, $optional = false)
...
+function installer_msg_handler($errno, $msg_text, $errfile, $errline)
+{
+	switch ($errno)
+	{
+		case E_NOTICE:
+		case E_WARNING:
+		case E_USER_WARNING:
+		case E_USER_NOTICE:
+			$msg = '[phpBB debug] "' . $msg_text . '" in file ' . $errfile . ' on line ' . $errline;
+			throw new \phpbb\exception\runtime_exception($msg);
```
CHItA: Mostly this is what seems to be the problem. We could log them to a file or just use the iohandler to add a warning, not really sure what would be the best solution.

marc1706: Having the iohandler display a warning would IMHO be the best solution.

Nicofuma: for the debug maybe yes. Even if will often trigger another error later.

You can do that?

phpBB/install/startup.php
```diff
@@ -47,6 +47,48 @@ function phpbb_include_updated($path, $phpbb_root_path, $optional = false)
...
+
+			try
+			{
+				/** @var \phpbb\install\helper\iohandler\iohandler_interface $iohandler */
+				$iohandler = $phpbb_installer_container->get('installer.helper.iohandler');
+				$iohandler->add_warning_message($msg);
+			}
+			catch (\phpbb\install\helper\iohandler\exception\iohandler_not_implemented_exception $e)
+			{
+				print ($msg);
```
marc1706: Why is it now using print() instead of throwing the runtime exception? Does that prevent breaking the page?

CHItA: it will try to use the iohandler, if it is not set, it is more then likely an HTML page with no AJAX so it should be fine.

PR #4054, opened 2015-11-30
[ticket/14325] Preserve event variable BC

CHItA:
[PHPBB3-14325](https://tracker.phpbb.com/browse/PHPBB3-14325)

### DIFF
--- a/phpBB/includes/acp/acp_forums.php
+++ b/phpBB/includes/acp/acp_forums.php
@@ -944,63 +944,62 @@ class acp_forums
 	/**
 	* Update forum data
 	*/
-	function update_forum_data(&$forum_data)
+	function update_forum_data(&$forum_data_ary)
 	{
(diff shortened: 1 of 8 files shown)


CHItA:
@VSEphpbb

VSEphpbb:
Actually should be @Nicofuma who looks at this

CHItA:
I meant that you could check if all event variable names are changed 
back correctly now.

VSEphpbb:
Oh, yeah, you caught all of them.

Nicofuma:
:+1: that's the way to go

marc1706:
Any reason why the tests were cancelled?

naderman:
If we change all of this code already why not change it to no longer use references at all?

CHItA:
The ACP should be rewritten from scratch anyway and this was way faster to do

naderman:
Heh fair enough

PR #4055, opened 2015-11-30
[ticket/14326] Decode diffed file's content during update

CHItA:
[PHPBB3-14326](https://tracker.phpbb.com/browse/PHPBB3-14326)


phpBB/phpbb/install/module/update_filesystem/task/update_files.php
```diff
@@ -164,20 +164,20 @@ public function run()
 					{
 						case 'delete':
 							$this->file_updater->delete_file($path);
-							break;
+						break;
```
Nicofuma: shuldn't them be indented?

marc1706: No, per our coding guidelines these shouldn't be indented.

PR #4056, opened 2015-12-02
[ticket/14333] Change br tags in acp files

RMcGirr83:
Start of changing break tags to be HTML 5 compliant

PHPBB3-14333

### DIFF
--- a/phpBB/adm/style/acp_attachments.html
+++ b/phpBB/adm/style/acp_attachments.html
@@ -28,9 +28,9 @@
 	<h2>{L_UPLOADING_FILES}</h2>
 
 	<!-- BEGIN upload -->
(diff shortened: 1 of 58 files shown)


RMcGirr83:
!set WIP

marc1706:
Is this actually being flagged as invalid html5? If not, then this should be against 3.2 and not 3.1.

VSEphpbb:
`<br />` is HTML5 valid. Technically it's correct too, since it is a self-closing tag.

I'm against this change as well since its unnecessary and will lead to a lot of discrepancy with styles and extensions..

RMcGirr83:
Simple enough close the ticket and mark as invalid

marc1706:
Closed the ticket and therefore closing the PR.

PR #4067, opened 2015-12-07
[ticket/14345] Check if description exists in message ary

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-14345

### DIFF
--- a/phpBB/phpbb/install/helper/iohandler/cli_iohandler.php
+++ b/phpBB/phpbb/install/helper/iohandler/cli_iohandler.php
@@ -126,7 +126,8 @@ class cli_iohandler extends iohandler_base
 		$this->io->newLine();
 
 		$message = $this->translate_message($error_title, $error_description);
-		$this->io->error($message['title'] . "\n" . $message['description']);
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/install/helper/iohandler/cli_iohandler.php
```diff
@@ -126,7 +126,8 @@ public function add_error_message($error_title, $error_description = false)
 		$this->io->newLine();
 
 		$message = $this->translate_message($error_title, $error_description);
-		$this->io->error($message['title'] . "\n" . $message['description']);
+		$message_string = $message['title'] . "\n" . (!empty($message['description']) ? $message['description'] : '');
```
Nicofuma: the EOL char is needed only when there is a description IIRC. You should move it to the ternary

marc1706: Ok, will do.

marc1706:
@Nicofuma updated

Nicofuma:
:+1:

PR #4068, opened 2015-12-07
[ticket/14310] Put progress percentage inside span and make more visible

marc1706:
The progress bar percentage is hardly readable during the installation once the filler passes the text.
This change will take care of this (see pictures below).
install_percent_50

install_percent_80
The text will always be inside the filler.

Ticket: https://tracker.phpbb.com/browse/PHPBB3-14310


Nicofuma:
:+1:

PR #4072, opened 2015-12-07
[ticket/14344] Improve output of HTML errors trigged during install

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-14344

### DIFF
--- a/phpBB/phpbb/install/helper/iohandler/cli_iohandler.php
+++ b/phpBB/phpbb/install/helper/iohandler/cli_iohandler.php
@@ -125,6 +125,10 @@ class cli_iohandler extends iohandler_base
 	{
 		$this->io->newLine();
 
+		if (strpos($error_title, '<br />') !== false)
+		{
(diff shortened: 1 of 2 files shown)


phpBB/phpbb/install/helper/iohandler/cli_iohandler.php
```diff
@@ -125,6 +125,10 @@ public function add_error_message($error_title, $error_description = false)
 	{
 		$this->io->newLine();
 
+		if (strpos($error_title, '<br />') !== false)
+		{
+			$error_title = strip_tags(str_replace('<br />', "\n", $error_title));
```
Nicofuma: from where is the new line coming?

marc1706: SQL errors will be shown using HTML. That way the <br /> tags will be converted to new lines and then all other html tags will be removed.

PR #4077, opened 2015-12-11
[ticket/14306] Automatically enable a safe mode when container building fails

Nicofuma:
Automatically try to use a safe mode (no cached container, no extensions) when we fail to build the container and display the exception on every page in the ACP (do not display anything outside of the ACP).

PHPBB3-14306

### DIFF
--- a/phpBB/adm/style/overall_header.html
+++ b/phpBB/adm/style/overall_header.html
@@ -149,8 +149,7 @@ function popup(url, width, height, name)
 							<!-- IF CONTAINER_EXCEPTION !== false -->
 							<div class="errorbox">
 								<p>{L_CONTAINER_EXCEPTION}</p>
-								<p>{L_CONTAINER_EXCEPTION_DETAIL}</p>
-								<p>Exception message: {{ CONTAINER_EXCEPTION.getMessage() }}</p>
+								<p>{L_EXCEPTION}{L_COLON} {{ CONTAINER_EXCEPTION.getMessage() }}</p>
 								<pre>{{ CONTAINER_EXCEPTION.getTraceAsString() }}</pre>
 							</div>
 							<!-- ENDIF -->
--- a/phpBB/adm/style/overall_header.html
+++ b/phpBB/adm/style/overall_header.html
@@ -49,7 +49,7 @@ function marklist(id, name, state)
 	}
 
 	var rb = parent.getElementsByTagName('input');
-	
+
 	for (var r = 0; r < rb.length; r++)
 	{
 		if (rb[r].name.substr(0, name.length) == name)
@@ -103,7 +103,7 @@ function popup(url, width, height, name)
 		<p><a href="{U_ADM_INDEX}">{L_ADMIN_INDEX}</a> &bull; <a href="{U_INDEX}">{L_FORUM_INDEX}</a></p>
 		<p id="skip"><a href="#acp">{L_SKIP}</a></p>
 	</div>
-	
+
 	<div id="page-body">
 		<div id="tabs">
 			<ul>
@@ -120,7 +120,7 @@ function popup(url, width, height, name)
 						<!-- DEFINE $LI_USED = 0 -->
 						<!-- BEGIN l_block1 -->
 							<!-- IF l_block1.S_SELECTED -->
-	
+
 						<!-- BEGIN l_block2 -->
 							<!-- IF .l_block1.l_block2.l_block3 -->
 							<!-- IF $LI_USED --></ul></div><!-- ENDIF -->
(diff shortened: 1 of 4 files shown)


Nicofuma:
A few work has to be done when the exception is displayed in the ACP but it is working :)

Nicofuma:
@VSEphpbb any idea on how to display the error and what to say in the ACP?

VSEphpbb:
@Nicofuma use try catch and maybe feed the exception error message to an errors array or trigger_error, since I think exceptions just don't work still in the ACP.

VSEphpbb:
@Nicofuma if you start writing what it should say, since you know better what you're doing here, I can help clarify it up if needed.

Nicofuma:
I already have the exception and pass it to the template: https://github.com/phpbb/phpbb/pull/4077/files#diff-45715d6c5044a8cd533bdccc6422d61eR149 I just wonder if we can do something better then getTraceAsString()

About the message what I want to say is that an error occurred when building the container due to an extension and for this reasons they have been temporarily disabled (they will be reenabled again as soon as the error is gone).

VSEphpbb:
> phpBB encountered an error building the container due to an installed extension. For this reason, all extensions have been temporarily disabled. Please try purging your forum cache. All extensions will automatically be re-enabled once the container error resolves itself. If this error continues, please visit phpBB.com for support.

phpBB/phpbb/di/container_builder.php
```diff
@@ -125,62 +125,88 @@ function __construct($phpbb_root_path, $php_ext)
...
-		{
-			require($config_cache->getPath());
-			$this->container = new \phpbb_cache_container();
-		}
-		else
-		{
-			$this->container_extensions = array(new extension\core($this->get_config_path()));
-
-			if ($this->use_extensions)
+		try {
```
Elsensee: Coding style ;P

phpBB/phpbb/di/container_builder.php
```diff
@@ -125,62 +125,88 @@ function __construct($phpbb_root_path, $php_ext)
...
+					->with_custom_parameters(array_merge($this->custom_parameters, [
+						'container_exception' => $e,
+					]))
+					->get_container();
+			}
+			catch (\Exception $_)
+			{
+				// Rethrow the original exception if it's still failing
+				throw $e;
+			}
```
Elsensee: If the container is rebuilt and still throws an exception you won't get here. Instead it will try to build the container again and again.
You should tell the container builder after the first try that there already was a first try and it should not attempt to rebuild the container but instead throw the exception.

phpBB/adm/style/overall_header.html
```diff
@@ -143,6 +143,13 @@
 							</div>
 						<!-- ENDIF -->
 					</div>
-	
+
 					<div id="main">
 						<div class="main">
+							<!-- IF CONTAINER_EXCEPTION !== false -->
```
marc1706: I don't really like that you are mixing phpBB template syntax with twig syntax here. Why not just use twig for the if, too?

Nicofuma:
Updated and rebased

Elsensee:
I already know that our users will love this, but there are few things, I'd like to mention:

1. Probably not related to this PR since this is only about a safemode when container building fails, but I think it might be a good idea as well to turn off extensions if loading an extension-class fails.. if this is somehow possible. You know, sometimes users delete files of extensions without disabling them properly and then the cached container tries to access these classes but can't find them so this is not happening in container building... But as I said: Probably nothing for this PR.
2. I would put a newline there between message and call stack or somehow mention that "Prepare, information about your mistakes will be delivered in 3... 2... 1..."
bildschirmfoto 2016-01-09 um 17 47 28 387
3. What do you think about hyperlinking "phpBB.com" in the message?

Nicofuma:
1) Would be nice yes but not that easy with the current setup I think. I have an idea to do that in 3.3 though
2) done
3) done
4)Honestly, path disclosure in the ACP... I don't think that's an issue

marc1706:
Your ACP message states that the admin should purge the cache. However, based on your code it looks like the "fallback" container without extensions is not being cached. That makes this message kind of wrong and might not be the best idea as we'll try to rebuild the container every time.

Nicofuma:
The here is many reasons that can explain the error:
 - The cache container may be outdated
 - An extension (or the core) may have an incorrect service definition
 - An extension (or the core) may have an invalid php file
 - 2 extensions may be incompatibles
 - ...

For the first case (the most common one) purging the cache should be enough.
For the second part (rebuilding the container every time), it is necessary I think. Don't forget we are trying to recover from a major error. It's just a way to let the administrator able to use the ACP in order to purge the cache or disable the last extension per example. It will be slow but I don't think it is important, it is an exceptional case and we don't want to stay in this mode longer than necessary.

marc1706:
Ok, fair enough.

PR #4082, opened 2015-12-15
[ticket/14361] Prevent PM count from going negative

ajquick:
3.1.x Branch

Cosmetic fix to prevent the unread / new PM counts from going negative.

PHPBB3-14361

### DIFF
--- a/phpBB/includes/functions.php
+++ b/phpBB/includes/functions.php
@@ -5076,7 +5076,7 @@ function page_header($page_title = '', $display_online_list = false, $item_id = 
 		'LOGGED_IN_USER_LIST'			=> $online_userlist,
 		'RECORD_USERS'					=> $l_online_record,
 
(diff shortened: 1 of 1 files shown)


n-aleha:
@RMcGirr83 
The PM counter can display as negative. I've seen it on a board I run.

marc1706:
The PM counter shouldn't go negative. Instead of suppressing the issue we should rather fix the issue if there is one.

exussum12:
Did a little bit of digging in to this, When generating the number the unique messages are counted (for example 2 groups of people are messaged - person is in both groups)

then when deleted the cache takes away the number of deleted rows. The problem this is trying to solve I assume is not doing another query to properly update the messages.

This fix is incorrect really, the number will still be wrong when its above 0 if this bug has effected it.

marc1706:
Closing this as the PR only fixes the manifestation of the issue instead of the actual issue. Please reply to the ticket if you have more input: https://tracker.phpbb.com/browse/PHPBB3-14361

PR #4086, opened 2015-12-17
[ticket/14366] Add core events to the function decode_message()

rxu:
Add core event to the function decode_message() in
includes/functions_content.php to allow extensions performing additional
message handling before/after decoding.

<a href="https://tracker.phpbb.com/browse/PHPBB3-14366">PHPBB3-14366</a>.

### DIFF
--- a/phpBB/includes/functions_content.php
+++ b/phpBB/includes/functions_content.php
@@ -392,7 +392,7 @@ function phpbb_clean_search_string($search_string)
 */
 function decode_message(&$message, $bbcode_uid = '')
 {
-	global $config;
+	global $config, $phpbb_dispatcher;
 
 	if ($bbcode_uid)
 	{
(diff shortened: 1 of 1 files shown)


rxu:
!set Event

phpBB/includes/functions_content.php
```diff
@@ -405,12 +405,38 @@ function decode_message(&$message, $bbcode_uid = '')
...
+	* Use this event to modify the message after it is decoded
+	*
+	* @event core.decode_message_after
+	* @var string	message			The message content
+	* @var string	bbcode_uid		The message BBCode UID
+	* @var string	match			Match pattern to replace
+	* @var int		replace			Replacement for the matched text
+	* @since 3.1.8-RC1
+	*/
+	$vars = array('message', 'bbcode_uid', 'match', 'replace');
```
marc1706: Same as above. Also, do you really need the match and replace here?

phpBB/includes/functions_content.php
```diff
@@ -405,12 +405,38 @@ function decode_message(&$message, $bbcode_uid = '')
...
+	* Use this event to modify the message before it is decoded
+	*
+	* @event core.decode_message_before
+	* @var string	message			The message content
+	* @var string	bbcode_uid		The message BBCode UID
+	* @var string	match			Match pattern to replace
+	* @var int		replace			Replacement for the matched text
+	* @since 3.1.8-RC1
+	*/
+	$vars = array('message', 'bbcode_uid', 'match', 'replace');
```
marc1706: ```$message``` is passed by reference. Please use an alternative variable for the event. Also, the match and replace will no longer properly work for 3.2.

phpBB/includes/functions_content.php
rxu: Well, someone probably would use custom preg_expression and related match set.

phpBB/includes/functions_content.php
rxu: To be honest, I have no idea what to do with 3.2 text parser here. That's a real pita in regard of BC.

rxu:
Updated #4086.

PR #4090, opened 2015-12-20
[ticket/14372] Include functions_content.php into phpBB/bin/phpbbcli.php

rxu:
functions_content.php is needed for the cases phpbbcli is used to run
cron tasks from extensions when the latter are using methods from it.

<a href="https://tracker.phpbb.com/browse/PHPBB3-14372">PHPBB3-14372</a>.

### DIFF
--- a/phpBB/bin/phpbbcli.php
+++ b/phpBB/bin/phpbbcli.php
@@ -35,6 +35,7 @@ extract($phpbb_config_php_file->get_all());
 require($phpbb_root_path . 'includes/constants.' . $phpEx);
 require($phpbb_root_path . 'includes/functions.' . $phpEx);
 require($phpbb_root_path . 'includes/functions_admin.' . $phpEx);
+require($phpbb_root_path . 'includes/functions_content.' . $phpEx);
 require($phpbb_root_path . 'includes/utf/utf_tools.' . $phpEx);
 
 $phpbb_container_builder = new \phpbb\di\container_builder($phpbb_config_php_file, $phpbb_root_path, $phpEx);


marc1706:
Shouldn't cron tasks as part of extensions include that file themselves if needed?

rxu:
F.e. I didn't know ```set_content_visibility``` needs ```functions_content.php``` until I had to debug PHP fatal error.

marc1706:
To be honest, I think we should rather get rid of calling ```truncate_string()``` inside content_visibility instead of including functions_content.php in the cli.

rxu:
Well, this is just an example I faced, but whatever you decide. It's no problem with fixing it in the extension.

marc1706:
The point here is kind of that this has always been the case for MODs back in the days and also still for extensions with functions. Obviously, the goal for us is to get rid of functions in classes to prevent these kind of cases. Putting more includes is IMHO not the way to solve this.

Nicofuma:
The thing here is that truncate_strin() shouldn't be declared in functions_content but in a proper helper.

About the current patch... I don't really mind.

Nicofuma:
@rxu @marc1706 so it's a won't fix, do we agree?

marc1706:
Yes, that's what I'd go for.

rxu:
So, how do we handle the issue if won't fix it?

marc1706:
truncate_string() should be moved to a proper helper and used in the content visibility class. Should use a new ticket for that. If you want to use it right now, include functions_content.php if you want to use content visibility.

marc1706:
Closing this as previously stated. Feel free to reword the ticket to state the changes needed for content visibility.

PR #4096, opened 2015-12-22
[ticket/14380] Maintain consistent template var paths in 3.2 installer

CHItA:
[PHPBB3-14380](https://tracker.phpbb.com/browse/PHPBB3-14380)


marc1706:
Just saw this PR seconds after I noticed this was missing. Thx.

PR #4097, opened 2015-12-23
[ticket/14382] Specify the branch-alias in the composer.json files

Nicofuma:
PHPBB3-14382


Nicofuma:
@naderman

naderman:
Shouldn't that be 3.1.x-dev?

Nicofuma:
3.1-dev I think. Otherwise it's not a valid version number?

marc1706:
```.x-dev``` is allowed according to https://getcomposer.org/doc/articles/aliases.md#branch-alias

Nicofuma:
updated

PR #4101, opened 2015-12-27
[ticket/14343] Add missing $phpbb_dispatcher

kasimi:
[PHPBB3-14343](https://tracker.phpbb.com/browse/PHPBB3-14343)

PR #4066 is missing the $phpbb_dispatcher

### DIFF
--- a/phpBB/includes/mcp/mcp_main.php
+++ b/phpBB/includes/mcp/mcp_main.php
@@ -262,7 +262,7 @@ class mcp_main
 */
 function lock_unlock($action, $ids)
 {
-	global $auth, $user, $db, $phpEx, $phpbb_root_path, $request;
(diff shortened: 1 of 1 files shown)


Nicofuma:
@nickvergessen I think it's needed to rebuild the 3.1.7 packages with this fix. Otherwise locking/unlocking may throw fatal errors, right?

kasimi:
Unfortunately, yes.

nickvergessen:
Well let's not repackage, but also increase the version then, because packages are public quite a few days already

marc1706:
@kasimi Builds fails because you were using an incorrect ticket id in your commit message heading and were missing ```PHPBB3-14343``` in your commit message after the heading.

marc1706:
Replaced by #4109

PR #4104, opened 2015-12-30
[ticket/14386] Fix for open_basedir restriction in effect with remote upload avatar

ErnadoO:
PHPBB3-14386


marc1706:
Please modify your commit message and add a blank line between the heading and ```PHPBB3-14386```

ErnadoO:
I think commit message is good, now

Yes, i will send another PR for master

PR #4105, opened 2016-01-02
[ticket/14387] Allow extending avatars by extensions in ACP

marc1706:
Create a driver method to provide the driver config name, and use it
within the manager method. Default driver config name is the same as now.
But new drivers are able to override the config name with their own.

Ticket: https://tracker.phpbb.com/browse/PHPBB3-14387

PHPBB3-14387


javiexin:
@marc1706  Thanks a lot for updating the tests.

PR #4106, opened 2016-01-03
[3.2][ticket/14386] Fix for open_basedir restriction in effect with remote upload avatar

ErnadoO:
PHPBB3-14386

### DIFF
--- a/phpBB/phpbb/files/types/remote.php
+++ b/phpBB/phpbb/files/types/remote.php
@@ -200,8 +200,7 @@ class remote extends base
 			return $this->factory->get('filespec')->set_error($this->upload->error_prefix . 'EMPTY_REMOTE_DATA');
 		}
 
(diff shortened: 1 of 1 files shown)


marc1706:
Previously we used the cache directory with activated safe_mode setting. Your patch changes this to a different behavior. Please fix this the same way as for 3.1.x.

ErnadoO:
Yes but phpBB 3.2 require php5.4, or safe mod has been removed in this version, so this condition is obsolete.
I'm wrong ?

marc1706:
Oh right, forgot about that. Then this should be fine.

Nicofuma:
:+1:

marc1706:
Replacing this with a fast-forward compatible branch

marc1706:
Replaced by #4110

PR #4109, opened 2016-01-06
[ticket/14343] Added missing $phpbb_dispatcher

marc1706:
Replaced #4101 

Ticket: https://tracker.phpbb.com/browse/PHPBB3-14343

### DIFF
--- a/phpBB/includes/mcp/mcp_main.php
+++ b/phpBB/includes/mcp/mcp_main.php
@@ -262,7 +262,7 @@ class mcp_main
 */
 function lock_unlock($action, $ids)
 {
-	global $auth, $user, $db, $phpEx, $phpbb_root_path, $request;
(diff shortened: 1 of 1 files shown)


Nicofuma:
:+1:

kasimi:
Does this actually have to wait for 3.1.8?

If so, though locking/unlocking posts/topics in 3.1.7 will work, it'll raise a fatal error right [here](https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/includes/mcp/mcp_main.php#L339) and the JS loading indicator is never closed and the page is not reloaded.

marc1706:
We are aware of this. As you might have noticed, phpBB 3.1.7 has not been released yet.

kasimi:
I was just wondering because 3.1.8 is assigned to this issue.

PR #4113, opened 2016-01-07
[ticket/14395] Add event core.viewtopic_add_quickmod_option_before

LaxSlash:
Moved the call to function phpbb_add_quickmod_option to being under the core
event core.viewtopic_assign_template_vars_before so as to enable the
modification of elements in the  BEFORE they are assigned to the
template, so that removed elements don't show up in a non-functional state.
Allows the user to perform additional permissions checks on the items within the
array.

PHPBB3-14395

### DIFF
--- a/phpBB/viewtopic.php
+++ b/phpBB/viewtopic.php
@@ -592,7 +592,7 @@ $quickmod_array = array(
 * Event to modify data in the quickmod_array before it gets sent to the
 * phpbb_add_quickmod_option function.
 *
-* @event core.viewtopic_add_quickmod_option_before
+* @event core.viewtopic_add_quickmod_option_after
 * @var	int		forum_id			Forum ID
 * @var	int		post_id				Post ID
 * @var	array	quickmod_array		Array with quick moderation options data
--- a/phpBB/viewtopic.php
+++ b/phpBB/viewtopic.php
@@ -588,14 +588,6 @@ $quickmod_array = array(
 	'topic_logs'			=> array('VIEW_TOPIC_LOGS', $auth->acl_get('m_', $forum_id)),
 );
 
-foreach ($quickmod_array as $option => $qm_ary)
-{
-	if (!empty($qm_ary[1]))
-	{
-		phpbb_add_quickmod_option($s_quickmod_action, $option, $qm_ary[0]);
-	}
-}
-
 // Navigation links
 generate_forum_nav($topic_data);
 
@@ -666,6 +658,14 @@ $vars = array(
 	'viewtopic_url',
 );
 extract($phpbb_dispatcher->trigger_event('core.viewtopic_assign_template_vars_before', compact($vars)));
+
+foreach ($quickmod_array as $option => $qm_ary)
+{
+	if (!empty($qm_ary[1]))
+	{
+		phpbb_add_quickmod_option($s_quickmod_action, $option, $qm_ary[0]);
+	}
+}
 
 $pagination->generate_template_pagination($base_url, 'pagination', 'start', $total_posts, $config['posts_per_page'], $start);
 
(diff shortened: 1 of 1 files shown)


rxu:
Adding new event is preferrable to the moving code around, as nobody knows how the latter will affect things in the future.
So, I'd say, just add new event instead.

LaxSlash:
Okay, I moved the code back to its original spot and named the new event core.viewtopic_add_quickmod_option_after.

(For the reason that it's being inserted after the original assignment of the array - did I get that right, or should it be "before?")

LaxSlash:
Okay, I'm lost...... Travis failed with the slowtests exiting with error code 2.... so what do I need to fix?

rxu:
If you look into the tests log:
> LogicException: Event name does not match '@event' tag for event 'core.viewtopic_add_quickmod_option_after'

The event names in the dicblock and in its call mismatch (after/before).

LaxSlash:
Yup, was looking at the wrong thing. Gravitated towards the red text. Fixed now.... fingers crossed that this one passes inspection.

LaxSlash:
I'll take a look at it again tomorrow. I'm running the patch on my dev VM with no problems whatsoever and I just double-checked the tabs vs. space thing on all my lines - no idea why this one failed now, and all the Travis build says this time around is something about a timeout.

Un1matr1x:
if i read travis correct you just have to restart the one failed test - since it's only a timeout:

    No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

    The build has been terminated

rxu:
Yeah, the fail is unrelated. Failed job just needs to be restarted by developer.

rxu:
Also, it would be wise to change the PR title :wink:

LaxSlash:
Ah, that makes sense then. I changed the PR title here as well... I thought that the Git would've changed that when I re-committed. Weird. Glad to hear everything passed, though. :)

phpBB/viewtopic.php
```diff
@@ -588,6 +588,33 @@
...
+* @event core.viewtopic_add_quickmod_option_after
+* @var	int		forum_id			Forum ID
+* @var	int		post_id				Post ID
+* @var	array	quickmod_array		Array with quick moderation options data
+* @var	array	topic_data			Array with topic data
+* @var	int		topic_id			Topic ID
+* @var	array	topic_tracking_info	Array with topic tracking data
+* @var	string	viewtopic_url		URL to the topic page
+* @var  bool  allow_change_type   Topic change permissions check
+* @since 3.1.8-RC1
```
marc1706: Please change this to 3.1.9-RC1

LaxSlash:
Apologies for the late commit on this, I never got an e-mail telling me that I had a notification. I also changed the name of the event as to being _before, as I believe that naming it _after was a little bit misleading, due to where it was compared to the actual function that's called to make the QuickMod menu.

phpBB/viewtopic.php
```diff
@@ -588,6 +588,33 @@
...
+*
+* @event core.viewtopic_add_quickmod_option_before
+* @var	int		forum_id			Forum ID
+* @var	int		post_id				Post ID
+* @var	array	quickmod_array		Array with quick moderation options data
+* @var	array	topic_data			Array with topic data
+* @var	int		topic_id			Topic ID
+* @var	array	topic_tracking_info	Array with topic tracking data
+* @var	string	viewtopic_url		URL to the topic page
+* @var  bool  allow_change_type   Topic change permissions check
```
VSEphpbb: fix spacing here please

LaxSlash:
How's that looking, any better?

LaxSlash:
Fixed it finally, but I need to have a dev restart the Travis build. Looks like it timed out.

PR #4117, opened 2016-01-10
[ticket/14403] Don't expect user_id and user_ip in phpbb\log

Elsensee:
[PHPBB3-14403](https://tracker.phpbb.com/browse/PHPBB3-14403)

### DIFF
--- a/phpBB/bin/phpbbcli.php
+++ b/phpBB/bin/phpbbcli.php
@@ -59,6 +59,8 @@ $phpbb_container->get('request')->enable_super_globals();
 require($phpbb_root_path . 'includes/compatibility_globals.' . $phpEx);
 
 $user = $phpbb_container->get('user');
+$user->data['user_id'] = ANONYMOUS;
+$user->ip = '127.0.0.1';
 $user->add_lang('acp/common');
 $user->add_lang('cli');
 
--- a/phpBB/phpbb/log/log.php
+++ b/phpBB/phpbb/log/log.php
@@ -229,8 +229,8 @@ class log implements \phpbb\log\log_interface
 		}
 
 		$sql_ary = array(
-			'user_id'		=> $user_id,
-			'log_ip'		=> $log_ip,
+			'user_id'		=> !empty($user_id) ? $user_id : ANONYMOUS,
+			'log_ip'		=> !empty($log_ip) ? $log_ip : '',
 			'log_time'		=> $log_time,
 			'log_operation'	=> $log_operation,
 		);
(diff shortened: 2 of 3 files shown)


rxu:
This don't actually solve the issue, as accessing the ```$user->data['user_id']``` which is not set in outer script (like cron called via CLI) will throw error itself.
Thus an outer script will require the same workaround with id/ip still.

Elsensee:
Indeed, but E_NOTICE won't exit the script.

Nicofuma:
For that we need to fake the $user values when we are using the CLI. It can be done using the event ConsoleEvents::COMMAND

marc1706:
Why do you want to fake that with an event? The user object already exists in the cli and a session is not being created. I don't see a reason why we should not just inject the data there.

Nicofuma:
I want to fake the user data (data['user_ip'], data['user_id'] and eventually other data that are not present in the CLI context because of the lack of session) because otherwise we will throw notices (or other errors later) in many places (it is already the case actually). And also because I think it's better to have a global workaround for this kind of issue instead of having a different workaround everywhere.

Elsensee:
Marc proposed the same thing actually but he wanted to change these variables in bin/phpbbcli.php already, if I got him correct.

Nicofuma:
Yep but I don't like putting more stuff in phpbbcli and the other frontends (especially because I'm planning to nuke a lot of things (like common) and wrap everything with a single "kernel" object (but that's for 3.3 so we have time to see ;)))

marc1706:
Using an event for that just seems like a huge workaround for something that would just require 2 lines of code in the cli. Also, you'd have to add that event, too. IMHO there is no point for now in using an event for that really.

Nicofuma:
(just to say, the event already exists. It's one of Symfony's dedicated to the CLI workflow)

Elsensee:
So how do you want me to fake it, then? :D

Nicofuma:
just set the data in the user object :)

Elsensee:
Oh, err, btw: Updated. ;)

phpBB/bin/phpbbcli.php
```diff
@@ -59,6 +59,7 @@
 require($phpbb_root_path . 'includes/compatibility_globals.' . $phpEx);
 
 $user = $phpbb_container->get('user');
+$user->data['user_id'] = ANONYMOUS;
 $user->add_lang('acp/common');
```
marc1706: Please also add ```$user->ip = '127.0.0.1';```

Elsensee: Ah, I saw that $user->ip was already set to an empty string and thought "no work necessary here" but I guess that makes more sense. :D

PR #4119, opened 2016-01-13
[ticket/14407] Fix removing users from Newly Registered Users group

Zoddo:
[PHPBB3-14407](https://tracker.phpbb.com/browse/PHPBB3-14407)


Zoddo:
Sorry, caused by #4008

PR #2064, opened 2014-02-27
[WIP] [3.2][ticket/11875] Set default data type from MEDIUMINT (3-bytes) to INT(10) (4-bytes)

hjpotter92:
Use `INT(10)` data types for `user_id`, `post_id` and `topic_id` across all tables

Resending pull request with `develop` as the base branch.

[PHPBB3-11875]

  [PHPBB3-11875]: http://tracker.phpbb.com/browse/PHPBB3-11875 "mediumint(8) too small for post_id"

PR #4121, opened 2016-01-13
[ticket/11875] Use 4-byte integer instead of 3-byte integer for ID columns

marc1706:
Replaces #2064 

Ticket: https://tracker.phpbb.com/browse/PHPBB3-11875


Nicofuma:
:+1:

PR #4122, opened 2016-01-14
[ticket/12441] Correct database size detection in MariaDB 10

lavigor:
Code changes were taken from [this post](https://www.phpbb.com/community/viewtopic.php?p=13611421#p13611421).

PHPBB3-12441

### DIFF
--- a/phpBB/includes/functions_admin.php
+++ b/phpBB/includes/functions_admin.php
@@ -3101,9 +3101,9 @@ function get_database_size()
 			{
 				$version = $row['mysql_version'];
 
-				if (preg_match('#(3\.23|[45]\.)#', $version))
+				if (preg_match('#(3\.23|[45]\.|10\.[0-9]\.[0-9]{1,2}-+Maria)#', $version))
 				{
-					$db_name = (preg_match('#^(?:3\.23\.(?:[6-9]|[1-9]{2}))|[45]\.#', $version)) ? "`{$db->get_db_name()}`" : $db->get_db_name();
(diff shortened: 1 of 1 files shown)


lavigor:
Should I also create PR for 3.2?

phpBB/includes/functions_admin.php
```diff
@@ -3101,9 +3101,9 @@ function get_database_size()
 			{
 				$version = $row['mysql_version'];
 
-				if (preg_match('#(3\.23|[45]\.)#', $version))
+				if (preg_match('#(3\.23|[45]\.|10\..+Maria)#', $version))
```
marc1706: I'd suggest a more restrictive regex:
```#(3\.23|[45]\.|10\.[0-9]\.[0-9]{1,2}-+Maria)#````

marc1706:
We should be able to merge this into master so you don't need to submit a PR for the master branch.

lavigor:
@marc1706 updated to your regex, but I haven't found a mess on GitHub.
Should this code be updated again if MariaDB gets the 10.10.x update or MySQL is updated?
Or maybe it can be changed to set the minimum requirement?

marc1706:
With github messing around I meant that github changes the regex if I post it here.
This regex should work for a rather extended timeframe. Both MySQL 6.x and MariaDB 10.10.x are a long time away. This regex is to ensure that the specified command properly works. Having this more or less restrictive is intended.

lavigor:
OK. Thank you for the explanation.

PR #4126, opened 2016-01-15
[ticket/13733] Allow non-migration files inside migrations folder

marc1706:
Currently any file in an extension's migration folder is considered a migration.
This means extensions can't keep migration helpers / abstract classes they may want to use in their migrations folder (like the way phpBB is able to do).
Only the classes implementing a migration interface should be loaded by the migrator.

Replaces #3626 

Ticket: https://tracker.phpbb.com/browse/PHPBB3-13733

### DIFF
--- a/phpBB/phpbb/extension/base.php
+++ b/phpBB/phpbb/extension/base.php
@@ -24,7 +24,7 @@ class base implements \phpbb\extension\extension_interface
 	protected $container;
 
 	/** @var \phpbb\finder */
-	protected $finder;
+	protected $extension_finder;
 
 	/** @var \phpbb\db\migrator */
 	protected $migrator;
@@ -148,7 +148,7 @@ class base implements \phpbb\extension\extension_interface
 				if (class_exists($migration))
 				{
 					$reflector = new \ReflectionClass($migration);
-					if ($reflector->isSubclassOf('\phpbb\db\migration\migration') && $reflector->isInstantiable())
+					if ($reflector->implementsInterface('\phpbb\db\migration\migration_interface') && $reflector->isInstantiable())
 					{
 						continue;
 					}
--- a/phpBB/phpbb/extension/base.php
+++ b/phpBB/phpbb/extension/base.php
@@ -121,11 +121,9 @@ class base implements \phpbb\extension\extension_interface
 	/**
 	* Get the list of migration files from this extension
 	*
-	* @var bool $validate_classes Whether or not to check that the migration
-	*		class exists and extends the base migration class.
 	* @return array
 	*/
-	protected function get_migration_file_list($validate_classes = true)
+	protected function get_migration_file_list()
 	{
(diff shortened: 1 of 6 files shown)


marc1706:
Ready for review @nickvergessen @Nicofuma

phpBB/phpbb/extension/base.php
```diff
@@ -121,9 +121,11 @@ public function purge_step($old_state)
 	/**
 	* Get the list of migration files from this extension
 	*
+	* @var bool $validate_classes Whether or not to check that the migration
+	*		class exists and extends the base migration class.
 	* @return array
 	*/
-	protected function get_migration_file_list()
+	protected function get_migration_file_list($validate_classes = true)
```
Nicofuma: I wonder if there is a case where $validate_classes will be set to false

phpBB/phpbb/extension/base.php
```diff
@@ -137,6 +139,26 @@ protected function get_migration_file_list()
...
 
+		if ($validate_classes)
+		{
+			// Unset classes that do not exist or do not extend the
+			// abstract class phpbb\db\migration\migration
+			foreach ($migrations as $key => $migration)
+			{
+				if (class_exists($migration))
+				{
+					$reflector = new \ReflectionClass($migration);
```
Nicofuma: Reflection or class_implements ? What is the fastest

Nicofuma:
could you add a test?

phpBB/phpbb/extension/base.php
marc1706: Based on my tests class_implements is about 35% faster (on average 0.9 µs for reflection class vs. 0.7 µs for class_implements). However, class_implements does not allow us to check if the class is instantiable.

phpBB/phpbb/extension/base.php
marc1706: Andreas wondered that, too.

marc1706:
What do you think about adding a testing ext for that to the fixtures?

phpBB/phpbb/extension/base.php
Nicofuma: ok

Nicofuma:
fine with a testing ext

marc1706:
@Nicofuma updated

Nicofuma:
and it's failing because the array order :/

marc1706:
Should be taken care of by using ```sort()```

phpBB/phpbb/extension/base.php
Elsensee: ... so? Anything against removing that?

VSEphpbb: i wondered why this was added too

marc1706: Sure, will remove it then.

marc1706:
Updated

PR #4127, opened 2016-01-17
[ticket/14416] footer nav links have incorrect tooltip styling applied

RMcGirr83:
https://tracker.phpbb.com/browse/PHPBB3-14416

PHPBB3-14416

### DIFF
--- a/phpBB/styles/prosilver/template/navbar_footer.html
+++ b/phpBB/styles/prosilver/template/navbar_footer.html
@@ -4,10 +4,22 @@
 	<ul id="nav-footer" class="nav-footer linklist" role="menubar">
 		<li class="breadcrumbs">
 			<!-- IF U_SITE_HOME -->
-				<span class="crumb"><a href="{U_SITE_HOME}" data-navbar-reference="home"><i class="icon fa-home fa-fw" aria-hidden="true"></i><span>{L_SITE_HOME}</span></a></span>
+				{% spaceless %}
+				<span class="crumb">
+					<a href="{U_SITE_HOME}" data-navbar-reference="home">
+						<i class="icon fa-home fa-fw" aria-hidden="true"></i><span>{L_SITE_HOME}</span>
+					</a>
+				</span>
+				{% endspaceless %}
 			<!-- ENDIF -->
 			<!-- EVENT overall_footer_breadcrumb_prepend -->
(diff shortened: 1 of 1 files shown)


RMcGirr83:
!set Styles

Nicofuma:
Just removing the spaces should fix the issue?

phpBB/styles/prosilver/template/navbar_footer.html
```diff
@@ -4,18 +4,10 @@
...
-				</span>
+				<span class="crumb"><a href="{U_SITE_HOME}" data-navbar-reference="home"><i class="icon fa-home fa-fw" aria-hidden="true"></i><span>{L_SITE_HOME}</span></a></span>
 			<!-- ENDIF -->
 			<!-- EVENT overall_footer_breadcrumb_prepend -->
-			<span class="crumb">
-				<a href="{U_INDEX}" data-navbar-reference="index">
-					<!-- IF not U_SITE_HOME --><i class="icon fa-home fa-fw" aria-hidden="true"></i><!-- ENDIF --><span>{L_INDEX}</span>
-				</a>
-			</span>
+				<span class="crumb"><a href="{U_INDEX}" accesskey="h" data-navbar-reference="index"><!-- IF not U_SITE_HOME --><i class="icon fa-home fa-fw"></i><!-- ENDIF --><span>{L_INDEX}</span></a></span>
```
Nicofuma: `accesskey="h"` on the `<a>` tag is new and `aria-hidden="true"` is on the `li` tag

Elsensee: Indeed, I remember that more than one accesskey attribute pointing to the same key will break that key in firefox.

RMcGirr83: Oddly enough, yes

Nicofuma:
in this case, instead of inlining everything, you should wrap it by the spaceless tag `{% spaceless %} {% endspaceless %}`

Nicofuma:
http://twig.sensiolabs.org/doc/tags/spaceless.html

hanakin:
wow twig made it into 3.2 :+1:  lol

Elsensee:
@hanakin We have a whole twig template already in 3.2. It's for the feeds and in styles/all/template ;)

PR #4128, opened 2016-01-17
[ticket/13717] Set the assets after rendering the whole template

Nicofuma:
The goal being to be able to call INCLUDECSS/JS from anywhere in any tempalte or event

PHPBB3-13717


Nicofuma:
@VSEphpbb

phpBB/phpbb/template/assets_bag.php
```diff
@@ -0,0 +1,93 @@
...
+* the docs/CREDITS.txt file.
+*
+*/
+
+namespace phpbb\template;
+
+class assets_bag
+{
+	/** @var asset[] */
+	protected $stylesheets = [];
```
VSEphpbb: Need to be 5.3.3 compatible arrays

phpBB/phpbb/template/assets_bag.php
```diff
@@ -0,0 +1,93 @@
...
+
+	/**
+	 * Returns the HTML code to includes all css assets
+	 *
+	 * @return string
+	 */
+	public function get_stylesheets_content()
+	{
+		$output = '';
+		foreach ($this->stylesheets as $stylesheet) {
```
VSEphpbb: braces on newline

phpBB/phpbb/template/assets_bag.php
```diff
@@ -0,0 +1,93 @@
...
+
+	/**
+	 * Returns the HTML code to includes all js assets
+	 *
+	 * @return string
+	 */
+	public function get_scripts_content()
+	{
+		$output = '';
+		foreach ($this->scripts as $script) {
```
VSEphpbb: braces

phpBB/phpbb/template/assets_bag.php
Nicofuma: :cry:

nickvergessen:
Not in 3.1 please, already broke enough with "improvements on the session". Yes, it might be faster or lighter or whatever, but maintenance means less risky.

VSEphpbb:
Not having this in 3.1 will break BC for the many extensions that would take advantage of this. So this has no point IMO if it is not in 3.1 and 3.2.

nickvergessen:
What? We can't break BC, because this simply **never** worked?

I understand your wish to use the same extension in 3.2 as you do in 3.1, then simply you have to use the 3.1 way. BC break would be the other way around (something that worked in 3.1 does not work in 3.2 anymore), but I don't read, that this is the case?

Nicofuma:
replaced by https://github.com/phpbb/phpbb/pull/4131 (for 3.2)

lavigor:
2548 commits here caused that 58 tickets were opened after this PR had been closed.
Only 40 of them were closed after this.

Nicofuma:
do you know which tickets are still open?

ErnadoO:
For exemple: https://tracker.phpbb.com/browse/PHPBB3-14386

Nicofuma:
Found the missing tickets, thanks

PR #4129, opened 2016-01-18
[ticket/10824] Use composer.json in styles/languages

marc1706:
Replaces #3882 

Ticket: https://tracker.phpbb.com/browse/PHPBB3-10824


Nicofuma:
@marc1706 status?

marc1706:
I was thinking of implementing a finder for these so we don't have to have languages and/or styles in their respective directories.

Nicofuma:
ok, lets keep it for 3.3 though

marc1706:
Yes, probably a better idea to have this for 3.3.

phpBB/phpbb/cache/service.php
```diff
@@ -354,7 +354,8 @@ function obtain_cfg_items($style)
 		if (!isset($parsed_array['filetime']) || (($this->config['load_tplcompile'] && @filemtime($filename) > $parsed_array['filetime'])))
 		{
 			// Re-parse cfg file
-			$parsed_array = parse_cfg_file($filename);
+			$json = file_get_contents($filename);
+			$parsed_array = json_decode($json, true);
```
marc1706: Should probably use the json_sanitizer as well.

phpBB/phpbb/db/migration/data/v310/style_update_p1.php
```diff
@@ -69,10 +69,11 @@ public function styles_update()
 		$skip_dirs = array('.', '..', 'prosilver');
 		foreach ($iterator as $fileinfo)
 		{
-			if ($fileinfo->isDir() && !in_array($fileinfo->getFilename(), $skip_dirs) && file_exists($fileinfo->getPathname() . '/style.cfg'))
```
marc1706: Should probably be rewritten to both handle pre 3.3 and 3.3 styles, so both cfg and json.

phpBB/phpbb/db/migration/data/v31x/style_update.php
```diff
@@ -49,11 +49,8 @@ public function update_installed_styles()
 		// Install prosilver if no style is available and prosilver can be installed
 		if (empty($style_paths) && in_array('prosilver', $styles))
 		{
-			// Try to parse config file
-			$cfg = parse_cfg_file($this->phpbb_root_path . 'styles/prosilver/style.cfg');
-
-			// Stop running this if prosilver cfg file can't be read
-			if (empty($cfg))
+			// Stop running this if prosilver composer.json file can't be read
+			if (file_exists($this->phpbb_root_path . 'styles/prosilver/composer.json'))
```
marc1706: Should also handle both cfg and json

phpBB/phpbb/language/language_file_helper.php
```diff
@@ -45,28 +45,66 @@ public function get_available_languages()
...
+	/**
+	 * Collect some data from the composer.json file
+	 *
+	 * @param string $path
+	 * @return array
+	 */
+	public function get_language_data_from_composer_file($path)
+	{
+		$json_data = file_get_contents($path);
+		return $this->get_language_data_from_json(json_decode($json_data, true));
```
marc1706: Should use json sanitizer.

phpBB/phpbb/json_sanitizer.php
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	/**
+	 * Decode and sanitize json data
+	 *
+	 * @param string $json JSON data string
+	 *
+	 * @return array Data array
+	 */
+	static public function decode($json)
```
rubencm: If the json isn't correct this will return null, not an array, i think this should thrown an exception

phpBB/styles/prosilver/composer.json
```diff
@@ -0,0 +1,28 @@
...
+		}
+	],
+	"support": {
+		"issues": "https://tracker.phpbb.com",
+		"forum": "https://www.phpbb.com/community/",
+		"wiki": "https://wiki.phpbb.com",
+		"irc": "irc://irc.freenode.org/phpbb"
+	},
+	"require": {
+		"php": ">=5.3.3"
```
vinny: While not mandatory for styles, the minimum recommended [version](https://github.com/phpbb/phpbb/blob/master/phpBB/docs/INSTALL.html#L150) should be used.

DavidIQ: Do we really need this for styles? If this can maybe be `phpbb` instead it would probably be more useful.

marc1706: Agreed and removed.

phpBB/phpbb/json_sanitizer.php
marc1706: I have covered this possibility now.

By-oussef:
+1

phpBB/phpbb/db/migration/data/v310/style_update_p1.php
```diff
@@ -69,13 +71,26 @@ public function styles_update()
...
+			if ($fileinfo->isDir() && !in_array($fileinfo->getFilename(), $skip_dirs))
 			{
-				$style_cfg = parse_cfg_file($fileinfo->getPathname() . '/style.cfg');
-				if (isset($style_cfg['phpbb_version']) && version_compare($style_cfg['phpbb_version'], '3.1.0-dev', '>='))
+				if (file_exists($fileinfo->getPathname() . '/style.cfg'))
+				{
+					$style_cfg = parse_cfg_file($fileinfo->getPathname() . '/style.cfg');
+					if (isset($style_cfg['phpbb_version']) && version_compare($style_cfg['phpbb_version'], '3.1.0-dev', '>='))
+					{
+						// 3.1 & 3.2 style
```
marc1706, line 81: 3.x style

phpBB/phpbb/db/migration/data/v310/style_update_p1.php
```diff
@@ -69,13 +71,26 @@ public function styles_update()
...
+				}
+				else if (file_exists($fileinfo->getPathname() . '/composer.json'))
 				{
-					// 3.1 style
-					$available_styles[] = $fileinfo->getFilename();
+					$json = file_get_contents($fileinfo->getPathname() . '/composer.json');
+					$style_data = sanitizer::decode($json);
+					if (isset($style_data['extra']['phpbb-version']) && version_compare($style_data['extra']['phpbb-version'], '3.3.0-dev', '>='))
+					{
+						// 3.3 style
```
marc1706, line 91: 4.0 style

phpBB/phpbb/db/migration/data/v310/style_update_p1.php
```diff
@@ -69,13 +71,26 @@ public function styles_update()
...
+						$available_styles[] = $fileinfo->getFilename();
+					}
+				}
+				else if (file_exists($fileinfo->getPathname() . '/composer.json'))
 				{
-					// 3.1 style
-					$available_styles[] = $fileinfo->getFilename();
+					$json = file_get_contents($fileinfo->getPathname() . '/composer.json');
+					$style_data = sanitizer::decode($json);
+					if (isset($style_data['extra']['phpbb-version']) && version_compare($style_data['extra']['phpbb-version'], '3.3.0-dev', '>='))
```
marc1706, line 89: Needs to be updated.

phpBB/phpbb/language/language_file_helper.php
```diff
@@ -45,28 +46,67 @@ public function get_available_languages()
...
+			foreach ($data['authors'] as $author)
+			{
+				if (isset($author['name']) && $author['name'] !== '')
+				{
+					$authors[] = $author['name'];
+				}
+			}
+		}
+
+		return array(
```
marc1706, line 103: Use short array syntax

marc1706:
@phpbb/development-team can be reviewed again

phpBB/language/en/acp/styles.php
```diff
@@ -54,6 +54,7 @@
 	'INSTALL_STYLES'							=> 'Install styles',
 	'INSTALL_STYLES_EXPLAIN'					=> 'Here you can install new styles.<br>If you cannot find a specific style in list below, check to make sure style is already installed. If it is not installed, check if it was uploaded correctly.',
 	'INVALID_STYLE_ID'							=> 'Invalid style ID.',
+	'INVALID_STYLE_MESSAGE'						=> 'The directory <strong><i>%1$s</i></strong> does not contain a valid style.',
```
mrgoldy, line 57: Perhaps quote the directory, as done in other language strings: `“%s”`

phpBB/phpbb/json/sanitizer.php
```diff
@@ -0,0 +1,58 @@
...
+	 *
+	 * @param array $data Data to sanitize
+	 *
+	 * @return array Sanitized data
+	 */
+	static public function sanitize(array $data) : array
+	{
+		if (!empty($data))
+		{
+			$json_sanitizer = function (&$value) {
```
mrgoldy, line 36: `{` should be on a new line

mrgoldy (commented):
Just a code review, I'll try to test it as soon as possible.

phpBB/adm/style/acp_language.html
```diff
@@ -24,6 +24,14 @@ <h1>{L_LANGUAGE_PACK_DETAILS}</h1>
 		<dt><label>{L_LANG_ISO_CODE}{L_COLON}</label></dt>
 		<dd><strong>{LANG_ISO}</strong></dd>
 	</dl>
+	<dl>
+		<dt><label>{{ lang('PHPBB_VERSION') }}{{ lang('COLON') }}</label></dt>
```
3D-I, line 28: `{{ lang('') ~ lang('COLON') }}`

Everywhere here.

phpBB/includes/acp/acp_language.php
```diff
@@ -19,130 +19,212 @@
...
 				{
-					trigger_error($user->lang['NO_LANG_ID'] . adm_back_link($this->u_action), E_USER_WARNING);
+					trigger_error($this->language->lang('NO_LANG_ID') . adm_back_link($this->u_action), E_USER_WARNING);
 				}
 
 				$this->page_title = 'LANGUAGE_PACK_DETAILS';
 
 				$sql = 'SELECT *
 					FROM ' . LANG_TABLE . '
 					WHERE lang_id = ' . $lang_id;
```
3D-I, line 181: Could be casted to `int) $lang_id`

phpBB/styles/prosilver/composer.json
```diff
@@ -0,0 +1,25 @@
...
+	"support": {
+		"issues": "https://tracker.phpbb.com",
+		"forum": "https://www.phpbb.com/community/",
+		"wiki": "https://wiki.phpbb.com",
+		"irc": "irc://irc.freenode.org/phpbb"
+	},
+	"extra": {
+		"display-name": "prosilver",
+		"phpbb-version": "4.0.0-a1-dev"
+	}
```
dark-1, line 24: May I know how will Style Parent will be Handled ???
Unless you are missing this part :
```suggestion
		"style-parent": "prosilver"
	}
```

Best regards 👍

phpBB/includes/acp/acp_language.php
```diff
@@ -19,130 +19,212 @@
...
+	 */
+	public function __construct()
 	{
 		global $config, $db, $user, $template, $phpbb_log, $phpbb_container;
 		global $phpbb_root_path, $phpEx, $request, $phpbb_dispatcher;
 
+		$this->config = $config;
+		$this->db = $db;
+		$this->dispatcher = $phpbb_dispatcher;
+		$this->language = $phpbb_container->get('language');
```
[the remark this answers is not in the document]
marc1706: I think in general the idea should be to move away from using global.

phpBB/includes/functions_compatibility.php
```diff
@@ -881,3 +881,65 @@ function phpbb_check_and_display_sql_report(\phpbb\request\request_interface $re
...
+ * Parse cfg file
+ * @param string $filename
+ * @param bool|array $lines
+ * @return array
+ *
+ * @deprecated 4.0.0-a1 (To be removed: 5.0.0)
+ */
+function parse_cfg_file($filename, $lines = false)
+{
+	$parsed_items = array();
```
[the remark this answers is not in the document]
marc1706: This is the old method still used in 3.3.x so I would refrain from modifying it.

phpBB/includes/acp/acp_language.php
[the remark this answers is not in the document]
mrgoldy: I completely agree, don't get me wrong.
But in the end we have to move away from the container aswell, so I that's why I am in favor of injecting the container as less as possible. 
Anyway, in the end it is a rather redundant remark but just something to think about.

phpBB/language/en/acp/styles.php
marc1706: I think it's usually either in bold or with quotes, not both at the same time. Do you think using quotes would be better here instead of using bold text?

phpBB/styles/prosilver/composer.json
marc1706: That's actually a good point, I think it might make sense to explicitly state the parent here. Of course this would be empty in the case of prosilver as prosilver does not have a parent.

@mrgoldy thoughts?

phpBB/includes/acp/acp_language.php
marc1706: Also safe without that as it's already casted to int further above.

phpBB/styles/prosilver/composer.json
dark-1: > ... Of course this would be empty in the case of prosilver as prosilver does not have a parent. ...

Yea!
I went as-per `style.cfg` which has parent as it's self.
Either should do as long as it is mentioned.

Best regards 👍

marc1706:
Updated

phpBB/includes/acp/acp_styles.php
```diff
@@ -1135,43 +1186,45 @@ public function sort_styles($style1, $style2)
...
+	 * @param string $dir style directory
+	 *
+	 * @return array Style data
+	 * @throws \DomainException in case of error
+	 */
+	protected function read_style_composer_file($dir)
 	{
 		// This should never happen, we give them a red warning because of its relevance.
-		if (!file_exists($this->styles_path . $dir . '/style.cfg'))
+		if (!file_exists($this->styles_path . $dir . '/style.json'))
```
marc1706, line 1199: Should be composer.json

phpBB/includes/acp/acp_language.php
```diff
@@ -19,130 +19,212 @@
...
+use phpbb\language\language_file_helper;
+use phpbb\log\log_interface;
+use phpbb\request\request_interface;
+use phpbb\template\template;
+use phpbb\user;
+
 class acp_language
 {
 	var $u_action;
 	var $main_files;
```
rubencm, line 35: This variable isn't used

phpBB/includes/functions_compatibility.php
```diff
@@ -881,3 +881,65 @@ function phpbb_check_and_display_sql_report(\phpbb\request\request_interface $re
...
 	$controller_helper->display_sql_report();
 }
+
+/**
+ * Parse cfg file
+ * @param string $filename
+ * @param bool|array $lines
+ * @return array
+ *
+ * @deprecated 4.0.0-a1 (To be removed: 5.0.0)
```
rubencm, line 891: 5.0.0?

marc1706: Could also be removed in 4.1

rubencm: I think is better tu put it deprecated in 3.3 (or 3.4) and remove it in 4.0

marc1706: Well it imho should be deprecated in the next feature or major version updated.

phpBB/phpbb/language/language_file_helper.php
```diff
@@ -45,28 +46,67 @@ public function get_available_languages()
...
 		}
 
 		return $available_languages;
 	}
+
+	/**
+	 * Collect some data from the composer.json file
+	 *
+	 * @param string $path
+	 * @return array
```
CHItA, line 70: Maybe `@throws` here? Also you can alias it if you want to drop the leading backslash from `DomainException`.

phpBB/phpbb/version_helper.php
```diff
@@ -389,17 +389,8 @@ public function get_versions($force_update = false, $force_cache = false)
...
 			// Sanitize any data we retrieve from a server
-			if (!empty($info))
-			{
-				$json_sanitizer = function (&$value, $key) {
-					$type_cast_helper = new \phpbb\request\type_cast_helper();
-					$type_cast_helper->set_var($value, $value, gettype($value), true);
-				};
-				array_walk_recursive($info, $json_sanitizer);
-			}
+			$info = json\sanitizer::decode($info);
```
CHItA, line 393: Maybe add a `use` for this. What if we move the file to a different namespace?

phpBB/includes/acp/acp_language.php
```diff
@@ -19,130 +19,211 @@
...
-
-					'LANG_LOCAL_NAME'	=> $lang_entries['lang_local_name'],
-					'LANG_ENGLISH_NAME'	=> $lang_entries['lang_english_name'],
-					'LANG_ISO'			=> $lang_iso,
-					'LANG_AUTHOR'		=> $lang_entries['lang_author'],
-					'L_MISSING_FILES'			=> $user->lang('THOSE_MISSING_LANG_FILES', $lang_entries['lang_local_name']),
-					'L_MISSING_VARS_EXPLAIN'	=> $user->lang('THOSE_MISSING_LANG_VARIABLES', $lang_entries['lang_local_name']),
+				try
+				{
+					$lang_cfg = $this->language_helper->get_language_data_from_composer_file("{$this->phpbb_root_path}language/$lang_iso/composer.json");
```
CHItA, line 194: My preference generally is concatenation by `.`, but this rather a small thing, so feel free to ignore me.

phpBB/includes/acp/acp_language.php
```diff
@@ -160,21 +241,21 @@ function main($id, $mode)
 						$relative_path = $iterator->getInnerIterator()->getSubPathname();
 						$relative_path = str_replace(DIRECTORY_SEPARATOR, '/', $relative_path);
 
-						if (file_exists($phpbb_root_path . 'language/' . $lang_iso . '/' . $relative_path))
+						if (file_exists($this->phpbb_root_path . 'language/' . $lang_iso . '/' . $relative_path))
 						{
-							if (substr($relative_path, 0 - strlen($phpEx)) === $phpEx)
+							if (substr($relative_path, 0 - strlen($this->php_ext)) === $this->php_ext)
```
CHItA, line 246: Can't we just `-strlen()`, this looks weird.

CHItA (commented):
LGTM in general.

Nicofuma:
looks good

phpBB/includes/acp/acp_language.php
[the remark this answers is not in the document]
Nicofuma: to be honnest, I think it is better to have only one global (the container) instead of 30

phpBB/phpbb/language/language_file_helper.php
marc1706: Added those.

marc1706:
Updated again based on latest comments

marc1706:
And rebased again. :)

CHItA (approved):
LGTM

PR #4130, opened 2016-01-18
[ticket/14419] Update composer dependencies and lock file

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-14419


marc1706:
@Nicofuma I guess symfony causes password-compat to be added?

Nicofuma:
yes and the next release will introduce paragonie-random (or something like that) because they had a sec issue in their own random_bytes implementation

PR #4131, opened 2016-01-18
[ticket/13717] Set the assets after rendering the whole template

Nicofuma:
The goal being to be able to call INCLUDECSS/JS from anywhere in any tempalte or event

PHPBB3-13717


Nicofuma:
and it's green :)

PR #4133, opened 2016-01-19
 twig extension for template name

agrisvv:
Allow twig extension for template name for better IDE support.

https://tracker.phpbb.com/browse/PHPBB3-14421

### DIFF
--- a/phpBB/includes/functions_module.php
+++ b/phpBB/includes/functions_module.php
@@ -953,7 +953,7 @@ class p_master
 	*/
 	function get_tpl_name()
 	{
-		return $this->module->tpl_name . '.html';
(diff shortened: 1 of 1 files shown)


VSEphpbb:
phpBB does not use .twig files as part of its core. We should not change core code just to make an IDE's syntax highlighter happy. Besides, my IDE highlights twig syntax in .html fine, but I set it up to do so, and most should allow that if desired.

agrisvv:
i understand you. but i know what some day phpbb code will go to full names of template files, because, templates could bee different type- xml json html xsd.... 
i think it is not hard to change core code, just allow template names with extensions not just *.html.
thanks.

VSEphpbb:
When/if phpBB does switch to twig only templates we will handle the naming recognition properly at that time.

marc1706:
Closing as per explanation of @VSEphpbb

PR #4141, opened 2016-01-24
[ticket/14431] Remote avatar uploading

exussum12:
Allow HTTPS images to be remotely uploaded.
Also includes support for redirects (currently up to 5)

PHPBB3-14431

### DIFF
--- a/phpBB/includes/functions_upload.php
+++ b/phpBB/includes/functions_upload.php
@@ -752,10 +752,11 @@ class fileupload
 	*
 	* @param string $upload_url URL pointing to file to upload, for example http://www.foobar.com/example.gif
 	* @param \phpbb\mimetype\guesser $mimetype_guesser Mimetype guesser
+	* @param int $redirect_count the current count of redirects
 	* @return object $file Object "filespec" is returned, all further operations can be done with this object
 	* @access public
 	*/
-	function remote_upload($upload_url, \phpbb\mimetype\guesser $mimetype_guesser = null)
+	function remote_upload($upload_url, \phpbb\mimetype\guesser $mimetype_guesser = null, $redirect_count = 0)
 	{
 		global $user, $phpbb_root_path;
 
@@ -776,9 +777,18 @@ class fileupload
 
 		$url = parse_url($upload_url);
 
+		$default_port = 80;
+		$hostname = $url['host'];
+
+		if ($url['scheme'] == 'https')
+		{
+			$default_port = 443;
+			$hostname = 'tls://' . $url['host'];
+		}
+
 		$host = $url['host'];
 		$path = $url['path'];
-		$port = (!empty($url['port'])) ? (int) $url['port'] : 80;
+		$port = (!empty($url['port'])) ? (int) $url['port'] : $default_port;
 
 		$upload_ary['type'] = 'application/octet-stream';
 
@@ -818,7 +828,7 @@ class fileupload
 		$errno = 0;
 		$errstr = '';
 
-		if (!($fsock = @fsockopen($host, $port, $errno, $errstr)))
+		if (!($fsock = @fsockopen($hostname, $port, $errno, $errstr)))
 		{
 			$file = new fileerror($user->lang[$this->error_prefix . 'NOT_UPLOADED']);
 			return $file;
@@ -898,6 +908,21 @@ class fileupload
 					{
 						$file = new fileerror($user->lang[$this->error_prefix . 'URL_NOT_FOUND']);
 						return $file;
+					}
+					else if (stripos($line, 'location: ') !== false)
+					{
+						//there is a redirect, follow up to 5
+						if ($redirect_count >= 5)
+						{
+							$file = new fileerror($user->lang[$this->error_prefix . 'URL_NOT_FOUND']);
+							return $file;
+						}
+
+						$upload_url = rtrim(str_replace('location: ', '', strtolower($line)));
+						//close the current connection, lets not leave dangeling connections open
+						@fclose($fsock);
+
+						return $this->remote_upload($upload_url, $mimetype_guesser, ++$redirect_count);
 					}
 				}
 			}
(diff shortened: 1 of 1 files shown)


Nicofuma:
looks good to me, even if at some point we should start using a proper http client/lib.
@marc1706 thoughts about the current patch?

exussum12:
Agreed. I assume curl / guzzle are not used due to support on shared hosting platforms ?

phpBB/includes/functions_upload.php
```diff
@@ -899,6 +909,18 @@ function remote_upload($upload_url, \phpbb\mimetype\guesser $mimetype_guesser =
 						$file = new fileerror($user->lang[$this->error_prefix . 'URL_NOT_FOUND']);
 						return $file;
 					}
+					else if (stripos($line, 'location: ') !== false)
+					{
+						//there is a redirect, follow up to 5
+						if ($redirect_count > 5)
```
ErnadoO: If we want up to 5 redirects, it should be >= 5 ;)

exussum12: Missing the initial request. The initial request is 0, 1 means currently processing the first redirect. 5 means currently processing the 5th redirect. So it seems right to me

Nicofuma:
We can use guzzle actually. The lib provides a lot of handlers and doesn't rely on cURL being available

marc1706:
I think we should go for guzzle instead of trying to reinvent the wheel here.

exussum12:
For 3.1 too or just master ?

Nicofuma:
master

exussum12:
As commented in the main ticket. Guzzle is only a dev requirement currently. Is it OK to move guzzle to a main dependecy ?

Nicofuma:
but honestly I think we should take our time and use guzzle properly everywhere for 3.3.
In the meantime this solution is good.

exussum12:
Ah OK. This is a pr for 3.1. I'll update masters pull to use guzzle

phpBB/includes/functions_upload.php
ErnadoO: Exactly, so this condition is true for $redirect_count = 6, that mean after the 6th redirect, or we want to stop before the 6th.

phpBB/includes/functions_upload.php
```diff
@@ -899,6 +909,18 @@ function remote_upload($upload_url, \phpbb\mimetype\guesser $mimetype_guesser =
...
+						{
+							$file = new fileerror($user->lang[$this->error_prefix . 'URL_NOT_FOUND']);
+							return $file;
+						}
+
+						$upload_url = rtrim(str_replace('location: ', '', strtolower($line)));
+						return $this->remote_upload($upload_url, $mimetype_guesser, ++$redirect_count);
```
bantu: No recursion please.

bantu:
One important thing that has to be kept in mind here: The web server accessing the specified URL might have access to resources on the network that can not be accessed by the client requesting the remote URL to be downloaded. This may be a security issue.

Nicofuma:
That's right... Do you see any way to prevent the issue?

exussum12:
That really is a non issue. They could put the very same URL directly in the box and have it download directly rather than have it redirect first

phpBB/includes/functions_upload.php
exussum12: Generally I agree. But there is a fair amount of logic would either need to be repeated or wrapping the whole lot in a loop. Personally I think the recursion is the best way here

phpBB/includes/functions_upload.php
exussum12: Apologies your correct, I was getting confused with the original request

exussum12:
Added some code to close the connection, 5 redirects means 5 potentially open connections to the server

Nicofuma:
looks good to me. @nickvergessen

marc1706:
This breaks the upload time limit we initially set with the timer stop variable. That means that for every redirect a new timeout is started. Not sure how desirable this is.

marc1706:
It also opens concurrent connections without closing the previous one. To be honest, I'd rather not implement this for 3.1.x as it's a new feature and might cause more issues.

Nicofuma:
Didn't think about that... and I tend to agree with you.

@exussum12 But it is still valid for 3.2, but we need to hurry with your patch if we want it in 3.2 :)

nickvergessen:
I'd say 3.2 as well

exussum12:
Worth adding in the checks for the timeout ? or just add as a 3.2 only feature ?

exussum12:
@marc1706  it does close the connections. Added that in 610048c

exussum12:
Or I can just remove the redirect code and have the https stuff ? Guzzle will have the same issue in the master branch. I'm not 100% sure it's an issue really though ? Do you want the call to return in a Max of 5 seconds ?

phpBB/includes/functions_upload.php
```diff
@@ -899,6 +909,21 @@ function remote_upload($upload_url, \phpbb\mimetype\guesser $mimetype_guesser =
...
+					{
+						//there is a redirect, follow up to 5
+						if ($redirect_count >= 5)
+						{
+							$file = new fileerror($user->lang[$this->error_prefix . 'URL_NOT_FOUND']);
+							return $file;
+						}
+
+						$upload_url = rtrim(str_replace('location: ', '', strtolower($line)));
+						//close the current connection, lets not leave dangeling connections open
```
cs278: 'dangling'

phpBB/includes/functions_upload.php
```diff
@@ -752,10 +752,11 @@ function local_upload($source_file, $filedata = false, \phpbb\mimetype\guesser $
 	*
 	* @param string $upload_url URL pointing to file to upload, for example http://www.foobar.com/example.gif
 	* @param \phpbb\mimetype\guesser $mimetype_guesser Mimetype guesser
+	* @param int $redirect_count the current count of redirects
 	* @return object $file Object "filespec" is returned, all further operations can be done with this object
 	* @access public
 	*/
-	function remote_upload($upload_url, \phpbb\mimetype\guesser $mimetype_guesser = null)
+	function remote_upload($upload_url, \phpbb\mimetype\guesser $mimetype_guesser = null, $redirect_count = 0)
```
cs278: I'd implement this by setting `$redirect_count` to 5 by default and then counting down to zero, means a caller may disable redirect processing or decrease the number of follows.

exussum12:
Removed the redirect, This is now just allowing the HTTPS downloads

PR #4142, opened 2016-01-25
[ticket/14129] Caches extensions autoloaders

Nicofuma:
Replaces #3885

https://tracker.phpbb.com/browse/PHPBB3-14129

PHPBB3-14129

### DIFF
--- a/phpBB/phpbb/di/container_builder.php
+++ b/phpBB/phpbb/di/container_builder.php
@@ -135,6 +135,11 @@ class container_builder
 			$config_cache = new ConfigCache($container_filename, defined('DEBUG'));
 			if ($this->use_cache && $config_cache->isFresh())
 			{
+				if ($this->use_extensions)
+				{
+					require($this->get_autoload_filename());
+				}
+
 				require($config_cache->getPath());
 				$this->container = new \phpbb_cache_container();
 			}
@@ -405,6 +410,15 @@ class container_builder
 			$extensions = $ext_container->get('ext.manager')->all_enabled();
 
 			// Load each extension found
+			$autoloaders = '<?php
+/**
+ * Loads all extensions custom auto-loaders.
+ *
+ * This file has been auto-generated
+ * by phpBB while loading the extensions.
+ */
+
+';
 			foreach ($extensions as $ext_name => $path)
 			{
 				$extension_class = '\\' . str_replace('/', '\\', $ext_name) . '\\di\\extension';
@@ -420,9 +434,14 @@ class container_builder
 				$filename = $path . 'vendor/autoload.php';
 				if (file_exists($filename))
 				{
-					require $filename;
+					$autoloaders .= "require('{$filename}');\n";
 				}
 			}
+
+			$configCache = new ConfigCache($this->get_autoload_filename(), false);
+			$configCache->write($autoloaders);
+
+			require($this->get_autoload_filename());
 		}
 		else
 		{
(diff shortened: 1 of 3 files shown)


Nicofuma:
@marc1706 thoughts?

rxu:
Got this error while tried testing it:
> Catchable fatal error: Object of class phpbb\config_php_file could not be converted to string in C:\Program Files\Apache\htdocs\306\phpbb\di\container_builder.php on line 385

PHP 7.0.1.

phpBB/phpbb/di/container_builder.php
```diff
@@ -135,6 +135,11 @@ public function get_container()
 			$config_cache = new ConfigCache($container_filename, defined('DEBUG'));
 			if ($this->use_cache && $config_cache->isFresh())
 			{
+				if ($this->use_extensions)
+				{
+					require $this->get_autoload_filename();
```
Elsensee: Parenthesis, I guess.. Or is it okay like that? I don't know the coding guidelines here.

Nicofuma: me neither...

phpBB/phpbb/di/container_builder.php
```diff
@@ -540,6 +559,16 @@ protected function get_container_filename()
 	}
 
 	/**
+	 * Get the filename under which the dumped extensions autoloader will be stored.
+	 *
+	 * @return string Path for dumped extensions autoloader
+	 */
+	protected function get_autoload_filename()
+	{
+		return $this->get_cache_dir() . 'autoload_' . md5($this->phpbb_root_path) . '.' . $this->php_ext;
```
Elsensee: This needs to be added to $cache->purge(), then

Nicofuma: well, it's not necessary (it will always be erased by the container builder) but I can add it

phpBB/phpbb/di/container_builder.php
```diff
@@ -420,9 +434,14 @@ protected function load_extensions()
 				$filename = $path . 'vendor/autoload.php';
 				if (file_exists($filename))
 				{
-					require $filename;
+					$autoloaders .= "require '{$filename}';\n";
 				}
 			}
+
+			$configCache = new ConfigCache($this->get_autoload_filename(), false);
```
marc1706: Shouldn't this also handle debug mode?

phpBB/phpbb/di/container_builder.php
marc1706: We usually use ```require(...)```. Also see the piece of code right below this. I suggest we use require() everywhere.

phpBB/phpbb/di/container_builder.php
```diff
@@ -420,9 +434,14 @@ protected function load_extensions()
 				$filename = $path . 'vendor/autoload.php';
 				if (file_exists($filename))
 				{
-					require $filename;
+					$autoloaders .= "require '{$filename}';\n";
```
marc1706: Should be ```require('{$filename}')```

phpBB/phpbb/di/container_builder.php
Nicofuma: I'm using ConfigCache just to not have to write the dumpFile logic myself but I don't need the others features (including debug versus not debug)

phpBB/phpbb/di/container_builder.php
Nicofuma: ok

Nicofuma:
updated require() calls

phpBB/phpbb/di/container_builder.php
marc1706: Please also add it.

Nicofuma: updated

PR #4144, opened 2016-01-26
[ticket/14434] Allow non-migration files inside migrations folder (continued)

VSEphpbb:
https://tracker.phpbb.com/browse/PHPBB3-14434

PHPBB3-14434

### DIFF
--- a/phpBB/phpbb/console/command/db/list_command.php
+++ b/phpBB/phpbb/console/command/db/list_command.php
@@ -39,6 +39,12 @@ class list_command extends \phpbb\console\command\db\migration_command
 
 		foreach ($this->load_migrations() as $name)
 		{
+			// Ignore non-migration files
+			if (\phpbb\db\migrator::is_migration($name) === false)
+			{
+				continue;
+			}
+
 			if ($this->migrator->migration_state($name) !== false)
 			{
 				$installed[] = $name;
--- a/phpBB/phpbb/console/command/db/list_command.php
+++ b/phpBB/phpbb/console/command/db/list_command.php
@@ -39,12 +39,6 @@ class list_command extends \phpbb\console\command\db\migration_command
 
 		foreach ($this->load_migrations() as $name)
 		{
-			// Ignore non-migration files
-			if (\phpbb\db\migrator::is_migration($name) === false)
-			{
-				continue;
-			}
-
 			if ($this->migrator->migration_state($name) !== false)
 			{
 				$installed[] = $name;
--- a/phpBB/phpbb/console/command/db/migration_command.php
+++ b/phpBB/phpbb/console/command/db/migration_command.php
@@ -45,7 +45,7 @@ abstract class migration_command extends \phpbb\console\command\command
 
 		$this->migrator->set_migrations($migrations);
 
-		return $migrations;
+		return $this->migrator->get_migrations();
 	}
 
 	protected function finalise_update()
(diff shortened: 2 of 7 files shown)


VSEphpbb:
@marc1706 Found this problem also happens, again, in the phpbbcli
```
PHP Fatal error:  Cannot instantiate abstract class vse\abbc3\migrations\bbcode_migration_base in /phpBB/phpbb/db/migrator.php on line 799
PHP Stack trace:
PHP   1. {main}() /phpBB/bin/phpbbcli.php:0
PHP   2. Symfony\Component\Console\Application->run() /phpBB/bin/phpbbcli.php:84
PHP   3. phpbb\console\application->doRun() /phpBB/vendor/symfony/console/Application.php:123
PHP   4. Symfony\Component\Console\Application->doRun() /phpBB/phpbb/console/application.php:122
PHP   5. Symfony\Component\Console\Application->doRunCommand() /phpBB/vendor/symfony/console/Application.php:192
PHP   6. Symfony\Component\Console\Command\Command->run() /phpBB/vendor/symfony/console/Application.php:857
PHP   7. phpbb\console\command\db\migrate->execute() /phpBB/vendor/symfony/console/Command/Command.php:259
PHP   8. phpbb\db\migrator->finished() /phpBB/phpbb/console/command/db/migrate.php:61
PHP   9. phpbb\db\migrator->unfulfillable() /phpBB/phpbb/db/migrator.php:756
PHP  10. phpbb\db\migrator->get_migration() /phpBB/phpbb/db/migrator.php:727
```

Maybe we need a better solution since this seems to be cropping up in multiple areas?

VSEphpbb:
Updated. Maybe a little ugly, but this now checks the migration files when migrations are installed in the ACP, via CLI, and during unit testing.

@marc1706 
@Nicofuma

Elsensee:
I don't like this patch. This method is okay-ish but it doesn't belong into the migration class itself IMHO.
Also it's **just** okay-ish because the method name isn't accurate if an array was given. I would actually expect a boolean then and not to remove migrations from that array that cannot be instantiated.

VSEphpbb:
Updated

VSEphpbb:
Cleaned up a little more

VSEphpbb:
Found/Fixed another spot.

phpBB/phpbb/console/command/db/list_command.php
```diff
@@ -39,6 +39,12 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
 		foreach ($this->load_migrations() as $name)
 		{
+			// Ignore non-migration files
+			if (\phpbb\db\migrator::is_migration($name) === false)
```
Elsensee: Would rather put this into migration_command::load_migration

VSEphpbb: that'll get ugly..will be another loop

VSEphpbb: It would also run twice when running the `migrate` cli command. Once inside load_migrations, then again inside the migration class

Nicofuma:
Couldn't we have the checks in migrator::set_migrations() to fix most of the issues?

VSEphpbb:
@Nicofuma It looks like doing that would break the functionality in the various migrator methods where it returns the name if the class is not valid, ie: https://github.com/phpbb/phpbb/blob/master/phpBB/phpbb/db/migrator.php#L722-L725

VSEphpbb:
^ It also won't change the issues being addressed in base or schema_generator classes.

VSEphpbb:
@Elsensee and @Nicofuma I have an idea to address both your comments but will need to discuss on IRC :smile:

VSEphpbb:
Updated so migrations are checked within the migrator now (except schema generator which is a special case)

PR #4153, opened 2016-02-01
[ticket/14443] Introduce additional parameter to messenger

gn36:
See https://tracker.phpbb.com/browse/PHPBB3-14443

PHPBB3-14443

### DIFF
--- a/phpBB/includes/functions_messenger.php
+++ b/phpBB/includes/functions_messenger.php
@@ -214,6 +214,8 @@ class messenger
 	{
 		global $config, $phpbb_root_path, $phpEx, $user, $phpbb_extension_manager;
 
+		$template_dir_prefix = (!$template_dir_prefix || substr($template_dir_prefix, 0, 1) === '/') ? $template_dir_prefix : '/' . $template_dir_prefix;
+
 		$this->setup_template();
 
 		if (!trim($template_file))
@@ -232,7 +234,7 @@ class messenger
 		if ($template_path)
 		{
 			$template_paths = array(
-				$template_path . '/' . $template_dir_prefix,
+				$template_path . $template_dir_prefix,
 			);
 		}
 		else
@@ -241,7 +243,7 @@ class messenger
 			$template_path .= $template_lang . '/email';
 
 			$template_paths = array(
-				$template_path . '/' . $template_dir_prefix,
+				$template_path . $template_dir_prefix,
 			);
 
 			// we can only specify default language fallback when the path is not a custom one for which we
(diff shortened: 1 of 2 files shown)


Elsensee:
I have checked and phpBB itself doesn't use the `$template_path` parameter and to be honest, I doubt that any extension will use it. I guess we keep it anyway to keep some sort of BC? Somebody agreeing/disagreeing here?

marc1706:
It's called a prefix but to me it seems like it's a suffix?

gn36:
I actually thought the same thing. It is a suffix to the path, but a prefix to the template name, which is probably the reason it was named this way everywhere it is used originally (note that the original use was actually prefixing this path to the template name). The only reason I kept the name was to keep it the same as in the methods that call this one, so it is easier to track the function of the variable.

I can rename it if you want. Since this parameter is handed down from other methods, I would then suggest renaming them all, though.

phpBB/includes/functions_messenger.php
```diff
@@ -232,7 +232,7 @@ function template($template_file, $template_lang = '', $template_path = '')
 		if ($template_path)
 		{
 			$template_paths = array(
-				$template_path,
+				$template_path . '/' . $template_dir_prefix,
```
Nicofuma: Shouldn't it be a ternary to avoid an extra `/` if `$template_dir_prefix` is empty?

marc1706: Yes, should only be added if the prefix is not empty.

marc1706:
Ah ok, that's the thinking behind calling it a prefix. You can leave it at that, then.

phpBB/includes/functions_messenger.php
gn36: Oops, I will take care of that.

Elsensee:
:+1:

phpBB/includes/functions_messenger.php
```diff
@@ -210,10 +210,12 @@ function set_mail_priority($priority = MAIL_NORMAL_PRIORITY)
 	/**
 	* Set email template to use
 	*/
-	function template($template_file, $template_lang = '', $template_path = '')
+	function template($template_file, $template_lang = '', $template_path = '', $template_dir_prefix = '')
 	{
 		global $config, $phpbb_root_path, $phpEx, $user, $phpbb_extension_manager;
 
+		$template_dir_prefix = (!$template_dir_prefix || substr($template_dir_prefix, 0, 1) === '/') ? $template_dir_prefix : '/' . $template_dir_prefix;
```
nickvergessen: `substr($template_dir_prefix, 0, 1)` is equivalent to `$template_dir_prefix[0]`, which should be faster, but in this case it is negligible in this case I guess

gn36: @nickvergessen: I can certainly change that if you want me to.

marc1706: I'd prefer that change, too. Please modify it when you have some time and we'll review the PR again once you've updated it.

gn36: @marc1706: Done.

PR #4157, opened 2016-02-02
[ticket/14447] Remove trailing white spaces in Prosilver

VSEphpbb:
https://tracker.phpbb.com/browse/PHPBB3-14447

PHPBB3-14447


bantu:
We should detect those ...

marc1706:
I think the code sniffer only checks PHP files right now. We definitely might want to check for trailing spaces in html files, too.

bantu:
@marc1706 PHP Code Sniffer does not seem to like being applied to HTML files. Hummm.

bantu:
This PR looks good. :+1:

Nicofuma:
I don't really like this sort of PR (poor style team and contributors...) but... it's kinda necessary.

Thanks @VSEphpbb

PR #4158, opened 2016-02-02
[ticket/14321] Clean up arcihve controller logic

CHItA:
[PHPBB3-14321](https://tracker.phpbb.com/browse/PHPBB3-14321)

This bug should have been solved by the solution of [14269](https://tracker.phpbb.com/browse/PHPBB3-14269) so this is more of a clean up then anything else.

### DIFF
--- a/phpBB/phpbb/install/controller/archive_download.php
+++ b/phpBB/phpbb/install/controller/archive_download.php
@@ -46,9 +46,9 @@ class archive_download
 	 */
 	public function conflict_archive()
 	{
-		$filename = $this->installer_config->get('update_file_conflict_archive', false);
+		$filename = $this->installer_config->get('update_file_conflict_archive', '');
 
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/install/controller/archive_download.php
```diff
@@ -46,9 +46,9 @@ public function __construct(config $config)
 	 */
 	public function conflict_archive()
 	{
-		$filename = $this->installer_config->get('update_file_conflict_archive', false);
+		$filename = $this->installer_config->get('update_file_conflict_archive', '');
 
-		if (!$filename)
+		if (!empty($filename))
```
Elsensee: You sure? Not something like `if (empty($filename))`?

CHItA: yep, thanks

PR #4159, opened 2016-02-03
[ticket/14448] Use guzzle for remote files uploading

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-14448


Elsensee:
Don't forget to cleanup the vendor directory, since you added a new dependency. I think we forgot this the past few times, we updated the composer.json or .lock.

marc1706:
Already on it

marc1706:
@Elsensee done

exussum12:
Will this break what ever guzzle was used for before? II moved guzzle from dev to the main block. Will the old code still work ( I didn't actually search where it was used )

marc1706:
Guzzle was and is used by Goutte for functional tests. As Goutte was updated to version 2.0, the fitting guzzle version is 5.3.0 which is also why some of the previous implementation changed.

phpBB/phpbb/files/types/remote.php
```diff
@@ -106,108 +99,55 @@ protected function remote_upload($upload_url)
...
-		$filesize = 0;
 
 		$remote_max_filesize = $this->get_max_file_size();
 
-		$errno = 0;
-		$errstr = '';
+		$guzzle_options = [
+			'timeout'			=> $this->upload->upload_timeout,
+			'connect_timeout'	=> $this->upload->upload_timeout,
+			'verify'			=> !empty($this->config['remote_upload_verify']),
```
Nicofuma: shouldn't be the result of !empty but the actual value of the setting cast to boolean

marc1706: If you cast the setting to boolean it can be either false if you have it set to false, 0, etc. or true if you have anything like 1 or true set. This will result in exactly the same.

tests/files/types_remote_test.php
```diff
@@ -112,22 +117,9 @@ public function test_get_max_file_size($max_file_size, $link, $expected = array(
 		$this->assertSame($expected, $file->error);
 	}
 
-	public function test_upload_timeout()
```
Nicofuma: why did you remove this test?

marc1706: Before we were able to specify a negative timeout and therefore hit the timeout limit before even uploading the file. With guzzle, the smallest timeout one can specify is 0.001s which is 1 ms. Anything below that or negative will be ignored, hence why this test is no longer possible. A 1 ms timeout still sometimes does not trigger the timeout during upload.

Nicofuma: I see :/

PR #4166, opened 2016-02-05
[ticket/14315] Correctly set permission roles for multiple groups

marc1706:
The previous code was based on multiple false implications. These changes should fix those.

Ticket: https://tracker.phpbb.com/browse/PHPBB3-14315

### DIFF
--- a/phpBB/adm/style/admin.css
+++ b/phpBB/adm/style/admin.css
@@ -2471,7 +2471,7 @@ fieldset.permissions .padding {
 
 .roles-options > .dropdown {
 	left: auto;
-	top: 3em;
+	top: 3.2em;
 	width: 250px;
 }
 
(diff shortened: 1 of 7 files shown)


Nicofuma:
Should be possible to write a test for that, right?

Nicofuma:
@marc1706 bump for a test?

marc1706:
Hardly testable thx to all that javascript. I'll probably have to take a look at the ui tests ...

Nicofuma:
IIRC the JS is just to hide/show the tabs using `display: none`. Rigth? Because if it is the case you don't need the UI tests, the form is present in the DOM.

marc1706:
Turns out it didn't actually work without javascript. I've added back the previous select. It will get hidden by the javascript code and replaced by the span that allows us to show tooltips.

marc1706:
This also means I'll probably have to add both normal functional tests and ui tests for the javascript part.

marc1706:
@Nicofuma added ui and functional tests

marc1706:
Having some issues with ui tests randomly acting up

marc1706:
@Nicofuma UI tests seem to be fixed now. Session ID for ACP was incorrectly retrieved from cookies.

Nicofuma:
nice, thanks

PR #4171, opened 2016-02-06
[ticket/14462] Try to prevent timeouts in the installer

CHItA:
[PHPBB3-14462](https://tracker.phpbb.com/browse/PHPBB3-14462)

### DIFF
--- a/phpBB/assets/javascript/installer.js
+++ b/phpBB/assets/javascript/installer.js
@@ -12,6 +12,7 @@
 	var progressTimer = null;
 	var currentProgress = 0;
 	var refreshRequested = false;
+	var transmissionOver = false;
 
 	// Template related variables
 	var $contentWrapper = $('.install-body').find('.main');
@@ -329,6 +330,12 @@
 		if (responseObject.hasOwnProperty('redirect')) {
 			redirect(responseObject.redirect.url, responseObject.redirect.use_ajax);
 		}
+
+		if (responseObject.hasOwnProperty('over')) {
+			if (responseObject.over) {
+				transmissionOver = true;
+			}
+		}
 	}
 
 	/**
@@ -357,9 +364,20 @@
 			$('#loading_indicator').css('display', 'none');
 			resetPolling();
 
+			var timeoutDetected = !transmissionOver;
+
 			if (refreshRequested) {
 				refreshRequested = false;
 				doRefresh();
+			}
+
+			if (timeoutDetected) {
+				addMessage('error',
+					[{
+						title:'The installer detected a timeout.',
+						description: 'The installer has detected a timeout, you may try to refresh the page, that may lead to data corruption. We suggest that you either increase your timeout settings or try to use the CLI.'
+					}]
+				);
 			}
 		}
 	}
@@ -420,6 +438,7 @@
 	 */
 	function startPolling(xhReq) {
 		resetPolling();
+		transmissionOver = false;
 		pollTimer = setInterval(function () {
 			pollContent(xhReq);
 		}, 250);
--- a/phpBB/assets/javascript/installer.js
+++ b/phpBB/assets/javascript/installer.js
@@ -374,8 +374,8 @@
 			if (timeoutDetected) {
 				addMessage('error',
 					[{
-						title:'The installer detected a timeout.',
-						description: 'The installer has detected a timeout, you may try to refresh the page, that may lead to data corruption. We suggest that you either increase your timeout settings or try to use the CLI.'
+						title: installLang.title,
+						description: installLang.msg
 					}]
 				);
 			}
--- a/phpBB/config/installer/container/services_install_database.yml
+++ b/phpBB/config/installer/container/services_install_database.yml
@@ -19,7 +19,7 @@ services:
             - '@installer.helper.iohandler'
             - '%core.root_path%'
         tags:
-            - { name: install_database_install, order: 13 }
+            - { name: install_database_install, order: 20 }
 
     installer.install_database.add_tables:
         class: phpbb\install\module\install_database\task\add_tables
@@ -29,7 +29,7 @@ services:
             - '@filesystem'
             - '%core.root_path%'
         tags:
-            - { name: install_database_install, order: 17 }
+            - { name: install_database_install, order: 30 }
 
     installer.install_database.add_default_data:
         class: phpbb\install\module\install_database\task\add_default_data
@@ -41,7 +41,7 @@ services:
             - '@language'
             - '%core.root_path%'
         tags:
-            - { name: install_database_install, order: 20 }
+            - { name: install_database_install, order: 40 }
 
     installer.install_database.add_config_settings:
         class: phpbb\install\module\install_database\task\add_config_settings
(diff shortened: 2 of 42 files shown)


CHItA:
@phpbb/development-team there are some tasks where this could be added but i don't think it would make too much sense. if you disagree, please let me know

phpBB/phpbb/install/module/install_database/task/add_config_settings.php
```diff
@@ -320,6 +324,15 @@ public function run()
 				$error = $this->db->sql_error($this->db->get_sql_error_sql());
 				$this->iohandler->add_error_message('INST_ERR_DB', $error['message']);
 			}
+
+			$i++;
```
Elsensee: I don't see how `$i` affects anything? (Same for add_bots)

CHItA: true, i forgot about the `array_slice()`... not my day

phpBB/phpbb/install/module/install_database/task/add_tables.php
```diff
@@ -0,0 +1,158 @@
...
+		{
+			$db_table_schema = array_slice($db_table_schema, $i);
+
+			$fp = @fopen($this->schema_file_path, 'wb');
+			if (!$fp)
+			{
+				throw new \Exception('INST_SCHEMA_FILE_NOT_WRITABLE');
+			}
+
+			fwrite($fp, $db_table_schema);
```
Elsensee: json_encode()?

phpBB/assets/javascript/installer.js
```diff
@@ -357,10 +364,21 @@
...
 			if (refreshRequested) {
 				refreshRequested = false;
 				doRefresh();
 			}
+
+			if (timeoutDetected) {
+				addMessage('error',
+					[{
+						title:'The installer detected a timeout.',
+						description: 'The installer has detected a timeout, you may try to refresh the page, that may lead to data corruption. We suggest that you either increase your timeout settings or try to use the CLI.'
```
Elsensee: Should be translatable

phpBB/phpbb/install/helper/config.php
```diff
@@ -95,8 +95,9 @@ public function __construct(\phpbb\filesystem\filesystem_interface $filesystem,
 		$this->installer_config	= array();
 		$this->system_data		= array();
 		$this->progress_data	= array(
-			'last_task_module_name'		=> '', // Stores the service name of the latest finished module
-			'last_task_name'			=> '', // Stores the service name of the latest finished task
+			'last_task_module_neme'		=> '', // Stores the service name of the latest finished module
```
Elsensee: Typo -> "name"

phpBB/assets/javascript/installer.js
CHItA: I would pass on that for now if possible. i don't think that really matters + pretty hard to do.

phpBB/config/installer/container/services_install_database.yml
```diff
@@ -1,16 +1,36 @@
...
+    installer.install_database.set_up_database:
+        class: phpbb\install\module\install_database\task\set_up_database
+        arguments:
+            - '@installer.helper.config'
+            - '@installer.helper.database'
+            - '@filesystem'
+            - '@installer.helper.iohandler'
+            - '%core.root_path%'
+        tags:
+            - { name: install_database_install, order: 13 }
```
marc1706: I'd prefer to have at least a step of 5 and not uneven steps (in this case 3 and 4).

phpBB/phpbb/install/installer.php
```diff
@@ -244,6 +275,17 @@ public function run()
...
+
+				try
+				{
+					/** @var \phpbb\cache\driver\driver_interface $cache */
+					$cache = $this->container_factory->get('cache.driver');
+					$cache->purge();
+				}
+				catch (cannot_build_container_exception $e)
+				{
+					// Do not do anything, this is just means there is no config.php yet
```
marc1706: this just means there is no

phpBB/phpbb/install/module/install_data/task/add_bots.php
```diff
@@ -221,6 +226,21 @@ public function run()
 			));
 
 			$this->db->sql_query($sql);
+
+			$i++;
+
+			// Run until there are available resources
```
marc1706: These don't really make much sense as you do the exact opposite. Should probably say something like:
```Stop execution if resource limit is reached```

File: phpBB/phpbb/install/module/install_finish/task/populate_migrations.php [hunk 8]
```diff
@@ -46,6 +57,15 @@ public function __construct(\phpbb\install\helper\container_factory $container)
 	 */
 	public function run()
 	{
+		if (!$this->config->get('populate_migration_refresh_before', false))
+		{
+			if ($this->config->get_time_remaining() < 1)
```
marc1706: Why is this set to ```< 1``` while the other ones are `<= 0`? Seems a bit inconsistent

CHItA: because the installer does not have finer controll over that, so basically it just forces a refresh if the remaining time is smaller than 1.

phpBB/config/installer/container/services_install_database.yml
CHItA: i don't think that there will be way more then that number of tasks having to be in between these, but if you insist i can fix it

phpBB/phpbb/install/module/install_finish/task/populate_migrations.php
Elsensee: which is equal to <= 0..?

phpBB/assets/javascript/installer.js
Elsensee: Why do you think that doesn't matter?
Also, pass the lang vars to the installer.html-whatever-template in some javascript code and then access them from the installer.js? That's how I'd approach this.

phpBB/phpbb/install/module/install_finish/task/populate_migrations.php
CHItA: nope

Elsensee: Ah, I see. PHP doesn't cast it into an integer.

phpBB/assets/javascript/installer.js
marc1706: I'd definitely prefer this being properly translated and as Elsensee mentioned this shouldn't be that hard to do.

phpBB/config/installer/container/services_install_database.yml
marc1706: I'd rather have this evenly spaced. Otherwise my OCD might act up ;)

phpBB/assets/javascript/installer.js
CHItA: okay, but no cookies for Tristan [from now on](https://github.com/phpbb/phpbb/pull/3514#discussion-diff-35011176)

phpBB/config/installer/container/services_install_database.yml
CHItA: okay, but no cookies for Tristan [from now on](https://github.com/phpbb/phpbb/pull/3514#discussion-diff-35011176)

phpBB/assets/javascript/installer.js
CHItA: I cannot get this to work

PR #4173, opened 2016-02-07
[ticket/14466] Add an event to cron.php

lavigor:
Should replace [PHPBB3-14012](https://tracker.phpbb.com/browse/PHPBB3-14012).

[PHPBB3-14466](https://tracker.phpbb.com/browse/PHPBB3-14466)

### DIFF
--- a/phpBB/cron.php
+++ b/phpBB/cron.php
@@ -59,7 +59,7 @@ if ($cron_lock->acquire())
 		 * This event enables you to catch the task before it runs
 		 *
 		 * @event core.cron_run_before
-		 * @var	\phpbb\cron\task\task	task	Current Cron task
+		 * @var	\phpbb\cron\task\wrapper	task	Current Cron task
 		 * @since 3.1.8-RC1
 		 */
 		$vars = array(
--- a/phpBB/cron.php
+++ b/phpBB/cron.php
@@ -55,6 +55,18 @@ if ($cron_lock->acquire())
 	$task = $cron->find_task($cron_type);
 	if ($task)
 	{
+		/**
+		 * This event enables you to catch the task before it runs
+		 *
+		 * @event core.cron_run_before
+		 * @var	\phpbb\cron\task\task	task	Current Cron task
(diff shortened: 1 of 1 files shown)


Nicofuma:
You want to use this event to change which task should be run? why?

lavigor:
!set Event

lavigor:
@Nicofuma No, I want to write its name to the database so that if the task isn't finished, the user will know what to check.

Nicofuma:
ok

lavigor:
@Nicofuma [here](https://github.com/phpbb/phpbb/pull/4173/files#diff-ba8bda93858e9f8b9f5cc68ddca1ad5cR62) I wrote `\phpbb\cron\task\task` because of [this line](https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/phpbb/cron/manager.php#L113).
However, I noticed that all tasks in the manager are returned wrapped because of [this line](https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/phpbb/cron/manager.php#L61).

Then I think that `return` statement should be changed to `\phpbb\cron\task\wrapper` in the manager and this PR. Should I fix it here or create a separate PR?

marc1706:
Please create another ticket for other incorrect docblocks.

phpBB/cron.php
```diff
@@ -55,6 +55,18 @@ function output_image()
 	$task = $cron->find_task($cron_type);
 	if ($task)
 	{
+		/**
+		 * This event enables you to catch the task before it runs
+		 *
+		 * @event core.cron_run_before
+		 * @var	\phpbb\cron\task\task	task	Current Cron task
```
marc1706: Should be `\phpbb\cron\task\wrapper` as you pointed out.

lavigor:
!unset WIP

lavigor:
@marc1706 are events possible in CLI?
Or should they be avoided there?

I ask because Cron can be run through CLI too and this event won't be executed in that case.

Nicofuma:
Yes you can

PR #4176, opened 2016-02-13
[ticket/14409] Update session page info before displaying online list

Elsensee:
[PHPBB3-14409](https://tracker.phpbb.com/browse/PHPBB3-14409)

### DIFF
--- a/phpBB/includes/functions.php
+++ b/phpBB/includes/functions.php
@@ -4969,6 +4969,8 @@ function page_header($page_title = '', $display_online_list = false, $item_id = 
 		}
 	}
 
+	$user->update_session_infos();
+
 	// Generate logged in/logged out status
 	if ($user->data['user_id'] != ANONYMOUS)
 	{
@@ -5394,8 +5396,6 @@ function page_footer($run_cron = true, $display_template = true, $exit_handler =
 	{
 		return;
 	}
-
-	$user->update_session_infos();
 
 	phpbb_check_and_display_sql_report($request, $auth, $db);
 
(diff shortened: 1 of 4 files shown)


phpBB/viewonline.php
```diff
@@ -21,6 +21,7 @@
 
 // Start session management
 $user->session_begin();
+$user->update_session_infos();
```
nickvergessen: Move after 
```php
// Whois requested
```

phpBB/includes/functions.php
```diff
@@ -4969,6 +4969,8 @@ function page_header($page_title = '', $display_online_list = false, $item_id =
 		}
 	}
 
+	$user->update_session_infos();
```
marc1706: Didn't we say wrap this in ```if ($display_online_list)```?

nickvergessen: Well then we still need the call in the footer, where we don't know whether we already did this?

phpBB/includes/functions.php
```diff
@@ -5395,8 +5397,6 @@ function page_footer($run_cron = true, $display_template = true, $exit_handler =
 		return;
 	}
 
-	$user->update_session_infos();
```
nickvergessen: This needs to stay, and the header call should be wrapped as mentioned

phpBB/includes/functions.php
marc1706: Ok, I guess we'll have to prevent session refresh (if wanted) using the session time then. I guess that's fine, too.

Elsensee: `$user->update_session_page` will be checked in `update_session_infos();`
Also there is many other variables you could use to prevent that from happening but whatever. If it's in page_header you'll have to do the same anyway - it really is not that much of a difference.

Elsensee:
Updated. ;)

nickvergessen:
Same here, merging via prep-release

PR #4177, opened 2016-02-13
[ticket/14475] Do not log removal of users from newly registered group

marc1706:
PHPBB3-14475


nickvergessen:
Will merge via prep-release for 3.1.8-RC1

PR #4181, opened 2016-02-15
Ticket/14185

v12mike:
Change the ACP whois pane to use a new tyle that is left-aligned with neutral background colour

https://tracker.phpbb.com/browse/PHPBB3-14479


CHItA:
Please create one PR per ticket.

PR #4182, opened 2016-02-15
[ticket/14481] Respect HTTP_X_FORWARDED headers for implying https

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-14481

### DIFF
--- a/phpBB/phpbb/auth/provider/oauth/oauth.php
+++ b/phpBB/phpbb/auth/provider/oauth/oauth.php
@@ -271,7 +271,13 @@ class oauth extends \phpbb\auth\provider\base
 		}
 
 		$uri_factory = new \OAuth\Common\Http\Uri\UriFactory();
-		$current_uri = $uri_factory->createFromSuperGlobalArray($this->request->get_super_global(\phpbb\request\request_interface::SERVER));
+		$super_globals = $this->request->get_super_global(\phpbb\request\request_interface::SERVER);
+		if (!empty($super_globals['HTTP_X_FORWARDED_PROTO']) && $super_globals['HTTP_X_FORWARDED_PROTO'] === 'https')
+		{
+			$super_globals['HTTPS'] = 'on';
+			$super_globals['SERVER_PORT'] = 443;
+		}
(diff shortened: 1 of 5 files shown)


marc1706:
Ready for review

phpBB/phpbb/auth/provider/oauth/oauth.php
```diff
@@ -271,7 +271,13 @@ protected function get_current_uri($service_name, $query)
 		}
 
 		$uri_factory = new \OAuth\Common\Http\Uri\UriFactory();
-		$current_uri = $uri_factory->createFromSuperGlobalArray($this->request->get_super_global(\phpbb\request\request_interface::SERVER));
+		$super_globals = $this->request->get_super_global(\phpbb\request\request_interface::SERVER);
+		if (!empty($super_globals['HTTP_X_FORWARDED_PROTO']) && $super_globals['HTTP_X_FORWARDED_PROTO'] === 'https')
+		{
+			$super_globals['HTTPS'] = 'on';
+			$super_globals['SERVER_PORT'] = 443;
```
Nicofuma: We should use the value of HTTP_X_FORWARDED_PORT here I think

Nicofuma: nevermind, it should be used like this :)

Nicofuma:
I can see an issue because any proxy (trusted or not) could set this header... But we don't have any mechanism of trusted proxy so... Something to introduce when we will merge or request class and Symfony's one I guess

Nicofuma:
Maybe we should add the support of HTTP_X_FORWARDED_PORT at the same time (should have an higher precedence than (HTTP_X_FORWARDED_PROTO)

marc1706:
Right now, a possible side effect that might be introduced by a proxy is "maliciously" upgrading to https requests even though the server does not support that (will result in broken URLs). As demonstrated by the tests, a downgrade attack using the headers is not possible. Supporting the PORT field will nonetheless result in downgrade attacks being possible by specifying a different port than the default.

Nicofuma:
I see. So forget about the port, we will introduce that later along a "trusted proxies list" like in Symfony if we want  to support the feature

PR #4184, opened 2016-02-18
[ticket/14486] Add an event and fix an event in login_box()

lavigor:
[PHPBB3-14486](https://tracker.phpbb.com/browse/PHPBB3-14486)

### DIFF
--- a/phpBB/includes/functions.php
+++ b/phpBB/includes/functions.php
@@ -2821,6 +2821,21 @@ function login_box($redirect = '', $l_explain = '', $l_success = '', $admin = fa
 		$user->setup();
 	}
 
+	/**
+	 * This event allows an extension to modify the login process
+	 *
+	 * @event core.login_box_before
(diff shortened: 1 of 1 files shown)


lavigor:
!set Event

phpBB/includes/functions.php
```diff
@@ -2833,7 +2848,7 @@ function login_box($redirect = '', $l_explain = '', $l_success = '', $admin = fa
 		trigger_error('NO_AUTH_ADMIN');
 	}
 
-	if ($request->is_set_post('login') || ($request->is_set('login') && $request->variable('login', '') == 'external'))
+	if (($request->is_set_post('login') || ($request->is_set('login') && $request->variable('login', '') == 'external')) && !$err)
```
marc1706: Please use empty() and move the check for err being set in front of the other checks.

lavigor: Done.

PR #4186, opened 2016-02-18
[ticket/14488] Delete cookies to get correct session ID during login

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-14488


marc1706:
Restarted tests. One test failure was caused by a UTC offset glitch.

PR #4187, opened 2016-02-19
[ticket/14489] Fix bug where extension custom compiler pass cannot be found

markshawtoronto:
This is returning an extra / on my machine which throws off the next code block, and then the class can't be found.

PHPBB3-14489


marc1706:
Thank you for your contribution

PR #4189, opened 2016-02-20
[ticket/14492] Improve send stats page and include VigLink

marc1706:
Todo:
- [x] Improve console command for install/update incl. info on commands
- [x] Checkout tags of ext in build script
- [x] Make sure to allow install when updating from pre-PR version
- [x] Redirect to help phpBB page after install

Ticket: https://tracker.phpbb.com/browse/PHPBB3-14492


phpBB/phpbb/install/module/install_finish/task/install_extensions.php
```diff
@@ -0,0 +1,173 @@
...
+
+		parent::__construct(true);
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function run()
+	{
+		if (defined('PHPBB_ENVIRONMENT') && PHPBB_ENVIRONMENT === 'test')
```
CHItA: Probably would be better to do this in the test (as it is done with notify_user).

phpBB/phpbb/install/module/install_finish/task/install_extensions.php
```diff
@@ -0,0 +1,173 @@
...
+			}
+		}
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	static public function get_step_count()
+	{
+		return 0;
```
CHItA: `return 1;`

phpBB/phpbb/install/module/update_database/task/update_extensions.php
```diff
@@ -0,0 +1,196 @@
...
+			}
+		}
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	static public function get_step_count()
+	{
+		return 0;
```
CHItA: `return 1;`

phpBB/phpbb/install/module/install_finish/task/install_extensions.php
```diff
@@ -0,0 +1,173 @@
...
+	 * {@inheritdoc}
+	 */
+	public function run()
+	{
+		if (defined('PHPBB_ENVIRONMENT') && PHPBB_ENVIRONMENT === 'test')
+		{
+			return;
+		}
+
+		$this->user->session_begin();
```
CHItA: Not sure, but this might cause conflicts as it is called in the next task as well.

phpBB/config/installer/container/services_install_finish.yml
```diff
@@ -31,4 +41,4 @@ services:
         arguments:
             - '@installer.module.install_finish_collection'
         tags:
-            - { name: installer_install_module, order: 60 }
+            - { name: installer_install_module, order: 80 }
```
CHItA: Did you add any new modules?

phpBB/config/installer/container/services_update_database.yml
```diff
@@ -26,4 +37,4 @@ services:
             - true
             - false
         tags:
-            - { name: installer_update_module, order: 40 }
+            - { name: installer_update_module, order: 60 }
```
CHItA: I cannot see why this was done

Nicofuma:
A few things: 

- When installing we need a way to specify the list of extensions to enable (option configurable on CLI only) with a default value saying: enable anything
- When updating we definitely not want to enable all extensions that was previously disabled. Only Viglink and only if we are updating from <3.2.0 to >=3.2.0

build/build.xml
```diff
@@ -277,6 +277,28 @@
...
+		<if>
+			<and>
+				<equals arg1="${add-viglink-ext}" arg2="1" trim="true" />
+				<equals arg1="${revision}" arg2="HEAD" trim="true" />
+			</and>
+			<then>
+				<exec dir="${dir}/ext" command="mkdir phpbb" passthru="true" />
+
+				<exec dir="${dir}/ext/phpbb" command="git clone https://github.com/phpbb-extensions/viglink.git viglink" passthru="true" checkreturn="true" />
+				<exec dir="${dir}/ext/phpbb/viglink" command="git checkout master" />
```
VSEphpbb: Hmm, maybe we should checkout tagged release versions instead? Since development progress will probably take place on master.

phpBB/config/installer/container/services_install_finish.yml
marc1706: I did add the install extensions module.

phpBB/phpbb/install/module/install_finish/task/install_extensions.php
```diff
@@ -0,0 +1,178 @@
...
+				$this->extension_manager->enable($ext_name);
+				$extensions = $this->get_extensions();
+
+				if (isset($extensions[$ext_name]) && $extensions[$ext_name]['ext_active'])
+				{
+					// Create log
+					$this->log->add('admin', ANONYMOUS, '', 'LOG_EXT_ENABLE', time(), array($ext_name));
+				}
+			}
+			else
```
Elsensee: I'd put this else one level deeper, so it will add an error to the log if it *really* tried to enable the extension but couldn't do it.

phpBB/phpbb/install/module/update_database/task/update_extensions.php
```diff
@@ -0,0 +1,201 @@
...
+
+				// Disable enabled extensions in order to run migrations if needed
+				if ($this->extension_manager->is_enabled($ext_name))
+				{
+					$this->extension_manager->disable($ext_name);
+				}
+
+				if ($this->extension_manager->is_available($ext_name))
+				{
+					$this->extension_manager->enable($ext_name);
```
Elsensee: We might not want to enable extensions on our own that the user has disabled, I guess

phpBB/phpbb/install/module/update_filesystem/task/file_check.php
```diff
@@ -138,6 +138,15 @@ function ($filename) use ($root_path)
 			$progress_count++;
 			$this->iohandler->set_progress('UPDATE_CHECK_FILES', $progress_count);
 
+			// Do not copy viglink again if the previous version was packaged
+			// with it but it does not exist (e.g. deleted by admin)
+			if (strpos($file, $this->phpbb_root_path . 'ext/phpbb/viglink') !== false &&
+				$this->update_helper->phpbb_version_compare($update_info['version']['from'], '3.2.0', '>=') &&
+				!$this->filesystem->exists($this->phpbb_root_path . 'ext/phpbb/viglink'))
```
Elsensee: If there is a new file in viglink, that will just skip that file.
I suggest extending `!$this->filesystem->exists($this->phpbb_root_path . 'ext/phpbb/viglink')` to `!$this->filesystem->exists($this->phpbb_root_path . 'ext/phpbb/viglink/composer.json')` (which we theoretically could apply generally for all extensions with a bit more work, I guess)

phpBB/phpbb/install/module/install_finish/task/install_extensions.php
```diff
@@ -0,0 +1,178 @@
...
+		$input = new ArgvInput();
+		$install_extensions = explode(',', $input->getArgument('install-extensions'));
+
+		// Find available extensions
+		foreach ($this->finder as $file)
+		{
+			/** @var \SplFileInfo $file */
+			$ext_name = preg_replace('#(.+[\\/\\\]ext[\\/\\\])(\w+)[\\/\\\](\w+)#', '$2/$3', dirname($file->getRealPath()));
+
+			if (!empty($install_extensions) && !in_array($ext_name, $install_extensions))
```
marc1706: Might need to use count() instead of empty() here.

phpBB/phpbb/install/module/install_finish/task/install_extensions.php
```diff
@@ -0,0 +1,178 @@
...
+	public function run()
+	{
+		$this->user->session_begin();
+		$this->user->setup(array('common', 'acp/common', 'cli'));
+
+		$input = new ArgvInput();
+		$install_extensions = explode(',', $input->getArgument('install-extensions'));
+
+		// Find available extensions
+		foreach ($this->finder as $file)
```
Nicofuma: couldn't you use the extension manager here?

phpBB/phpbb/install/module/install_finish/task/install_extensions.php
```diff
@@ -0,0 +1,178 @@
...
+	/**
+	 * {@inheritdoc}
+	 */
+	public function run()
+	{
+		$this->user->session_begin();
+		$this->user->setup(array('common', 'acp/common', 'cli'));
+
+		$input = new ArgvInput();
+		$install_extensions = explode(',', $input->getArgument('install-extensions'));
```
Nicofuma: don't do that but directly use the installer config instead

phpBB/phpbb/install/module/install_finish/task/install_extensions.php
Nicofuma: (with $manager->all_available())

phpBB/phpbb/install/module/update_database/task/update_extensions.php
```diff
@@ -0,0 +1,201 @@
...
+		{
+			// Find available extensions
+			foreach ($this->finder as $file)
+			{
+				/** @var \SplFileInfo $file */
+				$ext_name = preg_replace('#(.+[\\/\\\]ext[\\/\\\])(\w+)[\\/\\\](\w+)#', '$2/$3', dirname($file->getRealPath()));
+
+				// Skip extensions that were not added or updated during update
+				if (!count(preg_grep('#ext/' . $ext_name . '#', $update_info['files'])) &&
+					!in_array($ext_name, $update_extensions) && $ext_name !== 'phpbb/viglink')
```
Nicofuma: if we always ant to update viglink, it should be better to ensure it is always in `$update_extensions` this way it is scalable if there is other extensions we always want to update

Nicofuma: by the way, why do we want to always disable/re-enable viglink (even if its files didn't change) ?

phpBB/phpbb/install/module/update_database/task/update_extensions.php
Nicofuma: that's the difficult part yes. And indeed I think we don't want to enable an extensions which wasn't previously enabled. Except if it is in a list of newly extensions enabled by default (imho, the list should be provided by the update package, like we provide the list of changed files)/

build/build.xml
marc1706: That will be the goal

phpBB/phpbb/install/module/update_filesystem/task/file_check.php
marc1706: Why would that skip a file if it's a new file in viglink? It will however also copy / compare files even if composer.json does not exist, which means it might update files if only the folder is left. Checking for composer.json is probably better.

phpBB/phpbb/install/module/update_database/task/update_extensions.php
marc1706: This is more about extensions in general. By disabling and enabling extensions again we will make sure that any migration is ran (and therefore extensions that were possibly replaced by newer versions prior to running the updater are updated).

Nicofuma: ok, in this case no exception :) just disable and re-enable all extensions (and later we could even go a step further by building the container and checking a few services to ensure the extension doesn't break anything crucial. But if we do something like that it will be for later)

phpBB/phpbb/install/module/update_filesystem/task/file_check.php
Elsensee: I don't know anymore why I thought files would be skipped. Let's just say it was because it was 2am

phpBB/phpbb/install/module/update_database/task/update_extensions.php
```diff
@@ -0,0 +1,228 @@
...
+				}
+				else
+				{
+					$this->iohandler->add_log_message('CLI_EXTENSION_ENABLE_FAILURE', array($ext_name));
+				}
+
+				// Disable extensions if it was disabled by the admin before
+				if (!$extension_enabled && !in_array($ext_name, $default_update_extensions))
+				{
+					$this->extension_manager->disable($ext_name);
```
Elsensee: We shouldn't waste time enabling and disabling a previously disabled extension here. If the administrator re-enables an extension it will be updated automatically. No reason to do anything here.

Also, what's that other condition? What's the reason for that?

marc1706:
Ready for review

CHItA:
looks good to me :+1:

build/build.xml
```diff
@@ -277,6 +277,43 @@
 			</else>
 		</if>
 
+		<!-- Checkout latest viglink to ext folder -->
+		<available file="${dir}/ext" type="dir" property="add-viglink-ext" />
```
Nicofuma: Shouldn't you check the presence of the viglink directory? Or I am missing something?

build/build.xml
```diff
@@ -277,6 +277,43 @@
...
+			checkreturn="true"
+			outputProperty='viglink-available' />
+		<if>
+			<and>
+				<equals arg1="${add-viglink-ext}" arg2="1" trim="true" />
+				<or>
+					<equals arg1="${revision}" arg2="HEAD" trim="true" />
+					<equals arg1="${viglink-available}" arg2="1" trim="true" />
+				</or>
+			</and>
```
Nicofuma: If I understand correctly, viglink will be present in the package if the version is >= 3.2.0-b3 or if viglink is present in the directory?

I'm not sure that's what we want.

IIRC, we said: always on install
For the update I don't remember, and I don't really know what to do:
 - I wouldn't reinstall it (even if not enabled) if the user have deleted the files
 - I wouldn't update the files and create conflicts if the user has updated (or reinstalled) the extension manually
 - Could be nice to update the extension if the user didn't touch it...

So... I don't know

phpBB/phpbb/install/module/install_finish/task/install_extensions.php
```diff
@@ -0,0 +1,175 @@
...
+	 */
+	public function run()
+	{
+		$this->user->session_begin();
+		$this->user->setup(array('common', 'acp/common', 'cli'));
+
+		$install_extensions = $this->iohandler->get_input('install-extensions', array());
+
+		// Find available extensions
+		foreach ($this->finder as $file)
```
Nicofuma: why are you using the finder for that? You can't just use `$this->extension_manager->all_available()` ?

phpBB/phpbb/install/module/install_finish/task/install_extensions.php
```diff
@@ -0,0 +1,175 @@
...
+
+			if ($this->extension_manager->is_available($ext_name))
+			{
+				$this->extension_manager->enable($ext_name);
+				$extensions = $this->get_extensions();
+
+				if (isset($extensions[$ext_name]) && $extensions[$ext_name]['ext_active'])
+				{
+					// Create log
+					$this->log->add('admin', ANONYMOUS, '', 'LOG_EXT_ENABLE', time(), array($ext_name));
```
Nicofuma: could be nice to log that to the iohandler too

phpBB/phpbb/install/module/install_finish/task/install_extensions.php
```diff
@@ -0,0 +1,175 @@
...
+
+		$install_extensions = $this->iohandler->get_input('install-extensions', array());
+
+		// Find available extensions
+		foreach ($this->finder as $file)
+		{
+			/** @var \SplFileInfo $file */
+			$ext_name = preg_replace('#(.+[\\/\\\]ext[\\/\\\])(\w+)[\\/\\\](\w+)#', '$2/$3', dirname($file->getRealPath()));
+
+			if (!empty($install_extensions) && !in_array($ext_name, $install_extensions))
```
Nicofuma: I think I would add an extra special case: if $install_extensions === ['all'] (or if $install_extensions contains 'all' ?) enable all available extensions. And it would be the default value when not overridden by the cli config.

build/build.xml
Nicofuma: Or... we could say: "viglink is always part of phpbb.com packages, so the files will always be reinstalled when updating the board (but never re-enabled) and you may have some conflicts if you update the extension yourselves"
could be the easier and best solution I think

phpBB/phpbb/install/module/update_filesystem/task/file_check.php
```diff
@@ -138,6 +138,15 @@ function ($filename) use ($root_path)
 			$progress_count++;
 			$this->iohandler->set_progress('UPDATE_CHECK_FILES', $progress_count);
 
+			// Do not copy viglink again if the previous version was packaged
+			// with it but it does not exist (e.g. deleted by admin)
+			if (strpos($file, $this->phpbb_root_path . 'ext/phpbb/viglink') !== false &&
+				$this->update_helper->phpbb_version_compare($update_info['version']['from'], '3.2.0', '>=') &&
+				!$this->filesystem->exists($this->phpbb_root_path . 'ext/phpbb/viglink/composer.json'))
+			{
+				continue;
```
Nicofuma: I missed this part, so it answers to one of my upper questions (about reinstalling viglink if removed) :)

phpBB/phpbb/install/module/update_database/task/update_extensions.php
```diff
@@ -0,0 +1,239 @@
...
+
+		if (!empty($version_from))
+		{
+			$update_extensions = $this->iohandler->get_input('update-extensions', []);
+
+			// Create list of default extensions that need to be enabled in update
+			$default_update_extensions = [];
+			foreach ($this->default_update as $version => $extensions)
+			{
+				if ($this->update_helper->phpbb_version_compare($version_from, $version, '<'))
```
Nicofuma: should be <= I think. It's easier to understand if we say: starting from 3.2.0-b3 viglink is installed.

phpBB/phpbb/install/module/install_finish/task/install_extensions.php
```diff
@@ -0,0 +1,175 @@
...
+			{
+				continue;
+			}
+
+			if ($this->extension_manager->is_available($ext_name))
+			{
+				$this->extension_manager->enable($ext_name);
```
Nicofuma: I think we might want to catch the exception here. So if we fail at installing an extension it will not fail the whole installation.

build/build.xml
marc1706: No, this check just makes sure we don't try to add an extensions to an old version that does not have an extension directory. VigLink gets added to the folders after this.

Nicofuma: ok

build/build.xml
Nicofuma: actually you already solve the first and third point. About the second... I think we can say it's up to the user if he has updated the extension manually

marc1706: No, the viglink extension will be included in the package when packaging the latest version and it will be added to old versions that should contain viglink (that is 3.2.0-b3 and newer). Based on the latest package and these "old versions", the automatic update packages will be created. That means that for example and update from 3.1.8 will include the viglink extension in the new files inside the installer but not in the old files of the installer. That way the installer will properly know that there are new files.

During update the extension is treated as follow:
- copy extension if we're updating from a version prior to 3.2.0
- do not copy the extension if it was removed and we're updating from a version newer than 3.2.0

Elsensee: Well, how else would you include it? We decide later if we update the files.. ;)

build/build.xml
```diff
@@ -277,6 +277,43 @@
...
+				<exec dir="${dir}/ext" command="mkdir phpbb" passthru="true" />
+
+				<exec dir="${dir}/ext/phpbb" command="git clone https://github.com/phpbb-extensions/viglink.git viglink" passthru="true" checkreturn="true" />
+				<if>
+					<equals arg1="${revision}" arg2="HEAD" trim="true" />
+					<then>
+						<exec dir="${dir}/ext/phpbb/viglink" command="git checkout master" passthru="true" />
+					</then>
+					<else>
+						<exec dir="${dir}/ext/phpbb/viglink" command="git checkout release-phpbb-${version}" passthru="true" />
```
Elsensee: This means we have to create a tag in viglink for every phpbb release right? I don't know, if that's such a good solution...
But tbh I can't think of a better one..

phpBB/install/schemas/schema_data.sql
```diff
@@ -142,6 +142,7 @@ INSERT INTO phpbb_config (config_name, config_value) VALUES ('fulltext_postgres_
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('fulltext_sphinx_indexer_mem_limit', '512');
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('fulltext_sphinx_stopwords', '0');
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('gzip_compress', '0');
+INSERT INTO phpbb_config (config_name, config_value) VALUES ('help_send_statistics', '1');
```
Elsensee: What about help_send_statistics_time?

build/build.xml
Nicofuma: ok, that's good

build/build.xml
Nicofuma: I think it is a good solution yes. Doesn't have to interfer with viglink lifecycle. Just that when we want to release the core we create a tag in viglink (which will be an alias for another tag)

phpBB/phpbb/install/module/update_database/task/update.php
```diff
@@ -187,8 +187,6 @@ public function run()
 
 		$this->iohandler->add_success_message('INLINE_UPDATE_SUCCESSFUL');
 
-		$this->config->delete('version_update_from');
```
Elsensee: Why did you delete that one?

Elsensee:
Out of curiosity, what went with the plan of just enabling/updating all extensions in the extension folder?

phpBB/phpbb/install/module/update_database/task/update_extensions.php
```diff
@@ -0,0 +1,239 @@
...
+						$this->log->add('admin', ANONYMOUS, '', 'LOG_EXT_ENABLE', time(), array($ext_name));
+					} else
+					{
+						$this->iohandler->add_log_message('CLI_EXTENSION_ENABLE_FAILURE', array($ext_name));
+					}
+
+					// Disable extensions if it was disabled by the admin before
+					if (!$extension_enabled && !in_array($ext_name, $default_update_extensions))
+					{
+						$this->extension_manager->disable($ext_name);
```
Elsensee: I am still not sure why you enable and disable extensions again that are not added during the update but disabled by the admin before.

Also, `try { enable/disable }` (as Tristan already pointed out for the update)

Nicofuma:
We can't do that (updating all extensions) because we can't enable an extension which wasn't already enabled

phpBB/phpbb/install/module/update_filesystem/task/file_check.php
```diff
@@ -138,6 +138,15 @@ function ($filename) use ($root_path)
 			$progress_count++;
 			$this->iohandler->set_progress('UPDATE_CHECK_FILES', $progress_count);
 
+			// Do not copy viglink again if the previous version was packaged
+			// with it but it does not exist (e.g. deleted by admin)
+			if (strpos($file, $this->phpbb_root_path . 'ext/phpbb/viglink') !== false &&
+				$this->update_helper->phpbb_version_compare($update_info['version']['from'], '3.2.0', '>=') &&
```
Elsensee: Maybe you can add that array from the updater here too, so we can easily modify this?

phpBB/install/schemas/schema_data.sql
marc1706: Missed that one and should add it.

phpBB/phpbb/install/module/update_database/task/update_extensions.php
marc1706: This will only be the case if the admin "forces" a set of extensions to update during an update using the CLI config. In that case the admin is explicitly stating which extensions should be updated. Therefore this is no case of accidental enabling / disabling.

phpBB/phpbb/install/module/install_finish/task/install_extensions.php
marc1706: done

marc1706:
While installing with the installer and not specifying which extensions to install (which is only possible via the CLI), all available extensions will be installed. This is obviously not the case for the updater as previously outlined and discussed.

phpBB/phpbb/install/module/install_finish/task/install_extensions.php
marc1706: As I'm currently treating that as an array that will require some extra logic. Maybe use `array('all')` instead of the string `all` in the config?

phpBB/install/schemas/schema_data.sql
marc1706: Added.

build/build.xml
marc1706: I'd also like to note that while we do have to create a tag for the release, this doesn't have to be done prior to the release of a version. For example, a 3.2.0-rc1 release after 3.2.0-b3 will require us to have a release-phpbb-3.2.0-b3 tag but not a 3.2.0-rc1 tag as the latest version is always HEAD and therefore a special case.
This does not mean we should not tag stuff but it takes a bit of pressure away from us. However, this should of course still be added to the release to-do list.

phpBB/phpbb/install/module/update_database/task/update_extensions.php
```diff
@@ -0,0 +1,244 @@
...
+						// Disable extensions if it was disabled by the admin before
+						if (!$extension_enabled && !in_array($ext_name, $default_update_extensions))
+						{
+							$this->extension_manager->disable($ext_name);
+						}
+					}
+					catch (\Exception $e)
+					{
+						// Add fail log and continue
+						$this->iohandler->add_log_message('CLI_EXTENSION_ENABLE_FAILURE', array($ext_name));
```
Elsensee: Not sure if we should use another log message here like "Updating extension %s failed" but I leave this up to you, just mentioning it.

phpBB/phpbb/install/module/update_database/task/update_extensions.php
```diff
@@ -0,0 +1,244 @@
...
+			{
+				// Update extensions if:
+				//	1) Extension is currently enabled
+				//	2) Extension was implicitly defined as needing an update
+				//	3) Extension was newly added as default phpBB extension in
+				//		this update and should be enabled by default.
+				if ($this->extension_manager->is_enabled($ext_name) ||
+					in_array($ext_name, $update_extensions) ||
+					in_array($ext_name, $default_update_extensions)
+				)
```
Elsensee: Looks so much cleaner now! :)
Anyway, since some administrators tend to install as many extensions as possible it might be cleverer to handle timeouts. I'm not sure how fast disabling and enabling an extension is without any migrations changes (which will be the case for most of the extensions during an update) but this is probably just an extreme case.

Elsensee:
Otherwise it looks good to me. :+1:

phpBB/phpbb/install/module/update_database/task/update_extensions.php
Nicofuma: could be a good idea to handle timeouts indeed

marc1706: Timeouts are now handled.

phpBB/phpbb/install/module/install_finish/task/install_extensions.php
Nicofuma: sounds good and that's a wrote actually but you couldn't see that because I forgot the backticks :p

Marshalrusty:
Logos:
https://www.dropbox.com/s/rxzeson3sieulw1/VigLink_Primary_Logo_cmyk.eps?dl=0
https://www.dropbox.com/s/qem2vy3ubkuv826/VigLink_Primary_Logo_rgb.png?dl=0
https://www.dropbox.com/s/6adbww3fk22ikgz/VigLink_Stacked_Logo_cmyk.eps?dl=0
https://www.dropbox.com/s/e208iyyim0ujgbg/VigLink_Stacked_Logo_rgb.png?dl=0

VSEphpbb:
Looking good to me 👍

Marshalrusty:
Okay, this is good to go from their end. We just need to implement the sub-tracking stuff to differentiate between different sponsors.

Attaching relevant specs so everyone can look at them at the same time I do.


[Sub-User Implementation Doc.pdf](https://github.com/phpbb/phpbb/files/403204/Sub-User.Implementation.Doc.pdf)
[SubUserReportingAPI.docx](https://github.com/phpbb/phpbb/files/403203/SubUserReportingAPI.docx)
[SubUserReportingAPI.txt](https://github.com/phpbb/phpbb/files/403202/SubUserReportingAPI.txt)

Marshalrusty:
The subID can be a max of 32 alphanumeric characters. My suggestion would be to make it a combination of a random 16 character immutable token that is generated when the extension is installed, and a 16 character hash of the forum's domain name.

There might even be something better than that, but it was my first thought.

Elsensee:
Amazing!
Can't we just use the installation_id? I think it was made for these kind of things.

marc1706:
The install_id is only 16 characters long. We could of course use a hash of the forum's name but that would result in it changing once the forum is renamed for whatever reason. Are we sure we want that? Otherwise we could also just concat two calls to unique_id() resulting in 32 unique characters.

Nicofuma:
`The subID can be a max of 32 alphanumeric characters` 16 isn't enough?

marc1706:
Well if 16 is enough that would also work. @Marshalrusty any reason against just using that?

phpBB/includes/acp/acp_help_phpbb.php
```diff
@@ -0,0 +1,146 @@
...
+
+		$collector = new phpbb_questionnaire_data_collector($install_id);
+
+		// Add data provider
+		$collector->add_data_provider(new phpbb_questionnaire_php_data_provider());
+		$collector->add_data_provider(new phpbb_questionnaire_system_data_provider());
+		$collector->add_data_provider(new phpbb_questionnaire_phpbb_data_provider($config));
+
+		/**
+		 * Event to add and/or modify acp_board configurations
```
Elsensee: Probably misleading description.

build/build.xml
Elsensee: The version has to be fixed here.

phpBB/phpbb/install/module/update_database/task/update_extensions.php
```diff
@@ -0,0 +1,262 @@
...
+						}
+						$this->extension_manager->enable($ext_name);
+						$extensions = $this->get_extensions();
+
+						if (isset($extensions[$ext_name]) && $extensions[$ext_name]['ext_active'])
+						{
+							// Create log
+							$this->log->add('admin', ANONYMOUS, '', 'LOG_EXT_UPDATE', time(), array($ext_name));
+							$this->iohandler->add_success_message(array('CLI_EXTENSION_UPDATE_SUCCESS', $ext_name));
+						} else
```
Elsensee: Code Style. Why did the sniffer didn't catch this?

marc1706:
Updated the version and fixed the mentioned issues.

Nicofuma:
So, now there isn't any blocker anymore? @Marshalrusty

Elsensee:
@Nicofuma The extension isn't ready yet... That's what we still have to fix. Otherwise we are good to go.

phpBB/adm/style/ajax.js
```diff
@@ -4,6 +4,101 @@
 
 'use strict';
 
+
+phpbb.prepareSendStats = function () {
+	var $form = $('#acp_help_phpbb');
+	var $dark = $('#darkenwrapper');
+	var $loadingIndicator;
+
+	$form.on('submit', function () {
```
marc1706: `function (event) {`

Crizz0: With that change the "send statistics" ajax works fine.

build/build.xml
```diff
@@ -287,6 +288,54 @@
...
+					</else>
+				</if>
+				<delete dir="${dir}/ext/phpbb/viglink/.git" />
+				<delete dir="${dir}/ext/phpbb/viglink/tests" />
+				<delete dir="${dir}/ext/phpbb/viglink/travis" />
+				<delete file="${dir}/ext/phpbb/viglink/.gitattributes" />
+				<delete file="${dir}/ext/phpbb/viglink/.travis.yml" />
+				<delete file="${dir}/ext/phpbb/viglink/phpunit.xml.dist" />
+				<delete file="${dir}/ext/phpbb/viglink/README.md" />
```
Elsensee: We should think about adding a .gitattributes to the viglink repo, so we can remove this. ;)

phpBB/docs/events.md
```diff
@@ -184,6 +184,18 @@ acp_groups_position_teampage_add_button_before
 * Since: 3.1.7-RC1
 * Purpose: Add content before adding group to teampage submit button
 
+acp_help_phpbb_stats_after
+===
+* Location: adm/style/acp_help_phpbb.html
+* Since: 3.2.0-b2
```
Elsensee: Version needs to be fixed here.

phpBB/docs/events.md
```diff
@@ -184,6 +184,18 @@ acp_groups_position_teampage_add_button_before
...
+* Purpose: Add content after send statistics tile
+
+acp_help_phpbb_stats_before
+===
+* Location: adm/style/acp_help_phpbb.html
+* Since: 3.2.0-b2
```
Elsensee: And here, too.

phpBB/phpbb/db/migration/data/v320/add_help_phpbb.php
```diff
@@ -0,0 +1,50 @@
...
+*/
+
+namespace phpbb\db\migration\data\v320;
+
+class add_help_phpbb extends \phpbb\db\migration\migration
+{
+	static public function depends_on()
+	{
+		return array(
+			'\phpbb\db\migration\data\v320\v320a2',
```
Elsensee: I guess that's okay?

phpBB/phpbb/install/module/update_database/task/update_extensions.php
```diff
@@ -0,0 +1,263 @@
...
+
+	/** @var \phpbb\db\driver\driver_interface */
+	protected $db;
+
+	/**
+	 * @var array	List of default extensions to update, grouped by version
+	 *				they were added
+	 */
+	static public $default_extensions_update = [
+		'3.2.0-b3' => ['phpbb/viglink']
```
Elsensee: Version also needs to be fixed here.

phpBB/includes/acp/acp_help_phpbb.php
```diff
@@ -0,0 +1,146 @@
...
+		$collector->add_data_provider(new phpbb_questionnaire_php_data_provider());
+		$collector->add_data_provider(new phpbb_questionnaire_system_data_provider());
+		$collector->add_data_provider(new phpbb_questionnaire_phpbb_data_provider($config));
+
+		/**
+		 * Event to modify ACP help phpBB page and/or listen to submit
+		 *
+		 * @event core.acp_help_phpbb_submit_before
+		 * @var	boolean	submit			Do we display the form or process the submission
+		 * @since 3.2.0-b2
```
Elsensee: Version fix.

build/build.xml
VSEphpbb: Still need this, for offline builds.

marc1706:
Rebased on top of latest 3.2.x and fixed an issue with submitting the data.

Nicofuma:
You have a conflict now

PR #4190, opened 2016-02-21
[ticket/14437] Correctly assume index from attachment display order

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-14437

### DIFF
--- a/phpBB/includes/functions_content.php
+++ b/phpBB/includes/functions_content.php
@@ -974,6 +974,9 @@ function parse_attachments($forum_id, &$message, &$attachments, &$update_count, 
 		unset($new_attachment_data);
 	}
 
+	// Make sure attachments are properly ordered
+	ksort($attachments);
+
 	foreach ($attachments as $attachment)
 	{
 		if (!sizeof($attachment))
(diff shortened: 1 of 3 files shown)


phpBB/includes/functions_content.php
```diff
@@ -1248,6 +1236,18 @@ function parse_attachments($forum_id, &$message, &$attachments, &$update_count,
 
 	$unset_tpl = array_unique($unset_tpl);
 
+	// Sort correctly
+	if ($config['display_order'])
+	{
+		// Ascending sort
```
rxu: Descending sort

phpBB/includes/functions_content.php
```diff
@@ -1248,6 +1236,18 @@ function parse_attachments($forum_id, &$message, &$attachments, &$update_count,
...
+		krsort($attachments);
+	}
+	else
+	{
+		// Descending sort
+		ksort($attachments);
```
rxu: Ascending sort

phpBB/includes/functions_content.php
marc1706: This is the same as under the acp settings for the display order. Why do you think it should be the other way round?

rxu: Dunno, but isn't it what ```krsort``` does?

rxu: ok, probably despite it's descending technically, it's acsending in respect of the attachments order.

marc1706: Yeah, the idea behind this might probably be somewhat confusing. ;)

PR #4192, opened 2016-02-25
[ticket/14496] Cache last update of update list instead of using filemtime

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-14496


Elsensee:
This creates an extra file just for the time. I would rather not do that and just call it `update_list_time` so it gets into the global cache. Or is the global one non existent in the installer?

marc1706:
Yes, should definitely go into the global cache file.

marc1706:
@Elsensee turns out that the global cache does not work as expected in the installer. Went back to using the underscore prefix.

Elsensee:
Oh. Then maybe we can write it to installer_config instead?

marc1706:
As this is for 3.1.x, there is not installer config.

PR #4194, opened 2016-02-25
[ticket/14499] Add command to update the board from CLI

Nicofuma:
https://tracker.phpbb.com/browse/PHPBB3-14499

PHPBB3-14499

### DIFF
--- a/phpBB/docs/update-config.sample.yml
+++ b/phpBB/docs/update-config.sample.yml
@@ -1,2 +1,2 @@
-udater:
+updater:
     type: all
--- a/phpBB/docs/update-config.sample.yml
+++ b/phpBB/docs/update-config.sample.yml
@@ -1,2 +1,2 @@
 udater:
-    type: db_only 
+    type: all
--- a/phpBB/docs/update-config.sample.yml
+++ b/phpBB/docs/update-config.sample.yml
@@ -1,0 +1,2 @@
+udater:
+    type: db_only 
--- a/phpBB/phpbb/install/console/command/update/update.php
+++ b/phpBB/phpbb/install/console/command/update/update.php
@@ -87,7 +87,7 @@ class update extends \phpbb\console\command\command
 	/**
 	 * Executes the command update.
 	 *
-	 * Install the board
+	 * Update the board
 	 *
 	 * @param InputInterface  $input  An InputInterface instance
 	 * @param OutputInterface $output An OutputInterface instance
--- a/phpBB/phpbb/install/console/command/update/update.php
+++ b/phpBB/phpbb/install/console/command/update/update.php
@@ -11,14 +11,14 @@
 *
 */
 
-namespace phpbb\install\console\command\install;
+namespace phpbb\install\console\command\update;
 
 use phpbb\install\exception\installer_exception;
 use phpbb\install\helper\install_helper;
 use phpbb\install\helper\iohandler\cli_iohandler;
 use phpbb\install\helper\iohandler\factory;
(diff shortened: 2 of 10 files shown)


phpBB/phpbb/install/console/command/update/update.php
```diff
@@ -0,0 +1,176 @@
...
+				InputArgument::REQUIRED,
+				$this->language->lang('CLI_CONFIG_FILE'))
+			->setDescription($this->language->lang('CLI_UPDATE_BOARD'))
+		;
+	}
+
+	/**
+	 * Executes the command update.
+	 *
+	 * Install the board
```
marc1706: Update ;)

phpBB/phpbb/install/console/command/update/update.php
```diff
@@ -0,0 +1,176 @@
...
+	 *
+	 * @param cli_iohandler $iohandler
+	 * @param array $config
+	 */
+	private function register_configuration(cli_iohandler $iohandler, $config)
+	{
+		$iohandler->set_input('update_type', $config['type']);
+		$iohandler->set_input('submit_update', 'submit');
+
+		$iohandler->set_input('method', '.tar');
```
marc1706: Why do you default to .tar. Shouldn't we use .zip for better cross-platform compatibility?

marc1706:
Haven't tested the update yet but the rest looks good.

phpBB/phpbb/install/console/command/update/update.php
Nicofuma: it's just to have something valid (tar is always available). But it will not be used: we are in CLI, you can't have a link to DL the archive. Reminds me I still need to check what will happen in this case

marc1706:
Checking the updater config with ```update:config:validate``` I get:
```
 [OK] CONFIGURATION_VALID
```
The config show does work fine. Instead of the default sample file, I decided to try to run the full update via cli (type: all). This only shows me this though:
```
php install/phpbbcli.php update docs/update-config.sample.yml 
       0/0   ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░    0%
             
     87/87  ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓  100%
             Check files to update                                            
                                                                              
                                                                                
 [ERROR] Config file is missing data or might contain invalid settings.         
                                                                                

     87/87  ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓  100%
             Check files to update                                            

```

marc1706:
Yes, works fine now.

phpBB/docs/update-config.sample.yml
```diff
@@ -0,0 +1,2 @@
+udater:
```
marc1706: udater? Did you mean update or updater?

phpBB/phpbb/install/console/command/update/update.php
marc1706: Bump, please change this, too.

PR #4196, opened 2016-02-27
[ticket/14487] Try to handle connection timeouts

CHItA:
[PHPBB3-14487](https://tracker.phpbb.com/browse/PHPBB3-14487)

### DIFF
--- a/phpBB/assets/javascript/installer.js
+++ b/phpBB/assets/javascript/installer.js
@@ -13,6 +13,9 @@
 	var currentProgress = 0;
 	var refreshRequested = false;
 	var transmissionOver = false;
+	var status = '';
+	var statusCount = 0;
+	var statusTimeout = null;
 
 	// Template related variables
 	var $contentWrapper = $('.install-body').find('.main');
@@ -372,12 +375,45 @@
 			}
 
 			if (timeoutDetected) {
-				addMessage('error',
-					[{
-						title: installLang.title,
-						description: installLang.msg
-					}]
-				);
+				status = queryInstallerStatus();
+				statusCount = 0;
+
+				if (status === 'continue') {
+					refreshRequested = false;
+					doRefresh();
+				} else if (status === 'running')	{
+					statusTimeout = setTimeout(function() {
+						var s = queryInstallerStatus();
+						if (statusCount === 12) { // 1 minute hard cap
+							s = 'fail';
+						}
+
+						if (s === 'continue') {
+							refreshRequested = false;
+							doRefresh();
+							clearTimeout(statusTimeout);
+						} else if (s === 'fail') {
+							addMessage('error',
+								[{
+									title: installLang.title,
+									description: installLang.msg
+								}]
+							);
+							clearTimeout(statusTimeout);
+						}
+
+						statusCount++;
+						},
+						5000
+					);
+				} else {
+					addMessage('error',
+						[{
+							title: installLang.title,
+							description: installLang.msg
+						}]
+					);
+				}
 			}
 		}
 	}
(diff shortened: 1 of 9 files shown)


CHItA:
This PR has multiple problems beyond that it is not tested at all.

First, it is relying on `flock` which relies on the OS to clean up locks, so as far as i can tell there are no insurence that it would work on all platforms (the Linux kernel and Windows seems to handle it okay, but those are just my tests, so...).

I haven't tested the JS, also not sure how to do that.

Elsensee:
Is there a reason you don't use `\phpbb\lock\flock`?

CHItA:
Because it would have made things more complicated.

marc1706:
While using phpBB's lock implementation might make things more complicated it might be the better option if relying on OS support is unreliable.

CHItA:
It doesn't change the fact that we relying on the OS. As of `flock`'s documentation it seems, that the locks have to be removed manually, which probably means that PHP doesn't lift the locks automatically when the PHP process ends, therefore the OS has to do it (if it is implemented). 

But `\phpbb\lock\flock` has the same problem, so that's that.

Nicofuma:
There is also an flock class in symfony/process did you look at it?

CHItA:
nope, but we have a problem with `flock`'s implementation (or not, i haven't looked at its implementation, but this is a PHP limitation if anything).

phpBB/assets/javascript/installer.js
```diff
@@ -13,6 +13,9 @@
 	var currentProgress = 0;
```
CHItA: @callumacrae if you could quickly estimate the damages, it would be much appreciated.

phpBB/assets/javascript/installer.js
```diff
@@ -564,4 +600,18 @@
...
 		});
 	}
+
+	/**
+	 * Queries the installer's status
+	 */
+	function queryInstallerStatus()
+	{
+		var data = null;
+		$.ajax({url: "../installer/status", dataType: "json"})
```
callumacrae: single quotes. Also, [`$.getJSON()`](http://api.jquery.com/jquery.getjson/).

phpBB/assets/javascript/installer.js
```diff
@@ -564,4 +600,18 @@
...
+		.done(function(d) {
+			data = d;
+		});
+
+		return data.status;
```
callumacrae: This won't work. `data` will be `null` because the request won't yet have returned. Use a callback instead.

Have you tested this? This will throw "Cannot call property status of null".

CHItA: Thanks for the feedback, now all should be fixed.

phpBB/assets/javascript/installer.js
```diff
@@ -339,6 +340,57 @@
...
+		} else if (status === 'running') {
+			statusCount++;
+			$('#loading_indicator').css('display', 'block');
+			setTimeout(queryInstallerStatus, 5000);
+		} else {
+			$('#loading_indicator').css('display', 'none');
+			addMessage('error',
+				[{
+					title: installLang.title,
+					description: installLang.msg
```
callumacrae: I feel that with a consistent API, this could be simplified to `addMessage('error', [ installLang ])` :wink:

callumacrae:
JS LGTM :+1:

phpBB/assets/javascript/installer.js
CHItA: Yeah, but then it would be harder to extend it (the `installLang` object that is), so that's why it was done this way.

callumacrae: Why, does `addMessage()` not like it if you pass in extra options? I would have thought that it would just ignore them, in which case it would be fine.

PR #4198, opened 2016-02-28
[ticket/14459] Check language inpunt for group

phpbb-es:
https://tracker.phpbb.com/browse/PHPBB3-14459

### DIFF
--- a/phpBB/includes/acp/acp_attachments.php
+++ b/phpBB/includes/acp/acp_attachments.php
@@ -1228,7 +1228,7 @@ class acp_attachments
 						'FILETIME'			=> $user->format_date((int) $row['filetime']),
 						'REAL_FILENAME'		=> (!$row['in_message']) ? utf8_basename((string) $row['real_filename']) : '',
 						'PHYSICAL_FILENAME'	=> utf8_basename((string) $row['physical_filename']),
-						'EXT_GROUP_NAME'	=> (!empty($extensions[$row['extension']]['group_name'])) ? $user->lang['EXT_GROUP_' . $extensions[$row['extension']]['group_name']] : '',
+						'EXT_GROUP_NAME'   	=> (!empty($user->lang[$extensions[$row['extension']]['group_name']])) ? $user->lang['EXT_GROUP_' . $extensions[$row['extension']]['group_name']] : $extensions[$row['extension']]['group_name'],
 						'COMMENT'			=> $comment,
 						'TOPIC_TITLE'		=> (!$row['in_message']) ? (string) $row['topic_title'] : '',
 						'ATTACH_ID'			=> (int) $row['attach_id'],


phpBB/includes/acp/acp_attachments.php
```diff
@@ -1228,7 +1228,7 @@ function main($id, $mode)
 						'FILETIME'			=> $user->format_date((int) $row['filetime']),
 						'REAL_FILENAME'		=> (!$row['in_message']) ? utf8_basename((string) $row['real_filename']) : '',
 						'PHYSICAL_FILENAME'	=> utf8_basename((string) $row['physical_filename']),
-						'EXT_GROUP_NAME'	=> (!empty($extensions[$row['extension']]['group_name'])) ? $user->lang['EXT_GROUP_' . $extensions[$row['extension']]['group_name']] : '',
+						'EXT_GROUP_NAME'   	=> (!empty($user->lang[$extensions[$row['extension']]['group_name']])) ? $user->lang['EXT_GROUP_' . $extensions[$row['extension']]['group_name']] : $extensions[$row['extension']]['group_name'],
```
Nicofuma: please use the lang function here (and not the array access)

Elsensee: I think we could modify the group_helper a bit and then patch this in 3.2.x?

phpbb-es: @Nicofuma but it seems that the problem comes from the name of the group, I'm not sure for this.

I only reported it, the team member "tower" can not access the tracker, GitHub, phpBB official site, etc., he lives in Cuba and access to Internet is complicated.

@Elsensee , "tower" commented that he look the same problem in 3.1 and 3.2

CHItA: I think what @Nicofuma means is `$user->lang($extensions[$row['extension']]['group_name'])` will produce the same result, as what your modification does.

Elsensee: It doesn't which is why I was referring to the group_helper which was added in 3.2.x

I know this problem might occur in 3.1 too, I'm just telling you that there should be a rather easy fix, if you'll try to fix it in 3.2 only. :)

rubencm:
```suggestion
						'EXT_GROUP_NAME'	=> $this->language->is_set('EXT_GROUP_' . $extensions[$row['extension']]['group_name']) ?  $this->language->lang('EXT_GROUP_' . $extensions[$row['extension']]['group_name']) : $extensions[$row['extension']]['group_name'],
```

They are asking you to do something like this. (If you use this code you will have to add the language property in the top of the class and define it in the top of the main() method, like this:
`$this->language = $phpbb_container->get('language');`

rubencm: @phpbb-es

phpbb-es: @rubencm sorry, now I see this message ... I see that another ticket has been created and PR for you, this was commented by a Moderator of my forum some time ago.

rubencm: Yes, i forked your repository and fixed it, so your commit is still there and your contribution will be included in the repository, but if you still want to fix it here, i will close my pull request

phpbb-es:
Thanks @rubencm

PR #4199, opened 2016-02-29
[ticket/12684] Add console command user:add

VSEphpbb:
This replaces https://github.com/phpbb/phpbb/pull/2585

https://tracker.phpbb.com/browse/PHPBB3-12684
PHPBB3-12684

### DIFF
--- a/phpBB/language/en/cli.php
+++ b/phpBB/language/en/cli.php
@@ -82,6 +82,11 @@ $lang = array_merge($lang, array(
 	'CLI_DESCRIPTION_THUMBNAIL_GENERATE'	=> 'Generate all missing thumbnails.',
 	'CLI_DESCRIPTION_THUMBNAIL_RECREATE'	=> 'Recreate all thumbnails.',
 
+	'CLI_DESCRIPTION_USER_ADD'					=> 'Add a new user',
+	'CLI_DESCRIPTION_USER_ADD_OPTION_USERNAME'	=> 'Username of the new user',
+	'CLI_DESCRIPTION_USER_ADD_OPTION_PASSWORD'	=> 'Password of the new user',
+	'CLI_DESCRIPTION_USER_ADD_OPTION_EMAIL'		=> 'E-mail address of the new user',
+
 	'CLI_EXTENSION_DISABLE_FAILURE'		=> 'Could not disable extension %s',
 	'CLI_EXTENSION_DISABLE_SUCCESS'		=> 'Successfully disabled extension %s',
 	'CLI_EXTENSION_ENABLE_FAILURE'		=> 'Could not enable extension %s',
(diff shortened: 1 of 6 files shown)


phpBB/phpbb/console/command/user/add.php
```diff
@@ -0,0 +1,311 @@
...
+			'group_id'      => $group_id,
+			'user_timezone' => $this->config['board_timezone'],
+			'user_lang'     => $this->config['default_lang'],
+			'user_type'     => USER_NORMAL,
+			'user_regdate'  => time(),
+		);
+
+		$user_id = (int) user_add($user_row);
+
+		if ($input->getOption('send-email') && $this->config['email_enable'])
```
marc1706: Shouldn't you also check if user_add() actually returns a user id? Just to be safe that there was no error while adding the user to the database.

phpBB/language/en/cli.php
```diff
@@ -82,6 +82,12 @@
 	'CLI_DESCRIPTION_THUMBNAIL_GENERATE'	=> 'Generate all missing thumbnails.',
 	'CLI_DESCRIPTION_THUMBNAIL_RECREATE'	=> 'Recreate all thumbnails.',
 
+	'CLI_DESCRIPTION_USER_ADD'					=> 'Add a new user.',
+	'CLI_DESCRIPTION_USER_ADD_OPTION_USERNAME'	=> 'Username of the new user',
+	'CLI_DESCRIPTION_USER_ADD_OPTION_PASSWORD'	=> 'Password of the new user',
+	'CLI_DESCRIPTION_USER_ADD_OPTION_EMAIL'		=> 'E-mail address of the new user',
+	'CLI_DESCRIPTION_USER_ADD_OPTION_NOTIFY'	=> 'Send account activation email to the new user',
```
marc1706: Should probably add that it does not send the mail by default.

VSEphpbb: Hmm, well it is under options, so isn't it kinda given that you need to specify this option to send them?

VSEphpbb: @marc1706 see new extended help message below for this.

marc1706: The CLI does show each option with a help message right next to it. I think people will look directly at those infos before reading a longer text. Therefore I'd recommend stating this here, too. This is also in line with the man pages for most executables on linux.

phpBB/phpbb/console/command/user/add.php
```diff
@@ -0,0 +1,318 @@
...
+			$question->setMaxAttempts(5);
+
+			$data['new_password'] = $helper->ask($input, $output, $question);
+		}
+
+		if (!$data['email'])
+		{
+			$question = new Question($this->ask_user('EMAIL_ADDRESS'));
+			$data['email'] = $helper->ask($input, $output, $question);
+		}
```
Nicofuma: you should move this whole block (from line 130) to the method 

```
	protected function interact(InputInterface $input, OutputInterface $output)
	{
	}
```

And you should also mark the email as required I think

VSEphpbb: Currently the email is just an option so you add it to the command if you want to send the email. It sounds like what you're asking for would only make sense if the email option is interactive, which currently is not  but it Can be if you prefer.

Nicofuma: nevermind for the email part. I was talking about email which is already required

VSEphpbb:
Updated

VSEphpbb:
I have a bunch of other user commands to add in a separate PR after this one is merged.

PR #4200, opened 2016-03-02
[ticket/14508] Change language notice on account activation

RMcGirr83:
https://tracker.phpbb.com/browse/PHPBB3-14508

PHPBB3-14508


marc1706:
I'd change it to:
```
Your account has been created. However, this board requires account activation. An activation key has been sent to the email address you provided. Please check your email for further information and also be sure to check your junk mail box. It may take a while to get the email depending on your email provider.
```
Mainly cause there are other factors than server load and there is really no point in overstating that people need to be patient.

PR #4202, opened 2016-03-04
[ticket/13972] Change hard coded time check to use flood interval

RMcGirr83:
https://tracker.phpbb.com/browse/PHPBB3-13972

PHPBB3-13972

### DIFF
--- a/phpBB/posting.php
+++ b/phpBB/posting.php
@@ -104,7 +104,7 @@ $vars = array(
 extract($phpbb_dispatcher->trigger_event('core.modify_posting_parameters', compact($vars)));
 
 // Was cancel pressed? If so then redirect to the appropriate page
-if ($cancel || ($current_time - $lastclick < 2 && $submit))
+if ($cancel || ($current_time - $lastclick < intval($config['flood_interval']) && $submit))
 {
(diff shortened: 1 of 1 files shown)


RMcGirr83:
Maybe the check should be removed completely with the exception of someone pressing the cancel button?  Flood check is handled later on in the file.

ElooKoN:
See also: https://www.phpbb.com/community/viewtopic.php?p=14370186#p14370186

RMcGirr83:
Bump @marc1706 @Nicofuma @CHItA

marc1706:
This check doesn't actually make much sense. The point of it is some kind of flood control. As a result of that, one should see an error message. It does however redirect to viewtopic without no apparent reason. Removing the second part (``` || ($current_time - $lastclick < intval($config['flood_interval']) && $submit)```) will properly trigger the flood control so I'd rather go with checking that.
As this is a change of behavior I'd vote for merging this into 3.2 only though.
@Nicofuma ^

Nicofuma:
Fine with 3.2

RMcGirr83:
https://github.com/phpbb/phpbb/pull/4210

PR #4204, opened 2016-03-06
[ticket/14483] Do not send headers by default on access via controller

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-14483

### DIFF
--- a/phpBB/includes/functions.php
+++ b/phpBB/includes/functions.php
@@ -4909,7 +4909,7 @@ function phpbb_get_avatar($row, $alt, $ignore_config = false, $lazy = false)
 /**
 * Generate page header
 */
-function page_header($page_title = '', $display_online_list = false, $item_id = 0, $item = 'forum')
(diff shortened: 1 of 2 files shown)


phpBB/includes/functions.php
```diff
@@ -5249,17 +5249,22 @@ function page_header($page_title = '', $display_online_list = false, $item_id =
...
-	$http_headers = array(
-		// application/xhtml+xml not used because of IE
-		'Content-type' => 'text/html; charset=UTF-8',
-		'Cache-Control' => 'private, no-cache="set-cookie"',
-		'Expires' => gmdate('D, d M Y H:i:s', time()) . ' GMT',
-	);
-	if (!empty($user->data['is_bot']))
+	$http_headers = array();
+
+	if ($send_headers)
```
Elsensee: Don't see why this was done.

marc1706: Excerpt from IRC:
I removed the second check for send_headers before the foreach.
that way extension authors will both know if send_headers() is set to false by looking at the empty http_headers and they'll still be able to add headers

PR #4214, opened 2016-03-11
[ticket/14241] Prevent empty q&a as a result of improper settings

marc1706:
Improper setup of the q&a captcha in combination with the admin choosing a
default language that does not have any questions and answers set might
result in the user being presented empty questions and answers. This
change will try to fall back to any question in case the admin incorrectly
set the default language and has no questions & answers set. If that does
not work, the captcha will not allow passing it and suggest to contact the
board admin to resolve this issue.

Ticket: https://tracker.phpbb.com/browse/PHPBB3-14241

### DIFF
--- a/phpBB/phpbb/captcha/plugins/qa.php
+++ b/phpBB/phpbb/captcha/plugins/qa.php
@@ -220,10 +220,12 @@ class qa
 	*/
 	function get_template()
 	{
-		global $template;
+		global $phpbb_log, $template, $user;
 
-		if ($this->is_solved() || !count($this->question_ids))
+		if ($this->is_solved() || empty($this->question_text) || !count($this->question_ids))
 		{
+			/** @var \phpbb\log\log_interface $phpbb_log */
+			$phpbb_log->add('critical', $user->data['user_id'], $user->ip, 'LOG_ERROR_CAPTCHA', time(), array($user->lang('CONFIRM_QUESTION_MISSING')));
 			return false;
 		}
 		else
@@ -386,13 +388,15 @@ class qa
 	*/
 	function validate()
 	{
-		global $user;
+		global $phpbb_log, $user;
 
 		$error = '';
 
 		if (!sizeof($this->question_ids))
 		{
-			return $user->lang['CONFIRM_QUESTION_MISSING'];
+			/** @var \phpbb\log\log_interface $phpbb_log */
+			$phpbb_log->add('critical', $user->data['user_id'], $user->ip, 'LOG_ERROR_CAPTCHA', time(), array($user->lang('CONFIRM_QUESTION_MISSING')));
+			return $user->lang('CONFIRM_QUESTION_MISSING');
 		}
 
 		if (!$this->confirm_id)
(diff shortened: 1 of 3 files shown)


phpBB/phpbb/captcha/plugins/qa.php
```diff
@@ -100,6 +100,28 @@ function init($type)
 			$db->sql_freeresult($result);
 		}
 
+		// final fallback to any language
+		if (!sizeof($this->question_ids))
```
Nicofuma: ``count() === 0` should be proper but...

phpBB/phpbb/captcha/plugins/qa.php
```diff
@@ -100,6 +100,28 @@ function init($type)
...
+
+			while ($row = $db->sql_fetchrow($result))
+			{
+				if (empty($this->question_lang))
+				{
+					$this->question_lang = $row['lang_iso'];
+				}
+				$this->question_ids[$row['question_id']] = $row['question_id'];
+			}
+			$db->sql_freeresult($result);
```
Nicofuma: there should be something wrong here... you are selecting all questions and you set the language to the language of the first question? sounds a bit odd to me.

phpBB/phpbb/captcha/plugins/qa.php
marc1706: Fits nicely with the one above

phpBB/phpbb/captcha/plugins/qa.php
marc1706: We only do this if:
- the user selected language does not have questions available
- the default language does not have questions available

In order to be able to properly select questions and fitting answers, we need to have a valid language selected (no array, just a lang iso). We could of course also select a language with picking a random question using array_lang() and then select its lang iso.

phpBB/phpbb/captcha/plugins/qa.php
```diff
@@ -100,6 +100,28 @@ function init($type)
...
 
+		// final fallback to any language
+		if (!sizeof($this->question_ids))
+		{
+			$this->question_lang = '';
+
+			$sql = 'SELECT q.question_id, q.lang_iso
+				FROM ' . $this->table_captcha_questions . ' q, ' . $this->table_captcha_answers . ' a
+				WHERE q.question_id = a.question_id
+				GROUP BY lang_iso';
```
gooof: GROUP BY q.lang_iso

Derky: @gooof did you got an error or are you just suggesting this because you think it looks nicer?
(The answers table [doesn't have a lang_iso](https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/phpbb/captcha/plugins/qa.php#L351-L359) so it shouldn't be necessary)

gooof: Because it is now incorrect. 
mysql searching through both tables for this column whether exists or not.

gooof: not really important below 1000 rows, but this should not matter.

Nicofuma: anyway the PR has already been merged. You should create a ticket and send a PR to fix the issue

PR #4217, opened 2016-03-12
[ticket/14533] U_NOTIFICATION_SETTINGS doesn't return the correct URL

Skouat:
PHPBB3-14533


Skouat:
@marc1706 do I need to create a PR for 3.2 ?

marc1706:
No, this was also merged into 3.2.x and master.

Skouat:
ok

PR #3747, opened 2015-07-07
[ticket/13377] Fix decoding for www type URLs in function decode_message()

rxu:
<a href="https://tracker.phpbb.com/browse/PHPBB3-13377">PHPBB3-13377</a>.

PR #4218, opened 2016-03-13
[ticket/14527] Retrieve link URL from href and not link text

marc1706:
Ticket: https://tracker.phpbb.com/browse/PHPBB3-14527


rxu:
This reverts the fix introduces here https://github.com/phpbb/phpbb/pull/3747 for the bug https://tracker.phpbb.com/browse/PHPBB3-13377
Why?

rxu:
Ok, that fix can work wrong for long URLs of ```MAGIC_URL_FULL``` type as the link text got shortened in this case.
I'd suggest just splitting regexes from ```[mw]``` to separate ```m``` and ```w``` to fix the issue correctly.

phpBB/includes/functions.php
```diff
@@ -2745,7 +2745,8 @@ function get_preg_expression($mode)
 			return array(
 				'#<!\-\- e \-\-><a href="mailto:(.*?)">.*?</a><!\-\- e \-\->#',
 				'#<!\-\- l \-\-><a (?:class="[\w-]+" )?href="(.*?)(?:(&amp;|\?)sid=[0-9a-f]{32})?">.*?</a><!\-\- l \-\->#',
-				'#<!\-\- ([mw]) \-\-><a (?:class="[\w-]+" )?href="(.*?)">(.*?)</a><!\-\- \1 \-\->#',
+				'#<!\-\- ([mw]) \-\-><a (?:class="[\w-]+" )?href="http://(.*?)">\2</a><!\-\- \1 \-\->#',
```
JoeUser78: What is the point of hardcoding the http:// here? What about https|ftp|ftps|whatever://?

marc1706: This is a new regex that is being introduced to take care of links that are automatically prefixed with http://. It is unrelated to other prefixes like https, ftp, etc. That's what the regex below this is for.

JoeUser78: I realy don't get the point why to hardcode a http:// prefix at all? Especially as the world is switching from http:// to https:// for good and nowadays http(S):// is not the only prefixes out there.
Why not getting rid of any hardcoded prefix?
And why prefixing anything with http:// when you don't know if it's intended to be http:// and not https:// or something else or even realy nothing?
This leads to unexpected behavior and false positives.

When i post www.phpbb.com which prefix did i mean? http or https or ftp or git or ssh or imap or foo?

marc1706: I get the feeling that you don't exactly know what this part actually does. The regex is used in the decode_message() function. This means it's only used to decode messages that were saved in the database prior to 3.2. In addition to that, it only matches URLs that are automatically parsed as URLs (magic URLs) and that have automatically been prefixed with the `http://` prefix by phpBB (which can be assumed by the `http://` prefix missing in the URL title.

Why we're not getting rid of prefixes? The scheme is obligatory in the href of hyperlinks. They should not be omitted (although some browsers do not care about the scheme missing).
Prefixing with `http://` is something that should always work if the scheme is not known. Servers should properly take care of redirecting to e.g. `https://` if that is the preferred scheme.

If you post bar.phpbb.com you didn't specify a prefix. One can assume that it's supposed to be http as that is the most widely used scheme. If you really mean any non standard scheme then you should specify that.

JoeUser78: OK, hopefully this fixes the bug as expected.

Will you provide some instructions for users already bitten by this bug?

marc1706: You'll have to reset post text or signatures that were affected by this to their state before they were reparsed (as in pre 3.2.x state).

This bug is a perfect example of why we discourage the use of beta software in production environments as it might result in some data loss or corruption.

JoeUser78: Which Version of phpBB should be used in production with PHP7?
The only PHP7-compatible Version of phpBB is 3.2.x...

marc1706: Until a stable version of phpBB 3.2 is released, one should stay with phpBB 3.1 and one of the maintained versions of PHP 5 (e.g. PHP 5.6).

JoeUser78:
Did not fix all problems: Links like ftps://www.phpbb.com/ are still broken.

marc1706:
Define what you mean by broken? Just calling things "broken" does not help with figuring things out or solving issues.

PR #4219, opened 2016-03-13
[ticket/12479] Remove deprecated methods

CHItA:
[PHPBB3-12479](https://tracker.phpbb.com/projects/PHPBB3/issues/PHPBB3-12479)


Nicofuma:
Actually I don't know if it is a good idea to remove all of these functions so early. I don't think I would do that before 4.0 or something like that (it doesn't change anything for us if we keep them).

By the way, it could be good to be able to know which of them are still used by extensions (and the number of extensions using them)

VSEphpbb:
I'm afraid...very afraid.

PR #4215, opened 2016-03-11
 Structured data - breadcrumbs error 

C1n:
PHPBB3-14528

PR #4220, opened 2016-03-13
[ticket/14533] Update navbar_header.html for valid microdata

marc1706:
Replaces #4215 

Ticket: https://tracker.phpbb.com/browse/PHPBB3-14528


gooof:
same on line 183

marc1706:
Hm, that also seems to have been removed by accident. Thanks for the heads up.

Skouat:
@marc1706 
The ticket number is not correct in the title of the PR and on the commit message.

You must used `14528` and not `14533`

PR #4222, opened 2016-03-14
[ticket/14536] Force unix time stamps to be integer

marc1706:
This will ensure to prevent PHP fatal errors in case the passed timestamp
is an empty string or does not evaluate to an integer (i.e. strings like
foobar).

Ticket: https://tracker.phpbb.com/browse/PHPBB3-14536


marc1706:
Build error was solved by restarting tests :+1:

marc1706:
Also see https://tracker.phpbb.com/browse/PHPBB3-13977 as a possible side effect of this PR.

PR #4227, opened 2016-03-18
[ticket/14318] Fix SQL error in Notifications Board migration

Senky:
[PHPBB3-14318](https://tracker.phpbb.com/browse/PHPBB3-14318)

### DIFF
--- a/phpBB/phpbb/db/migration/data/v320/notifications_board.php
+++ b/phpBB/phpbb/db/migration/data/v320/notifications_board.php
@@ -34,7 +34,7 @@ class notifications_board extends \phpbb\db\migration\migration
 	public function update_module()
 	{
 		$sql = 'UPDATE ' . MODULES_TABLE . "
-			SET auth = 'cfg_allow_board_notifications'
+			SET module_auth = 'cfg_allow_board_notifications'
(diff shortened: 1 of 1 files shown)


RMcGirr83:
You should prepend your commit with the ticket number per the guidelines.

Senky:
Well, I believe that `[ticket/14318]` prefix is just what you mean.

RMcGirr83:
Correct I guess but as far as I'm aware your initial commit message
`Fix SQL error in Notifications Board migration`
should be like
`[ticket/14318] Fix SQL error in Notifications Board migration`

Mauron:
That is not a commit message, that's the PR title: thatìs why TravisCI check is successful.

RMcGirr83:
If one looks at all the other commit messages the PR title, or whatever you want to call it, all reference the original ticket.  Pretty sure that is a requirement per the guidelines no matter what travis does or does not do.

Senky:
Oh, you meant PR title? Commit message already includes ticket ID. Now, PR title too.

Mauron:
As far as I recall, PR title has not to have the ticket ID.
I read this in a previous PR where the first commit was actually without ticket ID and the author added it to PR title (but the build still failed).

PR #4228, opened 2016-03-18
[ticket/9435] Replace BBcode magic numbers with constants

Senky:
[PHPBB3-9435](https://tracker.phpbb.com/browse/PHPBB3-9435)

### DIFF
--- a/phpBB/includes/bbcode.php
+++ b/phpBB/includes/bbcode.php
@@ -219,7 +219,7 @@ class bbcode
 		{
 			switch ($bbcode_id)
 			{
-				case 0:
+				case BBCODE_QUOTE_ID:
 					$this->bbcode_cache[$bbcode_id] = array(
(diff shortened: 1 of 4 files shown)


RMcGirr83:
You should prepend your commit message with the ticket number

Senky:
Well, I believe that `[ticket/9435]` prefix is just what you mean.

VSEphpbb:
I'd prefer to see BBCODE_ID_XXX. Looks better and more readable that way.

Elsensee:
I don't see the point in this patch. You're patching an old, abandoned part which will likely be removed in one of the next versions. And the new BBCode Parser doesn't need any IDs.

Nicofuma:
I agree but it doesn't harm.
Looks acceptable to me, @marc1706 what do you think?

PR #4229, opened 2016-03-18
[ticket/13013] Explain to user explicitl what "participate more" means

Senky:
[PHPBB3-13013](https://tracker.phpbb.com/browse/PHPBB3-13013)

### DIFF
--- a/phpBB/language/en/common.php
+++ b/phpBB/language/en/common.php
@@ -822,7 +822,7 @@ $lang = array_merge($lang, array(
 	'USERS'					=> 'Users',
 	'USE_PERMISSIONS'		=> 'Test out user’s permissions',
 
-	'USER_NEW_PERMISSION_DISALLOWED'	=> 'We are sorry, but you are not authorised to use this feature. You may have just registered here and may need to participate more to be able to use this feature.',
+	'USER_NEW_PERMISSION_DISALLOWED'	=> 'We are sorry, but you are not authorised to use this feature. You may have just registered here and may need to participate more in discussions to be able to use this feature.',
 
 	'VARIANT_DATE_SEPARATOR'	=> ' / ',	// Used in date format dropdown, eg: "Today, 13:37 / 01 Jan 2007, 13:37" ... to join a relative date with calendar date
 	'VIEWED'					=> 'Viewed',


VSEphpbb:
I don't see this as really needed, but this is not good English. If any thing, it should be more like "participate more within this board's community'

Mauron:
I'd go for
> _participate more to board activities_

DavidIQ:
That doesn't sound right either.  If this does get reviewed for merging VSE looks to be the closest in providing the proper English for this.

Senky:
The general purpose of the ticket is, that new users might be confused about what exactly "participating more" means. The idea is to explicitly state, that "participating more" means creating more posts. Thus, none of two corrections above sound sufficient to me.

RMcGirr83:
How about 
`may need to participate more by posting in the forum to be able to use this feature.`

rxu:
```may need to [leave|create|have] more posts```

marc1706:
As previously stated this is not proper English. You also don't participate to sth but participate in sth.

@VSEphpbb the idea behind this seems to be to actually tell people that they need to have more posts instead of being vague.

I'd recommend:
`You may have just registered here and may need to participate more in discussions to be able to use this feature.`
or possibly:
`You may have just registered here and may need to participate more in topics to be able to use this feature.`

marc1706:
@VSEphpbb mind reading over the text agan? Thx

RMcGirr83:
FWIW, looks fine to me.

VSEphpbb:
Fine I guess. phpBB is saved.

RMcGirr83:
lol

On Thu, Mar 24, 2016 at 12:07 PM, Matt Friedman <notifications@github.com>
wrote:

> Fine I guess. phpBB is saved.
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> <https://github.com/phpbb/phpbb/pull/4229#issuecomment-200904143>
>



-- 
Rich McGirr

marc1706:
Last questions: Shouldn't we maybe add this to 3.1.x, too? :)
In which case I'd appreciate a new PR against 3.1.x and not 3.2.x.

Nicofuma:
I think it should be against 3.1.x

Senky:
https://github.com/phpbb/phpbb/pull/4250

PR #4231, opened 2016-03-18
[ticket/13630] Prevent empty parameter select_single

Senky:
[PHPBB3-13630](https://tracker.phpbb.com/browse/PHPBB3-13630)

Please note, that this is probably only one of many problematic parameters. For example, when you submit member search for every user that starts with "a", this URL is generated:
`http://localhost/phpbb/phpBB/memberlist.php?sk=c&sd=a&username=a&email=&search_group_id=0&joined_select=lt&active_select=lt&count_select=eq&joined=&active=&count=&ip=`
Apache's mod_security has no knowledge, that this URL was generated by browser's form instead of classical anchor. Further inspection is required.

### DIFF
--- a/phpBB/memberlist.php
+++ b/phpBB/memberlist.php
@@ -1277,7 +1277,8 @@ switch ($mode)
 			}
 
 			$param = call_user_func_array('request_var', $call);
-			$param = urlencode($key) . '=' . ((is_string($param)) ? urlencode($param) : $param);
+			// Encode strings, convert everything else to int in order to prevent empty parameters.
+			$param = urlencode($key) . '=' . ((is_string($param)) ? urlencode($param) : (int) $param);
 			$params[] = $param;
 
 			if ($key != 'first_char')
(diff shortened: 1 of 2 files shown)


phpBB/memberlist.php
```diff
@@ -1277,7 +1277,9 @@
 			}
 
 			$param = call_user_func_array('request_var', $call);
-			$param = urlencode($key) . '=' . ((is_string($param)) ? urlencode($param) : $param);
+			// Encode strings, convert ints and bools to strings in order to prevent empty parameters.
+			// See: https://tracker.phpbb.com/browse/PHPBB3-13630
+			$param = urlencode($key) . '=' . ((is_string($param)) ? urlencode($param) : var_export($param, true));
```
marc1706: Why overcomplicate things with var_export()? This might actually cause issues due to the use of output buffering which can interfere with ob_start(). Converting things to int will also take care of the described issue.

phpBB/memberlist.php
```diff
@@ -1277,7 +1277,9 @@
 			}
 
 			$param = call_user_func_array('request_var', $call);
-			$param = urlencode($key) . '=' . ((is_string($param)) ? urlencode($param) : $param);
+			// Encode strings, convert ints and bools to strings in order to prevent empty parameters.
+			// See: https://tracker.phpbb.com/browse/PHPBB3-13630
```
marc1706: Please don't add links to tickets in the issue tracker in source code comments.

PR #4232, opened 2016-03-18
[ticket/14542] Move cron to controller

CHItA:
[PHPBB3-14542](https://tracker.phpbb.com/browse/PHPBB3-14542)

### DIFF
--- a/phpBB/config/default/routing/routing.yml
+++ b/phpBB/config/default/routing/routing.yml
@@ -8,6 +8,10 @@
 # instantiate the 'foo_service' service and call the 'method' method.
 #
 
+phpbb_cron_routing:
+    resource: cron.yml
+    prefix: /cron
+
 phpbb_feed_routing:
     resource: feed.yml
     prefix: /feed
--- a/phpBB/cron.php
+++ b/phpBB/cron.php
@@ -11,10 +11,11 @@
 *
 */
 
+use Symfony\Component\HttpFoundation\RedirectResponse;
+
 /**
 */
 define('IN_PHPBB', true);
-define('IN_CRON', true);
 $phpbb_root_path = (defined('PHPBB_ROOT_PATH')) ? PHPBB_ROOT_PATH : './';
 $phpEx = substr(strrchr(__FILE__, '.'), 1);
 include($phpbb_root_path . 'common.' . $phpEx);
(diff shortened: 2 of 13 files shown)


phpBB/config/default/routing/routing.yml
```diff
@@ -8,6 +8,9 @@
 # instantiate the 'foo_service' service and call the 'method' method.
 #
 
+phpbb_cron_routing:
+    resource: cron.yml
```
Nicofuma: you should move the /cron/ prefix here

phpBB/cron.php
```diff
@@ -23,50 +22,29 @@
...
+
+	$get_params .= $separator . $name . '=' . $value;
+	$separator = '&';
 }
 
-garbage_collection();
+/** @var \phpbb\controller\helper $controller_helper */
+$controller_helper = $phpbb_container->get('controller.helper');
+$response = new RedirectResponse(
+	$controller_helper->route('phpbb_cron_run', array(
```
Nicofuma: you can pass the whole parameters array array to the route() method. They will be added as query string by the UrlGenerator if required

phpBB/phpbb/cron/event/cron_runner_listener.php
```diff
@@ -0,0 +1,101 @@
...
+	private $request;
+
+	/**
+	 * Constructor
+	 *
+	 * @param db 				$lock
+	 * @param manager			$manager
+	 * @param request_interface	$request
+	 */
+	public function __construct(db $lock, manager $manager, request_interface $request)
```
Nicofuma: instead of injecting the request you could add a state to the event_listener and subscribe to Kernel::Request too (you will have the symfony request but I don't think that's an issue here, rigth?)

Nicofuma: nevermind... I forgot that `parse_parameters` expects the phpBB request :/

We should merge them at some point

Nicofuma:
please add a test to ensure that the old endpoint (cron.php) still works (and redirects to the good page)

CHItA:
All green, all good, really ugly tests, but that's that.

phpBB/phpbb/cron/controller/cron.php
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Controller for running cron jobs
+ */
+class cron
+{
+	/**
+	 * Handles CRON requests
+	 *
+	 * @return Response
```
marc1706: Is missing the `$cron_type` param in the docblock

phpBB/phpbb/cron/event/cron_runner_listener.php
```diff
@@ -0,0 +1,101 @@
...
+	{
+		$this->cron_lock	= $lock;
+		$this->cron_manager	= $manager;
+		$this->request		= $request;
+	}
+
+	/**
+	 * Runs the cron job after the response was sent
+	 */
```
marc1706: Missing `@param` for `$event` in docblock.

PR #4233, opened 2016-03-19
[ticket/14547] Add Vagrant support to phpBB

VSEphpbb:
https://tracker.phpbb.com/browse/PHPBB3-14547

This should be tested and evaluated first, to determine if it's a useful development addition/tool for phpBB to have. I made a pretty in depth documentation for it, the idea being to make coding, testing, debugging easier for contributors, especially the more casual contributors.

### DIFF
--- a/phpBB/docs/vagrant.md
+++ b/phpBB/docs/vagrant.md
@@ -2,7 +2,7 @@
 
 phpBB includes support for Vagrant. This allows developers and contributors to run phpBB without the need to set up their own local web server with traditional WAMP/MAMP stacks. It also provides a consistent environment between developers for writing and debugging code changes more productively.
 
-phpBB uses the [Laravel/Homestead](https://laravel.com/docs/5.2/homestead) Vagrant box. It runs a Linux server with Ubuntu 14.04, PHP 7, Nginx, SQLite3, MySQL, and a whole lot more (complete specs below).
+phpBB uses the [Laravel/Homestead](https://laravel.com/docs/5.1/homestead) Vagrant box. It runs a Linux server with Ubuntu 14.04, PHP 5.6, Nginx, SQLite3, MySQL, and a whole lot more (complete specs below).
 
 ## Get Started
 
@@ -14,7 +14,7 @@ phpBB uses the [Laravel/Homestead](https://laravel.com/docs/5.2/homestead) Vagra
 $ vagrant up
 ```
 
-* Access phpBB at `http://192.168.33.10/phpBB/`
+* Access phpBB at `http://192.168.33.10/`
 * Username: **admin**
 * Password: **adminadmin**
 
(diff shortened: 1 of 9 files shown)


VSEphpbb:
Damn, looks like the package I want to use requires PHP 5.5+

marc1706:
You should be able to use `~2.2` for having PHP 5.4 compatibility.

VSEphpbb:
Yeah, but the the point was to use the box with PHP 7. The 2.2 boxes are 5.6

marc1706:
Maybe something for master then? @Nicofuma

Nicofuma:
If we bump the php req in master yes (and I think we should). But we need to announce that first

VSEphpbb:
You going to bump it all the way up to PHP 5.6?

Nicofuma:
You have a conflict

marc1706:
@VSEphpbb instead of merging 3.2.x into your branch, just rebase your branch onto 3.2.x next time:
```git rebase upstream/3.2.x```
Make sure to run `git fetch upstream` before that.

**Comment by VSEphpbb** [comment]
My only question in this PR to you dev team guys is about the backup strategy being used to backup config.php files. Since a new config.php is generated for the vagrant environment, it backups up any existing config.php files, so as not to blow out any settings a user has if they also use the board in local AMP stacks. 

The current strategy is to have it backed up during the shell scripts, which result in:
config.php.bak~1
config.php.bak~2
etc..

Hence the modification to the gitignore file.

File: phpBB/docs/vagrant.md [hunk 1]
```diff
@@ -0,0 +1,123 @@
...
+> Note: destroying the vagrant server will remove all traces of the VM from your computer, reclaiming any disk space used by it. However, it also means the next time you vagrant up, you will be creating a brand new VM with a fresh install of phpBB and a new database.
+
+## Customising the phpBB configuration
+
+By default, phpBB is pre-configured to install with a MySQL database. You can, however, switch to PostegreSQL or SQLite3 by editing the `phpbb-install-config.yml` file in the vagrant directory. The next time you run `vagrant up` (or `vagrant provision`) it will be installed under the new configuration.
+
+If you prefer to access phpBB from the more friendly URL `http://phpbb.app` then you must update your computer's hosts file. This file is typically located at `/etc/hosts` for Mac/Linux or `C:\Windows\System32\drivers\etc\hosts` for Windows. Open this file and add the following line to it, at the very bottom, and save.
+
+```
+192.168.10.10  phpbb.app
```
marc1706: Seems like a different IP than above?

vagrant/after.sh
```diff
@@ -0,0 +1,33 @@
...
+# Update sqlite db file permissions
+sudo chown -R vagrant /tmp/phpbb.sqlite3
+
+# Add DEBUG mode to phpBB to remove annoying installer warnings
+sed -i "$ a @define('DEBUG', true);" ${PHPBB_CONFIG}
+
+# Update the PHP memory limits (enough to allow phpunit tests to run)
+sed -i "s/memory_limit = .*/memory_limit = 768M/" /etc/php5/fpm/php.ini
+
+echo "Your board is ready at http://192.168.10.10/"
```
marc1706: Also seems to be a different IP than mentioned at the beginning of the documentation.

vagrant/after.sh
```diff
@@ -0,0 +1,33 @@
...
+php ${PHPBB_PATH}/phpBB/install/phpbbcli.php install ${PHPBB_INSTALL}
+
+# Update sqlite db file permissions
+sudo chown -R vagrant /tmp/phpbb.sqlite3
+
+# Add DEBUG mode to phpBB to remove annoying installer warnings
+sed -i "$ a @define('DEBUG', true);" ${PHPBB_CONFIG}
+
+# Update the PHP memory limits (enough to allow phpunit tests to run)
+sed -i "s/memory_limit = .*/memory_limit = 768M/" /etc/php5/fpm/php.ini
```
marc1706: phpunit might actually grow above that so I'd recommend 1024M.

marc1706:
Rest looks good, only the IP at the beginning seems to be wrong.

VSEphpbb:
Updated

VSEphpbb:
Failure is unrelated

PR #4205, opened 2016-03-07
[ticket/13264] Do not force approved state if post visibility is known

marc1706:
Post visibility is used if it is set to a valid value.

Ticket: https://tracker.phpbb.com/browse/PHPBB3-13264

PR #4234, opened 2016-03-20
[ticket/9674] Use poster auth instead of current user auth

Senky:
[PHPBB3-9674](https://tracker.phpbb.com/browse/PHPBB3-9674)

Side effect of this modification is, that post awaiting approval edited by moderator does not get approved automatically upon editing. But imho this is desired behaviour, just ignored by everyone till now.

### DIFF
--- a/phpBB/includes/functions_posting.php
+++ b/phpBB/includes/functions_posting.php
@@ -1535,7 +1535,7 @@ function delete_post($forum_id, $topic_id, $post_id, &$data, $is_soft = false, $
 */
 function submit_post($mode, $subject, $username, $topic_type, &$poll, &$data, $update_message = true, $update_search_index = true)
 {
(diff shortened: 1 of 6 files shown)


VSEphpbb:
That's ugly...another query during posting !?! Any way, at the very least it should be checking if the current user is not the poster, so this is not always being run.

marc1706:
Posts awaiting approval will no longer be approved while being edited by moderators upon editing posts (as of https://github.com/phpbb/phpbb/pull/4205).

VSEphpbb:
Good that's a way better fix :)

marc1706:
This change poses a high risk of privilege escalation. Also, I'm not able to confirm this issue on the latest 3.1.x branch.

Senky:
> ...I'm not able to confirm this issue on the latest 3.1.x branch.

True.

PR #4235, opened 2016-03-23
[ticket/14550] Fix the number of characters returned by unique_id()

Nicofuma:
https://tracker.phpbb.com/browse/PHPBB3-14550

PHPBB3-14450


hanakin:
This is under the wrong ticket number you tagged it 14450 and it should be 14550 fyi

PR #4237, opened 2016-03-23
[ticket/14555] Uniformize cache directory usages

Nicofuma:
https://tracker.phpbb.com/browse/PHPBB3-14555

PHPBB3-14555


marc1706:
The check_ban_test still needs to be fixed apparently.

PR #4240, opened 2016-03-24
[ticket/13625] Correct types of properties

RMcGirr83:
https://tracker.phpbb.com/browse/PHPBB3-13625

PHPBB3-13625


marc1706:
Did you modify the complete file by accident? Make sure it's saved with Unix line endings and not windows line endings.

RMcGirr83:
Didn't think I did.  Farken smartgit messing around with line endings because I'm on windows. :angry: Will redo

marc1706:
Thanks, looks better now.

PR #4253, opened 2016-03-27
[ticket/14559] Remove attachment BBCode tags from the quoted message

rxu:
<a href="https://tracker.phpbb.com/browse/PHPBB3-14559">PHPBB3-14559</a>.

### DIFF
--- a/phpBB/posting.php
+++ b/phpBB/posting.php
@@ -1589,6 +1589,9 @@ $message_parser->decode_message($post_data['bbcode_uid']);
 
 if ($generate_quote)
 {
+	// Remove attachment bbcode tags from the quoted message to avoid mixing with the new post attachments if any
+	$message_parser->message = preg_replace('#\[attachment=([0-9]+)\](.*?)\[\/attachment\]#uis', '\\2', $message_parser->message);
+
 	if ($config['allow_bbcode'])
 	{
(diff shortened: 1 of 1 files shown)


rxu:
@JoshyPHP Not sure but I guess for 3.2 it should be handled in different way. I couldn't found the relevant textformatter util though (just the one which is removing the whole bbcode tag with its content).

JoshyPHP:
@rxu Yes, since you remove the markup after the message has been decoded, it works the same way in 3.1 and 3.2.

phpBB/posting.php
```diff
@@ -1589,6 +1589,9 @@
 
 if ($generate_quote)
 {
+	// Remove attachment bbcode tags from the quoted message to avoid mixing with the new post attachments if any
+	$message_parser->message = preg_replace('#\[attachment=([0-9]+)\](.*?)\[\/attachment\]#uis', '\\2', $message_parser->message);
```
Nicofuma: why do you keep the attachment description?

rxu: This is the file name unlinked, to keep the discussion context I'd say. And it's the same way it looks when you quote attachments without adding new ones to new post.

Nicofuma: ok

PR #4254, opened 2016-03-27
[ticket/14562] Language fallback for extension permissions

kasimi:
The code was copied and adjusted from `add_mod_info()` in `includes/functions_module.php` which already has language fallback.

PHPBB3-14562

### DIFF
--- a/phpBB/includes/functions_admin.php
+++ b/phpBB/includes/functions_admin.php
@@ -3081,17 +3081,46 @@ function tidy_database()
 */
 function add_permission_language()
 {
-	global $user, $phpEx, $phpbb_extension_manager;
+	global $config, $user, $phpEx, $phpbb_extension_manager;
 
 	// add permission language files from extensions
 	$finder = $phpbb_extension_manager->get_finder();
 
-	$lang_files = $finder
+	// We grab the language files from the default, English and user's language.
+	// So we can fall back to the other files like we do when using add_lang()
+	$default_lang_files = $english_lang_files = $user_lang_files = array();
+
+	// Search for board default language if it's not the user language
+	if ($config['default_lang'] != $user->lang_name)
+	{
+		$default_lang_files = $finder
+			->prefix('permissions_')
+			->suffix(".$phpEx")
+			->core_path('language/' . basename($config['default_lang']) . '/')
+			->extension_directory('/language/' . basename($config['default_lang']))
+			->find();
+	}
+
+	// Search for english, if its not the default or user language
+	if ($config['default_lang'] != 'en' && $user->lang_name != 'en')
+	{
+		$english_lang_files = $finder
+			->prefix('permissions_')
+			->suffix(".$phpEx")
+			->core_path('language/en/')
+			->extension_directory('/language/en')
+			->find();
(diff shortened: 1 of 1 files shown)


Nicofuma:
You have send your PR against the develop branch which is very old.
Could you open a new PR (with the same fix) against 3.2.x please?

phpBB/includes/functions_admin.php
```diff
@@ -3081,18 +3081,47 @@ function tidy_database()
...
+			->extension_directory('/language/en')
+			->find();
+	}
+
+	// Find files in the user's language
+	$user_lang_files = $finder
 		->prefix('permissions_')
 		->suffix(".$phpEx")
 		->core_path('language/' . $user->lang_name . '/')
 		->extension_directory('/language/' . $user->lang_name)
```
CHItA: For 3.2.x you probably achieve the same thing if you just remove the `$user->lang_name` from both places above.

kasimi: Wouldn't that load `permissions_*.php` files from *all* languages?

kasimi:
@Nicofuma Sorry, didn't pay attention there. Against 3.2.x? Does that mean the code won't make it into 3.1.x?


Btw, the deprecated branch names are still mention [here](https://www.phpbb.com/get-involved/).

phpBB/includes/functions_admin.php
CHItA: yes it would, but i don't see the problem with that.

CHItA:
Yep, that would mean it will be <=3.2.x only. Though i'm not sure whether it has to be excluded from 3.1. @nickvergessen?

phpBB/includes/functions_admin.php
kasimi: Maybe I see a problem where there is none... I think loading *all* language files would result in each language overwriting the previous one. We'd end up with always the same language being served to the user (probably the last one, alphabetically), ignoring the user-specific language preference.

CHItA: That is not really a problem, though this shouldn't be an issue for 3.2.x anyway. Did you tested it for 3.2.x or just 3.1?

kasimi: Now I have and I see why it works. I should probably submit 2 PRs, this one against 3.1.x and another one with your suggestion against 3.2.x. How should I name my branches? Or should I open another ticket?

nickvergessen:
Well this goes against the wrong branch, develop* branches are read only...

CHItA:
Yep, but the question is whether it should be against 3.1.x or not

marc1706:
@kasimi Thanks for the heads up regarding the info on the get involved page. This will be updated asap.

nickvergessen:
Looking at the state of maintenance and since this "extends functionality", I'd say 3.2+?

kasimi:
A user is complaining because he can't understand what the extension's permissions are for. Why is displaying proper language instead of `ACL_U_SOME_PERM` a feature?

CHItA:
i tend to agree with @kasimi.

PR #4256, opened 2016-03-27
[ticket/13683] Respects force_server_vars settings when generating URLs

Nicofuma:
https://tracker.phpbb.com/browse/PHPBB3-14134

PHPBB3-13683

### DIFF
--- a/phpBB/phpbb/controller/helper.php
+++ b/phpBB/phpbb/controller/helper.php
@@ -71,7 +71,8 @@ class helper
 	* @param \phpbb\template\template $template Template object
 	* @param \phpbb\user $user User object
 	* @param \phpbb\config\config $config Config object
-	* @param \phpbb\controller\provider $provider Path provider
+	 *
+	 * @param \phpbb\controller\provider $provider Path provider
 	* @param \phpbb\extension\manager $manager Extension manager object
 	* @param \phpbb\symfony_request $symfony_request Symfony Request object
 	* @param \phpbb\request\request_interface $request phpBB request object
@@ -146,7 +147,7 @@ class helper
 		if ($this->config['force_server_vars'])
 		{
 			$context->setHost($this->config['server_name']);
-			$context->setScheme(substr($this->config['server_protocol'], -4));
+			$context->setScheme(substr($this->config['server_protocol'], 0, -3));
 			$context->setHttpPort($this->config['server_port']);
 			$context->setHttpsPort($this->config['server_port']);
 			$context->setBaseUrl($this->config['script_path']);
--- a/phpBB/phpbb/controller/helper.php
+++ b/phpBB/phpbb/controller/helper.php
@@ -150,7 +150,7 @@ class helper
 			$context->setScheme(substr($this->config['server_protocol'], 0, -3));
 			$context->setHttpPort($this->config['server_port']);
 			$context->setHttpsPort($this->config['server_port']);
-			$context->setBaseUrl($this->config['script_path']);
+			$context->setBaseUrl(rtrim($this->config['script_path'], '/'));
 		}
 
 		$script_name = $this->symfony_request->getScriptName();
--- a/phpBB/phpbb/controller/helper.php
+++ b/phpBB/phpbb/controller/helper.php
@@ -143,6 +143,14 @@ class helper
 		$context = new RequestContext();
 		$context->fromRequest($this->symfony_request);
 
+		if ($this->config['force_server_vars']) {
+			$context->setHost($this->config['server_name']);
+			$context->setScheme(substr($this->config['server_protocol'], -4));
+			$context->setHttpPort($this->config['server_port']);
+			$context->setHttpsPort($this->config['server_port']);
+			$context->setBaseUrl($this->config['script_path'] . '/app.php');
+		}
+
 		$script_name = $this->symfony_request->getScriptName();
 		$page_name = substr($script_name, -1, 1) == '/' ? '' : utf8_basename($script_name);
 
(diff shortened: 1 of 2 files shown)


marc1706:
There seems to be an issue with your first commit's message.

phpBB/phpbb/controller/helper.php
```diff
@@ -143,6 +143,15 @@ public function route($route, array $params = array(), $is_amp = true, $session_
 		$context = new RequestContext();
 		$context->fromRequest($this->symfony_request);
 
+		if ($this->config['force_server_vars'])
+		{
+			$context->setHost($this->config['server_name']);
+			$context->setScheme(substr($this->config['server_protocol'], -4));
```
marc1706: Not sure if this potentially buggy implementation should be used. Server protocols are not limited to 4 characters. This is the definition by the RFC:
```
Scheme names consist of a sequence of characters beginning with a
   letter and followed by any combination of letters, digits, plus
   ("+"), period ("."), or hyphen ("-").  Although schemes are case-
   insensitive, the canonical form is lowercase and documents that
   specify schemes must do so with lowercase letters.  An implementation
   should accept uppercase letters as equivalent to lowercase in scheme
   names (e.g., allow "HTTP" as well as "http") for the sake of
   robustness but should only produce lowercase scheme names for
   consistency.

      scheme      = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )
```

tests/controller/common_helper_route.php
```diff
@@ -415,7 +428,40 @@ public function helper_url_data_network_with_rewrite()
...
+	{
+		return array(
+			array(false, true, 'my_server', 443, '/my/board', 'http', UrlGeneratorInterface::ABSOLUTE_URL, 'http://my_server:443/my/board/app.php/foo'),
+			array(true, true, 'my_server', 443, '/my/board', 'http', UrlGeneratorInterface::ABSOLUTE_URL, 'http://my_server:443/my/board/foo'),
+			array(false, true, 'my_server', 443, '/my/board', 'http', UrlGeneratorInterface::NETWORK_PATH, '//my_server:443/my/board/app.php/foo'),
+			array(true, true, 'my_server', 443, '/my/board', 'http', UrlGeneratorInterface::NETWORK_PATH, '//my_server:443/my/board/foo'),
+			array(false, true, 'my_server', 443, '/my/board', 'http', UrlGeneratorInterface::ABSOLUTE_PATH, '/my/board/app.php/foo'),
+			array(true, true, 'my_server', 443, '/my/board', 'http', UrlGeneratorInterface::ABSOLUTE_PATH, '/my/board/foo'),
+			array(false, true, 'my_server', 443, '/my/board', 'http', UrlGeneratorInterface::RELATIVE_PATH, 'app.php/foo'),
+			array(true, true, 'my_server', 443, '/my/board', 'http', UrlGeneratorInterface::RELATIVE_PATH, 'foo'),
```
marc1706: Please also add tests for non-standard ports (i.e. 444 instead of 443).

Nicofuma:
shoud be good now

phpBB/phpbb/controller/helper.php
```diff
@@ -71,7 +71,8 @@ class helper
 	* @param \phpbb\template\template $template Template object
 	* @param \phpbb\user $user User object
 	* @param \phpbb\config\config $config Config object
-	* @param \phpbb\controller\provider $provider Path provider
+	 *
```
marc1706: You have indented these with PHPStorm apparently. However, the indentation on these two lines is different than the other lines before and after it.

phpBB/phpbb/controller/helper.php
```diff
@@ -143,6 +144,15 @@ public function route($route, array $params = array(), $is_amp = true, $session_
 		$context = new RequestContext();
 		$context->fromRequest($this->symfony_request);
 
+		if ($this->config['force_server_vars'])
+		{
+			$context->setHost($this->config['server_name']);
+			$context->setScheme(substr($this->config['server_protocol'], 0, -3));
```
marc1706: I was actually thinking of something like ```preg_replace('/^([a-zA-Z][a-zA-Z+.-]+)(.+)?$/i', '$1', $this->config['server_protocol'])```

phpBB/phpbb/controller/helper.php
```diff
@@ -143,6 +144,15 @@ public function route($route, array $params = array(), $is_amp = true, $session_
...
+			$context->setHttpPort($this->config['server_port']);
+			$context->setHttpsPort($this->config['server_port']);
+			$context->setBaseUrl(rtrim($this->config['script_path'], '/'));
```
marc1706: Can this potentially be used to redirect or load other URLs (as in redirect users to other websites)?

phpBB/phpbb/controller/helper.php
Nicofuma: actually here we are just removing the 3 last characters because in the acp we require `http://` or `https://`

phpBB/phpbb/controller/helper.php
Nicofuma: Yes (if you are generating a full UR and not only a relative path) but that's exactly the purpose of `force_server_vars` and the current applied behaviour in generate_board_url()

phpBB/phpbb/controller/helper.php
Nicofuma: fixed

phpBB/phpbb/controller/helper.php
marc1706: Ok, fair enough.

PR #4259, opened 2016-03-27
[ticket/14565] Updates composer.phar to 1.0.0-beta2

Nicofuma:
https://tracker.phpbb.com/browse/PHPBB3-14565

PHPBB3-14565


marc1706:
@Nicofuma it seems like the openssl extension is not available on 5.3.3 builds. Can you check if this can be enabled and otherwise add a workaround to disable tls for composer on 5.3.3 builds as described in the error message?

PR #3196, opened 2014-11-29
Ticket/10748 Refactor db tools into multiple classes

nickvergessen:
https://tracker.phpbb.com/browse/PHPBB3-10748

PR #4261, opened 2016-03-28
[ticket/10436] Move get_database_size() into phpbb\db\tools\tools

Senky:
[PHPBB3-10436](https://tracker.phpbb.com/browse/PHPBB3-10436)


marc1706:
I don't think that coupling the database tools with the user class is a good idea. Especially since the language methods have been decoupled from the user class into the language class. Even then, the database tools shouldn't need to output language specific strings as that is not its purpose.
The complete function should be reworked and split for the specific db types. Some work has already been put into splitting the tools up per database type:
https://github.com/phpbb/phpbb/pull/3196

Nicofuma:
Just to complete what Marc said about language strings, the process now is to send an instance of `\phpbb\exception\runtime_exception` (or any exception class implementing the interface `\phpbb\exception\exception_interface`) every-time there is an error (instead of relying on trigger_error or returning an error?

Senky:
> I don't think that coupling the database tools with the user class is a good idea.

Me thinks the same, I just did not want to change the method much.

This PR depend on #3196 then. I will wait till it is merged.

VSEphpbb:
In phpBB 3.2 we do not use the user class for language translations anyways. There's a language class for that. The user has no reason to be used here.

PR #4263, opened 2016-03-29
[ticket/10809] Remove MSSQL Support

Senky:
[PHPBB3-10809](https://tracker.phpbb.com/browse/PHPBB3-10809)


marc1706:
I don't think such a large part should be removed from phpBB 3.1 as that could potentially break things and is a feature change. The PR should therefore be against 3.2. /cc @Nicofuma

Nicofuma:
not in 3.1 for sure. Maybe 3.2 I don't know. @Noxwizard what do you think?

By the way, it's sad to remove mssql support :/

DavidIQ:
mssqlnative works far better than the old mssql and mssql_odbc so I agree that this should be left for 3.2, not for current 3.1.

Noxwizard:
It's not a matter of one working better than the other, PHP dropped this module in 5.3+. Since phpBB 3.1.x only supports PHP 5.3.3+, you can't even use this in 3.1 if you wanted to.

marc1706:
@Noxwizard while this module certainly shouldn't exist on 5.3+, I can imagine some hosters having hacked this up for their sake. As this is a potentially breaking change, it shouldn't be added to 3.1.x now that it's been stable for 1.5 years.

PR #4271, opened 2016-04-01
[ticket/14562] Language fallback for extension permissions (3.2.x)

kasimi:
Continuation of https://github.com/phpbb/phpbb/pull/4254 for `3.2.x` using @CHItA's suggestion.

[PHPBB3-14562](https://tracker.phpbb.com/browse/PHPBB3-14562)

### DIFF
--- a/phpBB/includes/functions_admin.php
+++ b/phpBB/includes/functions_admin.php
@@ -2999,8 +2999,8 @@ function add_permission_language()
 	$lang_files = $finder
 		->prefix('permissions_')
 		->suffix(".$phpEx")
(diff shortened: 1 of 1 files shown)


CHItA:
@Nicofuma maybe the symfony finder would function better here? (we could exclude the `language/iso/` part)

CHItA:
nvm, that's probably will be the solution for 3.3, as that is somewhere between our finder and symfony's.

Nicofuma:
yeah, that's something for 3.3 because we need functionalities of both finders.

PR #4275, opened 2016-04-03
[ticket/14426] Fixed undefined offset error

JoshyPHP:
It's just an undefined offset error that was introduced in 3106195cddec48ec779282d03f2c54c88fc66511. The second capture is optional and since it's the last capture it will be omitted from `$match` rather than set to an empty string.

No test attached because there's no support for running "second pass" tests on the old BBCode engine.

PHPBB3-14426


marc1706:
@JoshyPHP please send PRs for 3.2.x against the 3.2.x branch next time. Saves me a bit of time when merging this into phpBB.

marc1706:
Merged via https://github.com/phpbb/phpbb/commit/3281826a5155f50a38d8534de6956e07d9e1b72d and https://github.com/phpbb/phpbb/commit/0eefec86a54d2f13a5d7742c067a7dd78fe60e29

PR #4278, opened 2016-04-03
[ticket/14581]  Add core events to content_visibility.

LaxSlash:
Events added for both changing the content visibility for both posts and topics,
executing right before and after the SQL UPDATE query for setting the
visibillity.

PHPBB3-14581

### DIFF
--- a/phpBB/phpbb/content_visibility.php
+++ b/phpBB/phpbb/content_visibility.php
@@ -459,36 +459,6 @@ class content_visibility
 			WHERE ' . $this->db->sql_in_set('post_id', $post_ids);
 		$this->db->sql_query($sql);
 
-		/**
-		 * Perform actions right after the query to change post visibility
-		 *
-		 * @event core.set_post_visibility_after_sql
-		 * @var			int			visibility		Element of {ITEM_APPROVED, ITEM_DELETED, ITEM_REAPPROVE}
-		 * @var			array		post_id			Array containing all post IDs to be modified. If blank, all posts within the topic are modified.
-		 * @var			int			topic_id		Topic of the post IDs to be modified.
-		 * @var			int			forum_id		Forum ID that the topic_id resides in.
-		 * @var			int			user_id			User ID doing this action.
-		 * @var			int			timestamp		Timestamp of this action.
-		 * @var			string		reason			Reason specified by the user for this change.
-		 * @var			bool		is_starter		Are we changing the topic's starter?
-		 * @var			bool		is_latest		Are we changing the topic's latest post?
-		 * @var			array		data			The data array for this action.
-		 * @since 3.1.9-RC1
-		 */
-		$vars = array(
-			'visibility',
-			'post_id',
-			'topic_id',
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/content_visibility.php
```diff
@@ -424,11 +424,71 @@ public function set_post_visibility($visibility, $post_id, $topic_id, $forum_id,
...
+			'topic_id',
+			'forum_id',
+			'user_id',
+			'timestamp',
+			'reason',
+			'is_starter',
+			'is_latest',
+			'data',
+		);
+		extract($this->phpbb_dispatcher->trigger_event('core.set_post_visibility_after_sql', compact($vars)));
```
Nicofuma: I don't really see the use case for this event. With the previous you can alter the SQL query. But with this one?

LaxSlash:
The _after events could be used for informational reasons for such a case where any further actions need to be performed on other tables. In my case, I have an SQL Query that grabs the count of topic rows that match a certain critera, and this would be useful so that I can run those queries around the same time that the queries to update the post counts would be run. Plus, it can be used for mods that have either delete_topics_aftery_query and/or delete_topics_before_query but don't have compatibillity for soft deletes.

Nicofuma:
But wouldn't it have more sense to trigger an event after the forums/users/topics count have been synced?

LaxSlash:
Fair enough point. I moved both events to after all the sync steps, and took out the _sql part to their names.

Also, sorry about that double commit..... completely forgot to prefix the title of it, but figured that I might as well clarify the comments before the events.

LaxSlash:
Looks like all the builds passed this time around... everything check out?

phpBB/phpbb/content_visibility.php
```diff
@@ -424,6 +424,36 @@ public function set_post_visibility($visibility, $post_id, $topic_id, $forum_id,
...
+		 * @var			array		post_id			Array containing all post IDs to be modified. If blank, all posts within the topic are modified.
+		 * @var			int			topic_id		Topic of the post IDs to be modified.
+		 * @var			int			forum_id		Forum ID that the topic_id resides in.
+		 * @var			int			user_id			User ID doing this action.
+		 * @var			int			timestamp		Timestamp of this action.
+		 * @var			string		reason			Reason specified by the user for this change.
+		 * @var			bool		is_starter		Are we changing the topic's starter?
+		 * @var			bool		is_latest		Are we changing the topic's latest post?
+		 * @var			array		data			The data array for this action.
+		 * @since 3.1.9-RC1
```
marc1706: Unfortunately, this didn't make it into 3.1.9. I'd very much appreciate it, if you could update this to say 3.1.10-RC1. Thank you.

rmcgirr83:
@marc1706 is this still a WIP?  #4406 replaces this PR.  Should it be closed?

LaxSlash:
Yes, this can be closed. Completely forgot about this... crazy Summer.

Thanks for including it in that PR!

PR #4282, opened 2016-04-06
[ticket/14548] Move deprecated globals and functions

CHItA:
[PHPBB3-14584](https://tracker.phpbb.com/browse/PHPBB3-14584)

### DIFF
--- a/phpBB/includes/functions_compatibility.php
+++ b/phpBB/includes/functions_compatibility.php
@@ -22,7 +22,7 @@ if (!defined('IN_PHPBB'))
 /**
 * Get user avatar
 *
-* @deprecated 3.1.0-a1 (To be removed: 3.3.0)
+* @deprecated 3.1.0-a1 (To be removed: 4.0.0)
 *
 * @param string $avatar Users assigned avatar name
 * @param int $avatar_type Type of avatar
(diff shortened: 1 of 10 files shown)


phpBB/includes/functions_compatibility.php
```diff
@@ -511,3 +511,93 @@ function phpbb_pcre_utf8_support()
 {
 	return true;
 }
+
+/**
+ * Casts a variable to the given type.
+ *
+ * @deprecated 3.3.0-dev (To be removed 3.4.0)
```
Nicofuma: deprecated in 3.1

CHItA:
@phpbb/development-team bump

phpBB/includes/functions_compatibility.php
```diff
@@ -162,7 +162,7 @@ function cache_moderators()
 /**
 * Removes moderators and administrators from foe lists.
 *
-* @deprecated 3.1.0 (To be removed: 3.3.0)
```
marc1706: I guess the reasoning for this change is to not cause BC breaks in the 3.x line?

CHItA: yeah, it was the idea of @Nicofuma iirc. And since it only means that we don't delete something i guess it is a fairly good one.

PR #4283, opened 2016-04-07
[ticket/14587] Remove sectionGroup from web.config

DavidIQ:
Section can cause errors on IIS installations.

PHPBB3-14587


marc1706:
@nickvergessen pls make sure to include this before packaging anything

PR #4286, opened 2016-04-12
[ticket/14569] Extract CLI progress bar creation to a method

VSEphpbb:
https://tracker.phpbb.com/browse/PHPBB3-14569

PHPBB3-14569

### DIFF
--- a/phpBB/phpbb/console/command/command.php
+++ b/phpBB/phpbb/console/command/command.php
@@ -13,6 +13,10 @@
 
 namespace phpbb\console\command;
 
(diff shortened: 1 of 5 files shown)


VSEphpbb:
Tests fail because phing-sniff kind of false-positives on a Use statement being used only in a docblock. The thing sniffer should be updated to not consider Use statements unused when they're only appearing in docblocks, and then ere-run the tests here I guess :construction:

VSEphpbb:
Depends on https://github.com/phpbb/phpbb/pull/4287

PR #4299, opened 2016-04-20
[ticket/14607] Missing Auto Increment in Report Table

JimMH:
Adding $report_id to the array in the report handler and adding auto_increment to the DB field report id

PHPBB3-14607

### DIFF
--- a/phpBB/install/schemas/schema.json
+++ b/phpBB/install/schemas/schema.json
@@ -2264,7 +2264,8 @@
         "COLUMNS": {
             "report_id": [
                 "ULINT",
-                0
+                null,
+				"auto_increment"
             ],
             "reason_id": [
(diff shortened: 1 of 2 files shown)


marc1706:
This will probably need a new migration file to correctly update previously installed boards.

rxu:
The question is whether this was done intentionally (non-autoincrement field) and why, if yes.

marc1706:
Caused by: https://github.com/phpbb/phpbb/commit/7cd7bb3c015be1355ab634d3cb704c1d764bf8b0

There needs to be a migration that reverts the incorrect change to not use auto_increment and the migration changing the default data types needs to be modified to use auto_increment for report_id.

marc1706:
Closing for incomplete fix

PR #4304, opened 2016-04-29
[3.1.x] Fixed auto-prune failing on large forums.

TheKigen:
Auto-prune would fail on very large forums.  This occurs in a normal environment at about 1000 topics.

PHPBB3-14616

### DIFF
--- a/phpBB/includes/functions_admin.php
+++ b/phpBB/includes/functions_admin.php
@@ -2531,7 +2531,7 @@ function sync($mode, $where_type = '', $where_ids = '', $resync_parents = false,
 /**
 * Prune function
 */
-function prune($forum_id, $prune_mode, $prune_date, $prune_flags = 0, $auto_sync = true)
+function prune($forum_id, $prune_mode, $prune_date, $prune_flags = 0, $auto_sync = true, $prune_limit = 0)
 {
 	global $db, $phpbb_dispatcher;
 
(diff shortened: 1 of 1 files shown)


TheKigen:
The travis build errors look like bugs in the testing environment and not with the actual commit.

marc1706:
The travis build failure is due to the commit message not fitting our pre-defined commit message format. Please take a look at this detailed explanation on how the commit message should be structured: https://wiki.phpbb.com/Git#Commit_Messages

marc1706:
Have you checked what is causing this limit? Is it maybe platform dependent (e.g. is it affecting MySQL on windows, which is incredibly slow).?

TheKigen:
The problem is its trying to delete way too many topics at once.  Generally what happens is it times out.  The system is CentOS 7.2.1511, Apache 2.4.6, MariaDB 10.1.13, PHP 5.6.21.

To test it just have a script generate about 5000 topics, then try to prune those topics.  Either automatically through the built-in cron or manually.  Its trying to do too much in one page load.  In my specific environment I run a honeypot for spanbots.  Their spam generates a lot of topics in one day.

marc1706:
This might potentially have an unexpected side effect. Due to the introduced prune limit. boards might end up not fully pruning forums anymore. After this change, the next prune is still set for the next day. If more topics are created per day than the prune limit covers, this might result in the topics building up in  a forum for no apparent reason.

TheKigen:
This does not set the "prune done" flag until the pruning is actually done for the cron job.  For manual pruning the limit is not set and thus it'll attempt to prune everything in one attempt.

n-aleha:
If you manually try to delete them all 5000 at once, then I had I similar issue (+explanation in my case) with "Remember me" keys.
https://area51.phpbb.com/phpBB/viewtopic.php?f=81&t=47421

TheKigen:
Tried to alter the commit message but it gives different errors.

https://travis-ci.org/phpbb/phpbb/builds/140481573

Inspecting commit message of commit 80c0416d139a1b96ddf0e00810997611985d133e
Unexpected EOF encountered 
Aborting commit.

TheKigen:
Fixed commit messages in https://github.com/phpbb/phpbb/pull/4374 and Travis accepts it.

PR #4306, opened 2016-04-30
[ticket/14618] Update s9e\TextFormatter to 0.5.0

JoshyPHP:
PHPBB3-14618

### DIFF
--- a/phpBB/composer.json
+++ b/phpBB/composer.json
@@ -33,7 +33,7 @@
 		"marc1706/fast-image-size": "1.1.*",
 		"paragonie/random_compat": "^1.2",
 		"patchwork/utf8": "1.1.*",
-		"s9e/text-formatter": "^0.4.2",
+		"s9e/text-formatter": "^0.5.0",
 		"symfony/config": "2.8.*",
 		"symfony/console": "2.8.*",
 		"symfony/debug": "2.8.*",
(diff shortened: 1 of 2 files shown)


phpBB/composer.json
```diff
@@ -33,7 +33,7 @@
 		"marc1706/fast-image-size": "1.1.*",
 		"paragonie/random_compat": "^1.2",
 		"patchwork/utf8": "1.1.*",
-		"s9e/text-formatter": "^0.4.2",
+		"s9e/text-formatter": "^0.5.0",
```
Nicofuma: if you are using `^` you don't need to change the version

JoshyPHP: Not in pre-1.0 versions, where Composer [*"acts with safety in mind and treats \^0.3 as >=0.3.0 <0.4.0."*](https://getcomposer.org/doc/articles/versions.md#caret)

Nicofuma: in this case, use the proper operator: `~`

JoshyPHP: Unfortunately I have no idea what's the difference between those operators. Since my intention was to update the version number without changing the semantics of the operator it is associated with, I will withdraw this pull request and let you sort it out at your convenience. This seems much more efficient this way.

Nicofuma: Do you want to restrict to 0.5.* or 0.* ?
If it is 0.5.*, then use `~0.5`. Otherwise `^0.5`

Elsensee:
The `^` operator was useful for the `lusitanian/oauth` package because there was a pretty big BC break between 0.2 and 0.8. Thus I used the operator for every package instead of `~` to be consistent. (and because of the advantages for pre-1.0 versions)
I think we should stick to the operator so we can prevent breaking phpBB because of any BC breaks within other dependencies.

PR #4307, opened 2016-05-01
[ticket/14614] Allow usage of language keys in composer.json

Skouat:
These changes allow to use language keys in composer.json to translate the name and description of the extension and also the Author Role.

https://tracker.phpbb.com/browse/PHPBB3-14614

PHPBB3-14614


VSEphpbb:
This is not a good idea. 

Composer.json will be used by packagist, so they must be valid composer.json files for one, and use normal English as well...like every single other composer project does (including phpBB).

marc1706:
I don't see the point of doing this and also agree with what @VSEphpbb said.

PR #4308, opened 2016-05-01
[ticket/14607] Add missing auto_increment to report_id

marc1706:
Replaces #4299 

Ticket: https://tracker.phpbb.com/browse/PHPBB3-14607

### DIFF
--- a/phpBB/phpbb/db/migration/data/v320/report_id_auto_increment.php
+++ b/phpBB/phpbb/db/migration/data/v320/report_id_auto_increment.php
@@ -1,0 +1,45 @@
+<?php
+/**
+*
+* This file is part of the phpBB Forum Software package.
+*
+* @copyright (c) phpBB Limited <https://www.phpbb.com>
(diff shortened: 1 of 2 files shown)


phpBB/phpbb/db/migration/data/v320/report_id_auto_increment.php
```diff
@@ -0,0 +1,45 @@
...
+			),
+		);
+	}
+
+	public function revert_schema()
+	{
+		return array(
+			'change_columns'	=> array(
+				$this->table_prefix . 'reports'				=> array(
+					'report_id'		=> array('ULINT', 0), 0),
```
lavigor: Sorry for the question, but is it intended?
It seems that the second zero is a typo.

marc1706: No, it isn't. Added pull request for it here: https://github.com/phpbb/phpbb/pull/4349

PR #4309, opened 2016-05-01
[ticket/14415] Adds data-ajax="mark_topics_read" to link "mark_all_read"

Crizz0:
The link in the search-results of "unread posts" will now work with AJAX. I think it is okay to use the existing function from "mark_topics_read" instead of creating a new, very similar function for "mark_all_read" instead. 

PHPBB3-14415 https://tracker.phpbb.com/projects/PHPBB3/issues/PHPBB3-14415

### DIFF
--- a/phpBB/styles/prosilver/template/search_results.html
+++ b/phpBB/styles/prosilver/template/search_results.html
@@ -33,7 +33,7 @@
 		<!-- EVENT search_results_searchbox_after -->
 
 		<div class="pagination">
-			<!-- IF U_MARK_ALL_READ --><a href="{U_MARK_ALL_READ}" class="mark-read" accesskey="m">{L_MARK_ALL_READ}</a> &bull;<!-- ENDIF -->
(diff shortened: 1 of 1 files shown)


Crizz0:
!rhead

Crizz0:
On the other hand, do you think it is senseful to do this? 

I mean, the user will be on a "unread posts"-search page, where he has marked all posts read and then he can navigate where he wants to go. 

At the moment he can click on the link and goes to the information page and gets redirected to the Board-Index. Where he starts navigating now. Not much difference.

Crizz0:
@marc1706 what do you think?

marc1706:
I do actually think this might add more confusion as it leaves the user in an incorrect state. The page won't be refreshed IIRC? Therefore the posts on the unread posts page will be read? Seems like weird behavior to me and the previous implementation works fine as well.

Crizz0:
The ajax will only refresh the topic-titles and icons from "unread" to "read". And won't display new unread topics and remove the read ones. 

So i think I'll close this.

PR #4310, opened 2016-05-01
[ticket/14619] Updates "docs" from 3.1.x/ascraeus to 3.2.x/rhea

Crizz0:
I just updated the head and some titles and a few section-titles and some "3.1.0" infos in the coding-guidelines, to keep the example up to date. 

The INSTALL.html and README.html need further work to describe the updates from 3.1 to 3.2, 3.0 to 3.2 and 2.0 to 3.2. 

I could do that, if you think i should and it should be done now. 

https://tracker.phpbb.com/projects/PHPBB3/issues/PHPBB3-14619
PHPBB3-14619


Nicofuma:
Please do it :) And then you could do the same for 3.3 (the master branch) :)

PR #4312, opened 2016-05-04
[ticket/14625] Event core.viewtopic_modify_tpl_ary

tas2580:
[ticket/14625] (https://tracker.phpbb.com/browse/PHPBB3-14625)


rxu:
Those template vars could be simply overwritten by another ```assign_vars``` call in appropriate listener, so no need in special event in common.

tas2580:
Oh @rxu you are right!
I hadn't thought of that, thank you

rmcgirr83:
ticket should be closed

PR #4313, opened 2016-05-05
[ticket/14626] Event core.make_jumpbox_modify_tpl_ary

tas2580:
PHPBB3-14626

### DIFF
--- a/phpBB/includes/functions_content.php
+++ b/phpBB/includes/functions_content.php
@@ -243,8 +243,8 @@ function make_jumpbox($action, $forum_id = false, $select_all = false, $acl_list
 			'tpl_ary',
 		);
 		extract($phpbb_dispatcher->trigger_event('core.make_jumpbox_modify_tpl_ary', compact($vars)));
-		 
-		foreach($tpl_ary as $i => $ary)
+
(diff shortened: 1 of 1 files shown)


rmcgirr83:
You have to unlink the tracker url from the PHPBB3-14626 I believe

Crizz0:
You have to change the first commits message, too. Then the error will disappear.

phpBB/includes/functions_content.php
```diff
@@ -227,7 +228,27 @@ function make_jumpbox($action, $forum_id = false, $select_all = false, $acl_list
...
+		 * @var	array	tpl_ary			Template data of the forum
+		 * @since 3.1.10-RC1
+		 */
+		$vars = array(
+			'row',
+			'tpl_ary',
+		);
+		extract($phpbb_dispatcher->trigger_event('core.make_jumpbox_modify_tpl_ary', compact($vars)));
+
+		foreach ($tpl_ary as $i => $ary)
```
marc1706: This is not needed. You should instead use $template->assign_block_vars_array().

marc1706:
Thanks for your pull request. I've taken the liberty to try and fix the commit message issue. See #4352

PR #4317, opened 2016-05-08
[ticket/14610] Don't log an error if the Q&A captcha has been solved

rmcgirr83:
https://tracker.phpbb.com/browse/PHPBB3-14610

PHPBB3-14610

### DIFF
--- a/phpBB/phpbb/captcha/plugins/qa.php
+++ b/phpBB/phpbb/captcha/plugins/qa.php
@@ -228,7 +228,7 @@ class qa
 			if ($this->is_solved())
 			{
 				return;
-			}			
+			}
 			/** @var \phpbb\log\log_interface $phpbb_log */
 			$phpbb_log->add('critical', $user->data['user_id'], $user->ip, 'LOG_ERROR_CAPTCHA', time(), array($user->lang('CONFIRM_QUESTION_MISSING')));
 			return false;
--- a/phpBB/phpbb/captcha/plugins/qa.php
+++ b/phpBB/phpbb/captcha/plugins/qa.php
@@ -224,6 +224,11 @@ class qa
 
 		if ($this->is_solved() || empty($this->question_text) || !count($this->question_ids))
 		{
+			// don't log an error if the captcha has been solved
+			if ($this->is_solved())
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/captcha/plugins/qa.php
```diff
@@ -224,6 +224,11 @@ function get_template()
 
 		if ($this->is_solved() || empty($this->question_text) || !count($this->question_ids))
 		{
+			// don't log an error if the captcha has been solved
```
lavigor: Maybe it will be better to move this check into a separate if branch?
Result:
```
		if ($this->is_solved())
		{
			return;
		}
		else if (empty($this->question_text) || !count($this->question_ids))
		{
			/** @var \phpbb\log\log_interface $phpbb_log */
			$phpbb_log->add('critical', $user->data['user_id'], $user->ip, 'LOG_ERROR_CAPTCHA', time(), array($user->lang('CONFIRM_QUESTION_MISSING')));
			return false;
		}
		else
		{
			$template->assign_vars(array(
				'QA_CONFIRM_QUESTION'		=> $this->question_text,
				'QA_CONFIRM_ID'				=> $this->confirm_id,
				'S_CONFIRM_CODE'			=> true,
				'S_TYPE'					=> $this->type,
			));

			return 'captcha_qa.html';
		}
```

Lady-G: Could the return value be stated, which will help for readability? Otherwise, a NULL return value will be used.

```
if ($this->is_solved())
{
		return false;
}
```

lavigor: @Lady-G my answer above also contains `false` value. 😉

rmcgirr83:
@marc1706 @Nicofuma @CHItA any insight?

phpBB/phpbb/captcha/plugins/qa.php
marc1706: I'd definitely prefer the version proposed by @lavigor

Nicofuma:
looks good

PR #4318, opened 2016-05-13
[ticket/14638] New UCP hooks for subscriptions

wesleyfok:
Adds various event hooks to the Manage subscriptions page of the User
Control Panel. core.ucp_subscribed_post_data allows us to handle
additional post data from the form that unwatches subscriptions.
core.ucp_main_subscribed_forums_modify_query,
core.ucp_main_topiclist_count_modify_query and
core.ucp_main_topiclist_modify_query all modify the queries used to get
the list of subscribed forums and subscribed/bookmarked topics.
core.ucp_main_subscribed_forum_modify_template_vars and
core.ucp_main_topiclist_topic_modify_template_vars modify the template
variable array associated with each subscribed forum and topic row.

PHPBB3-14638


CHItA:
@phpbb/development-team do we accept multiple events in one PR? I guess that shouldn't be a problem, but not sure if there is a policy about this.

nickvergessen:
No, it's fine. The reason we advice against it is, that when one needs more fixing the other ones can not be merged already to avoid conflicts later on. But its okay to merge a PR with multiple ones

phpBB/includes/ucp/ucp_main.php
```diff
@@ -213,7 +213,21 @@ function main($id, $mode)
 
 				add_form_key('ucp_front_subscribed');
 
-				$unwatch = (isset($_POST['unwatch'])) ? true : false;
+				$post_data = $_POST;
```
marc1706: It is generally discouraged to use $_POST data. This is only used for checking if it's set. Newer code will use the request class' is_set() and is_set_post() methods. Blindly passing on $_POST data to an event is potentially very dangerous and shouldn't be done. Events should also be able to use the request class themselves.

phpBB/includes/ucp/ucp_main.php
```diff
@@ -299,6 +313,22 @@ function main($id, $mode)
 						$tracking_topics = ($tracking_topics) ? tracking_unserialize($tracking_topics) : array();
 					}
 
+					/**
+					 * Modify the query used to retrieve a list of subscribed forums
+					 *
+					 * @event core.ucp_main_subscribed_forums_modify_query
+					 * @var array	sql_array	       The subscribed forums query
+					 * @var array   forbidden_forums   The list of forbidden forums
+					 * @var user    user               The current user
```
marc1706: Classes like the user class shouldn't be passed with an event. They will be copied which causes some overhead that is really not needed at all. Instead, these classes should be passed to the events via the container definition.

phpBB/includes/ucp/ucp_main.php
```diff
@@ -664,6 +722,25 @@ function assign_topiclist($mode = 'subscribed', $forbidden_forum_ary = array())
...
 				AND ' . $db->sql_in_set('t.forum_id', $forbidden_forum_ary, true, true),
 		);
+
+		/**
+		 * Modify the query used to retrieve the count of subscribed/bookmarked topics
+		 *
+		 * @event core.ucp_main_topiclist_count_modify_query
+		 * @var array	sql_array	          The subscribed/bookmarked topics query
+		 * @var array   forbidden_forum_ary   The list of forbidden forums
+		 * @var user    user                  The current user
```
marc1706: See above comment.

phpBB/includes/ucp/ucp_main.php
```diff
@@ -732,6 +809,24 @@ function assign_topiclist($mode = 'subscribed', $forbidden_forum_ary = array())
 			$sql_array['SELECT'] .= ', tp.topic_posted';
 		}
 
+		/**
+		 * Modify the query used to retrieve the list of subscribed/bookmarked topics
+		 *
+		 * @event core.ucp_main_topiclist_modify_query
+		 * @var array	sql_array	          The subscribed/bookmarked topics query
+		 * @var array   forbidden_forum_ary   The list of forbidden forums
+		 * @var user    user                  The current user
```
marc1706: See above

marc1706:
Is there a specific reason why you added those events to 3.2.x and not to 3.1.x? This will result in them not being available in 3.1.x versions and I don't really see why this was done.

wesleyfok:
Only reason is because we're developing off 3.2.x right now, as it has several features we need for our forum. If it's necessary to backport these events to 3.1.x as well I can try to do so.

Nicofuma:
It shouldn't be an issue to backport the changes to 3.1.x as this part of the code didn't change. So please do it

wesleyfok:
Sure, I can do that. What's the best way to modify this PR/create a new one? Should I rebase this branch onto 3.1.x or open a new ticket and PR?

Nicofuma:
The best is to create a new PR and keep the current ticket.

PR #4321, opened 2016-05-16
[ticket/14615] Fix HTML5 validation errors on avatar deletion

Senky:
[PHPBB3-14615](https://tracker.phpbb.com/browse/PHPBB3-14615)

Empty is default. So why do we default it "0"? Giving it proper default value fixes validation errors as well.

### DIFF
--- a/phpBB/phpbb/avatar/driver/gravatar.php
+++ b/phpBB/phpbb/avatar/driver/gravatar.php
@@ -52,8 +52,8 @@ class gravatar extends \phpbb\avatar\driver\driver
 	public function prepare_form($request, $template, $user, $row, &$error)
 	{
 		$template->assign_vars(array(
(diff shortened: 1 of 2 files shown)


DavidIQ:
This removes the protection given by casting the input to an int and can lead to more issues elsewhere if the administrator puts something other than a number.  I think this should be taken care of via the two template files by checking if these have a value before setting them as the minvalue.

Senky:
I don't get it. Validation is done in `process_form` method, where I kept input cast to int. `prepare_form` serves only to display the values, so no protection is removed.

DavidIQ:
You're right.  I didn't look at the rest of the file.  My bad.  Why are we even using the request object here then?

Senky:
Good point! According to my inspection, it is used there to ensure that data sent by the user will be displayed again even if form contains errors.

DavidIQ:
Kind of weird we aren't doing the same for email...

Senky:
I agree. Email should be kept as well. However, this is not part of this ticket. Should I create a new ticket so that we do not forget about it?

PR #4322, opened 2016-05-17
[ticket/14571] Reword egosearch label

Senky:
[PHPBB3-14571](https://tracker.phpbb.com/browse/PHPBB3-14571)


VSEphpbb:
Wow that's horribly verbose!

marc1706:
It's better to keep things simple and straightforward. The proposed change is neither of those which is why I disagree with it.

PR #4323, opened 2016-05-17
[ticket/13994] Make attachments page description clearer

Senky:
[PHPBB3-13994](https://tracker.phpbb.com/browse/PHPBB3-13994)

### DIFF
--- a/phpBB/language/en/acp/attachments.php
+++ b/phpBB/language/en/acp/attachments.php
@@ -37,7 +37,7 @@ if (empty($lang) || !is_array($lang))
 // in a url you again do not need to specify an order e.g., 'Click %sHERE%s' is fine
 
 $lang = array_merge($lang, array(
-	'ACP_ATTACHMENT_SETTINGS_EXPLAIN'	=> 'Here you can configure the main settings for attachments and the associated special categories.',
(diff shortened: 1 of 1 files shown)


phpBB/language/en/acp/attachments.php
```diff
@@ -37,7 +37,7 @@
 // in a url you again do not need to specify an order e.g., 'Click %sHERE%s' is fine
 
 $lang = array_merge($lang, array(
-	'ACP_ATTACHMENT_SETTINGS_EXPLAIN'	=> 'Here you can configure the main settings for attachments and the associated special categories.',
+	'ACP_ATTACHMENT_SETTINGS_EXPLAIN'	=> 'Here you can configure the main settings for attachments. The associated special categories settings that control which file types may be attached are found on the "Postings" tab under "Attachments".',
```
marc1706: `... found in the "Postings" category beneath the "Attachments" heading.` ?
@VSEphpbb what do you think?

VSEphpbb: I think this change is wholly unnecessary myself (aside from the fact that the change is not good English). This module is already under the Posting tab. 

Just because this module is mirrored for convenience under the general tab, doesn't mean we should start writing "where to find stuff" under the headlines, otherwise this is just setting a precedent that would be warranted on most modules that appear under the general tab, as well as under their proper tabs.

A proper solution is a proper redesign of the ACP layout IMHO.

marc1706: After further looking at this, I do have to agree with @VSEphpbb. This does not help much but it something we should worry about while restructuring the ACP:

PR #4329, opened 2016-05-22
[ticket/14429] Add core.obtain_users_online_string_before_modify

Senky:
[PHPBB3-14429](https://tracker.phpbb.com/browse/PHPBB3-14429)

### DIFF
--- a/phpBB/includes/functions.php
+++ b/phpBB/includes/functions.php
@@ -4334,6 +4334,30 @@ function obtain_users_online_string($online_users, $item_id = 0, $item = 'forum'
 			}
 		}
 	}
+
+	/**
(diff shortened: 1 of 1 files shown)


phpBB/includes/functions.php
```diff
@@ -4334,6 +4334,30 @@ function obtain_users_online_string($online_users, $item_id = 0, $item = 'forum'
 			}
 		}
 	}
+
+	/**
+	* Modify online userlist data
+	*
+	* @event core.modify_users_online_string_modify
```
marc1706: The name seems off. Did you mean `core.obtain_users_online_string_modify`?

PR #4332, opened 2016-05-23
[ticket/11446] Use sql_in_set properly in phpbb_notification_manager

Senky:
[PHPBB3-11446](https://tracker.phpbb.com/browse/PHPBB3-11446)

### DIFF
--- a/phpBB/phpbb/notification/manager.php
+++ b/phpBB/phpbb/notification/manager.php
@@ -967,11 +967,16 @@ class manager
 	/**
 	* Get notification type ids (as an array)
 	*
-	* @param array $notification_type_names Array of strings
+	* @param string|array $notification_type_names Notification type names
 	* @return array Array of integers
 	*/
-	public function get_notification_type_ids(array $notification_type_names)
+	public function get_notification_type_ids($notification_type_names)
 	{
+		if (!is_array($notification_type_names))
+		{
+			$notification_type_names = array($notification_type_names);
+		}
+
 		$notification_type_ids = array();
 
 		foreach ($notification_type_names as $name)
--- a/phpBB/phpbb/notification/manager.php
+++ b/phpBB/phpbb/notification/manager.php
@@ -192,7 +192,7 @@ class manager
 			$sql = 'SELECT n.*, nt.notification_type_name
 				FROM ' . $this->notifications_table . ' n, ' . $this->notification_types_table . ' nt
 				WHERE n.user_id = ' . (int) $options['user_id'] .
-					(($options['notification_id']) ? ((is_array($options['notification_id'])) ? ' AND ' . $this->db->sql_in_set('n.notification_id', $options['notification_id']) : ' AND n.notification_id = ' . (int) $options['notification_id']) : '') . '
+					(($options['notification_id']) ? ' AND ' . $this->db->sql_in_set('n.notification_id', $options['notification_id']) : '') . '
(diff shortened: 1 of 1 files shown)


marc1706:
Closed and reopened to have travis run.

phpBB/phpbb/notification/manager.php
```diff
@@ -275,10 +275,9 @@ public function mark_notifications_read($notification_type_name, $item_id, $user
 		$sql = 'UPDATE ' . $this->notifications_table . "
 			SET notification_read = 1
 			WHERE notification_time <= " . (int) $time .
-				(($notification_type_name !== false) ? ' AND ' .
-					(is_array($notification_type_name) ? $this->db->sql_in_set('notification_type_id', $this->get_notification_type_ids($notification_type_name)) : 'notification_type_id = ' . $this->get_notification_type_id($notification_type_name)) : '') .
-				(($user_id !== false) ? ' AND ' . (is_array($user_id) ? $this->db->sql_in_set('user_id', $user_id) : 'user_id = ' . (int) $user_id) : '') .
-				(($item_id !== false) ? ' AND ' . (is_array($item_id) ? $this->db->sql_in_set('item_id', $item_id) : 'item_id = ' . (int) $item_id) : '');
+				(($notification_type_name !== false) ? ' AND ' . $this->db->sql_in_set('notification_type_id', $this->get_notification_type_ids($notification_type_name)) : '') .
```
marc1706: This line was actually correct as proven by the now failing tests.

phpBB/phpbb/notification/manager.php
```diff
@@ -297,10 +296,9 @@ public function mark_notifications_read_by_parent($notification_type_name, $item
 		$sql = 'UPDATE ' . $this->notifications_table . "
 			SET notification_read = 1
 			WHERE notification_time <= " . (int) $time .
-				(($notification_type_name !== false) ? ' AND ' .
-					(is_array($notification_type_name) ? $this->db->sql_in_set('notification_type_id', $this->get_notification_type_ids($notification_type_name)) : 'notification_type_id = ' . $this->get_notification_type_id($notification_type_name)) : '') .
-				(($item_parent_id !== false) ? ' AND ' . (is_array($item_parent_id) ? $this->db->sql_in_set('item_parent_id', $item_parent_id, false, true) : 'item_parent_id = ' . (int) $item_parent_id) : '') .
-				(($user_id !== false) ? ' AND ' . (is_array($user_id) ? $this->db->sql_in_set('user_id', $user_id) : 'user_id = ' . (int) $user_id) : '');
+				(($notification_type_name !== false) ? ' AND ' . $this->db->sql_in_set('notification_type_id', $this->get_notification_type_ids($notification_type_name)) : '') .
```
marc1706: Same issue as with above line.

phpBB/phpbb/notification/manager.php
Senky: I think we should consider rewriting `phpbb\notification\manager::get_notification_type_ids` method. Almost all phpBB methods accept both string/int parameter as well as an array. If parameter is not array, it will cast it to it. So I propose change to `get_notification_type_ids` method.

marc1706: Sure, shouldn't add any BC breaks either. Just simply remove the array constraint and do something like:
```
if (!is_array($notification_type_names))
{
    $notification_type_names = array($notification_type_names);
}
```

PR #4333, opened 2016-05-25
[ticket/14648] Fix bug where default notifications stop working if another setting is set.

markshawtoronto:
When a new user signs up, they have an email preference set but no board notification preference set. The default board preference should work. Also when a user checks the email box for any notification preference, the default board preference should still work.

PHPBB3-14648

### DIFF
--- a/phpBB/phpbb/notification/type/base.php
+++ b/phpBB/phpbb/notification/type/base.php
@@ -449,7 +449,7 @@ abstract class base implements \phpbb\notification\type\type_interface
 			return array();
 		}
 
-		$rowset = $resulting_user_ids = array();
+		$rowset = $output = array();
 
 		$sql = 'SELECT user_id, method, notify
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/notification/type/base.php
```diff
@@ -471,22 +469,47 @@ protected function check_user_notification_options($user_ids = false, $options =
...
+			}
+			if ($row['notify'])
+			{
+				$output[$row['user_id']][] = $row['method'];
+			}
 		}
 
 		$this->db->sql_freeresult($result);
 
+		$default_methods = $this->notification_manager->get_default_methods();
```
marc1706: Can you move the assignment to default_methods outside the while loop? This should be called exactly ones and not multiple times inside a loop.

marc1706:
Thanks for the PR

PR #4335, opened 2016-05-26
[ticket/14650] Move CONTRIBUTING.md and add pull request template

marc1706:
https://tracker.phpbb.com/browse/PHPBB3-14650


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,10 @@
+Checklist:
+
+- [ ] Correct branch: master for new features / 3.2.x, 3.1.x for fixes
```
VSEphpbb: what about `;` instead of `/` ?

Just because slash used below is used to group 3.2 and master, but here it separates 3.2 and master

marc1706: sure

PR #4338, opened 2016-05-29
Add viewforum_body_topicrow_before template event

Senky:
[PHPBB3-14469](https://tracker.phpbb.com/browse/PHPBB3-14469)

### DIFF
--- a/phpBB/docs/events.md
+++ b/phpBB/docs/events.md
@@ -2227,6 +2227,14 @@ viewforum_body_topic_row_prepend
 * Since: 3.1.7-RC1
 * Purpose: Add content at the end of the topic list item.
 
+viewforum_body_topicrow_row_before
+===
+* Locations:
+    + styles/prosilver/template/viewforum_body.html
+    + styles/subsilver2/template/viewforum_body.html
+* Since: 3.1.10-RC1
+* Purpose: Add content before list of topics.
+
 viewforum_buttons_bottom_before
 ===
 * Locations:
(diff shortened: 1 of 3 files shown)


phpBB/docs/events.md
```diff
@@ -2227,6 +2227,14 @@ viewforum_body_topic_row_prepend
 * Since: 3.1.7-RC1
 * Purpose: Add content at the end of the topic list item.
 
+viewforum_body_topicrow_before
```
marc1706: To be consistent with the other events regarding the topicrow, we should probably change it to `viewforum_body_topic`_row_before` and move it accordingly before the entries above this one to have a proper alphabetical order.

Elsensee:
events.md is in alphabetical order. Thus `viewforum_body_topicrow_before` needs to be moved up, so it's before `viewforum_body_topicrow_prepend`.

phpBB/docs/events.md
```diff
@@ -2227,6 +2227,14 @@ viewforum_body_topic_row_prepend
 * Since: 3.1.7-RC1
 * Purpose: Add content at the end of the topic list item.
 
+viewforum_body_topicrow_row_before
```
3D-I: Shouldn't it be **viewforum_body_topicrow_before** ¿?

Applies to all of this files.

3D-I:
I added a comment to the commit.

It should be **viewforum_body_topicrow_before** and not **viewforum_body_topicrow_row_before**, I guess.

PR #4339, opened 2016-05-29
Use HTTP GET instead of POST in topics sort

Senky:
[PHPBB3-13476](https://tracker.phpbb.com/browse/PHPBB3-13476)


marc1706:
We're using POST in this case in order to prevent cluttering up the URL with random settings. I don't get why we should change it to GET? You didn't elaborate on why you think this should be merged.

Senky:
I think it's not cluttering to show user where he really is. It's not like there is any readable information in the URLs nowadays...
Few extra characters that allow user to bookmark list of topics in order he wishes might be one of pretty good reasons to switch to GET method.

rmcgirr83:
How are these commits passing if they aren't following the guidelines? (eg, missing [ticket/13476] in the commit title.

marc1706:
@rmcgirr83 the commit actually contains that?

rmcgirr83:
Ah, didn't see that I assumed that the ticket number had to be in the actual commit message like I see with all the others.

marc1706:
It has to be and it is. Don't confuse the title of the PR and the commit message ;)

phpBB/styles/prosilver/template/viewforum_body.html
```diff
@@ -218,13 +218,14 @@ <h2 class="forum-title"><!-- EVENT viewforum_forum_name_prepend --><a href="{U_V
...
 <!-- IF S_SELECT_SORT_DAYS and not S_DISPLAY_ACTIVE -->
-	<form method="post" action="{S_FORUM_ACTION}">
+	<form method="get" action="{S_FORUM_ACTION}">
 		<fieldset class="display-options">
 	<!-- IF not S_IS_BOT -->
 			<label>{L_DISPLAY_TOPICS}{L_COLON} {S_SELECT_SORT_DAYS}</label>
 			<label>{L_SORT_BY} {S_SELECT_SORT_KEY}</label>
 			<label>{S_SELECT_SORT_DIR}</label>
-			<input type="submit" name="sort" value="{L_GO}" class="button2" />
+			<input type="hidden" name="f" value="{FORUM_ID}" />
```
Elsensee: Why that? `S_FORUM_ACTION` already includes the forum id. See here: https://github.com/phpbb/phpbb/blob/d0a5f87a44d4a73a740fd5242bcd11b059c7369b/phpBB/viewforum.php#L390

Senky: `<form method="GET">` ignores parameters in it's `action`. Thus, they need to be moved to hidden inputs.

marc1706:
@Senky in reply to your note that the URLs aren't readable anyway. We will try to move to controllers for 3.3. That will result in a lot less cluttered URL. I don't see a reason to add this clutter here really. I have not encountered people using direct URLs to sort topics on viewforum so I don't see a reason to add this.

Senky:
Alright, if there will be controllers used since 3.3, we can forget about this :)

PR #4340, opened 2016-05-29
[ticket/10961] Send HTTP 403 when applicable

Senky:
[PHPBB3-10961](https://tracker.phpbb.com/browse/PHPBB3-10961)

### DIFF
--- a/phpBB/adm/index.php
+++ b/phpBB/adm/index.php
@@ -41,6 +41,7 @@ if (!isset($user->data['session_admin']) || !$user->data['session_admin'])
 // check specific permissions but this is a catchall
 if (!$auth->acl_get('a_'))
 {
+	send_status_line(403, 'Forbidden');
 	trigger_error('NO_ADMIN');
 }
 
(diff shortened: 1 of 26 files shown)


phpBB/includes/acp/acp_board.php
```diff
@@ -623,6 +623,7 @@ function main($id, $mode)
 				}
 				else
 				{
+					send_status_line(403, 'Forbidden');
```
marc1706: `E_USER_ERROR` will result in a 503 being sent. This will mean we'll have duplicate Status header fields in this case.

marc1706:
I'm also very unsure about this change since we're trying to move away from using trigger_error() anyway.
As this will also affect 3.2.x, what do you think @Nicofuma ?

Nicofuma:
It should be fine for 3.2 but it will indeed affect my existing PRs against master.
Not sure what is the best here...

Nicofuma:
I think it should be fine (will just makes the rebase more complicated :p)

marc1706:
Ok, I've moved this to a 3.2.x target.

PR #4341, opened 2016-05-29
[ticket/14655] Change CTRL to Ctrl

Senky:
[PHPBB3-14655](https://tracker.phpbb.com/browse/PHPBB3-14655)


Mauron:
Why?

Senky:
Because of consistency with other language strings in phpBB and, well, whole world.

marc1706:
I wouldn't propose changing this as it's trivial and for example Windows has only changed to also say Ctrl a while ago. Then again, it's a trivial change and I really don't care so might as well just merge it.

Mauron:
For consistency sake I'd like the _control_ button to be showed as **ctrl** in lowercase, since my keyboard shows it like that.

VSEphpbb:
Well my keyboard shows `control` so trying to show what a keyboard has is not helpful.

It needs to be clear this is referring to a keyboard command, and so IMO, the all caps was better at that, as it stands out better from standard English text.

Mauron:
My point was that there's no consistency issue in uppercase _CTRL_.

marc1706:
Ok, this is already taking too much time.
@Senky thank you for the PR but I'll close it and the ticket as a "Won't fix". This isn't fixing or improving anything at really.

PR #4347, opened 2016-06-06
[ticket/14631] Load truncate_string() if needed

Senky:
[PHPBB3-14631](https://tracker.phpbb.com/browse/PHPBB3-14631)

### DIFF
--- a/phpBB/phpbb/content_visibility.php
+++ b/phpBB/phpbb/content_visibility.php
@@ -417,6 +417,11 @@ class content_visibility
 			return array();
 		}
 
+		if (!function_exists('truncate_string'))
+		{
+			include($this->phpbb_root_path . 'includes/functions_content.' . $this->php_ext);
+		}
+
 		$data = array(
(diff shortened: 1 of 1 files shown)


rmcgirr83:
functions_content is loaded on every page per the code in common.php

`require($phpbb_root_path . 'includes/functions_content.' . $phpEx);`

Further `$phpbb_root_path` and `$phpEx` are not defined within the file.  Should be `$this->phpbb_root_path` and `$this->php_ext`

Senky:
Apparently, it is not loaded during update process via `phpbbcli`.

It's also really strange, that all tests passed. There is one missing to cover this scenario.

rxu:
@Senky It was discussed before in https://github.com/phpbb/phpbb/pull/4090.

Senky:
Alright, I thought about quickfix for this release. It could be rewritten to proper solution, no problem. But I have a question. Everyone talks about "proper helper class", but none of existing seem to be suitable. A new one, say `content_helper` could be created?

PR #4355, opened 2016-06-16
[ticket/14678] Forum title hidden on viewforum without f_list permission

markshawtoronto:
If a user has f_read permission, but doesn't have f_list permission, they can read a forum but not see it in lists.
However when they read a forum, the title and navigation is hidden. I think they should be able to read the forum title.

More info here: https://tracker.phpbb.com/browse/PHPBB3-14678

### DIFF
--- a/phpBB/includes/functions_display.php
+++ b/phpBB/includes/functions_display.php
@@ -754,7 +754,7 @@ function generate_forum_nav(&$forum_data_ary)
 	global $template, $auth, $config;
 	global $phpEx, $phpbb_root_path, $phpbb_dispatcher;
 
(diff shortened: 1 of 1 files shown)


marc1706:
Does this only affect 3.2.x but also 3.1.x?

markshawtoronto:
@marc1706 It affects both, where should I make the PR?

marc1706:
Changes that affect both 3.1.x and 3.2.x should be against the 3.1.x branch. The changes will then be merged upstream through the branches (3.1.x --> 3.2.x --> master).

markshawtoronto:
@marc1706 - OK I'm redoing the PR: https://github.com/phpbb/phpbb/pull/4359

Hope that's better!

PR #4358, opened 2016-06-19
[ticket/14683] Fix typos in email templates.

rxu:
Checklist:

- [x] Tests pass

<a href="https://tracker.phpbb.com/browse/PHPBB3-14683">PHPBB3-14683</a>.

### DIFF
--- a/phpBB/language/en/email/forum_notify.txt
+++ b/phpBB/language/en/email/forum_notify.txt
@@ -2,7 +2,7 @@ Subject: Forum post notification - "{FORUM_NAME}"
 
 Hello {USERNAME},
 
-You are receiving this notification because you are watching the forum "{FORUM_NAME}" at "{SITENAME}". This forum has received a new reply to the topic "{TOPIC_TITLE}"<!-- IF AUTHOR_NAME !== '' --> by {AUTHOR_NAME}<!-- ENDIF --> since your last visit. You can use the following link to view the last unread reply, no more notifications will be sent until you visit the topic.
+You are receiving this notification because you are watching the forum "{FORUM_NAME}" at "{SITENAME}". This forum has received a new reply to the topic "{TOPIC_TITLE}"<!-- IF AUTHOR_NAME != '' --> by {AUTHOR_NAME}<!-- ENDIF --> since your last visit. You can use the following link to view the last unread reply, no more notifications will be sent until you visit the topic.
 
 {U_NEWEST_POST}
 
--- a/phpBB/language/en/email/newtopic_notify.txt
+++ b/phpBB/language/en/email/newtopic_notify.txt
@@ -2,7 +2,7 @@ Subject: New topic notification - "{FORUM_NAME}"
 
 Hello {USERNAME},
 
-You are receiving this notification because you are watching the forum "{FORUM_NAME}" at "{SITENAME}". This forum has received a new topic<!-- IF AUTHOR_NAME !== '' --> by {AUTHOR_NAME}<!-- ENDIF --> since your last visit, "{TOPIC_TITLE}". You can use the following link to view the forum, no more notifications will be sent until you visit the forum.
+You are receiving this notification because you are watching the forum "{FORUM_NAME}" at "{SITENAME}". This forum has received a new topic<!-- IF AUTHOR_NAME != '' --> by {AUTHOR_NAME}<!-- ENDIF --> since your last visit, "{TOPIC_TITLE}". You can use the following link to view the forum, no more notifications will be sent until you visit the forum.
 
 {U_FORUM}
 
--- a/phpBB/language/en/email/short/newtopic_notify.txt
+++ b/phpBB/language/en/email/short/newtopic_notify.txt
@@ -2,7 +2,7 @@ Subject: New topic notification - "{FORUM_NAME}"
 
 Hello {USERNAME},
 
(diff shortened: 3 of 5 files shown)


lavigor:
@rxu, is it a reverse for https://github.com/phpbb/phpbb/pull/1685?
I mean, has it been tested for the username `0`?

phpBB adds the support for `===` and `!==` [here](https://github.com/phpbb/phpbb/tree/3.1.x/phpBB/phpbb/template/twig/node/expression/binary) - there are no typos here.

BTW Twig has `sameas` comparison that [is the equivalent to === in PHP](http://twig.sensiolabs.org/doc/tests/sameas.html).

rxu:
@lavigor I didn't see that PR before, but with ```!==``` comparison the test throws error
> phpbb_email_parsing_test::test_email_parsing
Twig_Error_Syntax: Unexpected token "operator" of value "=" in "newtopic_notify.txt" at line 5

which can be fixed by replacing ```!==``` with ```!=```. I have no idea for now why phpBB parser doesn't help that.

rxu:
@marc1706 ^

lavigor:
@rxu, maybe the test should be updated then?
Because according to the documentation, Twig supports custom operators in extensions.
That is the way used by phpBB [here](https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/phpbb/template/twig/extension.php#L113).

rxu:
Then that has probably something to do with lexer in #4357 again rather than with test.

lavigor:
@rxu, no, it is not related to lexer, because the extension is defined [here](https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/phpbb/template/twig/twig.php#L111-L116).

BTW where is `phpbb_email_parsing_test` defined?
I could not find it in the `tests` directory.

rxu:
See the PR I referred sbove.

lavigor:
Oh, sorry.
But the phpBB extension is not added there indeed.

rxu:
Isn't it added on the twig class initializing?

lavigor:
Yes, the method `$twig->addExtension` is not fired.
The extension is just being added to the phpBB container.

I haven't tested that, but probably we can try something like that:
```php
		$twig_extension = new \phpbb\template\twig\extension($context, $user);
		$context = new \phpbb\template\context();
		$phpbb_container->set('template.twig.extensions.collection', array($twig_extension), phpbb_mock_container_builder::SCOPE_PROTOTYPE);

		$twig = new \phpbb\template\twig\environment(
			$config,
			$filesystem,
			$phpbb_path_helper,
			$cache_path,
			null,
			new \phpbb\template\twig\loader($filesystem, ''),
			array(
				'cache'			=> false,
				'debug'			=> false,
				'auto_reload'	=> true,
				'autoescape'	=> false,
			)
		);
		$this->twig->addExtension($twig_extension);
```

rxu:
The point is that the test uses core methods to parse the email template and doesn't do it on its own. Thus, if the test fails, that means the core template parsing will fail and should be fixed, not the test. Do you get the point? :wink:

lavigor:
@rxu, core extension is just ignored in the test => test fails.
But it does not mean that the extension is not supported.
It is still a part of the phpBB core.

That's why those templates are parsed OK in the core (`!==` was not the reason of the problem in 3.2) but not in the test because those core methods do not know about the non-added extension in the test.

That's why it is worth trying the code above I think.

rxu:
I'm not talking about twig extensions but about the messenger class.

lavigor:
Messenger class does not know about Twig extensions because initializing them is outside of its responsibility.

rxu:
Please see its ```setup_template()``` method in 3.2.x branch.

lavigor:
@rxu, but `lexer` uses the environment initialized in the test, not in the messenger.
See `vendor\twig\twig\lib\Twig\Lexer.php`:
```php
    protected function getOperatorRegex()
    {
        $operators = array_merge(
            array('='),
            array_keys($this->env->getUnaryOperators()),
            array_keys($this->env->getBinaryOperators())
        );
```
The environment is being defined [here](https://github.com/phpbb/phpbb/pull/4357/files#diff-9c7e3c7cdaff8095daa6f7ae2c11c53aR92) - it does not have the extension initialized as I have described above.

So you were right about the lexer, but the problem comes from the test, not from the core code.

rxu:
Well, extensions collection is set several lines above here https://github.com/phpbb/phpbb/pull/4357/files#diff-9c7e3c7cdaff8095daa6f7ae2c11c53aR76

lavigor:
@rxu, that collection is not used anywhere before you send `$twig` to lexer, or after that.
It is used in messenger for its own environment, that is not passed to lexer.

Maybe I can create another PR to fix that?

rxu:
> Maybe I can create another PR to fix that?

Feel free to.

lavigor:
@rxu, hmm... As for the test itself, `!= ''` is also OK for `0` username.
But maybe it should be tested however.
I've just improved the test here: #4372.

Should we leave `!=` as it is now then?

rxu:
Well, currently I can't see why the result of ```!= ''``` should be different from ```!== ''```. Does the type comparison matter here?

lavigor:
Probably all AUTHOR_NAME assignments are strings - then it is OK.

rxu:
I'm sure username should always be string.

PR #4372, opened 2016-06-26
[ticket/14696] Fix email template test for '0' username

lavigor:
Checklist:

- [x] Tests pass

[PHPBB3-14696](https://tracker.phpbb.com/browse/PHPBB3-14696)

This PR adds the test for `0` username fixed in #1685.
See #4358 for more information.

### DIFF
--- a/tests/email/email_parsing_test.php
+++ b/tests/email/email_parsing_test.php
@@ -19,7 +19,7 @@ class phpbb_email_parsing_test extends phpbb_test_case
 
 	public function setUp()
 	{
-		global $phpbb_container, $config, $phpbb_root_path, $phpEx, $request, $user; 
+		global $phpbb_container, $config, $phpbb_root_path, $phpEx, $request, $user;
 
 		$phpbb_container = new phpbb_mock_container_builder;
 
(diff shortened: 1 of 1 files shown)


tests/email/email_parsing_test.php
```diff
@@ -73,7 +73,8 @@ public function setUp()
 		$phpbb_container->set('ext.manager', $extension_manager, phpbb_mock_container_builder::SCOPE_PROTOTYPE);
 
 		$context = new \phpbb\template\context();
-		$phpbb_container->set('template.twig.extensions.collection', array(new \phpbb\template\twig\extension($context, $user)), phpbb_mock_container_builder::SCOPE_PROTOTYPE);
+		$twig_extension = new \phpbb\template\twig\extension($context, $user);
```
rxu: Actually, this part is wrong here (as well as in the previous PR) as the 2nd parameter should be ```$lang``` for 3.2.

lavigor: @rxu yes, just left it as is.
`setParameter` also accepts 2 arguments instead of 3.
But tests pass OK.

Should I fix these issues too?

rxu:
Well, tests will pass while the ```lang``` method is not involved, but as far as othrt users may use this test as an example, I guess it would be better to fix it.

lavigor:
@rxu this test creates a new `cache` directory on one level with `build`, `phpBB` etc.
Maybe we can use `$phpbb_root_path.'cache/test/twig'` instead?

rxu:
Yes please, I guess it should use the same cache folder as all other tests.

rxu:
For me, test fails locally with errors like 
>Failed asserting that 'Subject: New topic notification - "Any forum"

[skip]
>contains "Author username".

or

>contains "0".

marc1706:
Thanks :+1:

PR #4374, opened 2016-06-27
[ticket/14616] Fixed auto-prune failing on large forums

TheKigen:
[PHPBB3-14616](https://tracker.phpbb.com/browse/PHPBB3-14616)

### DIFF
--- a/phpBB/includes/functions_admin.php
+++ b/phpBB/includes/functions_admin.php
@@ -2573,11 +2573,6 @@ function prune($forum_id, $prune_mode, $prune_date, $prune_flags = 0, $auto_sync
 		$sql_and .= ' AND topic_status = ' . ITEM_MOVED . " AND topic_last_post_time < $prune_date";
 	}
 
-	if ($prune_limit > 0)
-	{
-		$sql_and .= " LIMIT $prune_limit";
-	}
-
 	/**
 	* Use this event to modify the SQL that selects topics to be pruned
 	*
@@ -2608,7 +2603,7 @@ function prune($forum_id, $prune_mode, $prune_date, $prune_flags = 0, $auto_sync
 		WHERE ' . $db->sql_in_set('forum_id', $forum_id) . "
 			AND poll_start = 0
 			$sql_and";
-	$result = $db->sql_query($sql);
+	$result = $db->sql_query_limit($sql, $prune_limit);
 
 	$topic_list = array();
 	while ($row = $db->sql_fetchrow($result))
@@ -2625,7 +2620,7 @@ function prune($forum_id, $prune_mode, $prune_date, $prune_flags = 0, $auto_sync
 				AND poll_start > 0
 				AND poll_last_vote < $prune_date
 				$sql_and";
-		$result = $db->sql_query($sql);
+		$result = $db->sql_query_limit($sql, $prune_limit);
 
 		while ($row = $db->sql_fetchrow($result))
 		{
--- a/phpBB/includes/functions_admin.php
+++ b/phpBB/includes/functions_admin.php
@@ -2588,9 +2588,19 @@ function prune($forum_id, $prune_mode, $prune_date, $prune_flags = 0, $auto_sync
 	* @var int		prune_flags		The prune flags
 	* @var bool		auto_sync		Whether or not to perform auto sync
 	* @var string	sql_and			SQL text appended to where clause
+	* @var int		prune_limit		The prune limit
 	* @since 3.1.3-RC1
+	* @changed 3.1.10-RC1			Added prune_limit
 	*/
-	$vars = array('forum_id', 'prune_mode', 'prune_date', 'prune_flags', 'auto_sync', 'sql_and');
+	$vars = array(
+		'forum_id',
+		'prune_mode',
+		'prune_date',
+		'prune_flags',
+		'auto_sync',
+		'sql_and',
+		'prune_limit'
+	);
 	extract($phpbb_dispatcher->trigger_event('core.prune_sql', compact($vars)));
 
 	$sql = 'SELECT topic_id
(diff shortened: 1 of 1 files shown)


TheKigen:
Continuation of https://github.com/phpbb/phpbb/pull/4304 with fixed commit messages.

Apologies.

TheKigen:
Fixed the misplaced tab.

marc1706:
Closed and reopened to trigger travis.

phpBB/includes/functions_admin.php
```diff
@@ -2573,6 +2573,11 @@ function prune($forum_id, $prune_mode, $prune_date, $prune_flags = 0, $auto_sync
 		$sql_and .= ' AND topic_status = ' . ITEM_MOVED . " AND topic_last_post_time < $prune_date";
 	}
 
+	if ($prune_limit > 0)
+	{
+		$sql_and .= " LIMIT $prune_limit";
```
marc1706: Instead of manually adding the `LIMIT`, we do actually have a separate call `$db->sql_query_limit()` that would be better used here.
You could also additionally add `$prune_limit` to the below event if you want to. That would also require a `@changed 3.1.10-RC1 Added prune_limit to event`

TheKigen: I wished to avoid unnecessarily adding LIMIT if its not needed.  The current implementation for `$db->sql_query_limit()` for MySQL always adds LIMIT, even if LIMIT is 0.  If LIMIT is passed as 0 it changes $total to 18446744073709551615.

As far as adding $prune_limit to the event, I do believe it should be done.  I just am unclear on all the procedures in phpBB for altering events.

marc1706: I don't think we'll approach a prune limit this high so I do still think that using sql_query_limit() should be ok here and also be the better way. (better as in: the DBAL should handle the limit. some DBALs, maybe even future supported types, might handle the limit differently).
I'll try to write something about the events when I'm back home. ;)

marc1706: Updating the event should be fairly straightforward. Just add another line with
`@var int prune_limit The prune limit` and also add a `@changed 3.1.10-RC1 Added prune_limit` after the line with `@since`. Then you'll also have to add `'prune_limit'` to the `$vars` array and probably change it to be in multi-line notation due to the amount of variables like this:
```
$vars = array(
    'forum_id',
    'prune_mode',
    'prune_date',
    'prune_flags',
    'auto_sync',
    'sql_and',
    'prune_limit'
);
```
And that should be it.

TheKigen: What I'm talking about is the current implementation of sql_query_limit().

In `phpBB\phpbb\db\driver\mysql_base.php`
```
	/**
	* Build LIMIT query
	*/
	function _sql_query_limit($query, $total, $offset = 0, $cache_ttl = 0)
	{
		$this->query_result = false;

		// if $total is set to 0 we do not want to limit the number of rows
		if ($total == 0)
		{
			// MySQL 4.1+ no longer supports -1 in limit queries
			$total = '18446744073709551615';
		}

		$query .= "\n LIMIT " . ((!empty($offset)) ? $offset . ', ' . $total : $total);

		return $this->sql_query($query, $cache_ttl);
	}
```

I do not think this is an ideal implementation.  But I don't know if they had a valid reason for always adding the LIMIT to the query.  Because in my implementation of the limit, $prune_limit can be 0 to specify that it should attempt to prune everything in one transaction.

If you still feel I should use sql_query_limit(), I shall.  Because I understand where you are coming from for different DBs.  And it maybe better to address the unneeded addition of LIMIT in sql_query_limit() instead.

TheKigen:
I am quote confused on what Travis is complaining about.

Crizz0:
He says, that you introduce new variables, like: $sql_and and $prune_limit but they are invalid.

phpBB/includes/functions_admin.php
```diff
@@ -2583,9 +2588,19 @@ function prune($forum_id, $prune_mode, $prune_date, $prune_flags = 0, $auto_sync
...
 	*/
-	$vars = array('forum_id', 'prune_mode', 'prune_date', 'prune_flags', 'auto_sync', 'sql_and');
+	$vars = array(
+		'forum_id',
+		'prune_mode',
+		'prune_date',
+		'prune_flags',
+		'auto_sync',
+		'sql_and',
+		'prune_limit'
```
marc1706: Try if travis is happy if you add a comma after `'prune_limit'`

TheKigen:
Yeah.  It was simply the comma after 'prune_limit' in the array.  The test scripts expect it.  I was thinking it was when I viewed other event code I noticed the comma on the ending element in the array.  PHP doesn't care if the comma is there.  And in on my own code I never have a comma on the ending element.

phpBB/includes/functions_admin.php
marc1706: I'd rather fix this correctly in the mysql_base class and use the sql_limit() method here. Would you be so kind and create a ticket for that in the tracker as well?

TheKigen: Ok.

I'll try to get to it after I figure out another Cron bug.  This one seems unrelated to this patch at all as it appears to just crash before anything happens.  Its on queue_interval, but I disabled queue via blanking the functions in cron.  And since this is a fatal error its definitely a problem.

PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 72 bytes) in /*****/forums/phpbb/db/driver/driver.php on line 567

TheKigen: Figured out the other error.  Its with phpBB Fulltext Native search.  I'll probably explore properly fixing that later.  For now emptying the search index worked.

Testing the change to sql_query_limit().

PR #4377, opened 2016-07-02
[ticket/14703] Fix parent module selection for custom extension modules

rxu:
The PR is for 3.1.x, 3.2.x and master.

Checklist:

- [x] Tests pass

<a href="https://tracker.phpbb.com/browse/PHPBB3-14703">PHPBB3-14703</a>.

### DIFF
--- a/phpBB/phpbb/db/migration/tool/module.php
+++ b/phpBB/phpbb/db/migration/tool/module.php
@@ -90,7 +90,7 @@ class module implements \phpbb\db\migration\tool\tool_interface
 		$parent_sql = '';
 		if ($parent !== false)
 		{
-			$parent = $this->get_parent_module_id($parent);
+			$parent = $this->get_parent_module_id($parent, $module);
 			$parent_sql = 'AND parent_id = ' . (int) $parent;
 		}
 
@@ -336,7 +336,7 @@ class module implements \phpbb\db\migration\tool\tool_interface
 			$parent_sql = '';
 			if ($parent !== false)
 			{
-				$parent = $this->get_parent_module_id($parent);
+				$parent = $this->get_parent_module_id($parent, $module);
 				$parent_sql = 'AND parent_id = ' . (int) $parent;
 			}
 
@@ -463,14 +463,20 @@ class module implements \phpbb\db\migration\tool\tool_interface
 	* Get parent module id
 	*
 	* @param string|int $parent_id The parent module_id|module_langname
-	* @param array $data The module data array
+	* @param int|string|array $data The module_id, module_langname for existance checking or module data array for adding
 	* @return int The parent module_id
 	* @throws \phpbb\db\migration\exception
 	*/
-	public function get_parent_module_id($parent_id, $data = array())
+	public function get_parent_module_id($parent_id, $data = '')
 	{
 		// Allow '' to be sent as 0
 		$parent_id = $parent_id ?: 0;
+
+		// If automatic adding is in action, convert array back to string to simplify things
+		if (is_array($data) && sizeof($data) == 1)
+		{
+			$data = $data['module_langname'];
+		}
 
 		if (!is_numeric($parent_id))
 		{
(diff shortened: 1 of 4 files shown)


phpBB/phpbb/db/migration/tool/module.php
```diff
@@ -210,12 +210,15 @@ public function add($class, $parent = 0, $data = array())
...
 		// The "manual" way
+		// More than 1 module with the same module_basename may exist
+		// Thus use empty module_basename to select a category as a parent
 		if (!is_numeric($parent))
 		{
 			$sql = 'SELECT module_id
 				FROM ' . $this->modules_table . "
 				WHERE module_langname = '" . $this->db->sql_escape($parent) . "'
-					AND module_class = '" . $this->db->sql_escape($class) . "'";
+					AND module_class = '" . $this->db->sql_escape($class) . "';
```
VSEphpbb: semi colon at end should be concatenating

VSEphpbb:
Not sure this will work. for example, ACP_ATTACHMENT_SETTINGS appears twice and both times have a module_basename with data.

rxu:
ACP_ATTACHMENTS is the parent category, ACP_ATTACHMENT_SETTINGS is an actual module and shouldn't be parent whenever as far as I see it.

VSEphpbb:
I'm just saying I'm not 100% sure that the desired target module will always have an empty basename, which this change might be forcing to be the case.

rxu:
Well, this is indeed the case for UCP_PM category which has basename set (```ucp_pm```).
Then we need a method to extract categories from the modules table, but it's not clear how categories differ from real modules other than by basenames.

rxu:
Ok. For the start, the 1st-level categories have parent_id of 0 which makes it easier to select. Thus, to get the 2nd level cats, we could select ones which parents has parent_id of 0.

rxu:
!set WIP

rxu:
!unset WIP

rxu:
Updated. For the reference, tests already exist here https://github.com/phpbb/phpbb/blob/3.1.x/tests/dbal/migrator_tool_module_test.php

rxu:
!set WIP

marc1706:
Is the WIP label still correct?

rxu:
Yes, I'm planning to implement getting parent id when before/after module parameters are set.

marc1706:
Ok, thanks for the update.

rxu:
Finally it works.
The patch may seem hackish, but as long as the ```module_langname``` is not unigue, workarounding is the only way I see here.

rxu:
!unset WIP

tests/dbal/migrator_tool_module_test.php
```diff
@@ -118,6 +118,40 @@ public function test_add()
...
+		{
+			$this->tool->add('acp', 'ACP_FORUM_BASED_PERMISSIONS', array(
+				'module_basename'	=> 'acp_new_permissions_module',
+				'module_langname'	=> 'ACP_NEW_PERMISSIONS_MODULE',
+				'module_mode'		=> 'test',
+				'module_auth'		=> '',
+			));
+			$this->fail('Exception not thrown');
+		}
+		catch (Exception $e) {}
```
marc1706: You should probably use setExpectedException() to test this

marc1706: Please do check the actual contents of the exception (the message for example) though

phpBB/phpbb/db/migration/tool/module.php
```diff
@@ -487,4 +432,110 @@ protected function get_module_info($class, $basename)
...
+							WHERE module_langname = '" . $this->db->sql_escape($neighbour_module_langname) . "'
+								AND " . $this->db->sql_in_set('parent_id', $ids);
+						$result = $this->db->sql_query($sql);
+						$parent_id = (int) $this->db->sql_fetchfield('parent_id');
+						if (!$parent_id)
+						{
+							throw new \phpbb\db\migration\exception('PARENT_MODULE_FIND_ERROR', $data['parent_id']);
+						}
+					}
+					else if (!empty($data) && !is_array($data))
```
marc1706: it seems like this part wasn't covered in tests?

tests/dbal/migrator_tool_module_test.php
rxu: I have no idea how to do it.

marc1706: You should be able to check the contents and type of the exception, right?

phpBB/phpbb/db/migration/tool/module.php
marc1706: Bump.

tests/dbal/migrator_tool_module_test.php
rxu: Is it something like
```php
catch (Exception $e) {
	$this->assertInstanceOf('exception', get_class($e));
	$this->assertEquals('PARENT_MODULE_FIND_ERROR', $e->getMessage());
}
```

phpBB/phpbb/db/migration/tool/module.php
rxu: It should be covered by ```test_exists()``` as this part of code is involved for a module existance checking.

rxu:
Updated #4377.

Nicofuma:
LGTM @marc1706

PR #4405, opened 2016-08-09
[ticket/14733] Support increasing hashing cost factor

marc1706:
This will need another pull request for 3.2.x.

Checklist:
- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-14733

PHPBB3-14733

### DIFF
--- a/phpBB/phpbb/passwords/driver/base.php
+++ b/phpBB/phpbb/passwords/driver/base.php
@@ -13,7 +13,7 @@
 
 namespace phpbb\passwords\driver;
 
-abstract class base implements driver_interface
+abstract class base implements rehashable_driver_interface
 {
 	/** @var \phpbb\config\config */
 	protected $config;
@@ -21,7 +21,7 @@ abstract class base implements driver_interface
 	/** @var \phpbb\passwords\driver\helper */
 	protected $helper;
 
-	/** @var driver name */
+	/** @var string Driver name */
 	protected $name;
 
 	/**
--- a/phpBB/phpbb/passwords/driver/base.php
+++ b/phpBB/phpbb/passwords/driver/base.php
@@ -53,6 +53,14 @@ abstract class base implements driver_interface
 	}
 
 	/**
+	 * {@inheritdoc}
+	 */
+	public function needs_rehash($hash)
+	{
+		return false;
+	}
+
+	/**
 	* {@inheritdoc}
 	*/
 	public function get_settings_only($hash, $full = false)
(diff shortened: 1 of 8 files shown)


phpBB/phpbb/passwords/driver/driver_interface.php
```diff
@@ -30,6 +30,14 @@ public function is_supported();
 	public function is_legacy();
 
 	/**
+	 * Check if password needs to be rehashed
+	 *
+	 * @param string $hash Hash to check for rehash
+	 * @return bool True if password needs to be rehashed, false if not
+	 */
+	public function needs_rehash($hash);
```
bantu: Breaks other existing interface implementations, thus can not go into 3.1.x?

bantu:
This can not go into 3.1.x because of https://github.com/phpbb/phpbb/pull/4405#discussion_r75744572. Additionally, it should not go into a stable branch because it is basically a new feature (with the usual possibility of regression/breaks).

phpBB/phpbb/passwords/driver/driver_interface.php
marc1706: As all drivers are extending the base class this shouldn't cause any BC breaks.

marc1706:
I'm ok with moving this to 3.2.x as target. It shouldn't cause any BC breaks as it's using the abstract base class for an implementation of the de-facto state of the current convert flag. I do however understand that this is basically a new feature so I guess it's fair to have this moved to a 3.2.x target.

@Nicofuma agree?

Nicofuma:
:+1: for 3.2.1 or 3.3.0 (we really need to define what we accept in each kind of release)

bantu:
@marc1706 An extension might implement driver_interface, no?

marc1706:
@bantu I guess you'd say 3.3.0 target in this case, i.e. master?

bantu:
@marc1706 Yes, master, when 3.2.0 is already out.

marc1706:
Rebased on top of master branch.

marc1706:
@Nicofuma if you think this can go into 3.2 I'll rebase this on top of 3.2

Nicofuma:
can be 3.2.1 as the BC issue is solved

marc1706:
Updated base branch to 3.2.x

phpBB/phpbb/passwords/driver/base.php
```diff
@@ -13,15 +13,15 @@
 
 namespace phpbb\passwords\driver;
 
-abstract class base implements driver_interface
+abstract class base implements rehashable_driver_interface
```
Nicofuma: Why implementing it by default?

marc1706: To be able to increase hashing cost factor by default. Might be used for other and future hashing types, too.

PR #4409, opened 2016-08-10
[ticket/14742] Improvements to migrator

Elsensee:
Schema step will be "paused" now after every update step, so the code, that's running the migrator has better control over how much time the update takes.

Checklist:

- [x] Tests pass

Tracker ticket:

### DIFF
--- a/phpBB/phpbb/db/migration/helper.php
+++ b/phpBB/phpbb/db/migration/helper.php
@@ -81,4 +81,36 @@ class helper
 
 		return $steps;
 	}
+
+	/**
+	 * Reverse the update steps from an array of data changes
+	 *
+	 * 'If' statements and custom methods will be skipped, for all
+	 * other calls the reverse method of the tool class will be called
+	 *
+	 * @param array $steps Update changes from migration
+	 *
+	 * @return array
+	 */
+	public function reverse_update_data($steps)
+	{
+		$reversed_array = array();
+
+		foreach ($steps as $step)
+		{
+			$parts = explode('.', $step[0]);
+			$parameters = $step[1];
+
+			$class = $parts[0];
+			$method = isset($parts[1]) ? $parts[1] : false;
+
+			if ($class !== 'if' && $class !== 'custom')
+			{
+				array_unshift($parameters, $method);
+				$reversed_array[] = array($class . '.reverse', $parameters);
+			}
+		}
+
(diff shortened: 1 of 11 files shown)


phpBB/phpbb/db/migration/helper.php
```diff
@@ -81,4 +81,36 @@ public function get_schema_steps($schema_changes)
...
+	public function reverse_update_data($steps)
+	{
+		$reversed_array = array();
+
+		foreach ($steps as $step)
+		{
+			$parts = explode('.', $step[0]);
+			$parameters = $step[1];
+
+			$class = $parts[0];
```
marc1706: Isn't it more like a migration type?

phpBB/phpbb/db/migration/tool/config.php
```diff
@@ -150,6 +150,11 @@ public function reverse()
 					$arguments[0],
 				);
 			break;
+
+			case 'reverse':
+				// It's like double negative
```
marc1706: Not sure if that is a very helpful comment. Double negative of what?

phpBB/phpbb/db/migrator.php
```diff
@@ -488,18 +529,27 @@ protected function process_data_step($steps, $state, $revert = false)
 
 				// Set state to false since we reached the point we were at
 				$state = false;
+
+				// There is a tendency to get stuck in some cases
```
marc1706: Why is that?

phpBB/phpbb/db/migration/helper.php
Elsensee: In this case `$class` refers to the tool class. E.g. `config`, `permission`, ...

phpBB/phpbb/db/migration/tool/config.php
Elsensee: Reverse of the reverse of a call is just the call itself. That's what I mean with double negative - I will fix the comment, though.

phpBB/phpbb/db/migrator.php
Elsensee: If `dbtools.perform_schema_changes()` is called, this method will return an array, no matter what. That means, on the next call it will continue on the *same* step. Thus, we have to move forward if the last call succeeded and "not get stuck".
I will also fix this comment, then.

PR #4410, opened 2016-08-10
[ticket/14742-32x] Improvements to migrator

Elsensee:
**This is the patch for the 3.2.x branch.
To the PR for 3.1.x: #4409** 

Schema step will be "paused" now after every update step, so the code, that's running the migrator has better control over how much time the update takes.

Checklist:

- [x] Tests pass

Tracker ticket:

### DIFF
--- a/phpBB/language/en/migrator.php
+++ b/phpBB/language/en/migrator.php
@@ -51,6 +51,7 @@ $lang = array_merge($lang, array(
 	'MIGRATION_NOT_INSTALLED'			=> 'The migration "%s" is not installed.',
 	'MIGRATION_NOT_VALID'				=> '%s is not a valid migration.',
 	'MIGRATION_SCHEMA_DONE'				=> 'Installed Schema: %1$s; Time: %2$.2f seconds',
+	'MIGRATION_SCHEMA_IN_PROGRESS'		=> 'Installing Schema: %1$s; Time: %2$.2f seconds',
 	'MIGRATION_SCHEMA_RUNNING'			=> 'Installing Schema: %s.',
 
 	'MIGRATION_REVERT_DATA_DONE'		=> 'Reverted Data: %1$s; Time: %2$.2f seconds',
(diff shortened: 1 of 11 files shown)


Elsensee:
These PRs are both ready to review.

#### A few open things:
- I measured a pretty long run time once because update_data() is still done at once. (\phpbb\db\migration\data\v310\dev has many tasks in its update_data()) So, also apply the change for update_schema to this?
- The updater does now properly prevent timeouts for schema updates by refreshing the request, however there is no sign to the user that the update is still running - it could have stalled as well. First idea was to adjust the task_number for the progress bar, so that each call from update_schema would have its own step in the progress bar. It would work for the ajax output, but not so much for the cli output, since it would just output many many progress bars, I think. Any ideas?
- The CLI command (bin/phpbbcli.php, not install/phpbbcli.php) for updating the database, outputs much debug stuff. Is this on purpose or just a misconfiguration by me?
- Previously if statements returned by update_data() were somehow executed when reverting. I don't (wanna) know how exactly. These are now skipped so we might need to make a post telling this to our ext authors. Again, this returned unexpected results so I really hope, nobody did this.. ever. It was just a possbility.

#### Changes until now:
- Reverting data will now never result in a loop. If there was a custom method in revert_data() that returned a result that wasn't true or null, we would've gone back to reverting update_data(), thus probably resulting a loop between these both.
- There are no more public variables in the migrator class. Encapsulation ftw!
- The total time that it took to install a part (schema/data) is now displayed to the user instead of the time for the last method call.
- Schema updates will be done piece by piece, so that the running instance has more control over the max script runtime.
- The progress bar in the new updater has been fixed.
- There are less unnecessary notices to the user (no multiple "Installing Schema/Data for...")

marc1706:
- Yes, applying these to update_data(), too, sounds like a good idea.

- Also, I guess there is no way in disabling the huge amount of progress bar updates for CLI while keeping it for the AJAX one? Then again, I'm pretty sure we should reach a timeout message at some point in case it's stuck on the AJAX so the current way should be fine.

- Not sure if the debug output in the database updater is on purpose. It might be but I haven't looked at it in a while. I might have to check this out myself again.
- It's probably a good idea to inform ext authors of this change in case they relied on this incorrect behavior.

PR #4411, opened 2016-08-11
[ticket/12925] Pluralize permanent delete checkbox label

Senky:
[PHPBB3-12925](https://tracker.phpbb.com/browse/PHPBB3-12925)


Elsensee:
I think we should use plural rules then here.

PR #4412, opened 2016-08-15
[ticket/14573] Extend breadcrumb to support more pages

Senky:
[PHPBB3-14573](https://tracker.phpbb.com/browse/PHPBB3-14573)


CHItA:
i think this won't go into 3.2, maybe 3.3. @Nicofuma ?

marc1706:
I tend to agree.

PR #4414, opened 2016-08-17
Support pre-submit Spam checking

Zenju:
Checklist:
- [ ] Tests pass

Why is the message text being removed? The subsequent "core.posting_modify_submission_errors" then has no way to check the message for spam.
The next events, e.g. "core.posting_modify_submit_post_before" OTOH _have_ this message from "'message' => $message_parser->message,", but they are not allowed to notify an error condition. So it's a catch 22.
Best solution would be to allow "core.posting_modify_submission_errors" to check the message.

### DIFF
--- a/phpBB/posting.php
+++ b/phpBB/posting.php
@@ -572,7 +572,7 @@ $message_parser->set_mimetype_guesser($mimetype_guesser);
 if (isset($post_data['post_text']))
 {
 	$message_parser->message = &$post_data['post_text'];
-	unset($post_data['post_text']);
+	//unset($post_data['post_text']);
 }
 
 // Set some default variables


rmcgirr83:
You can use the event `core.posting_modify_message_text` to check the text from the message parser `$event['message_parser']->message`

Zenju:
Yes, but you cannot report errors from "core.posting_modify_message_text", can you?

Zenju:
Ah, so it seems that message parser warnings are treated as errors in posting.php later on, so this will probably work. Still this feels more like a hack, when there is a "core.posting_modify_submission_errors" which seems to be the right place to check for errors, but which is only missing the message text as input.
Maybe just adding a "msg_text" variable to this event, or even the full "message_parser" would be the best solution?

bantu:
By actually removing the line instead of commenting it out, you can increase your chances of getting this patch accepted.

PR #4420, opened 2016-08-20
[ticket/14742-32x] Further improvements to migrator

Elsensee:
**This is the PR for 3.2.x. The PR for 3.1 is #4419**

This Pull Request fixes some things that were forgotten in the last PRs (#4409 and #4410) since they were merged after some miscommunication.

Checklist:

- [x] Tests pass

Tracker ticket:

PHPBB3-14742

### DIFF
--- a/phpBB/phpbb/db/migration/tool/config.php
+++ b/phpBB/phpbb/db/migration/tool/config.php
@@ -152,7 +152,7 @@ class config implements \phpbb\db\migration\tool\tool_interface
 			break;
 
 			case 'reverse':
-				// It's like double negative
+				// Reversing a reverse is just the call itself
 				$call = array_shift($arguments);
 			break;
 		}
--- a/phpBB/phpbb/db/migration/tool/config_text.php
+++ b/phpBB/phpbb/db/migration/tool/config_text.php
@@ -117,7 +117,7 @@ class config_text implements \phpbb\db\migration\tool\tool_interface
 			break;
 
 			case 'reverse':
-				// It's like double negative
+				// Reversing a reverse is just the call itself
 				$call = array_shift($arguments);
 			break;
 		}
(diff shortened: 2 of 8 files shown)


phpBB/phpbb/console/command/db/console_migrator_output_handler.php
```diff
@@ -54,11 +54,11 @@ public function write($message, $verbosity)
 		{
 			$translated_message = call_user_func_array(array($this->user, 'lang'), $message);
 
-			if ($verbosity === migrator_output_handler_interface::VERBOSITY_NORMAL)
+			if ($verbosity === OutputInterface::VERBOSITY_NORMAL)
```
CHItA: what's the reason for all these changes?

Elsensee: Because the output handler for the CLI worked with symfony constants, the migrator however did with ours.
And since they would otherwise be the same I changed all of them to use Symfony constants.
Furthermore I changed ours to be like the Symfony constants for BC reasons. I suggest to remove them in the future, though.

CHItA: then i would say let's revert these changes.

Elsensee: So you'd say, just change the values of the constants we use?

CHItA: that or just map the ones in the console migrator output handler, or whatever. it makes no difference, and it is mostly just a question of style, but i don't see why it would be reasonable to include a class just for its constants. also the symfony console is designed for the console, while the output handler is not, and therefore their similarity should be treated as a coincidence imho (meaning only be used in the console output handler).

Elsensee: Okay, then I'll go with changing the constants' values.

Elsensee: Fixed now.

PR #4424, opened 2016-08-23
[ticket/14748] Make sure config values are casted to int

marc1706:
Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-14748

PHPBB3-14748

### DIFF
--- a/phpBB/phpbb/db/migration/data/v310/timezone.php
+++ b/phpBB/phpbb/db/migration/data/v310/timezone.php
@@ -103,7 +103,7 @@ class timezone extends \phpbb\db\migration\migration
 	*/
 	public function convert_phpbb30_timezone($timezone, $dst)
 	{
-		$offset = $timezone + $dst;
+		$offset = (int) $timezone + (int) $dst;
 
 		switch ($timezone)
 		{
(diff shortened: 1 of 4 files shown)


marc1706:
Remaining failures seem to be caused by a bug in PHP 7.1:
https://3v4l.org/nZEP9

marc1706:
This might actually finally be correct behavior as it does no longer ignore the backslash or no longer allows it inside the userinfo part.

phpBB/phpbb/db/migration/data/v310/timezone.php
```diff
@@ -103,7 +103,7 @@ public function update_timezones($start)
 	*/
 	public function convert_phpbb30_timezone($timezone, $dst)
 	{
-		$offset = $timezone + $dst;
+		$offset = (int) $timezone + (int) $dst;
```
marc1706: `(float)`

marc1706:
Closing in favor of #4457

rxu:
Regarding remaining failures, in PHP 7.1+ ```parse_url()``` was changed to be more restrictive and supports RFC3986 now, so it just returns false on urls like ```http://localhost.foobar.com\\@localhost/troll/http://localhost/``` (from our tests). That's why tests fail on expecting exceptions from ```redirect()``` because ```redirect()``` returns url to index page.
We might want to throw exception in that case as well for consistency instead of redirecting to the same page.

```parse_url()``` output in PHP 7.1+ and older versions:
parse_url

PR #4438, opened 2016-09-06
[ticket/12133] fix download file names for WebKit-derived browsers

danchr:
Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-12133

All browser engines that descend from WebKit declare themselves as
Safari in the user agent, including Chrome. Currently, the code
assumes that any Safari-like browser cannot handle the RFC-compliant
filename syntax. At least for recent versions of Safari, this is no
longer the case, and the legacy syntax results in URI-quoted
filenames. Using the standard syntax works as expected in both Safari
9 and Chrome 45.

The ticket reporting this issue is from January 2014, so we can safely
ignore compatibility: any browser still relying on the previous
behaviour is unlikely to receive security updates.

PHPBB3-12133

### DIFF
--- a/phpBB/includes/functions_download.php
+++ b/phpBB/includes/functions_download.php
@@ -284,7 +284,7 @@ function header_filename($file)
 
 	// There be dragons here.
 	// Not many follows the RFC...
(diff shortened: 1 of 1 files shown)


Nicofuma:
I kinda agree with the BC statement.
LGTM

PR #4439, opened 2016-09-06
[RFC] Add support for playing audio files directly in phpBB

danchr:
Checklist:
- [x] Tests pass
- [ ] Add migration support.

https://tracker.phpbb.com/browse/PHPBB3-14771

I've tested this with MP3 and M4A files with Firefox, Chrome and Safari.
Everything works as expected. Extending the change to support video
tags is trivial, but I left it out as I had no need for it.

Safari on OS X is quite picky about which files to play, and refuses
to play MP4 audio unless the mime type is set on the request.

The one thing missing is migration.

PHPBB3-14771


danchr:
I mainly posted this to get some feedback; is this something you'd be interested in?

Of note, I wasn't sure which number to use for ATTACHMENT_CATEGORY_AUDIO, and I'm also not entirely sure how I should approach migration.

n-aleha:
The link to the ticket is not the right one.

n-aleha:
The link to the ticket is not the right one.

n-aleha:
The link to the ticket is not the right one.

danchr:
Ah, sorry — fixed!

Nicofuma:
I think it could be a nice addition. About the browser compatibility, how it is displayed if the file extension or the audio tag is not supported?

VSEphpbb:
I never really cared for the audio HTML5 tag because depending on the browser, you need to use both mp3 AND ogg... Very silly implementation.

Wouldn't it be a whole lot smarter to just implement our new s9e text formatter's Media Embed option, so we can finally have mebeddable media content (video and audio) ??

rxu:
> Wouldn't it be a whole lot smarter to just implement our new s9e text formatter's Media Embed option, so we can finally have mebeddable media content (video and audio) ??

+1
Plus having ACP option to add more resources to mediaembed that would be really cool one.

VSEphpbb:
^ Added a ticket
https://tracker.phpbb.com/browse/PHPBB3-14775
http://area51.phpbb.com/phpBB/viewtopic.php?f=126&t=50196

danchr:
> I think it could be a nice addition. About the browser compatibility, how it is displayed if the file extension or the audio tag is not supported?

I deliberately didn't add any fallback code, so the audio tag simply isn't there when not supported. I could add a text message that the file cannot be played in that browser. Similarly, when the audio format isn't supported, the file just doesn't play. We could determine that in advance using JavaScript and suppress the audio element, but I'm not doing that yet.

> I never really cared for the audio HTML5 tag because depending on the browser, you need to use both mp3 AND ogg... Very silly implementation.

Well, the situation _has_ improved a bit since they introduced the tag. Firefox was the lone holdout not supporting either MP3 or MP4 AAC, but Mozilla decided to bite the bullet and delegate media decoding to the OS. The end result is that almost all browsers handle both MP3 and AAC audio.

A few screenshots:

screen shot 2016-09-07 at 20 45 20
screen shot 2016-09-07 at 20 42 50

VSEphpbb:
Thats better news but I think, before considering this, we should consider a more comprehensive, single, all-purpose Media bbcode option first, which is why I recommend we look at what we already have available in our new BBCode engine (it supports soundcloud, spotify and others).

 In all likelelyhood, forum users will more likely to want to post content from media sites, rather than host audio files themselves somewhere as this type of BBCode here would require.

danchr:
> In all likelihood, forum users will more likely to want to post content from media sites, rather than host audio files themselves somewhere as this type of BBCode here would require.

That's probably true in many cases, but not the one that motivated me to propose this. The forum I help manage already contains quite a lot of audio attachments — we don't have the rights to post our recordings to public cloud servers. Playing them within the post would be nice — for me, this change is about improving the experience for already existing audio attachments.

Nicofuma:
I agree, and I think we can do both as they don't overlap each other. They just handle 2 different use cases.

So, if you have time I think you can work on the migration support :) (And after that, as you said, video would be nice to have :))

danchr:
> So, if you have time I think you can work on the migration support :)

About that; I'd appreciate any pointers 😊 Could you perhaps think of a similar migration I could use as a starting point? Thanks!

Also, concerning `ATTACHMENT_CATEGORY_AUDIO` — should I just pick the next number in line, or avoid using one of the values that RM, WM or QUICKTIME took?

phpBB/includes/acp/acp_attachments.php
```diff
@@ -1374,6 +1374,7 @@ function category_select($select_name, $group_id = false, $key = '')
 			ATTACHMENT_CATEGORY_NONE		=> $user->lang['NO_FILE_CAT'],
 			ATTACHMENT_CATEGORY_IMAGE		=> $user->lang['CAT_IMAGES'],
 			ATTACHMENT_CATEGORY_FLASH		=> $user->lang['CAT_FLASH_FILES'],
+			ATTACHMENT_CATEGORY_AUDIO		=> $user->lang['CAT_AUDIO_FILES'],
```
Nicofuma: you need to add it around line 623 too (for the language key)

Nicofuma:
We did something similar in https://github.com/phpbb/phpbb/pull/3478 when we removed the RM,WM and QUICKTIME

You will need to remap mp3 and ogg and add the 2 other extensions. 
Took the next number in line while avoiding using a previously used value.

I don't know if there is extension out there adding a new category and using a number less than 10 (or even 100). Actually I don't know if that's even possible... but I don't think we should care

hubaishan:
What is the problem with, Can I help?

rubencm:
Close in favor of #5446

PR #4440, opened 2016-09-07
[ticket/14774] Support partial downloads of attachments

danchr:
Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-14774

phpBB already had limited support for partial downloads, but only for
resuming downloads, disregarding any range ending before EOF.

WebKit on iOS and OS X uses partial downloads when fetching media
files. Previously, only MP3 attachments could play directly in the
browser, reported as a live stream, but with this change, all
supported media formats should render as expected.

Tested using cURL by verifying that partial downloads give exactly the
same results compared to Apache.

PHPBB3-14774

### DIFF
--- a/phpBB/includes/functions_download.php
+++ b/phpBB/includes/functions_download.php
@@ -254,11 +254,21 @@ function send_file_to_browser($attachment, $upload_dir, $category)
 				send_status_line(206, 'Partial Content');
 				header('Content-Range: bytes ' . $range['byte_pos_start'] . '-' . $range['byte_pos_end'] . '/' . $range['bytes_total']);
 				header('Content-Length: ' . $range['bytes_requested']);
-			}
 
-			while (!feof($fp))
+				// First read chunks
+				while (!feof($fp) && ftell($fp) < $range['byte_pos_end'] - 8192)
+				{
+					echo fread($fp, 8192);
+				}
+				// Then, read the remainder
+				echo fread($fp, $range['bytes_requested'] % 8192);
+			}
+			else
 			{
-				echo fread($fp, 8192);
+				while (!feof($fp))
+				{
+					echo fread($fp, 8192);
+				}
 			}
 			fclose($fp);
 		}
@@ -529,6 +539,9 @@ function phpbb_find_range_request()
 */
 function phpbb_parse_range_request($request_array, $filesize)
 {
+	$first_byte_pos	= -1;
+	$last_byte_pos	= -1;
+
 	// Go through all ranges
 	foreach ($request_array as $range_string)
 	{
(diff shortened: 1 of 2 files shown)


Nicofuma:
Do you know if it is correctly handled by Symfony's BinaryFileResponse?

Nicofuma:
I'm not very familiar  with multiple ranges, do you have an explanation somewhere or a link to the corresponding RFC?

danchr:
It's mentioned in the [RFC 7233](https://tools.ietf.org/html/rfc7233#section-4.2) — the short version is that discontiguous ranges should yield a `multipart/byteranges` response. It seems a somewhat esoteric feature to me.

```console
$ curl -vr 0-10,20-30 http://localhost/~danchr/lorem.txt
*   Trying ::1...
* Connected to localhost (::1) port 80 (#0)
> GET /~danchr/lorem.txt HTTP/1.1
> Host: localhost
> Range: bytes=0-10,20-30
> User-Agent: curl/7.50.1
> Accept: */*
> 
< HTTP/1.1 206 Partial Content
< Date: Wed, 07 Sep 2016 19:49:21 GMT
< Server: Apache/2.4.23 (Unix)
< Last-Modified: Wed, 07 Sep 2016 19:48:58 GMT
< ETag: "1bd-53bf0343a9e80"
< Accept-Ranges: bytes
< Content-Length: 209
< Content-Type: multipart/byteranges; boundary=d4430f813d74b9b1
< 

--d4430f813d74b9b1
Content-type: text/plain
Content-range: bytes 0-10/445

Lorem ipsum
--d4430f813d74b9b1
Content-type: text/plain
Content-range: bytes 20-30/445

t amet, con
--d4430f813d74b9b1--
* Connection #0 to host localhost left intact
```

Nicofuma:
Well esoteric indeed. I can see how it could use used but not in our use cases... so it should be rare :)

Thanks

danchr:
I see you haven't added a label for this — would you consider it for 3.1.x?

Nicofuma:
I have added it to 3.1.x but actually I think it would be a better fit in 3.2.x (that's more a feature than a bug fix)

@bantu @marc1706 what do you think?

Note: no need to close the PR if it has to go in 3.2

danchr:
Well, it's _sort of_ a bug fix, in that the inability to respond to these requests prevents iOS from playing MP4 attachments and from adjusting the playback location in MP3 files. 3.2 seems pretty close to release, though, so it won't matter all that much 😊

bantu:
As the code was never intended to support arbitrary range queries, this is not a bug fix. Thus, I'd rather not put this into 3.1.x.

bantu:
As a workaround: This might actually already work when using X-Accel-Redirect and/or X-Sendfile.

PR #4443, opened 2016-09-12
[ticket/14780] Correction to let group setting overwrite global PM setting

Crizz0:
Checklist:
- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-14780

PHPBB3-14780

### DIFF
--- a/phpBB/includes/functions_privmsgs.php
+++ b/phpBB/includes/functions_privmsgs.php
@@ -2162,7 +2162,7 @@ function phpbb_get_max_setting_from_group(\phpbb\db\driver\driver_interface $db,
 	$max_setting = (int) $row['max_setting'];
 	$min_setting = (int) $row['min_setting'];
 
-	return ($min_setting > 0) ? $max_setting : 0;
+	return ($min_setting > 0) ? $min_setting : $max_setting;
 }
 
 /**
--- a/phpBB/includes/functions_privmsgs.php
+++ b/phpBB/includes/functions_privmsgs.php
@@ -2151,7 +2151,7 @@ function phpbb_get_max_setting_from_group(\phpbb\db\driver\driver_interface $db,
 	}
 
 	// Get maximum number of allowed recipients
-	$sql = 'SELECT MIN(g.group_' . $setting . ') as min_setting, MAX(g.group_' . $setting . ') as max_setting
+	$sql = 'SELECT MAX(g.group_' . $setting . ') as max_setting
 		FROM ' . GROUPS_TABLE . ' g, ' . USER_GROUP_TABLE . ' ug
 		WHERE ug.user_id = ' . (int) $user_id . '
 			AND ug.user_pending = 0
@@ -2160,9 +2160,8 @@ function phpbb_get_max_setting_from_group(\phpbb\db\driver\driver_interface $db,
 	$row = $db->sql_fetchrow($result);
 	$db->sql_freeresult($result);
 	$max_setting = (int) $row['max_setting'];
-	$min_setting = (int) $row['min_setting'];
 
-	return ($min_setting > 0) ? $min_setting : $max_setting;
+	return $max_setting;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


Nicofuma:
Could you add tests about that?

Crizz0:
I'm not sure, I'll try to figure out how the tests work here and why they fail and if they need an addition.

bantu:
Without looking too depp into it, I have the feeling you're simply changing the intended behaviour and this is why the test fails. I.e. someone intended that funcationality not to work as you want and you're changing that now, which potentially leads to further issues.

marc1706:
The question is also whether the goal should be to allow the group settings to supersede global ones. Maybe that's exactly what was intended here: global settings should have a higher priority than local ones. Since group settings are however usually rather explicit it might be better to move to to group settings overriding the global ones.

Crizz0:
The description of the setting in "group settings" says:

> This setting overrides the per-user folder message limit. The maximum for all groups of the user is used to determine the actual value.

So it is suppose to overwrite the global settings. This allows the Admins of a board to define bigger values for special groups, like mod and admin groups. 

The other way would be: increase the global setting to 50.000 and then give each group lower maximum-values. With that way, you have to define it in every group and not only in the few ones, who should have more PMs in the box.

Crizz0:
The expected results of the test-function were wrong in some cases. Should be fixed now, the old "if-sentence" led to a few other results.

@fix: 
Testing the fix on a 3.2. with some group and global settings, led to the desired result.

Crizz0:
Any more thoughts on this patch? @phpbb/development-team

Nicofuma:
I sent a few time in the code and I think there is a lot of confusion here.
For what I can tell:

3.1 behaviour: 
- use the max of the user groups, if it is 0, convert it to the config[pm_max_msgs] value
- pm_max_msgs is defined as: 'Users may receive no more than this many messages in each of their private message boxes. Set this value to 0 to allow unlimited messages.'

For me it means that pm_max_msgs should be a hard limit for both unlimited and limited groups

That's my proposal for 3.2:
- keep the existing phpbb_get_max_setting_from_group function but adds a comment saying that 0 means unlimited
- everywhere phpbb_get_max_setting_from_group() is used (at least ucp_pm_compose line 533 and set_user_message_limit())
   - If the group limit is 0, set it to config[pm_max_msgs]
   - After computing the user limit, take the config hard limit in account : limit = min(limit, config[pm_max_msgs])
- add a control in the group settings form and an error (or warning) message if the set value is higher than the hard limit defined in the global private messages config (not sure about this one)

migration:
 1. for each group, if the limit is 0, set it to the config[pm_max_msgs] value
 2. Set config[pm_max_msgs] limit to max(groups limit)

Note: depending on the context, it may be pm_max_recipients or pm_max_msgs

Crizz0:
I'll change the PR according to: https://wiki.phpbb.com/Maximum_private_messages_per_box

DavidIQ:
This is causing an issue on .com after the upgrade wherein team members have a limit of 5000 but the global setting of 400 is being used. This means that team members that are over the default 400 limit (most are over this) will not be able to receive any PMs for the time being unless they clear out most of their inbox.

DavidIQ:
I believe the correct behavior should be that the highest limit should be used throughout. So if the user is in 3 groups and the limits are 100, 500, 200 then the PM limit would be 500. I think in all cases the group setting should override the global setting. If it's 0 then, as the description for the setting says, use the global value.

Crizz0:
I did not change the PR according to the wiki article yet. I don't think I've time for it before end January. :(

Crizz0:
As a temp workaround this patch should work.

michaelcullum:
@marc1706 Please can this be considered a blocker.

Nicofuma:
@michaelcullum I will

marc1706:
I'm ok with having it behave like it will after merging this PR. Makes most sense right now IMHO.

PR #4453, opened 2016-09-18
[ticket/11063] Use HTTPS for phpBB's version check

Nicofuma:
Checklist:
- [X] Tests pass
- [X] Code follows coding guidelines: [master / 3.2.x](https://area51.phpbb.com/docs/master/coding-guidelines.html), [3.1.x](https://area51.phpbb.com/docs/31x/coding-guidelines.html)

https://tracker.phpbb.com/browse/PHPBB3-11063

PHPBB3-11063

### DIFF
--- a/phpBB/phpbb/version_helper.php
+++ b/phpBB/phpbb/version_helper.php
@@ -36,7 +36,7 @@ class version_helper
 	/**
 	 * @var bool Use SSL or not
 	 */
-	protected $use_ssl = false;
+	protected $use_ssl = true;
 
 	/**
 	 * @var string Current version installed


Nicofuma:
@michaelcullum a small gift

3D-I:
Does that means that we will be able to use our github repositories for the extensions' version-checkers instead of working-around creating a github'd website for that, as I (and others) did?

michaelcullum:
Hah. Thanks.

Nicofuma:
@3D-I it is already possible, you need to set the `ssl` key to `true` for that

3D-I:
@Nicofuma Yes, but I can't ask the end user to hard-code its own phpBB.

Nicofuma:
I was talking about the version check definition of your own extensions

3D-I:
@Nicofuma I don't think I am following you, care to explain? I know this is not the right place though.

Nicofuma:
```
		"version-check": {
			"host": "nicofuma.github.io",
			"directory": "/",
			"filename": "webprofiler.json",
			"ssl": true,
}
```

marc1706:
@Nicofuma label says 3.1 target while PR shows master branch. I take it this is intended for 3.1 then?

Nicofuma:
not sure actually, that's quite a big change

marc1706:
I guess one could still overrride it with setting ssl to false?
I'd aim this at 3.2 then though.

3D-I:
@Nicofuma thanks, got it.

marc1706:
Moving to 3.2 milestone

Nicofuma:
pushed back to 3.3 as I'd prefer to have a proper ssl support for that. (may be included in one of the 3.2.x version if the diff is not too big or complex)

michaelcullum:
Long-term: Require a PSR-7 http client abstraction, then use Guzzle
Short term why not enable ssl?

Elsensee:
What is stopping this from happening?

marc1706:
I think there were a few questions regarding the support for this on shared hosting.

rubencm:
I think this should be something like
`extension_loaded  ('openssl') && in_array("https", stream_get_wrappers())`
(I havent check, and i am not sure about the first condition)

PR #4458, opened 2016-09-23
Root path was not working for non root foruns

Leinad4Mind:
Checklist:

- [ ] Tests pass

I don't know why it was being created the variable: $root_path = (defined('PHPBB_USE_BOARD_URL_PATH') && PHPBB_USE_BOARD_URL_PATH) ? generate_board_url() . '/' : $phpbb_path_helper->get_web_root_path();

Since for foruns: website.com/forum, it was not working when using extensions that grabs the text with generate_text_for_display()

Use this information as you like for solving the issue.

### DIFF
--- a/phpBB/includes/functions_content.php
+++ b/phpBB/includes/functions_content.php
@@ -988,8 +988,7 @@ function smiley_text($text, $force_option = false)
 	}
 	else
 	{
-		$root_path = (defined('PHPBB_USE_BOARD_URL_PATH') && PHPBB_USE_BOARD_URL_PATH) ? generate_board_url() . '/' : $phpbb_path_helper->get_web_root_path();
-		return preg_replace('#<!\-\- s(.*?) \-\-><img src="\{SMILIES_PATH\}\/(.*?) \/><!\-\- s\1 \-\->#', '<img class="smilies" src="' . $root_path . $config['smilies_path'] . '/\2 />', $text);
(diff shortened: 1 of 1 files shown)


DavidIQ:
This change sounds more like a configuration issue on your board or something else. There are thousands of boards out there that are not in the site root, even our very own board on phpBB.com so this would have come up.

Leinad4Mind:
Let me explain better. I'm using an extension that shows on the viewforum.php a preview of the topic in a modal window. And the extensions uses that function generate to text display  for generating the text. And ofc if I've a Smilies. The url of the smile will be: ./../../images/ and should be: ./images and I've trace the problem down to that line. 

I've a fresh 3.1.9, im pretty sure that's the line that's causing the issue. 

If anyone saw this is just a matter of fact that they never use or done an extension that uses smiles on viewforum.php

DavidIQ:
That's fine but I think that's why there is a check for PHPBB_USE_BOARD_URL_PATH. Have you tried defining this constant in your extension before making the call to this function?

Leinad4Mind:
Hmm I see. Thanks for pointing that. I'll investigate.

Leinad4Mind:
I was able to fix by adding this:

				if (!defined('PHPBB_USE_BOARD_URL_PATH')) {
					 define('PHPBB_USE_BOARD_URL_PATH', true);
				}

after my sql code.

Thanks for the help David! We can closed this issue. :)

PR #4463, opened 2016-09-25
[ticket/14797] Support PHP 7.1 and clearly state support for latest versions

marc1706:
Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-14797

PHPBB3-14797

### DIFF
--- a/phpBB/composer.json
+++ b/phpBB/composer.json
@@ -25,7 +25,7 @@
 		"phpbb/phpbb-core": "self.version"
 	},
 	"require": {
-		"php": ">=5.4,<7.1",
+		"php": ">=5.4",
 		"bantu/ini-get-wrapper": "1.0.*",
 		"google/recaptcha": "~1.1",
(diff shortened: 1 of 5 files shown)


phpBB/composer.json
```diff
@@ -25,7 +25,7 @@
 		"phpbb/phpbb-core": "self.version"
 	},
 	"require": {
-		"php": ">=5.4,<7.1",
+		"php": ">=5.4",
```
Nicofuma: < 7.2?

marc1706: No, I actually did that on purpose. Previously we had to update this requirement every time there is a new version. Instead of explicitly stating no compatibility with newer versions (which we don't actually know), we now implicitly state support unless we discover something else (in which case we might just have to change a few lines of code).

Nicofuma: ok

PR #4479, opened 2016-10-09
[ticket/14814] Fix reparser cron

Nicofuma:
Checklist:
- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-14814

PHPBB3-14814

### DIFF
--- a/phpBB/phpbb/cron/task/text_reparser/reparser.php
+++ b/phpBB/phpbb/cron/task/text_reparser/reparser.php
@@ -99,7 +99,7 @@ class reparser extends \phpbb\cron\task\base
 			$this->reparser_manager->get_resume_data($this->reparser_name);
 		}
 
-		if (empty($this->resume_data['range-max']) || $this->resume_data['range-max'] >= $this->resume_data['range-min'])
+		if (!isset($this->resume_data['range-max']) || $this->resume_data['range-max'] >= $this->resume_data['range-min'])
 		{
 			return true;
 		}
@@ -147,9 +147,9 @@ class reparser extends \phpbb\cron\task\base
 			 */
 			$reparser = $this->reparsers[$this->reparser_name];
 
-			$min = !empty($this->resume_data['range-min']) ? $this->resume_data['range-min'] : self::MIN;
-			$current = !empty($this->resume_data['range-max']) ? $this->resume_data['range-max'] : $reparser->get_max_id();
-			$size = !empty($this->resume_data['range-size']) ? $this->resume_data['range-size'] : self::SIZE;
+			$min = isset($this->resume_data['range-min']) ? $this->resume_data['range-min'] : self::MIN;
+			$current = isset($this->resume_data['range-max']) ? $this->resume_data['range-max'] : $reparser->get_max_id();
+			$size = isset($this->resume_data['range-size']) ? $this->resume_data['range-size'] : self::SIZE;
 
 			if ($current >= $min)
 			{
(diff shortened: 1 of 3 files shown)


Nicofuma:
@JoshyPHP just in case you didn't know: `empty(0) === emty(false) === empty('') ===  empty([]) === empty("0") === false`

JoshyPHP:
@Nicofuma Sure, although I don't recall writing anything for the cron so it's probably not mine.

With that said, looking at the diff I don't understand why the cron task keeps running after completion. If that change fixes it I guess that's good but there may be another issue in the logic somewhere. The task shouldn't run if `range-max` is 0.

phpBB/phpbb/cron/task/text_reparser/reparser.php
```diff
@@ -99,7 +99,7 @@ public function is_runnable()
 			$this->reparser_manager->get_resume_data($this->reparser_name);
 		}
 
-		if (empty($this->resume_data['range-max']) || $this->resume_data['range-max'] >= $this->resume_data['range-min'])
+		if (!isset($this->resume_data['range-max']) || $this->resume_data['range-max'] >= $this->resume_data['range-min'])
```
JoshyPHP: I think this might need to be this:

```php
if (!empty($this->resume_data['range-max']) && $this->resume_data['range-max'] >= $this->resume_data['range-min'])
```

IOW, only run the task if `range-max` is positive (non-zero) and the range is valid.

@Elsensee does that make sense to you? I can't run it at the moment to verify it. That may be the only change required.

Nicofuma:
Right, Elsensee did... sorry

About the issue the issue is with empty because:
 - we want to run the cron when resume_data is empty (the cron has not be run yet)
 - at the end (when everything has been reparsed) resume_data['range-max'] === 0
 - empty(0) === true

So what happens is that is_runnable returns true because it thinks that we don't have the resume data and in `run()` $current is set to `$reparser->get_max_id()` because `!empty($this->resume_data['range-max']) === true`

phpBB/phpbb/cron/task/text_reparser/reparser.php
Elsensee: I think we get the same result as with the current change.

JoshyPHP:
The migration sets up `resume_data`, right? I assumed that if it's empty then nothing needs to be run. And if someone schedules a reparser, the manager will set up the data.

That's where the console command and cron task differ. If someone runs a reparser in CLI without resume data then it should restart from the beginning. If a cron task is called without any data set, I'd think it shouldn't run. To be honest I haven't looked closely how the current system works exactly so I don't have a strong opinion.

JoshyPHP:
By the way, while you're at it you may want to fix an identity comparison that accidentally got changed to an equality there: https://github.com/phpbb/phpbb/commit/900ccd79af816d8d54e6974e9163d9999a823513#diff-9bd3cfd40ad4b7ee819ad500fab66b1dR249 (line 249)

rxu:
> we want to run the cron when resume_data is empty (the cron has not be run yet)

I guess we don't want to run the cron on boards which were just installed but not converted from previous phpBB versions. So I tend to agree with @JoshyPHP here https://github.com/phpbb/phpbb/pull/4479#issuecomment-252512236

Nicofuma:
I agree it should be done like that. But the migration needs to be updated

marc1706:
Do I understand this correctly that we still need the reparser migraton changed then?

VSEphpbb:
Bumpity bump

marc1706:
@VSEphpbb that's my comment
image

Nicofuma:
Updated
I hope it will be enough

marc1706:
@VSEphpbb mind checking and confirming whether this also solves the issue with the media embed extension?

VSEphpbb:
@marc1706 it does

PR #4480, opened 2016-10-11
[ticket/14818] Optimize generate_board_url function

gmixo:
if $config['force_server_vars'] is enabled no need to calculate $request

That is a really useful for my sending PM extension - it is totally haven't $request object so without this changes my function is fails.

 https://tracker.phpbb.com/browse/PHPBB3-14818 is matching issue


JoshyPHP:
Closing that pull request got a bunch of tickets to be reopened on the tracker.

PR #4482, opened 2016-10-14
[ticket/14821] Replace escaped strong tags with actual tags

marc1706:
Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-14821

Everything else will still be escaped. This will however add back support for
bold text in exception messages.

PHPBB3-14821

### DIFF
--- a/phpBB/phpbb/event/kernel_exception_subscriber.php
+++ b/phpBB/phpbb/event/kernel_exception_subscriber.php
@@ -69,7 +69,7 @@ class kernel_exception_subscriber implements EventSubscriberInterface
 		}
 
 		// Show <strong> text in bold
-		$message = preg_replace('#&lt;(/?)(strong)&gt;#i', '<$1$2>', $message);
+		$message = preg_replace('#&lt;(/?strong)&gt;#i', '<$1>', $message);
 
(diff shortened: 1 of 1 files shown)


marc1706:
Other option would be to no longer support bold text or simply use a str_replace() which would do the same.

phpBB/phpbb/event/kernel_exception_subscriber.php
```diff
@@ -68,6 +68,9 @@ public function on_kernel_exception(GetResponseForExceptionEvent $event)
 			$message = call_user_func_array(array($this->user, 'lang'), array_merge(array($message), $exception->get_parameters()));
 		}
 
+		// Show <strong> text in bold
+		$message = preg_replace('#&lt;(/?)(strong)&gt;#i', '<$1$2>', $message);
```
JoshyPHP: You should be able to simplify it to this:

```php
$message = preg_replace('#&lt;(/?strong)&gt;#i', '<$1>', $message);
```

marc1706: Thanks, updated.

PR #4488, opened 2016-10-17
Add core.auth_oauth_login_after and core.auth_oauth_link_after

senky:
Checklist:
- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-14825

PHPBB3-14825

### DIFF
--- a/phpBB/phpbb/auth/provider/oauth/oauth.php
+++ b/phpBB/phpbb/auth/provider/oauth/oauth.php
@@ -98,6 +98,13 @@ class oauth extends \phpbb\auth\provider\base
 	protected $phpbb_container;
 
 	/**
+	* phpBB event dispatcher
+	*
+	* @var \phpbb\event\dispatcher_interface
+	*/
+	protected $dispatcher;
+
+	/**
 	* phpBB root path
 	*
 	* @var string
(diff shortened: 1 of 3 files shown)


marc1706:
Build errors are caused by this:
`ERROR: An error occurred while parsing "phpbb\auth\provider\oauth\oauth::__construct" line "138": "13" @param tags are expected but only "1" found`

senky:
@marc1706 could you give me a hand with fixing this issue? I don't even understand the error message. There ARE 13 `@param` tags in the constructor's docblock...

phpBB/phpbb/auth/provider/oauth/oauth.php
```diff
@@ -126,8 +133,9 @@ class oauth extends \phpbb\auth\provider\base
 	* @param	\Symfony\Component\DependencyInjection\ContainerInterface $phpbb_container DI container
 	* @param	string			$phpbb_root_path
 	* @param	string			$php_ext
+	* @param 	\phpbb\event\dispatcher_interface	$dispatcher phpBB Event dispatcher object
```
marc1706: Hard to track down but there is actually a space after `@param which the sami parse check does not like. If you remove it, the tests will run fine.

marc1706:
Sorry, I overlooked that the oauth class already gets passed the container. You don't need to pass the dispatch via the constructor but can use `$phpbb_container->get('dispatcher');`.

tests/functions/user_delete_test.php
```diff
@@ -59,6 +59,7 @@ protected function setUp()
 		$passwords_helper = new \phpbb\passwords\helper;
 		// Set up passwords manager
 		$passwords_manager = new \phpbb\passwords\manager($config, $passwords_drivers, $passwords_helper, array_keys($passwords_drivers));
+		$dispatcher = new phpbb_mock_event_dispatcher();
```
marc1706: The dispatcher is already defined as `$phpbb_dispatcher` a few lines above this. As mentioned in my comment, you don't actually need to pass the dispatcher to the class via the constructor though.

senky:
> Sorry, I overlooked that the oauth class already gets passed the container. You don't need to pass the dispatch via the constructor but can use $phpbb_container->get('dispatcher');.

I don't really understand why are not other services initialized by container as well, but OK.

senky:
As build fails revealed:
https://travis-ci.org/phpbb/phpbb/builds/175502165
I really have to use DI to inject dispatcher.

marc1706:
No, you just have to set the dispatcher service in the tests (which is what this is complaining about):
https://github.com/senky/phpbb/blob/df0388ccc5b24cfb67078dd2fa8cb2fc4f607dac/tests/functions/user_delete_test.php#L83

However, you can leave it like that now. Really not worth the trouble to change this all over again. ;)

PR #4490, opened 2016-10-19
[ticket/14827] Fix adding multiple form keys

kasimi:
I consider this a bug because extensions that use the `add_form_key()` function without thinking twice will run into issues if they use it on a page where phpBB (or another extension) already called that function. Forms will only validate for whoever called the function last.

The solution is for the extension to use the `$template_variable_suffix` argument that is added in this PR and set it to `_{VENDOR}_{EXT}`, for example, to ensure they don't overwrite anyone else's form elements.

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-14827

### DIFF
--- a/phpBB/includes/functions.php
+++ b/phpBB/includes/functions.php
@@ -2609,8 +2609,9 @@ function check_link_hash($token, $link_name)
 /**
 * Add a secret token to the form (requires the S_FORM_TOKEN template variable)
 * @param string  $form_name The name of the form; has to match the name used in check_form_key, otherwise no restrictions apply
+* @param string  $template_variable_suffix A string that is appended to the name of the template variable to which the form elements are assigned
 */
-function add_form_key($form_name)
+function add_form_key($form_name, $template_variable_suffix = '')
 {
 	global $config, $template, $user, $phpbb_dispatcher;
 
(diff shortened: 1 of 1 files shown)


kasimi:
I intended to add this to the 3.1.x branch, are there any reasons not to?

marc1706:
I think there is some confusion on how to create distinctive form keys. Extensions should NOT call the add_form_key() with the same form name again. That should also cause the HTML to be invalid due to the duplicate form name as it is usually just that, the form's name attribute. This can be seen equal to the id attribute and is actually the future replacement for the form name:
```
When used with the form element, the name attribute is deprecated. Use id instead.
```

kasimi:
The issue here is the name of the template variable that is used to assign the form elements. When phpBB calls the function, the elements are assigned to `S_FORM_TOKEN`. When an extension calls the function subsequently (with a different `$form_name` of course) it also uses the `S_FORM_TOKEN` and the previous elements cannot be access anymore. The phpBB form displays the extension's elements and will fail to validate when being submitted.

marc1706:
Ah, now I understand your intention. I think it would be ok to add this to 3.1.x.

@phpbb/development-team anyone against adding this to 3.1.x?

kasimi:
I'm not sure I understand why `check_form_key()` doesn't support it. Submitting a form sends the `creation_time` and `form_token` values that are found *inside the respective `<form>` element*. Depending on which form is submitted, they have different values, and `check_form_key()` doesn't care about from which form they come from.

CHItA:
Probably an okay solution until we have a better form implementation

marc1706:
Sorry, was confused and misread. You're right that this will be correctly handled. I also think this should go against 3.1.x.

kasimi:
Yes it's based on 3.1.x.

PR #4491, opened 2016-10-22
[ticket/14831] Fix extension modules installation issues

rxu:
Checklist:

- [x] Tests pass

<a href="https://tracker.phpbb.com/browse/PHPBB3-14831">PHPBB3-14831</a>.

### DIFF
--- a/phpBB/phpbb/db/migration/tool/module.php
+++ b/phpBB/phpbb/db/migration/tool/module.php
@@ -482,7 +482,7 @@ class module implements \phpbb\db\migration\tool\tool_interface
 	public function get_parent_module_id($parent_id, $data = '', $throw_exception = true)
 	{
 		// Initialize exception object placeholder
-		$e = false;
+		$exception = false;
 
 		// Allow '' to be sent as 0
 		$parent_id = $parent_id ?: 0;
@@ -505,7 +505,7 @@ class module implements \phpbb\db\migration\tool\tool_interface
 			{
 				// No parent with the given module_langname exist
 				case 0:
-					$e = new \phpbb\db\migration\exception('MODULE_NOT_EXIST', $parent_id);
+					$exception = new \phpbb\db\migration\exception('MODULE_NOT_EXIST', $parent_id);
 				break;
 
 				// Return the module id
@@ -527,7 +527,7 @@ class module implements \phpbb\db\migration\tool\tool_interface
 						$parent_id = (int) $this->db->sql_fetchfield('parent_id');
 						if (!$parent_id)
 						{
-							$e = new \phpbb\db\migration\exception('PARENT_MODULE_FIND_ERROR', $data['parent_id']);
+							$exception = new \phpbb\db\migration\exception('PARENT_MODULE_FIND_ERROR', $data['parent_id']);
 						}
 					}
 					else if (!empty($data) && !is_array($data))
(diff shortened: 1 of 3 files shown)


phpBB/phpbb/db/migration/tool/module.php
```diff
@@ -81,6 +81,9 @@ public function get_name()
 	*/
 	public function exists($class, $parent, $module)
 	{
+		// Clear the Modules Cache first to get actual data
+		$this->cache->destroy("_modules_$class");
```
VSEphpbb: This doesn't seem to work. Method "destroy" not found in \phpbb\cache\service

rxu: But how it worked before in the ```add()``` method and why tests didn't fail on that, just curious. These things with migrations are all weird 😕

VSEphpbb:
Maybe don't change the cache then... My IDE just pointed that out, but it could be a magic method. Seems tests fail after changing it.

rxu:
It looks like this needs more work as there're some more issues with migrations relying on ```if module.exists``` conditions.

rxu:
The particular problem is that in case ```parent_id``` is not found the ```get_parent_module_id()``` throws exception before the ```exists()``` method returns anything out.

phpBB/phpbb/db/migration/tool/module.php
lavigor: @VSEphpbb @rxu method `destroy` exists for each cache driver and works OK because of [this](https://github.com/rxu/phpbb3/blob/0cc092073405166a0da72009823256a00389d0b8/phpBB/phpbb/cache/service.php#L94-L97).
Probably IDEs cannot handle this case at the moment.

lavigor: @VSEphpbb @rxu method `destroy` exists for each cache driver and works OK because of [this](https://github.com/rxu/phpbb3/blob/0cc092073405166a0da72009823256a00389d0b8/phpBB/phpbb/cache/service.php#L94-L97).
Probably IDEs cannot handle this case at the moment.

lavigor: @VSEphpbb @rxu method `destroy` exists for each cache driver and works OK because of [this](https://github.com/rxu/phpbb3/blob/0cc092073405166a0da72009823256a00389d0b8/phpBB/phpbb/cache/service.php#L94-L97).
Probably IDEs cannot handle this case at the moment.

phpBB/phpbb/db/migration/tool/module.php
```diff
@@ -469,11 +475,15 @@ protected function get_categories_list()
...
-	* @return int The parent module_id
+	* @param bool $throw_exception The flag indicating if exception should be thrown on error
+	* @return mixed The int parent module_id or false
 	* @throws \phpbb\db\migration\exception
 	*/
-	public function get_parent_module_id($parent_id, $data = '')
+	public function get_parent_module_id($parent_id, $data = '', $throw_exception = true)
 	{
+		// Initialize exception object placeholder
+		$e = false;
```
VSEphpbb: Please a better name than just $e

rxu:
I hope this can be !unset WIP now.

phpBB/phpbb/db/migration/tool/module.php
```diff
@@ -90,8 +90,15 @@ public function exists($class, $parent, $module)
 		$parent_sql = '';
 		if ($parent !== false)
 		{
-			$parent = $this->get_parent_module_id($parent, $module);
-			$parent_sql = 'AND parent_id = ' . (int) $parent;
+			$parent = $this->get_parent_module_id($parent, $module, false);
+			if ($parent !== false)
```
rxu: Good catch, agreed.

rxu:
!unset WIP

PR #4492, opened 2016-10-22
[ticket/14831] Make sure migrations always start with backslash

marc1706:
Checklist:
- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-14831

The migration system expects every migration to start with a backslash.
If depends on definitions do not supply that leading backslash, we should
make sure it's added manually before calling the depends on migration.

PHPBB3-14831

### DIFF
--- a/phpBB/phpbb/db/migration/data/v31x/migrations_deduplicate_entries.php
+++ b/phpBB/phpbb/db/migration/data/v31x/migrations_deduplicate_entries.php
@@ -55,21 +55,21 @@ class migrations_deduplicate_entries extends \phpbb\db\migration\migration
 			$prepended_name = preg_replace('#^(?!\\\)#', '\\\$0', $name);
 			$prefixless_name = preg_replace('#(^\\\)([^\\\].+)#', '$2', $name);
 
-			if ($prepended_name !== $name && isset($migration_state[$prepended_name]) && $migration_state[$prepended_name] === $migration_state[$name])
+			if ($prepended_name != $name && isset($migration_state[$prepended_name]) && $migration_state[$prepended_name]['migration_depends_on'] == $migration_state[$name]['migration_depends_on'])
 			{
 				$duplicate_migrations[] = $name;
 				unset($migration_state[$prepended_name]);
 			}
-			else if ($prefixless_name !== $name && isset($migration_state[$prefixless_name]) && $migration_state[$prefixless_name] === $migration_state[$name])
+			else if ($prefixless_name != $name && isset($migration_state[$prefixless_name]) && $migration_state[$prefixless_name]['migration_depends_on'] == $migration_state[$name]['migration_depends_on'])
 			{
-				$duplicate_migrations[] = $name;
+				$duplicate_migrations[] = $prefixless_name;
 				unset($migration_state[$prefixless_name]);
 			}
 		}
 
 		if (count($duplicate_migrations))
 		{
-			$sql = 'DELETE *
+			$sql = 'DELETE
 				FROM ' . $this->table_prefix . 'migrations
 				WHERE '  . $this->db->sql_in_set('migration_name', $duplicate_migrations);
 			$this->db->sql_query($sql);
--- a/phpBB/phpbb/db/migration/data/v31x/migrations_deduplicate_entries.php
+++ b/phpBB/phpbb/db/migration/data/v31x/migrations_deduplicate_entries.php
@@ -1,0 +1,78 @@
+<?php
+
+/**
+ *
+ * This file is part of the phpBB Forum Software package.
+ *
+ * @copyright (c) phpBB Limited <https://www.phpbb.com>
+ * @license GNU General Public License, version 2 (GPL-2.0)
+ *
+ * For full copyright and license information, please see
+ * the docs/CREDITS.txt file.
+ *
+ */
+
+namespace phpbb\db\migration\data\v31x;
+
+class migrations_deduplicate_entries extends \phpbb\db\migration\migration
+{
+	static public function depends_on()
+	{
+		return array('\phpbb\db\migration\data\v31x\v3110');
+	}
+
+	public function update_data()
+	{
+		return array(
+			array('custom', array(array($this, 'deduplicate_entries'))),
+		);
+	}
+
+	public function deduplicate_entries()
+	{
+		$migration_state = array();
+		$duplicate_migrations = array();
+
+		$sql = "SELECT *
+			FROM " . $this->table_prefix . 'migrations';
+		$result = $this->db->sql_query($sql);
+
+		if (!$this->db->get_sql_error_triggered())
+		{
+			while ($migration = $this->db->sql_fetchrow($result))
+			{
+				$migration_state[$migration['migration_name']] = $migration;
+
+				$migration_state[$migration['migration_name']]['migration_depends_on'] = unserialize($migration['migration_depends_on']);
+				$migration_state[$migration['migration_name']]['migration_data_state'] = !empty($migration['migration_data_state']) ? unserialize($migration['migration_data_state']) : '';
+			}
+		}
+
+		$this->db->sql_freeresult($result);
+
+		foreach ($migration_state as $name => $migration)
+		{
+			$prepended_name = preg_replace('#^(?!\\\)#', '\\\$0', $name);
+			$prefixless_name = preg_replace('#(^\\\)([^\\\].+)#', '$2', $name);
+
+			if ($prepended_name !== $name && isset($migration_state[$prepended_name]) && $migration_state[$prepended_name] === $migration_state[$name])
+			{
+				$duplicate_migrations[] = $name;
+				unset($migration_state[$prepended_name]);
+			}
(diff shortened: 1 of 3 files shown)


michaelcullum:
👍

marc1706:
@VSEphpbb updated to be more robust than the previous change. Please try to test this as much as possible ;)

phpBB/phpbb/db/migrator.php
```diff
@@ -264,10 +277,22 @@ protected function try_apply($name)
...
 		foreach ($state['migration_depends_on'] as $depend)
 		{
-			// Make sure migration starts with backslash
-			$depend = $depend[0] == '\\' ? $depend : '\\' . $depend;
+			// Try falling back to a valid migration name with or without leading backslash
+			if (!isset($this->migration_state[$name]))
+			{
+				if (isset($this->migration_state[preg_replace('#^(?!\\\)#', '\\\$0', $name)]))
+				{
+					$name = preg_replace('#^(?!\\\)#', '\\\$0', $name);
```
rxu: To save some ```preg_replace``` calls maybe assign it to some variable before.

phpBB/phpbb/db/migrator.php
```diff
@@ -264,10 +277,22 @@ protected function try_apply($name)
...
+				}
+				else if (isset($this->migration_state[preg_replace('#(^\\\)([^\\\].+)#', '$2', $name)]))
+				{
+					$name = preg_replace('#(^\\\)([^\\\].+)#', '$2', $name);
```
rxu: Same as above.

VSEphpbb:
A little problem encountered with this.

If you already had an extension installed that added a migration dependency without the leading slash to your migrations table. This is what you end up with in the database:
screen shot 2016-10-23 at 9 01 11 am
You'll notice that the m1 migration appears twice (once without the slash) and all the others depend on the one without the slash.

Then you uninstall that extension (the migration w/o the slash remains in the database). 
screen shot 2016-10-23 at 9 07 47 am

With this PR's changes, if you try to reinstall the extension again, that "rogue" migration is now prevented from running.

I'm not sure if we can do anything about this other than through support and tell people to uninstall the ext, then delete that from their DB?

phpBB/phpbb/db/migration/data/v31x/migrations_deduplicate_entries.php
```diff
@@ -0,0 +1,78 @@
...
+ * @license GNU General Public License, version 2 (GPL-2.0)
+ *
+ * For full copyright and license information, please see
+ * the docs/CREDITS.txt file.
+ *
+ */
+
+namespace phpbb\db\migration\data\v31x;
+
+class migrations_deduplicate_entries extends \phpbb\db\migration\migration
```
VSEphpbb: remove_duplicate_migration_entries ?

phpBB/phpbb/db/migration/data/v31x/migrations_deduplicate_entries.php
```diff
@@ -0,0 +1,78 @@
...
+		$result = $this->db->sql_query($sql);
+
+		if (!$this->db->get_sql_error_triggered())
+		{
+			while ($migration = $this->db->sql_fetchrow($result))
+			{
+				$migration_state[$migration['migration_name']] = $migration;
+
+				$migration_state[$migration['migration_name']]['migration_depends_on'] = unserialize($migration['migration_depends_on']);
+				$migration_state[$migration['migration_name']]['migration_data_state'] = !empty($migration['migration_data_state']) ? unserialize($migration['migration_data_state']) : '';
```
VSEphpbb: Is this even used?

VSEphpbb:
tested and works

phpBB/phpbb/db/migration/data/v31x/migrations_deduplicate_entries.php
VSEphpbb: or `migrations_remove_duplicates`
or `remove_duplicate_migrations`
:)

phpBB/phpbb/db/migration/data/v31x/migrations_deduplicate_entries.php
```diff
@@ -0,0 +1,78 @@
...
+			}
+		}
+
+		$this->db->sql_freeresult($result);
+
+		foreach ($migration_state as $name => $migration)
+		{
+			$prepended_name = preg_replace('#^(?!\\\)#', '\\\$0', $name);
+			$prefixless_name = preg_replace('#(^\\\)([^\\\].+)#', '$2', $name);
```
Nicofuma: I'd prefer `$name = ($name[0] === '\' ? '' : '\').$name;` I find it a lot more readable

Nicofuma:
What do we have in database before this PR? Duplicated migrations or just migrations without the leading slash?

phpBB/phpbb/db/migration/data/v31x/migrations_deduplicate_entries.php
```diff
@@ -0,0 +1,78 @@
...
+
+			if ($prepended_name != $name && isset($migration_state[$prepended_name]) && $migration_state[$prepended_name]['migration_depends_on'] == $migration_state[$name]['migration_depends_on'])
+			{
+				$duplicate_migrations[] = $name;
```
JoshyPHP: I don't know what this code block really does so my apologies if I'm way off base but it looks like it should be `$duplicate_migrations[] = $prepended_name;`

VSEphpbb:
@Nicofuma It's possible the DB can have duplicate migrations that are missing a slash.. See the 1st image attachment above.

phpBB/phpbb/db/migration/data/v31x/migrations_deduplicate_entries.php
VSEphpbb: Nah it should be $name.

Nicofuma:
I mean, when they is a a migration with a missing slash it is always a duplicate?

phpBB/phpbb/db/migration/data/v31x/migrations_deduplicate_entries.php
marc1706: I'm guessing you don't like using deduplicate?

phpBB/phpbb/db/migration/data/v31x/migrations_deduplicate_entries.php
marc1706: Nope, will remove

marc1706:
Probably, but I don't really want to rely on that which is why I'm checking for both possibilities. Also, if you're having the same migration in the db twice, but with different depends_on() (which really shouldn't happen) it will now not remove those.

phpBB/phpbb/db/migration/data/v31x/migrations_deduplicate_entries.php
VSEphpbb: you are correct sir

Nicofuma:
I am asking that because the current migration only remove duplicate, it does not update existig migrations

VSEphpbb:
This should only remove duplicates. They don't need updating.

marc1706:
Updated.

hubaishan:
This commit is buggy and cause the error explain here https://tracker.phpbb.com/browse/PHPBB3-14935

hubaishan:
to correct the bug the function get_valid_name must edited to

`	protected function get_valid_name($name)
	{
		// Try falling back to a valid migration name with or without leading backslash
		if (!isset($this->migration_state[$name]))
		{
			$prepended_name = ($name[0] == '\\' ? '' : '\\') . $name;
			$prefixless_name = $name[0] == '\\' ? substr($name, 1) : $name;

			if (isset($this->migration_state[$prefixless_name]))
			{
				$name = $prefixless_name;
			}
			else 
			{
				$name = $prepended_name;
			}
		}

		return $name;
	}
`

marc1706:
Have you actually tried this and ran the migration? That was the initial bugfix but instead we added a migration that gets rid of the duplicate entries. The migrator was also modified to always find a depends_on() migration with or without leading backslash in order to prevent future duplicates. In the end you should make sure to create a valid migrations file and any extension submitted to the customisations database will be checked for valid definitions both by EPV and the Extensions Team.

hubaishan:
Yes I tried it.
the mistake will occur if $name doesn't started with backslash and also it isn't exit in migration table in this case get_valid_name will return the same in $name which doesn't started with backslash.

this will cause error only if a migration (b) depends on migration (a) not started with backslash so try_apply() will check if (a) is exist in migration table, in case (a) is before (b) in (a-z) order function  get_valid_name will give the name which in the migration table (prepended_name), but if  (a) is after (b) in (a-z) order then (a) will not be in the migration table so get_valid_name will return original (a) name (prefixless_name) and try_apply() will call try_apply(a) so (a) will inserted in migration table without pre slash, so system will go in non stop loop, and hanged. 

by editing  get_valid_name with my suggest it will return  $prepended_name as default if it is not exist in migration table without backslash.

PR #4497, opened 2016-11-02
[ticket/14849] Add core.acp_extensions_run_action

senky:
Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-14849

PHPBB3-14849

### DIFF
--- a/phpBB/includes/acp/acp_extensions.php
+++ b/phpBB/includes/acp/acp_extensions.php
@@ -34,7 +34,7 @@ class acp_extensions
 	function main()
 	{
 		// Start the page
-		global $config, $user, $template, $request, $phpbb_extension_manager, $db, $phpbb_root_path, $phpEx, $phpbb_log, $cache;
+		global $config, $user, $template, $request, $phpbb_extension_manager, $db, $phpbb_root_path, $phpEx, $phpbb_log, $cache, $phpbb_dispatcher;
 
 		$this->db = $db;
 		$this->config = $config;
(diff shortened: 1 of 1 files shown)


phpBB/includes/acp/acp_extensions.php
```diff
@@ -55,6 +56,21 @@ function main()
...
 		$start_time = time();
 
+		/**
+		* Event to run a specific action on extension
+		*
+		* @event core.acp_extensions_run_action
+		* @var	string	action			Action to run
+		* @var	string	u_action		Url we are at
+		* @var	string	ext_name		Extension name from request
+		* @var	string	safe_time_limit	Safe limit of execution time
```
marc1706: That's an integer

phpBB/includes/acp/acp_extensions.php
```diff
@@ -55,6 +56,21 @@ function main()
...
+		* @var	string	start_time		Start time
```
marc1706: Also an integer

javiexin:
Sorry to bother, but shouldn't the event be placed just before the switch($action) ?

Checking for the "cancel" action, validating form key, and getting extension metadata in case extension is specified all seem like "required" steps, for any action.  So, whatever extension uses the above event should reimplement all these...

In my opinion, the event would be better placed just before the switch, and would include $md_manager (if possible, as it is an object, and might not be effective to do so... but will most probably require extensions to repeat the code).

javiexin:
@senky One more comment: if the extension sets/changes u_action, this change is lost, and not actioned properly.

javiexin:
New proposal here https://github.com/phpbb/phpbb/pull/4578

PR #4501, opened 2016-11-03
[ticket/14716] Update dependencies to latest versions

marc1706:
Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-14716

This addresses some issues with symfony that resulted in
issues when open_basedir restrictions were enabled, as well
as issues with JPEG images in fast-image-size.
The phpBB class extending the twig lexer also had to be
modified to ensure compatibility after BC was broken by
the PR https://github.com/twigphp/Twig/pull/2182 for twig
1.27.

PHPBB3-14716

### DIFF
--- a/phpBB/composer.lock
+++ b/phpBB/composer.lock
@@ -346,16 +346,16 @@
         },
         {
             "name": "marc1706/fast-image-size",
-            "version": "v1.1.1",
+            "version": "v1.1.2",
             "source": {
                 "type": "git",
                 "url": "https://github.com/marc1706/fast-image-size.git",
-                "reference": "50ae9b13f862b0c7c6c9df057c396ea32837733a"
+                "reference": "8bb644f8f767c5763acd61140045a78d10e1eb06"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/marc1706/fast-image-size/zipball/50ae9b13f862b0c7c6c9df057c396ea32837733a",
-                "reference": "50ae9b13f862b0c7c6c9df057c396ea32837733a",
+                "url": "https://api.github.com/repos/marc1706/fast-image-size/zipball/8bb644f8f767c5763acd61140045a78d10e1eb06",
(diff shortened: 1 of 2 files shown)


marc1706:
Reported twig BC break as a separate issue: https://github.com/twigphp/Twig/issues/2218

blitze:
The update to the dependencies appears to have introduced some path issues with INCLUDECSS and INCLUDEJS when the page is accessed from an extension controller. For example, the official Pages extension adds css like so: . When I go to /phpBB/app.php/page/foo it is trying to include the css as /phpBB/**app.php/page/**ext/phpbb/pages/styles/prosilver/theme/pages_controller.css instead of /phpBB/ext/phpbb/pages/styles/prosilver/theme/pages_controller.css

Without  a full knowledge of how this all works, it appears that the fix is to precede the generated relative path of \phpbb\template\asset::set_path on line 188 with a leading './' like so: $this->components['path'] = './' . $path. This fixed the problem for me.

VSEphpbb:
@blitze I'm not able to reproduce that. However, Pages is not currently compatible with phpBB 3.2. Also `<!-- INCLUDECSS @phpbb_pages/../theme/pages_controller.css -->` is old and should instead try `<!-- INCLUDECSS @phpbb_pages/pages_controller.css -->`

VSEphpbb:
Oh wait, I am seeing this, with all INCLUDECSS, even just go to the FAQ page to see it happen.

marc1706:
Could either of you create a new ticket in the tracker (https://tracker.phpbb.com)? Then we can properly fix this before rolling out RC2.

marc1706:
Might actually be caused by the prior update to twig as that introduced relative path issues:
https://github.com/phpbb/phpbb/commit/7a5fbd0257d7b15b8230fa6c60581cc197cb1a47

marc1706:
We're reverting back to twig 1.26.* due to issues introduced in 1.27:
https://github.com/phpbb/phpbb/pull/4516

PR #4512, opened 2016-11-13
[ticket/14863] Properly treat plural rules in confirmation box title

senky:
Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-14863

PHPBB3-14863

### DIFF
--- a/phpBB/includes/functions.php
+++ b/phpBB/includes/functions.php
@@ -2776,7 +2776,7 @@ function confirm_box($check, $title = '', $hidden = '', $html_body = 'confirm_bo
 	$u_action .= ((strpos($u_action, '?') === false) ? '?' : '&amp;') . 'confirm_key=' . $confirm_key;
 
 	$template->assign_vars(array(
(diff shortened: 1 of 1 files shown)


Nicofuma:
I don't think you need the `1` argument. And at the same time, please replace all $user->lang array usage in the line

senky:
Argument is exactly the game changer here. Without it, you end up with situation ticket explains, or php warning about missing arguments to `sprintf` at best. While argument don't hinder strings without parameters, it resolves plural errors.

Nicofuma:
IIRC is already handled by the lang() method, even if you don't give any parameter

senky:
Method call without parameter:
```
[phpBB Debug] PHP Warning: in file [ROOT]/phpbb/user.php on line 513: sprintf(): Too few arguments
```

marc1706:
@Nicofuma I think this is only the case on 3.2.x and newer.

PR #4517, opened 2016-11-17
[ticket/14869] Allows DEBUG and DEBUG_ERROR_LEVEL to be used

milankubin:
Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-14869

Currently debug is enabled by default based off environment with no way to set log level

### DIFF
--- a/phpBB/common.php
+++ b/phpBB/common.php
@@ -96,9 +96,10 @@ include($phpbb_root_path . 'includes/functions_compatibility.' . $phpEx);
 require($phpbb_root_path . 'includes/constants.' . $phpEx);
 require($phpbb_root_path . 'includes/utf/utf_tools.' . $phpEx);
 
-if (PHPBB_ENVIRONMENT === 'development' || DEBUG === true)
+if (PHPBB_ENVIRONMENT === 'development' || (defined('DEBUG') && DEBUG === true) )
 {
-	\phpbb\debug\debug::enable(DEBUG_ERROR_LEVEL);
+	$log_level = (defined('DEBUG_PHP_ERROR_LEVEL') ? constant('DEBUG_PHP_ERROR_LEVEL') : null);
+	\phpbb\debug\debug::enable($log_level);
 }
 else
 {
--- a/phpBB/common.php
+++ b/phpBB/common.php
@@ -96,9 +96,9 @@ include($phpbb_root_path . 'includes/functions_compatibility.' . $phpEx);
 require($phpbb_root_path . 'includes/constants.' . $phpEx);
 require($phpbb_root_path . 'includes/utf/utf_tools.' . $phpEx);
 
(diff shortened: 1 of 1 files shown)


phpBB/common.php
```diff
@@ -96,9 +96,9 @@
 require($phpbb_root_path . 'includes/constants.' . $phpEx);
 require($phpbb_root_path . 'includes/utf/utf_tools.' . $phpEx);
 
-if (PHPBB_ENVIRONMENT === 'development')
+if (PHPBB_ENVIRONMENT === 'development' || DEBUG === true)
 {
-	\phpbb\debug\debug::enable();
+	\phpbb\debug\debug::enable(DEBUG_ERROR_LEVEL);
```
marc1706: I don't like passing possibly undefined constants to methods. How about replacing this with something like:
`defined('DEBUG_ERROR_LEVEL') ? DEBUG_ERROR_LEVEL : null`

marc1706:
Ỳour commit does not follow our commit message format. I'd appreciate it if you could fix that. Otherwise we'd change it upon merging in this case.

milankubin:
Hope it's ok now!

milankubin:
And all checks passed now. Have to remember its PHP >= 5.4 not >= 5.5 :)

marc1706:
I've talked with @Nicofuma about this change. The idea when adding the environments was to remove the use of constants for debugging etc. I do generally agree with that opinion. Is there a reason to not use the development in case you want to actually debug something?

milankubin:
This scenario happened when we got stuck debugging as something was happening on dev, but not showing on test.  This happened to be a NOTICE which at that time we'd like to have surpressed since it did not affect functionality.
So the need to be able to override the php log level was there.
I then found the function that turns debugging on also allows log level to be passed as an argument.
I think this is still a nice to have feature.

Maybe the need for DEBUG to be on or off based on constant is not so much there, but it is quite normal for configuration to dictate environment behavior and not for environment _**naming**_ to dictate configuration IMO.

VSEphpbb:
I agree with @marc1706 and @Nicofuma. This seems a step backwards.

milankubin:
I agree this was a quick hack, but it didn't seem far off the way common.php is coded. There is heavy use of config constants.
The obvious best would be to refactor that entire file and create a kernel loader class.
Run that in app.php
Then only load configs through the service container. 

You would be left with one constant: PHPBB_ENVIRONMENT, which should be set in $ENV

milankubin:
This feels very naked and 
includes/startup.php
''''php
// Report all errors, except notices and deprecation messages
$level = E_ALL & ~E_NOTICE & ~E_DEPRECATED;
error_reporting($level);
''''

marc1706:
Closing as dicussed earlier. Your note about startup.php is unrelated and can be discussed separately.

PR #4521, opened 2016-11-20
[ticket/11989] Rationalize overall_header_navigation* events

senky:
Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-11989

PHPBB3-11989

### DIFF
--- a/phpBB/docs/events.md
+++ b/phpBB/docs/events.md
@@ -1441,6 +1441,22 @@ overall_header_navigation_prepend
 * Since: 3.1.0-a1
 * Purpose: Add links before the navigation links in the header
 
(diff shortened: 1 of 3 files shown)


VSEphpbb:
This is a bad change. It's too late for this (it was proposed while 3.1 was still in development). 

You can't move template events around now. 

Especially one's as widely used and important as these. This will break a lot of extensions.

Instead you should add **new template events** in the locations where you moved the existing ones to.

DavidIQ:
+1 agree it's too late to be moving events around. 3.3 should bring about a new default template so it'll be more appropriate there.

PR #4530, opened 2016-11-20
[ticket/14557] Simplify updating overloaded events for extensions

senky:
Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-14557

PHPBB3-14557

### DIFF
--- a/phpBB/phpbb/event/data.php
+++ b/phpBB/phpbb/event/data.php
@@ -67,12 +67,12 @@ class data extends Event implements \ArrayAccess
 	/**
 	 * Returns data with updated key in specified offset.
 	 *
-	 * @param	string	$offset	Data array offset
-	 * @param	string	$key	Offset key
-	 * @param	mixed	$value	Value to update
+	 * @param	string	$subarray	Data array subarray
+	 * @param	string	$key		Subarray key
+	 * @param	mixed	$value		Value to update
 	 */
-	public function update_subarray($offset, $key, $value)
+	public function update_subarray($subarray, $key, $value)
 	{
-		$this->data[$offset][$key] = $value;
+		$this->data[$subarray][$key] = $value;
 	}
 }
--- a/phpBB/phpbb/event/data.php
+++ b/phpBB/phpbb/event/data.php
@@ -70,10 +70,9 @@ class data extends Event implements \ArrayAccess
 	 * @param	string	$offset	Data array offset
 	 * @param	string	$key	Offset key
 	 * @param	mixed	$value	Value to update
-	 * @return	mixed	Value of $value
 	 */
 	public function update_subarray($offset, $key, $value)
 	{
-		return $this->data[$offset][$key] = $value;
+		$this->data[$offset][$key] = $value;
 	}
 }
--- a/phpBB/phpbb/event/data.php
+++ b/phpBB/phpbb/event/data.php
@@ -63,4 +63,17 @@ class data extends Event implements \ArrayAccess
 	{
 		unset($this->data[$offset]);
 	}
+
+	/**
+	 * Returns data with updated key in specified offset.
+	 *
+	 * @param	string	$offset	Data array offset
+	 * @param	string	$key	Offset key
+	 * @param	mixed	$value	Value to update
+	 * @return	mixed	Value of $value
+	 */
+	public function update_subarray($offset, $key, $value)
+	{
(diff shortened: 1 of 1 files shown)


bantu:
I am not sure this is such a good idea. It seems to be inherently limited to one-dimensional arrays. Is there a solution that is not limited with regards to that?

senky:
You can always use `offsetSet()` to replace whole array. It's just helper. To make life easier at least a little bit...

Nicofuma:
I'm not against a such DX improvement, but it needs another name and a docblock imho (a name stating what it does).

senky:
`offsetKeyUpdate`?

3D-I:
offsetKeyUpdate looks fine to me, but that's me.
btw only one function in this file got to have the docblock,

Nicofuma:
yes, the others are either explicit (set_data) either magic methods from `\ArrayAccess` but they could have a docblock too. About the name I'd prefer a really explicit "phpBB" name, not a name looking like it's part of \ArrayAccess something like update_subarray or something like that. At the same time you will need to add some error control: `$this->data[$offset]` may not exist or not be an array

senky:
"phpBB" name and docblock added. I would rather not control any errors. If extension author tries to update subarray that is not array, it is always his bad, since event input is well documented and it's content is guaranteed.

VSEphpbb:
If you're gonna call it subarray then all occurrences of offset should be replaced by subarray (the var name and in the docblocks)

VSEphpbb:
Error control would be easy, something like:

```php
if (!is_array($this->data[$subarray])
{
    throw new \phpbb\exception\runtime_exception('INVALID_SUBARRAY', $subarray);
}
```
@Nicofuma ?

Even though it may ext author's bad, it helps if they can find out where they made a bad.

bantu:
Why not https://github.com/igorw/get-in/blob/master/src/get_in.php#L28 ?

phpBB/phpbb/event/data.php
```diff
@@ -63,4 +63,17 @@ public function offsetUnset($offset)
...
 		unset($this->data[$offset]);
 	}
+
+	/**
+	 * Returns data with updated key in specified offset.
+	 *
+	 * @param	string	$offset	Data array offset
+	 * @param	string	$key	Offset key
+	 * @param	mixed	$value	Value to update
+	 * @return	mixed	Value of $value
```
bantu: What is the purpose of this? Why is there a return value?

VSEphpbb: Read the ticket ;)

bantu: Ticket does not explain. Reason should be document in code anyway.

senky:
@bantu do you mean to rewrite to the same principle as `update_in` in igorw/get-in? `callable` instead of direct value?

phpBB/phpbb/event/data.php
senky: We either return nothing or `$value`. There is nothing more to that, I just followed what other methods in the class do.

VSEphpbb:
This should be in the event/data class, as that is already available. Shouldn't come from some external class or library that needs to be added.

phpBB/phpbb/event/data.php
VSEphpbb: It does not need to be returned @senky

bantu:
@VSEphpbb I am pretty sure adding get-in as a dependency is an option. I however mostly posted the link to show that it is not too hard to come up with an implementation that is not limited to one-dimensional arrays (as mentioned in my first post to this PR).

senky:
@bantu while adding get-in as a dependency is an option, adding new dependency should probably be discussed and approved. This ticket, however, just tries to make life a lot easier for simple tasks like replacing template variable. Maybe we could merge this with shallow subarray support and if need emerges in the future, we can refactor for multidimensional arrays.

VSEphpbb:
Adding get-in as a dependency is not an option @senky As @bantu said he posted it as a coding example

marc1706:
Bump @senky

senky:
@marc1706 I am not sure what's missing here from the discussion. Subarray support? Using get-in?

marc1706:
Was just wondering whether you wanted to change anything else since it seemed like the last comments kind of left things off.

PR #4537, opened 2016-11-22
Add ability for template events to be sortable

senky:
Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-11365

### DIFF
--- a/phpBB/includes/acp/acp_extensions.php
+++ b/phpBB/includes/acp/acp_extensions.php
@@ -34,7 +34,7 @@ class acp_extensions
 	function main($id, $mode)
 	{
 		// Start the page
-		global $config, $user, $template, $request, $phpbb_extension_manager, $db, $phpbb_root_path, $phpbb_log, $cache, $phpbb_dispatcher;
+		global $config, $user, $template, $request, $phpbb_extension_manager, $db, $phpbb_root_path, $phpbb_log, $cache, $phpbb_dispatcher, $table_prefix;
 
 		$this->db = $db;
 		$this->config = $config;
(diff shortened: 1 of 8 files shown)


senky:
!wip

senky:
!WIP

senky:
!set WIP

senky:
Comments, please :)

javiexin:
I like it although I have not tested it yet.
However, I think this should also be applied to core events, as it might be helpful on support situations.
One addition that I would suggest is a reverse search: clicking on an event should bring up the list of all extensions that are using such event.  Well, maybe I have a second suggestion: on each extension details page, add the list of subscribed events, both template and core.
Hope this helps,
-javiexin

Nicofuma:
TBH, I don't really like giving the user the ability to order the events. It is really powerful, gives a lot of power to the end user and I can see why they would do that. But.... there is a lot of drawbacks and the implementation won't be easy, especially if we want to be efficient.

And the UI can be challenging: the user needs to understand what is the purpose of all events and as there is a looots of them (371 right now) they need to be classified, ordered and clearly identified. Also not all events are good to be re-ordered (overall_header_head_append does not make sense per example). Basically we're gonna need to maintain a list of available events that can easily be parsed by the board and updated by extensions (an extension can add new events).

So to sum-up, I understand why users ask for that. But the implementation is gonna be challenging to be efficient and it will represent a lot of work to be really usable (especially at the beginning, once the system is done it can be easier).

@phpbb/development-team what do you think?

CHItA:
I agree with what Nicofuma said.

phpBB/includes/acp/acp_extensions.php
```diff
@@ -47,294 +47,309 @@ function main()
 
 		$user->add_lang(array('install', 'acp/extensions', 'migrator'));
```
3D-I: still the user object?

3D-I:
Guys, Javier is not anymore with us, life's sad sometimes. 
Please take care of his contributions here.

Thanks.
https://www.phpbb.com/community/viewtopic.php?p=14938226#p14938226

DavidIQ:
Yes we are aware and will be reviewing all of his pending prs and seeing if they can either be merged or taken over to be merged.

PR #4539, opened 2016-11-24
[ticket/14875] Add method for untrimmed input to ajax iohandler

marc1706:
Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-14875

Due to the pre-encoded input and the escaping of the input, the
string has to be decoded twice for the password.

PHPBB3-14875

### DIFF
--- a/phpBB/phpbb/install/helper/iohandler/ajax_iohandler.php
+++ b/phpBB/phpbb/install/helper/iohandler/ajax_iohandler.php
@@ -27,7 +27,7 @@ class ajax_iohandler extends iohandler_base
 	protected $path_helper;
 
 	/**
-	 * @var \phpbb\request\request_interface
+	 * @var \phpbb\request\request
 	 */
 	protected $request;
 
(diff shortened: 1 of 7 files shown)


Nicofuma:
Couldn't we have a `get_raw_input` or something like that?
(by the way, I wonder what happen if we have an UT8 chars in the password...)

phpBB/phpbb/install/helper/iohandler/ajax_iohandler.php
```diff
@@ -90,12 +90,12 @@ class ajax_iohandler extends iohandler_base
...
 	 *
 	 * @param path_helper						$path_helper
-	 * @param \phpbb\request\request_interface	$request	HTTP request interface
+	 * @param \phpbb\request\request	$request	HTTP request interface
 	 * @param \phpbb\template\template			$template	Template engine
 	 * @param router 							$router		Router
 	 * @param string 							$root_path	Path to phpBB's root
 	 */
-	public function __construct(path_helper $path_helper, \phpbb\request\request_interface $request, \phpbb\template\template $template, router $router, $root_path)
+	public function __construct(path_helper $path_helper, \phpbb\request\request $request, \phpbb\template\template $template, router $router, $root_path)
```
CHItA: Why exactly do you need this?

marc1706: The `raw_variable` method is only part of the request class but not of the request_interface. This was a conscious choice as some other variables of request also are not part of the request_interface, mostly ones that are usually not required for a normal page operation (which IMHO also applies to raw input variables).

Nicofuma: personally I think it could be part of the interface

Nicofuma: I don't thinks there is anyone implementing this interface. And actually I don't thinks we should have one but it does not really matter.

marc1706: Ok, moving it to the interface then.

PR #4547, opened 2016-11-30
[ticket/14883] Do not display empty forum rules (after they've been re-parsed)

VSEphpbb:
Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-14883

PHPBB3-14883

### DIFF
--- a/phpBB/includes/functions_display.php
+++ b/phpBB/includes/functions_display.php
@@ -739,9 +739,9 @@ function generate_forum_rules(&$forum_data)
 	global $template;
 
 	$template->assign_vars(array(
-			'S_FORUM_RULES'	=> true,
-			'U_FORUM_RULES'	=> $forum_data['forum_rules_link'],
-			'FORUM_RULES'	=> $forum_data['forum_rules'])
+		'S_FORUM_RULES'	=> true,
+		'U_FORUM_RULES'	=> $forum_data['forum_rules_link'],
+		'FORUM_RULES'	=> $forum_data['forum_rules'])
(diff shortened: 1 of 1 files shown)


VSEphpbb:
The text reparser will store empty messages as `<t></t>` and when parsing them, will convert them back to empty `''` for display.

The forum rules checked if their was content to display before parsing, so it considered `<t></t>` to be displayable text. So we need to check for empty forum rules after it has been parsed into displayable text.

phpBB/includes/functions_display.php
```diff
@@ -738,6 +738,11 @@ function generate_forum_rules(&$forum_data)
 		$forum_data['forum_rules'] = generate_text_for_display($forum_data['forum_rules'], $forum_data['forum_rules_uid'], $forum_data['forum_rules_bitfield'], $forum_data['forum_rules_options']);
 	}
 
+	if (!$forum_data['forum_rules'])
+	{
+		return;
```
marc1706: As far as I can see, this used to display the forum rules if rules_link is set. Maybe just check
```if ((!$forum_data['forum_rules'] || $forum_data['forum_rules'] === '<t></t>') && !$forum_data['forum_rules_link'])```
instead of modifying the logic here. This will now also return if forum link is set and forum rules isn't. Something it didn't use to do previously.

VSEphpbb:
@marc1706 I think this is a better way... no dependency on coding the `<t></t>` into the logic which you never know, could change one day.

PR #4552, opened 2016-12-02
[ticket/13757] Properly fix bug with negative PM count

viteksafronov:
Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-13757

The fix from PR #4548 doesn't work as expected because
in phpBB the sql_affectedrows() method always returns
a count of matched rows instead of affected rows.
This commit fixes bug properly.

PHPBB3-13757

### DIFF
--- a/phpBB/includes/functions_privmsgs.php
+++ b/phpBB/includes/functions_privmsgs.php
@@ -889,7 +889,8 @@ function update_unread_status($unread, $msg_id, $user_id, $folder_id)
 		SET pm_unread = 0
 		WHERE msg_id = $msg_id
 			AND user_id = $user_id
-			AND folder_id = $folder_id";
+			AND folder_id = $folder_id
+			AND pm_unread = 1";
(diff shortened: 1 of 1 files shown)


viteksafronov:
We should check if this additional condition breaks anything or not. I think it shouln't.

phpBB/includes/functions_privmsgs.php
```diff
@@ -888,6 +888,7 @@ function update_unread_status($unread, $msg_id, $user_id, $folder_id)
 	$sql = 'UPDATE ' . PRIVMSGS_TO_TABLE . "
 		SET pm_unread = 0
 		WHERE msg_id = $msg_id
+			AND pm_unread = 1
```
marc1706: Please move this below the `AND folder_id`. It should be marginally faster if it's at bottom cause it's faster at selecting the proper primary key in this query.

PR #3717, opened 2015-06-19
[ticket/13853] Flexible schema for profilefields step 1 configuration

javiexin:
Modify the way adm/style/acp_profile.html STEP 1 (lines 106 to 128)
is coded to make it flexible and expandable to other new types of
profile fields, rather than the current mechanism, hardcoded for
the default types.

A flexible mechanism similar to the one used of STEP 2 is used.

This required a slight modification of the type_interface, and
corresponding modifications to all implementations of that interface.

Includes a fix in the type_interface, missing a required method.

PHPBB3-13853

PR #4554, opened 2016-12-03
[ticket/13853] Flexible schema for profilefields step 1 configuration

javiexin:
Modify the way adm/style/acp_profile.html STEP 1 (lines 106 to 128)
is coded to make it flexible and expandable to other new types of
profile fields, rather than the current mechanism, hardcoded for
the default types.

A flexible mechanism similar to the one used of STEP 2 is used.

This required a slight modification of the type_interface, and
corresponding modifications to all implementations of that interface.

Includes a fix in the type_interface, missing a required method.

Checklist:

- [ ] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-13853

### DIFF
--- a/phpBB/phpbb/profilefields/type/type_bool.php
+++ b/phpBB/phpbb/profilefields/type/type_bool.php
@@ -395,7 +395,7 @@ class type_bool extends type_base
 	/**
 	* {@inheritDoc}
 	*/
-	public function display_options(&$template_vars, &$field_data)
+	public function display_options($action, &$field_data)
 	{
 		// Initialize these array elements if we are creating a new field
 		if (!sizeof($field_data['lang_options']))
@@ -405,11 +405,14 @@ class type_bool extends type_base
 			$field_data['lang_options'][1] = '';
 		}
 
-		$template_vars = array_merge($template_vars, array(
-			'S_BOOL'					=> true,
-			'L_LANG_OPTIONS_EXPLAIN'	=> $this->user->lang['BOOL_ENTRIES_EXPLAIN'],
-			'FIRST_LANG_OPTION'			=> $field_data['lang_options'][0],
-			'SECOND_LANG_OPTION'		=> $field_data['lang_options'][1],
-		));
+		$doptions = array(
+			0 => array(
+					'TITLE' => $this->user->lang['ENTRIES'],
+					'EXPLAIN' => $this->user->lang['BOOL_ENTRIES_EXPLAIN'],
+					'FIELD' => '<input class="medium" id="lang_options" name="lang_options[0]" value="' . $field_data['lang_options'][0] . '" /> ' . $this->user->lang['FIRST_OPTION'] . '</dd><dd><input class="medium" name="lang_options[1]" value="' . $field_data['lang_options'][1] . '" /> ' . $this->user->lang['SECOND_OPTION']
+				),
+		);
+
+		return $doptions;
 	}
 }
--- a/phpBB/phpbb/profilefields/type/type_dropdown.php
+++ b/phpBB/phpbb/profilefields/type/type_dropdown.php
@@ -307,7 +307,7 @@ class type_dropdown extends type_base
 	/**
 	* {@inheritDoc}
 	*/
-	public function display_options(&$template_vars, &$field_data)
+	public function display_options($action, &$field_data)
 	{
 		// Initialize these array elements if we are creating a new field
 		if (!sizeof($field_data['lang_options']))
@@ -316,10 +316,14 @@ class type_dropdown extends type_base
 			$field_data['lang_options'] = array();
 		}
 
-		$template_vars = array_merge($template_vars, array(
-			'S_DROPDOWN'				=> true,
-			'L_LANG_OPTIONS_EXPLAIN'	=> $this->user->lang['DROPDOWN_ENTRIES_EXPLAIN'],
-			'LANG_OPTIONS'				=> implode("\n", $field_data['lang_options']),
-		));
+		$doptions = array(
+			0 => array(
+					'TITLE' => $this->user->lang['ENTRIES'],
+					'EXPLAIN' => $this->user->lang[($action == 'edit') ? 'EDIT_DROPDOWN_LANG_EXPLAIN' : 'DROPDOWN_ENTRIES_EXPLAIN'],
+					'FIELD' => '<textarea id="lang_options" name="lang_options" rows="5" cols="80">' . implode("\n", $field_data['lang_options']) . '</textarea>'
+				),
+		);
+
+		return $doptions;
 	}
 }
--- a/phpBB/phpbb/profilefields/type/type_interface.php
+++ b/phpBB/phpbb/profilefields/type/type_interface.php
@@ -134,6 +134,14 @@ interface type_interface
 	public function get_field_ident($field_data);
 
 	/**
+	* Get the localized name of the field
+	*
+	* @param string $field_name		Unlocalized name of this field
+	* @return string 	Localized name of the field
+	*/
+	public function get_field_name($field_name);
+
+	/**
 	* Get the column type for the database
 	*
 	* @return string	Returns the database column type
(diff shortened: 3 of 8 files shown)


javiexin:
This is a cleanup of https://github.com/phpbb/phpbb/pull/3717 removing some commits that were creating errors.

Please, bear in mind that the original submission of this patch dates back to 19 Jun 2015, so 18 months ago.

DavidIQ:
Isn't this a new feature for a feature-frozen version of phpBB? Should go against 3.2 not 3.1, which was said in the old pr as well.

javiexin:
No, I do not consider this a new feature, as without these changes it is completely impossible to add custom profile field types because of the hardcoding involved in acp_profile.php.
For me, this is a blocking bug in the original profile field implementation of 3.1, that should be fixed.
And besides, it has been open for 18 months...  And the extension that uses it, also, for as long as that...

DavidIQ:
This seems more of a change just to make an extension work than a change to fix an actual bug. Extending code and then finding out the system does not like what you're doing because it was not designed for what you are doing is neither a bug nor a blocker. If you're changing function definitions I would consider that a blocker for the pull request as there is no guarantee that someone hasn't already been using these functions, either in an extension or in custom code. This is definitely a new feature since it is changing behavior of an existing feature (we've seen this sort of thing come back to haunt us before).

This might be swell for 3.2 but my opinion is that this should not go into 3.1 due to what was stated in the previous pr and the reasons stated above. I'll let someone else make the final decision.

javiexin:
Sorry, but I must disagree. This is more like another kind of "event", something that is added to allow injecting code in certain places, in a flow-based development thread, unlike this one, that fixes an ill-defined class definition in a class-based extension system.  It is required for ANY extension that adds a profile field type. There is NO way to add anything in step 1 beyond the original types. Just try to do it.

javiexin:
Looking at it again, I think this will also help in case someone wants to modify the way things are displayed, as the "options" are exposed to extensions through the core.acp_profile_create_edit_after event.  If the code remains as is, the variables are hardcoded to the template, and not exposed to extensions for potential modification.  The only way to modify them would be to reproduce the code in each and every type (that needs modification) within the extension...  This may be even more generalized, if required, by exposing the regular template variables (in addition to block/type dependent variables, already taken care of) before being sent to the template...

phpBB/phpbb/profilefields/type/type_bool.php
```diff
@@ -405,11 +405,14 @@ public function display_options(&$template_vars, &$field_data)
 			$field_data['lang_options'][1] = '';
 		}
 
-		$template_vars = array_merge($template_vars, array(
-			'S_BOOL'					=> true,
-			'L_LANG_OPTIONS_EXPLAIN'	=> $this->user->lang['BOOL_ENTRIES_EXPLAIN'],
-			'FIRST_LANG_OPTION'			=> $field_data['lang_options'][0],
-			'SECOND_LANG_OPTION'		=> $field_data['lang_options'][1],
-		));
+		$doptions = array(
```
marc1706: Reminds me a bit of hungarian notation. Just write out the actual name: `$display_options`

phpBB/phpbb/profilefields/type/type_bool.php
```diff
@@ -405,11 +405,14 @@ public function display_options(&$template_vars, &$field_data)
...
+			0 => array(
```
marc1706: You don't need to specify 0 here as that is the default PHP behavior.

phpBB/phpbb/profilefields/type/type_bool.php
```diff
@@ -405,11 +405,14 @@ public function display_options(&$template_vars, &$field_data)
...
+					'TITLE' => $this->user->lang['ENTRIES'],
+					'EXPLAIN' => $this->user->lang['BOOL_ENTRIES_EXPLAIN'],
+					'FIELD' => '<input class="medium" id="lang_options" name="lang_options[0]" value="' . $field_data['lang_options'][0] . '" /> ' . $this->user->lang['FIRST_OPTION'] . '</dd><dd><input class="medium" name="lang_options[1]" value="' . $field_data['lang_options'][1] . '" /> ' . $this->user->lang['SECOND_OPTION']
```
marc1706: The goal should be to get rid of HTML in PHP files and not the other way round. It would IMHO make more sense to be able to specify an HTML file that contains the base code for a profile field instead of outputting HTML from PHP.

phpBB/phpbb/profilefields/type/type_dropdown.php
```diff
@@ -316,10 +316,14 @@ public function display_options(&$template_vars, &$field_data)
...
-		$template_vars = array_merge($template_vars, array(
-			'S_DROPDOWN'				=> true,
-			'L_LANG_OPTIONS_EXPLAIN'	=> $this->user->lang['DROPDOWN_ENTRIES_EXPLAIN'],
-			'LANG_OPTIONS'				=> implode("\n", $field_data['lang_options']),
-		));
+		$doptions = array(
+			0 => array(
+					'TITLE' => $this->user->lang['ENTRIES'],
+					'EXPLAIN' => $this->user->lang[($action == 'edit') ? 'EDIT_DROPDOWN_LANG_EXPLAIN' : 'DROPDOWN_ENTRIES_EXPLAIN'],
+					'FIELD' => '<textarea id="lang_options" name="lang_options" rows="5" cols="80">' . implode("\n", $field_data['lang_options']) . '</textarea>'
```
marc1706: Same issue as explained above.

phpBB/phpbb/profilefields/type/type_interface.php
```diff
@@ -200,11 +208,11 @@ public function prepare_hidden_fields($step, $key, $action, &$field_data);
...
 	* Allows assigning of additional template variables
 	*
-	* @param array	$template_vars	Template variables we are going to assign
+	* @param string	$action			Currently performed action (create|edit)
 	* @param array	$field_data		Array with data for this field
-	* @return null
+	* @return array	with the acp options for step 1
 	*/
-	public function display_options(&$template_vars, &$field_data);
+	public function display_options($action, &$field_data);
```
marc1706: Both of these changes break BC with previous implementations. Have you assessed how many extensions are affected by this change?

phpBB/phpbb/profilefields/type/type_bool.php
javiexin: Ok

phpBB/phpbb/profilefields/type/type_bool.php
javiexin: I am doing the same as here:
https://github.com/phpbb/phpbb/blob/master/phpBB/phpbb/profilefields/type/type_bool.php#L85

I agree that that should be the goal, but I think that would need a bigger change in the profile fields code.  Also affecting the language specific part (step 3).

A similar schema as used here might be possible as well:
https://github.com/phpbb/phpbb/blob/master/phpBB/phpbb/profilefields/type/type_base.php#L185-L205

phpBB/phpbb/profilefields/type/type_interface.php
javiexin: Previous implementation of what?  With the proposed changes, the core code remains 100% functionally compatible, with no implications whatsoever, and no changes to the db either.  And as for the potential for some extensions to have used this...  the only one I have been able to find is my own Advanced Profile Field Types that triggered all these changes.  That does not mean that someone somewhere might have done something that might be affected, but up to what I have been able to find, nothing in phpbb forum is affected (again, as far as I have been able to see...).

javiexin:
@marc1706 Commented on your requested changes

phpBB/phpbb/profilefields/type/type_interface.php
javiexin: Ok, I can make a proposal that is 100% BC, but is a kind of spaguetti code. Stay tuned.

javiexin:
I have decided to withdraw this PR.

The changes to profile fields should be major, in order to avoid the use of HTML from PHP code (already existing), and even though this code is not "clean" in my view, it works.
So let's keep it from now, and maybe we will revisit it for 3.3 in a more complete way.

PR #4572, opened 2016-12-09
[ticket/14855] Update notification and PM alert bubbles

javiexin:
Checklist:

- [ ] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-14855

Alternative implementation without removing notification from
the DOM, but changing the class.  So that style designers
might have more options (show/not show).

PHPBB3-14855

### DIFF
--- a/phpBB/styles/prosilver/theme/common.css
+++ b/phpBB/styles/prosilver/theme/common.css
@@ -1269,6 +1269,10 @@ ul.linklist:after,
 	padding: 4px 6px;
 }
 
+.badge.hidden {
+	display: none;
+}
+
 /* Navbar specific list items
 ----------------------------------------*/
 
--- a/tests/functional/notification_test.php
+++ b/tests/functional/notification_test.php
@@ -82,6 +82,7 @@ class phpbb_functional_notification_test extends phpbb_functional_test_case
 		// Get form token
 		$link = $crawler->selectLink($this->lang('NOTIFICATIONS_MARK_ALL_READ'))->link()->getUri();
 		$crawler = self::request('GET', substr($link, strpos($link, 'ucp.')));
-		$this->assertEquals(0, $crawler->filter('#notification_list_button strong.badge.hidden')->text());
+		$this->assertCount(1, $crawler->filter('#notification_list_button strong.badge.hidden'));
+		$this->assertEquals("0", $crawler->filter('#notification_list_button strong.badge.hidden')->text());
 	}
 }
--- a/tests/functional/notification_test.php
+++ b/tests/functional/notification_test.php
@@ -82,6 +82,6 @@ class phpbb_functional_notification_test extends phpbb_functional_test_case
 		// Get form token
 		$link = $crawler->selectLink($this->lang('NOTIFICATIONS_MARK_ALL_READ'))->link()->getUri();
 		$crawler = self::request('GET', substr($link, strpos($link, 'ucp.')));
-		$this->assertCount(0, $crawler->filter('#notification_list_button strong'));
+		$this->assertEquals(0, $crawler->filter('#notification_list_button strong.badge.hidden')->text());
 	}
 }
--- a/phpBB/styles/prosilver/template/ajax.js
+++ b/phpBB/styles/prosilver/template/ajax.js
@@ -132,9 +132,10 @@ phpbb.markNotifications = function($popup, unreadCount) {
 
 	// Update the unread count.
 	$('strong', '#notification_list_button').html(unreadCount);
-	// Remove the Mark all read link & notification count if there are no unread notifications.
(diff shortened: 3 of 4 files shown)


javiexin:
Building on https://github.com/phpbb/phpbb/pull/4504 to allow for style desgners to change the way no-notifications are shown.  With the approved patch, the ONLY possible option is to NOT have an empty set of notifications (value 0).  This variant does NOT remove any object, but adds a class to facilitate controlling the display via CSS rules, much simpler than editing a complex JS code.
As presented, behaviour should be 100% identical to the former.  Only in the event someone wants to take a different route, it would be possible by changing CSS.

javiexin:
The logged errors are precisely the cause of the PR: in case no notifications, no mark exists, neither visible nor hidden, it was destroyed.  Should check if notification bubble is SHOWN, not if it exists at all.

DavidIQ:
You actually need to fix the tests yourself since you're changing how it works so your changes are what actually break the build.

javiexin:
Done, now builds correctly.  Please, review my change in tests, as it is my first one, and not sure if done the way it is supposed to be.

phpBB/styles/prosilver/theme/common.css
```diff
@@ -1269,6 +1269,10 @@ ul.linklist:after,
 	padding: 4px 6px;
 }
 
+.badge.hidden {
+	display: none;
+}
+
```
VSEphpbb: This is not needed. There's already a hidden CSS rule in common.css.

tests/functional/notification_test.php
```diff
@@ -82,6 +82,6 @@ public function test_mark_notifications_read()
 		// Get form token
 		$link = $crawler->selectLink($this->lang('NOTIFICATIONS_MARK_ALL_READ'))->link()->getUri();
 		$crawler = self::request('GET', substr($link, strpos($link, 'ucp.')));
-		$this->assertCount(0, $crawler->filter('#notification_list_button strong'));
+		$this->assertEquals(0, $crawler->filter('#notification_list_button strong.badge.hidden')->text());
```
javiexin: But then you would not be checking for the actual contents, just the existence of the element with the class, right?
What I wanted to achieve is testing both the right element and class, and the actual content (0).
What os the right way to do that?

phpBB/styles/prosilver/theme/common.css
javiexin: Yes, but it is overwritten by the "badge" rule.

VSEphpbb: Oh I see.

tests/functional/notification_test.php
javiexin: Ok, will fix and resubmit later today

phpBB/styles/prosilver/theme/common.css
javiexin: If you prefer, it is possible to move the badge rule before the hidden rule, and that way, we would not need the extra rule. But I figured this would provide a hint to developers not to override the generic hidden rule (used elsewhere) but to modify this one, much more specific.

VSEphpbb: No, just keep what you have. It's a separate matter but our .hidden class should probably have an !important attribute for these kinds of situations.

javiexin: Agree with both, but if the !important attribute is added, then this would be redundant

VSEphpbb: Right, that's why i said it's a separate issue... not for this PR. Personally I'd like to just use the existing hidden rule, but I think the styles team may not like using !important.

DavidIQ: Pretty sure @hanakin is trying to move away from those.

javiexin: Oops... And it would block the ability to customize... Unless another !important is thrown in, not so nice... Maybe use a different class name?

hanakin: it should be important ideally but the issue is where it falls in the layering of rules, all important rules should be the last loaded in a utilities or trumps layer to solve inheritance issues. The main problem with phpbb hidden class is that the class is spiderweb'd in weird uses throughout the theme and the js improperly so we can not use the hidden class with important without causing major issues. for now its fine as is all though not optimal until we relocate the utilities layer to the end of the stylesheets.css file there is no way around it

javiexin:
Changed the tests, but now one of the execution paths fails, while the other work as expected.  However, the error it seems unrelated.  Could please someone check? My knowledge about the test framework is soooo limited...

DavidIQ:
Let's just restart it as it was just a curl timeout error.

PR #4578, opened 2016-12-11
[ticket/14849] Add core.acp_extensions_run_action

javiexin:
Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-14849

Moved the event to a slightly modified position to perform common
checks before calling the event, added an extra variable to allow
execution control, honor the extension modified value of u_action
(if any)
Added an extra event after execution of action has completed
to avoid having to duplicate multiple code checks.

PHPBB-14849

### DIFF
--- a/phpBB/includes/acp/acp_extensions.php
+++ b/phpBB/includes/acp/acp_extensions.php
@@ -71,19 +71,20 @@ class acp_extensions
 		/**
 		* Event to run a specific action on extension
 		*
-		* @event core.acp_extensions_run_action
-		* @var	string	action			Action to run; if the event executes completely the action, should be set to 'none'
+		* @event core.acp_extensions_run_action_before
+		* @var	string	action			Action to run; if the event completes execution of the action, should be set to 'none'
 		* @var	string	u_action		Url we are at
 		* @var	string	ext_name		Extension name from request
 		* @var	int		safe_time_limit	Safe limit of execution time
 		* @var	int		start_time		Start time
 		* @var	string	tpl_name		Template file to load
 		* @since 3.1.11-RC1
+		* @changed 3.2.1-RC1			Renamed to core.acp_extensions_run_action_before, added tpl_name, added action 'none'
 		*/
 		$u_action = $this->u_action;
 		$tpl_name = '';
 		$vars = array('action', 'u_action', 'ext_name', 'safe_time_limit', 'start_time', 'tpl_name');
-		extract($this->phpbb_dispatcher->trigger_event('core.acp_extensions_run_action', compact($vars)));
+		extract($this->phpbb_dispatcher->trigger_event('core.acp_extensions_run_action_before', compact($vars)));
 
 		// In case they have been updated by the event
 		$this->u_action = $u_action;
--- a/phpBB/includes/acp/acp_extensions.php
+++ b/phpBB/includes/acp/acp_extensions.php
@@ -361,6 +361,7 @@ class acp_extensions
 		* @since 3.1.11-RC1
 		*/
 		$u_action = $this->u_action;
+		$tpl_name = $this->tpl_name;
 		$vars = array('action', 'u_action', 'ext_name', 'safe_time_limit', 'start_time', 'tpl_name');
 		extract($this->phpbb_dispatcher->trigger_event('core.acp_extensions_run_action_after', compact($vars)));
 
--- a/phpBB/includes/acp/acp_extensions.php
+++ b/phpBB/includes/acp/acp_extensions.php
@@ -72,12 +72,12 @@ class acp_extensions
 		* Event to run a specific action on extension
 		*
 		* @event core.acp_extensions_run_action
-		* @var	string	action			Action to run
+		* @var	string	action			Action to run; if the event executes completely the action, should be set to 'none'
 		* @var	string	u_action		Url we are at
 		* @var	string	ext_name		Extension name from request
 		* @var	int		safe_time_limit	Safe limit of execution time
 		* @var	int		start_time		Start time
-		* @var	string	tpl_name		Template file to load; leave empty to continue execution, filled in if ready to finish
+		* @var	string	tpl_name		Template file to load
 		* @since 3.1.11-RC1
 		*/
 		$u_action = $this->u_action;
@@ -88,12 +88,6 @@ class acp_extensions
 		// In case they have been updated by the event
 		$this->u_action = $u_action;
 		$this->tpl_name = $tpl_name;
-
-		// If tpl_name was set by the prior event, we are done
-		if ($tpl_name)
-		{
-			return;
-		}
 
 		// If they've specified an extension, let's load the metadata manager and validate it.
 		if ($ext_name)
@@ -113,6 +107,10 @@ class acp_extensions
 		// What are we doing?
 		switch ($action)
 		{
+			case 'none':
+				// Intentionally empty, used by extensions that execute additional actions in the prior event
+				break;
+
 			case 'set_config_version_check_force_unstable':
 				$force_unstable = $this->request->variable('force_unstable', false);
 
--- a/phpBB/includes/acp/acp_extensions.php
+++ b/phpBB/includes/acp/acp_extensions.php
@@ -56,21 +56,6 @@ class acp_extensions
 		$safe_time_limit = (ini_get('max_execution_time') / 2);
 		$start_time = time();
 
-		/**
-		* Event to run a specific action on extension
-		*
-		* @event core.acp_extensions_run_action
-		* @var	string	action			Action to run
-		* @var	string	u_action		Url we are at
-		* @var	string	ext_name		Extension name from request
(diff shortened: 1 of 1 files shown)


javiexin:
Slightly modified from https://github.com/phpbb/phpbb/pull/4497

javiexin:
!set Event

phpBB/includes/acp/acp_extensions.php
```diff
@@ -65,22 +77,22 @@ function main()
 		* @var	string	ext_name		Extension name from request
 		* @var	int		safe_time_limit	Safe limit of execution time
 		* @var	int		start_time		Start time
+		* @var	string	tpl_name		Template file to load; leave empty to continue execution, filled in if ready to finish
```
javiexin: Additional var to allow execution control: if ready to move to next action, it should be set from the extension.  Empty means continue executing here.

phpBB/includes/acp/acp_extensions.php
```diff
@@ -335,6 +347,27 @@ function main()
...
+		* @since 3.1.11-RC1
+		*/
+		$u_action = $this->u_action;
+		$tpl_name = $this->tpl_name;
+		$vars = array('action', 'u_action', 'ext_name', 'safe_time_limit', 'start_time', 'tpl_name');
+		extract($this->phpbb_dispatcher->trigger_event('core.acp_extensions_run_action_after', compact($vars)));
+
+		// In case they have been updated by the event
+		$this->u_action = $u_action;
+		$this->tpl_name = $tpl_name;
```
javiexin: Additional event, after the execution of the main action.  To enable the possibility to modify output, or perform additional actions.

phpBB/includes/acp/acp_extensions.php
marc1706: Please add a `@changed` tag to show what was added and how this event was changed (e.g. the location of the other checks).

phpBB/includes/acp/acp_extensions.php
```diff
@@ -65,22 +77,22 @@ function main()
...
-		{
-			$action = 'list';
-			$ext_name = '';
-		}
+		// In case they have been updated by the event
+		$this->u_action = $u_action;
+		$this->tpl_name = $tpl_name;
 
-		if (in_array($action, array('enable', 'disable', 'delete_data')) && !check_link_hash($request->variable('hash', ''), $action . '.' . $ext_name))
+		// If tpl_name was set by the prior event, we are done
```
marc1706: Somewhat unexpected side effect. Not sure I like this. Couldn't you simply modify the action to have pretty much the same happening?

javiexin: That is the whole intention of this new event var, to be able to tell how to move on to the next step.  If you look closely, the last thing that is done in all the cases is setting the tpl_name and template vars for the followon action.
With the current situation, whatever you do in the event, you will come back AND execute the main switch, whatever is the action (default is the same as 'list').  So, the only way to tell the method that it has finished is to do something like this.  Otherwise, the only way to produce the same effect would be to call a "refresh" directly from the event, not too nice in my point of view.
But if you propose an alternative as to how to completely avoid the execution of the main switch (set "default" switch action to break, for example, instead of 'list', or create a new action called 'noaction', or whatever), I would be fine as well.
To be honest, I do not need this in the extension I am planning, but might be useful for others...

javiexin: I don't know how to make an "alternative" proposal, so here it goes:
- Remove lines 92 to 96 of the already modified file
- Add in line 116:
// This action indicates nothing else to do; used from extensions willing to add extra actions, that are dealt with in the prior event
case 'none':
break;

javiexin:
@marc1706 Please review my comments and provide guidance as to how to progress.

phpBB/includes/acp/acp_extensions.php
Nicofuma: ok fot this one

phpBB/includes/acp/acp_extensions.php
javiexin: I will implement this case and submit a new commit.  It seems clearer than the other alternative.

phpBB/includes/acp/acp_extensions.php
```diff
@@ -65,22 +77,22 @@ function main()
...
 		* @var	int		safe_time_limit	Safe limit of execution time
 		* @var	int		start_time		Start time
+		* @var	string	tpl_name		Template file to load; leave empty to continue execution, filled in if ready to finish
 		* @since 3.1.11-RC1
 		*/
 		$u_action = $this->u_action;
-		$vars = array('action', 'u_action', 'ext_name', 'safe_time_limit', 'start_time');
+		$tpl_name = '';
+		$vars = array('action', 'u_action', 'ext_name', 'safe_time_limit', 'start_time', 'tpl_name');
 		extract($this->phpbb_dispatcher->trigger_event('core.acp_extensions_run_action', compact($vars)));
```
javiexin: Rename, or not?

javiexin:
@marc1706  Already implemented requested changes. Now that 3.2 is out, anything else needed for these to be accepted?

javiexin:
Now, with the publication of 3.2, we do have an issue with this event: it is already "public", but not on 3.1.11RC1, as indicated in the "since" clause, but from 3.2.0, that has come out before 3.1.11RC1 even.  However, the originally proposed changes have been merged.

So, now, we should document the changes in a "changed" clause, and maybe even reconsider changing the event at all...  We should keep the other event though (the _after event) and the new param for the original event, that would not break BC.

Opinions?  Should I submit a new commit?  What "changed" message should I include (changed 3.1.11RC1 seems kind of odd when the since 3.1.11RC1 is already included)?.  We should incorporate this asap, so that this inconsistency does not stay too long in the code, creating extra potential issues.

javiexin:
What is blocking this PR?
Can I help in any way?

phpBB/includes/acp/acp_extensions.php
marc1706: As it can be used to override the actions being run I'm ok with it staying like that.

marc1706: If you want to still change it that'll be alright, too. Will probably be slightly clearer with the second event further down.

javiexin: Ok, will change it, clearer, but, how do I document this change?
@since 3.2.0 original name (already published in its current form)
@changed 3.2.1 rename (would change for 3.2.1 with new name)
@since 3.1.11 new name (new name is first name for 3.1 branch)
Note that these also applies even if the name is not changed: the changes of the checks (before and after the event), additional param tpl_name, action none should be documented for 3.2.
How do I add them?

javiexin: Ok, will change it, clearer, but, how do I document this change?
@since 3.2.0 original name (already published in its current form)
@changed 3.2.1 rename (would change for 3.2.1 with new name)
@since 3.1.11 new name (new name is first name for 3.1 branch)
Note that these also applies even if the name is not changed: the changes of the checks (before and after the event), additional param tpl_name, action none should be documented for 3.2.
How do I add them?

marc1706: Use `@changed 3.2.1 Renamed to core.acp_extensions_run_action_before`

javiexin: Done

javiexin:
@senky Please, review, given that you submitted the original event.  Now renamed, slightly changed in position, additional parameter, new action, and new event to complement.

senky:
Looks good to me.

CHItA:
@marc1706 I guess if it cannot get merged in this state then it should be closed, so could you have a look?

javiexin:
The requested changes had been addressed more than a month ago...  The last change was to add an @changed clause in the docblock due to the rename of the event, accepted by Marc 9 days ago (it is in one of the outdated comments, as I pushed a commit with these changes that same day).

Elsensee:
Changing the name and location of an existing event that some extensions might use is something we should do with caution.
Okay, to be honest, the location change is not that bad, I think it's actually better and extensions might still work but the name change is... weird.

Not so sure, but just leave the name and otherwise it looks good to me? @marc1706 @CHItA

javiexin:
The name change is due to the fact that there are TWO events, one "_before" and one "_after" the actual ACP extension action.  The change was proposed well before the event was published, and anyhow, it was accepted by Marc 10 days ago, that is when the name was actually changed/pushed (AFTER Marc accepted the change).  Anyhow, I am fine if we go back to the old name.  It will be less clear, but fine as well.
By the way, the original proposer of this event (Senky) is OK with all these changes, so if he would be using the event, he is informed AND ok with the name change as well.  And the even was published with 3.2.0 only, and it only affects that version.  In fact, in its docblock says "since 3.1.11-RC1"... so formally, there is NO name change for 3.1 as it has not been published in the 3.1 branch yet...

PR #4592, opened 2016-12-26
[ticket/14938] Inconsistency in ext_mgr all_available vs is_available

javiexin:
Made is_available much more strict, in line with the checks in all_available

PHPBB3-14938

Checklist:

- [ ] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-14938

### DIFF
--- a/phpBB/phpbb/extension/manager.php
+++ b/phpBB/phpbb/extension/manager.php
@@ -433,25 +433,11 @@ class manager
 			if ($file_info->isFile() && $file_info->getFilename() == 'composer.json')
 			{
 				$ext_name = $iterator->getInnerIterator()->getSubPath();
-				$composer_file = $iterator->getPath() . '/composer.json';
-
-				// Ignore the extension if there is no composer.json.
-				if (!is_readable($composer_file) || !($ext_info = file_get_contents($composer_file)))
-				{
-					continue;
-				}
-
-				$ext_info = json_decode($ext_info, true);
 				$ext_name = str_replace(DIRECTORY_SEPARATOR, '/', $ext_name);
-
-				// Ignore the extension if directory depth is not correct or if the directory structure
-				// does not match the name value specified in composer.json.
-				if (substr_count($ext_name, '/') !== 1 || !isset($ext_info['name']) || $ext_name != $ext_info['name'])
+				if ($this->is_available($ext_name))
 				{
-					continue;
+					$available[$ext_name] = $this->phpbb_root_path . 'ext/' . $ext_name . '/';
 				}
-
-				$available[$ext_name] = $this->phpbb_root_path . 'ext/' . $ext_name . '/';
 			}
 		}
 		ksort($available);
@@ -530,7 +516,7 @@ class manager
 			return false;
 		}
 
-		$composer_file = $this->get_extension_path($name, true) . '/composer.json';
+		$composer_file = $this->get_extension_path($name, true) . 'composer.json';
 
 		// Not available if there is no composer.json.
 		if (!is_readable($composer_file) || !($ext_info = file_get_contents($composer_file)))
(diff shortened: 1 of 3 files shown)


javiexin:
In fact, the new is_available could be called from all available, saving the duplicate code as well...
Let me know what to do, and I will push another commit with the changes you recommend.

Nicofuma:
Do what you want, calling is_available() is fine too.

javiexin:
Refactored using is_available to implement all_available, saving duplicate code.

phpBB/phpbb/extension/manager.php
```diff
@@ -433,27 +433,13 @@ public function all_available()
...
-				if (substr_count($ext_name, '/') !== 1 || !isset($ext_info['name']) || $ext_name != $ext_info['name'])
+				if ($this->is_available($ext_name))
 				{
-					continue;
-				}
-
 				$available[$ext_name] = $this->phpbb_root_path . 'ext/' . $ext_name . '/';
 			}
 		}
+		}
```
Nicofuma: you have an indentation issue here

javiexin: Fixed, sorry.  Tried amending last commit, let's see if it works.

javiexin:
Again, timeout errors on quicklinks tests, unrelated to this PR.

Nicofuma:
that's expected, some tests are failing due to phpbb.com maintenance

javiexin:
When ýou look at it, take into consideration that for 3.2, the code for is_available could be greatly simplified by replacing all current code by a call to metadata_manager->get_metadata, with more complete results, and more consistent with the usage of the call.
This cannot be done in 3.1 as the metadata_manager instance requires the template object to be created (constructor param) while this requirement is removed in 3.2 (rightly in my view, makes more sense this way).
If you tell me how, I can prepare a new PR for 3.2 with this change.  BC is 100% unchanged.

javiexin:
Well, I have incorporated the above mentioned code change, using metadata_manager to check for extension availability.  For 3.1, there is an extra call to get the template from the container, that may be removed in 3.2.

javiexin:
The errros are due to a "Circular Dependency Injection" issue between ext.manager and template.
So, this method cannot be used for 3.1.
To avoid the circular dependency in 3.1 we need to give default parameter values to template in the metadata_manager constructor interface.

javiexin:
There was an error due to the some objects not being instantiated in one of the tests, required for metadata_manager (passed as null).  Not instantiated in the mock extension_manager object, now done.

The remaining error is a sniffer error, due to the use of a "default" value in an intermediate parameter.  However, this is CORRECT, as it is NOT a default value, but a "type hint", as per the PHP manual:

> To specify a type declaration, the type name should be added before the parameter name. The declaration can be made to accept NULL values if the default value of the parameter is set to NULL. 

[PHP Manual page](http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration)

Strangely enough, this only kicks of in PHP 5.5 tests, not in 5.4 or in 5.6.

This is the affected line:
https://github.com/phpbb/phpbb/pull/4592/files#diff-c281ac320a01294180fa4225455f5687R79
`public function __construct($ext_name, \phpbb\config\config $config, \phpbb\extension\manager $extension_manager, \phpbb\template\template $template = null, \phpbb\user $user, $phpbb_root_path)`

javiexin:
Options to remove this error:

1. Ignore it (maybe using @codingStandardsIgnore directives ?) - this is my preferred solution
2. Add "default values" to subsequent parameters (in this case, `$user = null` and `$phpbb_root_path = ''` ) - could be an option, although it might be misinterpreted by users
3. Remove the type-hinting for the template parameter - I do not like this too much
4. Reorder the parameters to have template be the last - I would prefer NOT to do this

What should I do?

javiexin:
The PR for ticket/15087 includes the 3.2 version of this PR.

PR #4598, opened 2016-12-28
[ticket/14944] Add search for template loop indexes by key

javiexin:
Adds a new function to the template interface, and implements it in the
context class.  The function returns the ordinal index for a specified key,
with the same structure that the key for alter_block_array.
Reuses same code.

PHPBB3-14944

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-14944

### DIFF
--- a/phpBB/phpbb/template/context.php
+++ b/phpBB/phpbb/template/context.php
@@ -305,16 +305,28 @@ class context
 				$name = $blocks[$i];
 				$index = sizeof($block[$name]) - 1;
 			}
+			if (!isset($block[$name]))
+			{
+				return false;
+			}
 			$block = $block[$name];
+			if (!isset($block[$index]))
+			{
+				return false;
+			}
 			$block = $block[$index];
 		}
 
+		if (!isset($block[$blocks[$i]]))
+		{
+			return false;
+		}
 		$block = $block[$blocks[$i]]; // Traverse the last block
 
 		// Change key to zero (change first position) if false and to last position if true
 		if ($key === false || $key === true)
 		{
-			return ($key === false) ? 0 : sizeof($block);
+			return ($key === false) ? 0 : sizeof($block) - 1;
 		}
 
 		// Get correct position if array given
@@ -331,7 +343,7 @@ class context
 			}
 		}
 
-		return is_int($key) ? $key : false;
+		return (is_int($key) && ((0 <= $key) && ($key < sizeof($block)))) ? $key : false;
 	}
 
 	/**
--- a/phpBB/phpbb/template/context.php
+++ b/phpBB/phpbb/template/context.php
@@ -264,6 +264,77 @@ class context
 	}
 
 	/**
+	* Find the index for a specified key in the innermost specified block
+	*
+	* @param	string	$blockname	the blockname, for example 'loop'
+	* @param	mixed	$key		Key to search for
+	*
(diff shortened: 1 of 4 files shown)


phpBB/phpbb/template/context.php
```diff
@@ -264,6 +264,77 @@ public function assign_block_vars_array($blockname, array $block_vars_array)
...
+	*
+	* @return mixed false if not found, index position otherwise; be sure to test with ===
+	*/
+	public function find_key_index($blockname, $key = false)
+	{
+		// For nested block, $blockcount > 0, for top-level block, $blockcount == 0
+		$blocks = explode('.', $blockname);
+		$blockcount = sizeof($blocks) - 1;
+
+		$block = &$this->tpldata;
```
Nicofuma: you don't need the references as you don't modify anything

Nicofuma:
This new function needs to be tested by unit tests

phpBB/phpbb/template/context.php
javiexin: You are of course right. The reason for this are two:
- I was duplicating the code in alter_block_array, https://github.com/javiexin/phpbb/blob/cff57f9076dd160b2185895f7298d4acbc5b5bb3/phpBB/phpbb/template/context.php#L367-L404, so that it is (almost) exactly the same.
- Using references does not copy the object, in this case, complex multilevel arrays, so execution time would be (much) faster, and the block is unchanged anyways...
If you still think it is better to remove the reference, I will do it from all places where it is used in this funciton (none is needed, there are 4 instances I think).

javiexin:
@Nicofuma If you point me in the right direction for testing docs and/or relevant example test cases, all the better, as I do not know the testing framework at all.

phpBB/phpbb/template/context.php
Nicofuma: Actually uses a copy-on-write strategy

javiexin: Understood.  Changes on its way.

javiexin:
I have tried to understand the tests in `tests\template\template_test.php` specially those related to `alter_block_array`, and
- I cannot understand them too well
- It seems to me that the case that I am trying to solve is not covered by test, no reference to 'outer[2].middle' or anything like that
- Testing finding a key-value pair is not done either, as far as I can tell
So, I would need some documentation or extra examples/pointers so that I can see if building these tests is feasible for me...

javiexin:
Tests have been added for the new function, please review.

phpBB/phpbb/template/context.php
```diff
@@ -264,6 +264,89 @@ public function assign_block_vars_array($blockname, array $block_vars_array)
...
+	* @return mixed false if not found, index position otherwise; be sure to test with ===
+	*/
+	public function find_key_index($blockname, $key)
+	{
+		// For nested block, $blockcount > 0, for top-level block, $blockcount == 0
+		$blocks = explode('.', $blockname);
+		$blockcount = sizeof($blocks) - 1;
+
+		$block = $this->tpldata;
+		for ($i = 0; $i < $blockcount; $i++)
```
marc1706: In its current state, this will not loop for the top-level block. I take it that is expected due to it not needing to traverse into the sub blocks?

javiexin: Exactly, this is the expected behavior: this loop will only be traversed for nested level loops, top level skips it completely, and goes directly to identify the element.

PR #4601, opened 2016-12-29
[ticket/14935] Correct get_valid_name to avoid system hang when migration dependency hasn't sarted with backslash

hubaishan:
Correct function get_valid_name to return $name started by backslash as
default unless it exist in migration db without backslash

PHPBB3-14935

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-14935

### DIFF
--- a/phpBB/phpbb/db/migrator.php
+++ b/phpBB/phpbb/db/migrator.php
@@ -215,13 +215,13 @@ class migrator
 			$prepended_name = ($name[0] == '\\' ? '' : '\\') . $name;
 			$prefixless_name = $name[0] == '\\' ? substr($name, 1) : $name;
 
-			if (isset($this->migration_state[$prepended_name]))
-			{
-				$name = $prepended_name;
-			}
-			else if (isset($this->migration_state[$prefixless_name]))
+			if (!isset($this->migration_state[$prepended_name]) && isset($this->migration_state[$prefixless_name]))
 			{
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/db/migrator.php
```diff
@@ -215,13 +215,13 @@ protected function get_valid_name($name)
 			$prepended_name = ($name[0] == '\\' ? '' : '\\') . $name;
 			$prefixless_name = $name[0] == '\\' ? substr($name, 1) : $name;
 
-			if (isset($this->migration_state[$prepended_name]))
+			if (!isset($this->migration_state[$prepended_name]) && isset($this->migration_state[$prefixless_name]))
```
Nicofuma: You can remove the first check and only keep `if (isset($this->migration_state[$prefixless_name]))` both shouldn't exists in the database at the same time.

hubaishan: Before the first check was not exist as in
https://github.com/phpbb/phpbb/pull/4492#issuecomment-269352344
but I want @marc1706  to be sure there are any side effects with this correction

marc1706:
Looks like this change adds an infinite loop which causes the tests to fail.

marc1706:
Closing as author hasn't responded in 2 months and change seems to break migrations instead of fixing them.

PR #4618, opened 2017-01-05
[ticket/14961] Add cookie notice as enablable feature

marc1706:
Uses Insites cookieconsent v3.0.3:
https://cookieconsent.insites.com

PHPBB3-14961

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-14961


marc1706:
Full view:
image
Responsive view:
image

ACP view (no cookie bar in ACP btw):
image

vinny:
maybe add a mention in [phpBB/docs/CREDITS.txt](https://github.com/phpbb/phpbb/blob/master/phpBB/docs/CREDITS.txt)

Nicofuma:
TBH, I'm not that sure it should be in core.... I think it would be better as an official ext because we definitely don't want to support any internet law of any country around the world... Too much work. (the same apply for the american law that is currently supported by the core)

marc1706:
I'd rather not ship another extension with something as simple as this cookie notice. Cookie notices have become an industry standard and I think we should have this in the core if people are wishing to enable it.

Crizz0:
I'm fine with that feature in the core. I think it is useful for a lot of phpBB-users and nearly everyone uses that thing at the moment. Personally I don't need it and really hate it, because it is like "attention, we use HTML, CSS, JS..., please accept that".

PR #4632, opened 2017-01-11
[ticket/14989] Allow more admin-configurable schemes in post links

VSEphpbb:
PHPBB3-14989

This ticket is restoring parts of ticket/13935 which seem to have been lost somewhere.

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-14989


JoshyPHP:
The explanation is there: https://github.com/phpbb/phpbb/pull/3698#discussion-diff-32025172

IMO if the option exists it should be configurable via web, but I don't really care either way.

VSEphpbb:
Well, we already came upon an issue where we found something missing (irc protocol).

There's no point to store a value as a config if it's not configurable (otherwise it should just be hardcoded). So we should bring it back or hardcode it and add irc protocol to it.

JoshyPHP:
I think it's better to have it configurable, with a small set of schemes known to be safe by default.

PR #4634, opened 2017-01-12
[ticket/14994] Refactor template->assign_block_var

javiexin:
Refactor assign_block_var to use the same block selection mechanism
as is used in alter_block_array.  This allows creating new blocks
at any position in the template structure, not only on the last block.
Allows selecting a block as outer[2].middle.

PHPBB3-14994

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-14994


phpBB/phpbb/template/context.php
```diff
@@ -179,70 +179,65 @@ public function &get_root_ref()
...
+			$vararray['S_FIRST_ROW'] = true;
+		}
+
+		// Assign S_BLOCK_NAME
+		$vararray['S_BLOCK_NAME'] = $name;
 
-			// Add a new iteration to this block with the variable assignments we were given.
-			$this->tpldata[$blockname][] = $vararray;
+		// Now the tricky part, we always assign S_LAST_ROW and remove the entry before
+		// This is much more clever than going through the complete template data on display (phew)
```
Elsensee: Is it? Is it really cleverer to set a value and unset the value before in each iteration?

javiexin: I don't know, but I guess so: this os the original unchanged code and comment...

javiexin: And by the way, it is in each nested block level, once per nesting level, so two/three times max.  That is the minimum possible if the block existed before.

Elsensee: Hm. Then I guess, it's fine.

marc1706:
I'm missing a few tests that cover this change of behavior, e.g. tests that would fail before this PR and that do not fail now (as in setting a nested block for a previous block).

javiexin:
@marc1706 Tests added per your request

javiexin:
@marc1706 Anything else left to do here? Is this going forward?

marc1706:
Replacing with refactored version without merge conflicts.

PR #4646, opened 2017-01-14
[ticket/15011] Error not checked on metadata load failure

javiexin:
PHPBB3-15011

Checklist:

- [ ] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-15011

### DIFF
--- a/phpBB/phpbb/extension/metadata_manager.php
+++ b/phpBB/phpbb/extension/metadata_manager.php
@@ -229,7 +229,10 @@ class metadata_manager
 			case 'all':
 				$this->validate('display');
 
-				$this->validate_enable();
+				if (!$this->validate_enable())
+				{
+					throw new \phpbb\extension\exception($this->user->lang('META_FIELD_NOT_SET', $name));
+				}
 			break;
 
 			case 'display':


phpBB/phpbb/extension/metadata_manager.php
```diff
@@ -229,7 +229,10 @@ public function validate($name = 'display')
 			case 'all':
 				$this->validate('display');
 
-				$this->validate_enable();
+				if (!$this->validate_enable())
+				{
+					throw new \phpbb\extension\exception($this->user->lang('META_FIELD_NOT_SET', $name));
```
Nicofuma: Do not translate the langage key here: `throw new \phpbb\extension\exception('META_FIELD_NOT_SET', array($name));`

marc1706: That doesn't make much sense right now? It will just say:
`Required meta field all has not been set.`

javiexin: Probably right, Marc...
The alternatives I see are:
1) Throw specific exceptions from `validate_dir`, `validate_require_phpbb` and `validate_require_php`; at least two could use the same exception message, but `_dir` might need a new language message.  Breaks the current interface (exception instead of return false).
2) Throw a generic exception from `validate_enable`, probably with new language key as well. Breaks the current interface in the same way.
3) Change the throw message in `validate('all')`, also with a new language key.  Keeps the current interface (throw vs return).
Which one is best?  I will implement whatever is chosen.  If a new language string is required, please, provide proper wording, as I am not native :)

marc1706: Create a patch that should output more meaningful error messages:
https://github.com/phpbb/phpbb/pull/4651

PR #4652, opened 2017-01-16
[ticket/15019] Smiley Limit to a Higher Value

Leinad4Mind:
Changing the value from 1000 to 2000 for the smiley limit.

Checklist:

- [ ] Tests pass

Tracker ticket: https://tracker.phpbb.com/browse/PHPBB3-15019

### DIFF
--- a/phpBB/includes/constants.php
+++ b/phpBB/includes/constants.php
@@ -200,7 +200,7 @@ define('BBCODE_ID_ATTACH', 12);
 define('BBCODE_LIMIT', 1511);
 
 // Smiley hard limit
-define('SMILEY_LIMIT', 1000);
+define('SMILEY_LIMIT', 2000);
 
 // Magic url types
 define('MAGIC_URL_EMAIL', 1);


Nicofuma:
You commit message is not correctly formatted, it must end with PHPBB3-15019 on a single line.
I also think it would be better to completely remove this limit as it does not really have any reason to be today.

Leinad4Mind:
Done. (I guess)

hanakin:
click the format link in you post at the top to read the rules for proper commits. You are missing the `PHPBB3-15019` in the commit message body

Elsensee:
I wonder why this was ever introduced. Looks fine to me as soon as the commit msg is fixed.

Leinad4Mind:
Dunno how to change it via website.

VSEphpbb:
@Elsensee I assume it was added because having more than 1000 smilies is absurd ;)

Leinad4Mind:
Many people wants to add all the emojione collection ~1800 emojis as smilies.
That way they can easely add a emoji system (extension) with a smilie system working very well together. ;)

VSEphpbb:
We already have emoji support since 3.2 AFAIK.

Leinad4Mind:
Exactly, and thats the main reason for this.

DavidIQ:
To duplicate the already existing and supported emojis into smileys?

marc1706:
I've looked into this further and it really seems like having an excessive amount of smilies can significantly slow down the board and cause further issues. The recommended way to go is using emoticons that allow a great range of possible icons. To extend that functionality 1000 smilies is more than enough. As already mentioned, duplicating emoticons to smilies is also not recommended as that doesn't just duplicate content but probably can cause even more issues.

Closing this and the ticket as such.

PR #4660, opened 2017-01-21
[ticket/13250] Replace slashes in cache file names

marc1706:
PHPBB3-13250

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-13250

### DIFF
--- a/phpBB/phpbb/cache/driver/file.php
+++ b/phpBB/phpbb/cache/driver/file.php
@@ -601,6 +601,6 @@ class file extends \phpbb\cache\driver\base
 	*/
 	protected function clean_varname($varname)
 	{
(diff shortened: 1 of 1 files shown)


bantu:
m d 5 ?

marc1706:
md5() would also be an ok solution but at the same time probably slower and will currently cause issues due to filenames being somewhat hardcoded in the file cache driver in serveral places. Would certainly require more changes than this.

Nicofuma:
I don't think it will be slower enough to be noticeable. And I agree that a proper solution should be used. Maybe keep it for later in order to have the fix in 3.2.1.

@marc1706 @bantu fine with you?

PR #4661, opened 2017-01-21
[ticket/15037] Make imagesets retina

hanakin:
Checklist:

- [x] Tests pass
- [x] Code follows coding guidelines: [master / 3.2.x](https://area51.phpbb.com/docs/master/coding-guidelines.html)

* read the ticket for more info

PHPBB3-15037

https://tracker.phpbb.com/browse/PHPBB3-15037


hanakin:
!set Styles

hanakin:
some screens

screen shot 2017-01-21 at 6 02 00 pm
screen shot 2017-01-21 at 6 02 13 pm

JimMH:
Looks very good. Wouldn't it be better though to replace the existing images instead of using new ones and changing the CSS? Otherwise this might break custom styles.

Crizz0:
He changed the file type to .png (was .gif). I thinkt this i neccessary for better quality? 

But why do you add this "-2x" to the filename?

hanakin:
@JimMH  I made two changes to the colors.css file of note.

1. I used PNG for 2 reasons

    1. PNG is the std format for all graphic editors
    2. GIF is a dead tech its not used anymore as its not scaler meaning it does not support scaling of the img as other formats do. You get artifacting and other weird issues sometimes.

I can try converting them to gifs after all the icons are good to go as they are all new from scratch.

screen shot 2017-01-21 at 6 27 51 pm


2. I also changed the names as well to remove the stupid fucking '_' these are highly problematic for text editors and image editors and makes it take 5 times longer to make changes as you can not chain with them. again I can rename them back easily if there are no issues found with any of the images

@Crizz0 its the way my sketch works it saves a normal version and a double sized version with that on the end I removed the normal versions I just did not remove the appending yet.

hanakin:
!set Wip

hanakin:
!set WIP

hanakin:
@JimMH here is the difference of png vs gif

screenshot 2017-01-21 19 35 33

marc1706:
Any reason the locks on stickies are further to the top of the image than on the other icons?

VSEphpbb:
No more GIFs. Lets stickw ith PNG. If anything I'd say use SVGs with PNG fallbacks ;)

hanakin:
@marc1706 the locks is due to the shape at the bottom being narrower and the icon floating. its this way on the current sticky icon for the same reason. 

@VSEphpbb SVGs will be used in 3.3 for sure problem is some aspects of core oes not support svg which will need fixed in 3.3. This is meant as a quick fix for 3.2 in the mean time. There is no need for a fallback anymore svg is supported in all browsers. Although not all features are supported across the board so careful design practices and the way we serve them need to be taken into account.

hanakin:
!unset WIP

marc1706:
Changes look good to me. @Nicofuma want to have this in 3.2.1 or wait for 3.2.2 ?

Nicofuma:
3.2.2 if we have gifs (ie: same image names as in 3.2.0), otherwise 3.3 because it will break almost all 3.2 styles.

hanakin:
@Nicofuma gif is not possible

hanakin:
@Nicofuma I may have found a way to get gifs working but it means recreating all the images from scratch again and will take me a little so stby. here is one for an example
topic-read

marc1706:
I think that would be better, especially regarding BC.

hanakin:
Should have this up for review tonight just need to render unread versions

Nicofuma:
Thanks a lot

hanakin:
k should be all good to go pending tests here is an updated screen

screenshot 2017-01-23 19 35 23

hanakin:
possible counter part to this

screenshot 2017-01-23 20 50 18

VSEphpbb:
I like the new topic icons, but I'd like to keep the old 2 smileys... They're classics.

marc1706:
What's the test file in develop for?

tests/functional/fileupload_remote_test.php
```diff
@@ -100,8 +100,8 @@ public function test_successful_upload()
 		$upload = new \phpbb\files\upload($this->filesystem, $this->factory, $this->language, $this->php_ini, $this->request, $this->phpbb_root_path);
 		$upload->set_error_prefix('')
 			->set_allowed_extensions(array('gif'))
-			->set_max_filesize(1000);
-		$file = $upload->handle_upload('files.types.remote', self::$root_url . 'styles/prosilver/theme/images/forum_read.gif');
```
hanakin: @marc1706 because for some reason we are testing upload of the forum_read.gif to the files folder which does not make sense as this will no longer be a a file in 3.3 anyway so I just uploaded a random gif for performing these kind of tests to develop

hanakin:
@VSEphpbb we can always just bump the size
mrgreen2

VSEphpbb:
Yeah, keep 'em but scale them up ;)

hanakin:
ok so the redface animation is not 100% the same i had to guess not having the actual layer style changes... Also I complete changed the thinking icon, but thoughts?

fire
heart
radioactive
star
thinking
alert
info
mrgreen
question
redface

VSEphpbb:
i like them all

marc1706:
Tests should be fixed now

hanakin:
@Nicofuma this should be good now

Note for release we should point out that to see the effect on custom themes they need to save 2x size versions of there icons

milescellar:
@hanakin @marc1706 these new retina imagesets looks really blurries on Chrome 59 (it's okay on Edge, and I don't know for other browsers). See: [http://imgur.com/a/TnmEm](http://imgur.com/a/TnmEm)

I tested on area51 and on a fresh phpBB 3.2.1-RC1 local installation. I cleared my entire browser cache, but the issue is still there.

hanakin:
@milescellar what is you OS and pls check to see if you graphics card drivers are up to date...I am assuming you are on a machine with a graphics card where you have access to update the drivers

milescellar:
@hanakin I'm on Windows 10 (up-to-date), and I'm using the latest graphics card drivers (I'm using an NVIDIA GeForce GTX 970 and I'm using the latest 384.76 GeForce drivers version). I don't think the graphic card drivers are the issue, I don't experienced any issue on other places on the Web or in games.

Crizz0:
Is this maybe caused by the background-size-property?
```
dl.row-item {
    background-position: 10px 50%;
    background-repeat: no-repeat;
    background-size: 32px auto;
}
```

hanakin:
It's related to it as that is how you do retina! This is not the problem it is most likely caused by some weird render issue in this specific broweser on this OS possibly due to the use of gif vs png. Ultimately fixed in 3.3 with svgs. But we will have to test 3.2.1 for this specific situation. As I have not seen this problem on mac or 7 I did not test anything but edge on win 10

hanakin:
There is also a usecase for IE 11 that needs added to tests on win 10 as well as edge has a bad adoption rate still.

Noxwizard:
For me (Chrome 59, Win 10), adding `image-rendering: -webkit-optimize-contrast;` removed the blurriness.

hanakin:
Interesting? Will test when I get a chance? If works that is the best named property ever!

Scanialady:
> For me (Chrome 59, Win 10), adding image-rendering: -webkit-optimize-contrast; removed the blurriness.

-----

Where did you add it, please?

------
I like the new images.
Will you release a set of these images for dark styles, too? Would be helpful.

JimMH:
@Scanialady See this new pull request for that: https://github.com/phpbb/phpbb/pull/4880

PR #4678, opened 2017-01-25
[ticket/15047] Various db tools fixes for MSSQL

DavidIQ:
The character limit for index names on MS SQL is 128 so there's not really a need to check it and throw an error if it exceeds 30 characters.

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15047

### DIFF
--- a/phpBB/phpbb/db/tools/mssql.php
+++ b/phpBB/phpbb/db/tools/mssql.php
@@ -597,7 +597,7 @@ class mssql extends tools
 		// Change the column
 		$statements[] = 'ALTER TABLE [' . $table_name . '] ALTER COLUMN [' . $column_name . '] ' . $column_data['column_type_sql'];
 
-		if (!empty($column_data['default']))
+		if (!empty($column_data['default']) && !$this->mssql_is_column_identity($table_name, $column_name))
 		{
 			// Add new default value constraint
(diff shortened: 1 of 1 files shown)


marc1706:
FWIW, PRs against 3.2 should go against 3.2.x.

marc1706:
It seems like you're using CRLF in your commit messages. Make sure your git is set up to not use CRLF. Also, one commit seems to have a line that is longer than 80 characters:
https://github.com/phpbb/phpbb/pull/4678/commits/cc8196c2b0efe7cb016e6f1a77851c8f12388147

It's also missing the ID at the bottom. I'd recommend using the git-tools as that will automatically add the required parts to your commit messages.

DavidIQ:
Oops. Must have had master selected when I opened this pr up... Thanks for switching it.

git-tools seems to be mostly unix centric. I'll see how to get it working on Windows, unless we have documentation for that.

3D-I:
@DavidIQ 
if we are talking about git tools for Win here's what I use in combination with tortoise git.
https://git-for-windows.github.io/

DavidIQ:
@3D-I no we're talking about what's in the git-tools directory of this project.

DavidIQ:
After some further research I've re-instated the name length check but only for SQL Server 2000 since we seem to still be supporting that version in code and there was actually a 30 character limit for object names in that version. These checks should be removed in the next phpBB version and we should specify the minimum version of SQL Server we support which should be SQL Server 2008 and up. 2008 is on extended support until 7/9/2019. I'll create a ticket for that.

PR #4686, opened 2017-02-05
[ticket/15073] Consolidate responsive utilities

hanakin:
PHPBB3-15073

Depends on https://github.com/phpbb/phpbb/pull/4685

Checklist:

- [x] Correct branch: master for new features; 3.3
- [x] Tests pass
- [x] Code follows coding guidelines
- [x] Commit follows commit message [format](https://wiki.phpbb.com/Git#Commit_Messages)

https://tracker.phpbb.com/browse/PHPBB3-15073

PR #4685, opened 2017-02-05
[ticket/15071] Consolidate all media queries

hanakin:
PHPBB3-15071

Checklist:

- [x] Correct branch: master for new features; 3.3
- [x] Tests pass
- [x] Code follows coding guidelines
- [x] Commit follows commit message [format](https://wiki.phpbb.com/Git#Commit_Messages)

https://tracker.phpbb.com/browse/PHPBB3-15071

PR #4687, opened 2017-02-05
[ticket/15072] Remove current responsiveness

hanakin:
Checklist:

PHPBB3-15072

Depends on 
https://github.com/phpbb/phpbb/pull/4685
https://github.com/phpbb/phpbb/pull/4686

- [x] Correct branch: master for new features; 3.3
- [ ] Tests pass
- [x] Code follows coding guidelines

https://tracker.phpbb.com/browse/PHPBB3-15072


Nicofuma:
I'm not sure that's something we want to do in prosilver. I already can imagine it breaking all style again and I don't think that's something we want to do.

hanakin:
@Nicofuma the benefits of this will out way the negatives. It will be far simpler to understand and make the changes needed then it currently is as most individuals have no idea how things are accomplished do to the crappy us hacks and use of the giant parser function to handle what should noteably be all CSS work for the majority. It's not actually that much work since i have the reference of the current version to work off for the look and feel. It's just a transposition. It's early on so we can see and compare once I get everything up. If nothing else all the us stuff needs to go.

hanakin:
!set WIP

hanakin:
!set Styles

PR #4691, opened 2017-02-10
[ticket/15058] Force browser to redownload stylesheets

DavidIQ:
Checklist:

- [x] Tests pass
- [X] Code follows coding guidelines: [master / 3.2.x](https://area51.phpbb.com/docs/master/coding-guidelines.html), [3.1.x](https://area51.phpbb.com/docs/31x/coding-guidelines.html)

https://tracker.phpbb.com/browse/PHPBB3-15058

### DIFF
--- a/phpBB/styles/prosilver/theme/stylesheet.css
+++ b/phpBB/styles/prosilver/theme/stylesheet.css
@@ -7,15 +7,15 @@
     --------------------------------------------------------------
 */
 
-@import url("normalize.css");
-@import url("base.css");
-@import url("utilities.css");
-@import url("common.css");
-@import url("links.css");
-@import url("content.css");
-@import url("buttons.css");
-@import url("cp.css");
-@import url("forms.css");
-@import url("icons.css");
-@import url("colours.css");
-@import url("responsive.css");
+@import url("normalize.css?v=3.2");
+@import url("base.css?v=3.2");
+@import url("utilities.css?v=3.2");
+@import url("common.css?v=3.2");
+@import url("links.css?v=3.2");
+@import url("content.css?v=3.2");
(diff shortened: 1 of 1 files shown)


marc1706:
Why don't you append the assets version to the stylesheet.css links in overall_header and simple_header ?

DavidIQ:
We do that already and it's not enough apparently.

javiexin:
What I have seen is that even though the asset version is appended to stylesheet.css, it then imports other files, that have the SAME name as before (no asset version appended), so the browser interprets that it is the same file it already has, and it does not reload them.
What I had to do in my own local testing bed was to remove "import" and either include  the nested files in overall_header directly (appending the asset version) or actually including the rules into stylesheet.css.
The resolution here is smarter, of course, but will require to bump the number in stylesheet.css for every change that happens in an stylesheet, at least from version to version, and that in case the user has not modified manually any stylesheet css file.
I think some automated way of doing it would be required.  
What about templated CSS for key elements, as _stylesheet.css_?
Something like replacing  in _overall_header.html_
`<link href="{T_STYLESHEET_LINK}" rel="stylesheet">`
by
`<!-- INCLUDE stylesheet_css.html -->`
And then, use in _stylesheet_css.html_
```
<link href="{T_THEME_PATH}/normalize.css?assets_version={T_ASSETS_VERSION}" rel="stylesheet">
...
```

marc1706:
The goal for future versions is to compile the stylesheet into one and minimize it as well. I don't think we should spend too much time on trying to split this into a separate file that will then load all stylesheets one by one again. We could try using assets management like symfony's assetic in 3.2 but I think for now this change is ok as a quick fix.

PR #4695, opened 2017-02-11
[ticket/15083] Fix invalid emoji sizing in ACP & print view

marc1706:
PHPBB3-15083

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15083

### DIFF
--- a/phpBB/adm/style/admin.css
+++ b/phpBB/adm/style/admin.css
@@ -2692,9 +2692,11 @@ fieldset.permissions .padding {
 	word-break: break-all;
 }
 
+/* stylelint-disable declaration-property-unit-whitelist */
 .emoji {
 	min-height: 18px;
 	min-width: 18px;
 	height: 1em;
 	width: 1em;
 }
+/* stylelint-enable declaration-property-unit-whitelist */
--- a/phpBB/adm/style/admin.css
+++ b/phpBB/adm/style/admin.css
@@ -2691,3 +2691,10 @@ fieldset.permissions .padding {
 	word-wrap: break-word;
 	word-break: break-all;
 }
+
+.emoji {
+	min-height: 18px;
+	min-width: 18px;
+	height: 1em;
+	width: 1em;
+}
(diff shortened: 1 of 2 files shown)


hanakin:
@marc1706 pls stick to px do not use ems

marc1706:
This is a direct copy from https://github.com/phpbb/phpbb/commit/f111e70fc3959d979e61e8134fcaefb670b32603

Not sure if it's better to hardcode px here (i.e. something like 12px). Suggestions welcome.

hanakin:
its always better to hardcode px over the use of em especially since you never define a font-size on the element. em is a compounded multiple of the font-size of its containing element if not then its parents. It should never be used to define a height or width. 

The entire inclusion method for the img is faulty and needs redone never use img tags instead we need to include it as an svg using the viewbox property to define the aspect and setting the width/height to 100%/auto.

https://css-tricks.com/scale-svg/
http://stackoverflow.com/questions/19484707/how-can-i-make-an-svg-scale-with-its-parent-container

hanakin:
actually the entire inclusion process is faulty svg direct inclusion is recommended over using img tag inclusion for this specific reason

https://demos.emojione.com/latest/sprites-svg.html

phpBB/styles/prosilver/theme/print.css
```diff
@@ -137,3 +137,10 @@ div.spacer { clear: both; }
 .skip_link { display: none; }
 
 .codebox p { display: none; }
+
+.emoji {
```
hanakin: this needs to be ignored by linting so wrap the class like so pls

/* stylelint-disable declaration-property-unit-whitelist */
.emoji {
	min-height: 18px;
 	min-width: 18px;
 	height: 1em;
 	width: 1em;
 }
/* stylelint-enable declaration-property-unit-whitelist */

marc1706:
Updated

phpBB/adm/style/admin.css
```diff
@@ -2691,3 +2691,10 @@ fieldset.permissions .padding {
 	word-wrap: break-word;
 	word-break: break-all;
 }
+
```
hanakin: might as well add the same stylelint comments to here

PR #4699, opened 2017-02-13
[ticket/15088] Avoid code repetitions in extension manager

javiexin:
There are multiple identical queries in the extension manager, that can 
be made into a separate function, reused everywhere it is needed. 
Smaller code, and easier to maintain (single change to fix all uses).

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-15088

### DIFF
--- a/phpBB/phpbb/extension/manager.php
+++ b/phpBB/phpbb/extension/manager.php
@@ -236,14 +236,7 @@ class manager
 			'ext_state'		=> serialize($state),
 		);
 
-		$sql = 'SELECT COUNT(ext_name) as row_count
-			FROM ' . $this->extension_table . "
-			WHERE ext_name = '" . $this->db->sql_escape($name) . "'";
-		$result = $this->db->sql_query($sql);
-		$count = $this->db->sql_fetchfield('row_count');
-		$this->db->sql_freeresult($result);
-
-		$this->update_state($name, $extension_data, ($count) ? 'update' : 'insert');
+		$this->update_state($name, $extension_data, ($this->is_configured($name)) ? 'update' : 'insert');
 
 		if ($active)
 		{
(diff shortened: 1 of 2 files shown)


javiexin:
The Travis error is coming from the testing framework itself, but I don't see where.
The issue is that it might not be initializing correctly the $ext_manager->extensions array, and there is a discrepancy between that and the DB contents; that, or the DB is not updated/initialized correctly.
I have changed the way the check is done in the code, and it fails in a similar way (duplicate key in the DB when the check is done in the extensions array, and trying to merge a non-array object when checking the DB, so the error is really a discrepancy between DB and extensions array).
This is not happening in the fully working case (the actual forum software, not the testing framework).  Or at least I haven't been able to replicate the issue in a clean system.
So, if someone with better knowledge about the testing framework could check it, I would be really thankful.

javiexin:
Found it.  The config object was initialized empty, and the phpbb version is required and checked in the extension manager load_extensions method, used to get data from the DB into the extensions array.

javiexin:
Would it be possible to merge this, at least for 3.2.1 (ideally for 3.1.11 as well)?
I am working on a couple of additions (PRs) that are based on/use this one (for 3.2/master), so it would be cleaner than this is merged before.
Thanks.

javiexin:
Any chance this will make it to 3.1.11 and 3.2.1?

marc1706:
We're currently looking at a 3.2.2 rather close after 3.2.1 in which we'd incorporate these changes. We'd like to review this more extensively and not fiddle too much with the 3.1 system before reaching EOM.

javiexin:
Fair enough, thanks a lot

marc1706:
I think it makes more sense to have refactoring in 3.3 instead of 3.2. Therefore changing the target

rubencm:
Close in favour of #5557

PR #4701, opened 2017-02-13
[ticket/14992] Improve user_notifications table performance

DavidIQ:
PHPBB3-14992

To see why these changes are needed just go to your notification settings on phpbb.com and notice the time waiting for page load.

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-14992


DavidIQ:
I've restarted this build 3 times now. I think the problem is slowness with execution in a high volume environment? These migrations will be kind of slow with each one taking more than a minute to complete so it's expected.

phpBB/phpbb/db/migration/data/v32x/user_notifications_table_remove_duplicates.php
```diff
@@ -0,0 +1,55 @@
...
+		{
+			// Delete the duplicate entries
+			$this->sql_query("DELETE FROM {$this->table_prefix}user_notifications
+								WHERE user_id = {$row['user_id']}
+								  AND item_type = '{$row['item_type']}'
+								  AND method = '{$row['method']}'");
+
+			// And re-insert as a single one
+			$insert_buffer->insert($row);
+		}
```
Elsensee: This looks to be doomed to time-out. Maybe just LIMIT 50 (you don't even need a start) or so and if there were exactly 50 duplicates, return false?

phpBB/phpbb/db/migration/data/v32x/user_notifications_table_unique_index.php
```diff
@@ -0,0 +1,46 @@
...
+			'\phpbb\db\migration\data\v32x\user_notifications_table_remove_duplicates',
+		);
+	}
+
+	public function update_schema()
+	{
+		return array(
+			'add_unique_index' => array(
+				$this->table_prefix . 'user_notifications' => array(
+					'itm_type_itm_id_usr_id_method'	=> array('item_type', 'item_id', 'user_id', 'method'),
```
Elsensee: I think this would be too long on *any* setup since even the length of the index itself as already at 29 chars.

phpBB/phpbb/db/migration/data/v32x/user_notifications_table_remove_duplicates.php
DavidIQ: The number of rows returned when running the select against .com data is 9. I don't think there'll be much more than that on other installations. This doesn't seem to be a common occurrence and adding a limit means it would be possible that the query will need to be run multiple times, each time clocking in at around a minute of execution time for the select.

If you are referring to the delete the where clause limits the results to 2 rows, at least against the .com data.

Elsensee: Oh, okay. I don't expect the delete clause to take a long time, I think the bottleneck here is the SELECt. Some other points:
1. What does MAX(notify) as notfiy do? Do you ever ned the result?
2. One minute is a *really* long time and I assume the GROUP BY is consuming *much* time - at least on heavy databases. I'm wondering: What do you think would happen if we limit it to 1? Will it still take one minute or so - can you quickly test that?

I'm just concerned if board administrators with a medium-sized board use the web updater, then run into timeouts.
So if it still takes a minute.. okay.. shit, but we can't do better obviously. If not, I would prefer to limit the query.

DavidIQ: The MAX(notify) is just to get the max value for the notify setting, be it 0 or 1 when there are actual duplicates.
Adding a limit clause to the query doesn't actually improve the execution time.

The best I could do to improve the performance here would be to retrieve the entire contents of the table and then de-duplicate through code. That might actually take a lot less than a minute.

DavidIQ:
!set WIP

DavidIQ:
Going to try separating the creation of some of the indexes out to different migrations. The first one takes a few seconds, the second takes around 25, and the third takes almost a minute so it turns out that the index creation is yet another bottleneck here.

Also creating a temporary index on the table then dropping it and making it a unique index might be the way to go since it actually drops execution of the grouping query down to around 15 seconds.

Elsensee:
> Going to try separating the creation of some of the indexes out to different migrations. The first one takes a few seconds, the second takes around 25, and the third takes almost a minute so it turns out that the index creation is yet another bottleneck here.

Well, the creation of indexes can hardly be improved as it's all done by the database, isn't it? If so, then seperating out the creation won't do anything as the updater will very likely make a break between the schema steps.

> Also creating a temporary index on the table then dropping it and making it a unique index might be the way to go since it actually drops execution of the grouping query down to around 15 seconds.

That sounds really great! Not so sure if that's the right way to go, but if it works and you have to do two migrations anyway I don't see a reason why not to do it.
However, how long does it take to create the unique index?

DavidIQ:
> Well, the creation of indexes can hardly be improved as it's all done by the database, isn't it? If so, then seperating out the creation won't do anything as the updater will very likely make a break between the schema steps.

It's not really about improving them but splitting them out so each one is measured on its own. It looks like it takes just over 2 minutes to create all 3 of them in one migration and that's not including the unique index which takes another minute and a half on its own. Still researching and testing out options here.

DavidIQ:
!unset WIP

DavidIQ:
I've changed this to separate the index creation out to 4 migrations to reduce the incidences of timeouts while also adding a temporary index in order to actually be able to complete the de-duplication of data. Without it the query takes a good minute and a half and doesn't actually complete and that migration never runs. The caveat here is going to be that the index creation might cause the update to time out. However the index does get created successfully and the user will end up having to refresh the page.

I'm not sure what effect this will have on future Travis builds. Does Travis run the migrations on every build? If that's the case then these might need to get added to some sort of ignore list.

SiteSplat:
What's the situation with this? This bug with the double notification has been present since phpBB 3.1.0 despite sever reports about it... @VSEphpbb

phpBB/phpbb/db/migration/data/v32x/user_notifications_table_remove_duplicates.php
DavidIQ: FYI the last part of my comment is not possible. You run out of memory on larger versions of this table, like phpbb.com's.

marc1706:
Error seems to be:
```
Index name 'ser_notifications_usr_id_itm_id' on table 'phpbb_user_notifications' is too long. The maximum is 30 characters.
```

phpBB/phpbb/db/migration/data/v32x/user_notifications_table_index_p2.php
```diff
@@ -0,0 +1,46 @@
...
+			'\phpbb\db\migration\data\v32x\user_notifications_table_index_p1',
+		);
+	}
+
+	public function update_schema()
+	{
+		return array(
+			'add_index' => array(
+				$this->table_prefix . 'user_notifications' => array(
+					'usr_id_itm_id'					=> array('user_id', 'item_id'),
```
marc1706: Change to `uid_itm_id`

phpBB/phpbb/db/migration/data/v32x/user_notifications_table_reduce_column_sizes.php
```diff
@@ -0,0 +1,48 @@
...
+			'\phpbb\db\migration\data\v32x\user_notifications_table_remove_duplicates',
+		);
+	}
+
+	public function update_schema()
+	{
+		return array(
+			'change_columns'	=> array(
+				$this->table_prefix . 'user_notifications'			=> array(
+					'item_type'		=> array('VCHAR:155', ''),
```
marc1706: Did you see if something like 200 would be enough or is this an actual value after testing things out?

DavidIQ: I went by what is stated [here](https://dev.mysql.com/doc/refman/5.7/en/string-type-overview.html):

> utf8 characters can require up to three bytes per character, so a VARCHAR column that uses the utf8 character set can be declared to be a maximum of 21,844 characters

Based on that and taking into account the 1000 max byte count error I just used simple math:
1000 bytes / 3 bytes per char = 333.33 chars / 2 columns = 166.67 chars per column

155 seemed a little more "common" a size when compared to other columns. I can't imagine either of these columns having data being even remotely close to that amount so that's what I went with.

tldr; I can probably go as high as `VARCHAR:166` based on MySQL docs so up to you if you want me to use that amount.

DavidIQ: I went ahead and changed it to 165 since it's not much of a difference from 155. I did try 200 but turned out to be too high, kind of validating the calculations:
https://travis-ci.org/phpbb/phpbb/jobs/236011337#L671

PR #4705, opened 2017-02-17
[ticket/15097] Add PHP version to Board statistics

nomind60s:
The ACP main page shows the phpBB version and the database version, but didn't show the PHP version. With this change the PHP version is displayed as part of the Board statistics.

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15097

### DIFF
--- a/phpBB/adm/style/acp_main.html
+++ b/phpBB/adm/style/acp_main.html
@@ -169,7 +169,7 @@
 	</tr>
         <tr>
         <td>{L_PHP_VERSION}{L_COLON} </td>
-        <td><strong>{PHP_VERSION}</strong></td>
+        <td><strong>{PHP_VERSION_INFO}</strong></td>
     </tr>
 	<!-- ENDIF -->
 	</tbody>
--- a/phpBB/adm/style/acp_main.html
+++ b/phpBB/adm/style/acp_main.html
@@ -94,15 +94,45 @@
 
 	<!-- EVENT acp_main_notice_after -->
 
+	<table class="table1 no-header" style="width: auto;" data-no-responsive-header="true">
+		<caption>{L_SOFTWARE_HEADING}</caption>
+		<col class="col1" /><col class="col2" />
+	<thead>
+	<tr>
+		<th>{L_SOFTWARE_PRODUCT}</th>
+		<th>{L_SOFTWARE_VERSION}</th>
+	</tr>
+	</thead>
+	<tbody>
+	<tr>
+		<td>{L_BOARD_VERSION}</td>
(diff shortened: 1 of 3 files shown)


DavidIQ:
You're targeting the wrong branch (master). Needs to be the 3.1.x branch, although this probably should go into 3.2 instead.

nomind60s:
I thought I targeted the 3.1.x branch. What did I do incorrect?

DavidIQ:
When you created the pull request you needed to change the target branch, which you should still be able to do.

nomind60s:
I believe all my work is done here. First phpBB bugfix.

nomind60s:
Oops, wrong button. Sorry.

nomind60s:
Closing and reopening to cause travis checks to re-run. They had previously passed.

nomind60s:
Reopening to cause travis-ci checks to repeat.

marc1706:
We can also restart if there are issues (which we usually do before merging).

Lady-G:
There is a dependency on the language pack. There are two new variables not contained in my default American English language pack. The ACP displays the variables literally instead of the text description. I've updated the bug tracker with my comments.

DavidIQ:
As per coding guidelines, developers and code contributors only need to worry about the default language pack, which has been taken care of here. Other languages are the responsibility of the language pack maintainers. There is already a process in place for them to be notified of changes and it is up to them to monitor for these changes and implement those themselves.

phpBB/adm/style/acp_main.html
```diff
@@ -94,15 +94,45 @@
...
 	<table class="table1 two-columns no-header" data-no-responsive-header="true">
 		<caption>{L_FORUM_STATS}</caption>
 		<col class="col1" /><col class="col2" /><col class="col1" /><col class="col2" />
 	<thead>
 	<tr>
-		<th>{L_STATISTIC}</th>
-		<th>{L_VALUE}</th>
-		<th>{L_STATISTIC}</th>
-		<th>{L_VALUE}</th>
+		<th style="width: 25%">{L_STATISTIC}</th>
```
CHItA: This should be moved to some css file i think.

nomind60s: In principle I would agree. However if you grep for style=\"width: in adm/style/*.html there are ~70 occurrences already, so it certainly seems this is not handled consistently. 

Will investigate resolution.

Nicofuma: you are right, it's not really consistent and the design of this part of the ACP is a bit unique in the ACP.

I think we can merge it like that.

PR #4709, opened 2017-02-23
Update from 1.11.0 to 2.2.4

fonziset:
Checklist:

- [ ] Tests pass


CHItA:
A new style is in the picture for 3.3 with possibly proper front end dependency management, so i'm unsure whether or not this will ever get merged, however, thank you for the contribution.

@hanakin

DavidIQ:
This was likely not tested with any of the current JavaScript included in the project so I wouldn't merge this at all as I don't see any changes or fixes to any of the core's JS, changes that are almost certain to be needed when going from jQuery 1.x to 2.x. Plus there's nothing here upgrading the CDN for jQuery.

CHItA:
Not sure what it would or wouldn't break as i haven't tested this yet (i could imagine that it doesn't break because we don't really do anything super advanced in js).

hanakin:
The entirety of the js will need to be rewritten with the new theme, leveraging Jquery where possible. I agree we should consider updating it when try but we are no where near try for the js aspects not to mention prosilver will need to still function so if the proper testing/changes to prosilver/admin themes  are taken into consideration then I see no issue

PR #4710, opened 2017-02-23
[ticket/15086] Replace quote.gif with Font Awesome icon

nomind60s:
Use the Font Awesome icon for a quote rather than the quote.gif file.

I believe I've handled the RTL case as well.

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15086

### DIFF
--- a/phpBB/styles/prosilver/theme/bidi.css
+++ b/phpBB/styles/prosilver/theme/bidi.css
@@ -502,15 +502,13 @@ li.breadcrumbs span:first-child > a {
 
 .rtl blockquote cite {
 	/* Username/source of quoter */
-	margin-right: 0;
 	margin-left: 0;
 }
 
 .rtl blockquote cite:before, .uncited:before {
-	/* Font Awesome quote-right */
-	content: '\f10e';
 	padding-right: 0;
 	padding-left: 5px;
+	content: '\f10e'; /* Font Awesome quote-right */
 }
 
 .rtl blockquote .codebox {
--- a/phpBB/styles/prosilver/theme/bidi.css
+++ b/phpBB/styles/prosilver/theme/bidi.css
@@ -492,7 +492,6 @@ li.breadcrumbs span:first-child > a {
 /* Quote block */
 .rtl blockquote {
 	margin: 0.5em 25px 0 1px;
-	background-position: 99% 8px;
 }
 
 .rtl blockquote blockquote {
--- a/phpBB/styles/prosilver/theme/bidi.css
+++ b/phpBB/styles/prosilver/theme/bidi.css
@@ -506,10 +506,9 @@ li.breadcrumbs span:first-child > a {
 	margin-left: 0;
 }
 
-.rtl blockquote cite:before {
+.rtl blockquote cite:before, .uncited:before {
 	/* Font Awesome quote-right */
 	content: '\f10e';
-	font-family: FontAwesome;
 	padding-right: 0;
 	padding-left: 5px;
 }
--- a/phpBB/styles/prosilver/theme/bidi.css
+++ b/phpBB/styles/prosilver/theme/bidi.css
@@ -505,6 +505,10 @@ li.breadcrumbs span:first-child > a {
 	margin-left: 0;
 }
 
+.rtl blockquote cite:before, .rtl .uncited:before {
+	padding-left: 5px;
+}
+
 .rtl blockquote .codebox {
 	margin-right: 0;
 }
--- a/phpBB/styles/prosilver/theme/bidi.css
+++ b/phpBB/styles/prosilver/theme/bidi.css
@@ -505,12 +505,6 @@ li.breadcrumbs span:first-child > a {
 	margin-left: 0;
 }
 
-.rtl blockquote cite:before, .uncited:before {
-	padding-right: 0;
-	padding-left: 5px;
-	content: '\f10e'; /* Font Awesome quote-right */
-}
-
 .rtl blockquote .codebox {
 	margin-right: 0;
 }
--- a/phpBB/styles/prosilver/theme/bidi.css
+++ b/phpBB/styles/prosilver/theme/bidi.css
@@ -502,15 +502,15 @@ li.breadcrumbs span:first-child > a {
 
 .rtl blockquote cite {
 	/* Username/source of quoter */
-	margin-right: 0px;
+	margin-right: 0;
 	margin-left: 0;
 }
 
-blockquote cite:before {
+.rtl blockquote cite:before {
 	/* Font Awesome quote-right */
 	content: '\f10e';
 	font-family: FontAwesome;
-	padding-right: 0px;
+	padding-right: 0;
 	padding-left: 5px;
 }
 
--- a/phpBB/styles/prosilver/theme/bidi.css
+++ b/phpBB/styles/prosilver/theme/bidi.css
@@ -502,8 +502,16 @@ li.breadcrumbs span:first-child > a {
 
 .rtl blockquote cite {
 	/* Username/source of quoter */
-	margin-right: 20px;
+	margin-right: 0px;
 	margin-left: 0;
+}
+
+blockquote cite:before {
+	/* Font Awesome quote-right */
+	content: '\f10e';
+	font-family: FontAwesome;
+	padding-right: 0px;
+	padding-left: 5px;
 }
 
 .rtl blockquote .codebox {
--- a/phpBB/styles/prosilver/theme/colours.css
+++ b/phpBB/styles/prosilver/theme/colours.css
@@ -409,12 +409,7 @@ dl.faq dt {
 /* Quote block */
 blockquote {
 	background-color: #EBEADD;
-	background-image: url("./images/quote.gif");
 	border-color:#DBDBCE;
-}
-
-.rtl blockquote {
-	background-image: url("./images/quote_rtl.gif");
 }
 
 blockquote blockquote {
--- a/phpBB/styles/prosilver/theme/content.css
+++ b/phpBB/styles/prosilver/theme/content.css
@@ -474,6 +474,10 @@ blockquote cite cite {
 	font-size: 1em;
 }
 
+blockquote cite:before, .uncited:before {
+	padding-right: 5px;
+}
+
 blockquote cite > div {
 	float: right;
 	font-weight: normal;
--- a/phpBB/styles/prosilver/theme/content.css
+++ b/phpBB/styles/prosilver/theme/content.css
@@ -466,9 +466,16 @@ blockquote cite {
 	/* Username/source of quoter */
 	font-style: normal;
 	font-weight: bold;
-	margin-left: 20px;
+	margin-left: 0px;
 	display: block;
 	font-size: 0.9em;
+}
+
+blockquote cite:before {
+	/* Font Awesome quote-left */
+	content: '\f10d';
+	font-family: FontAwesome;
+	padding-right: 5px;
 }
 
 blockquote cite cite {
--- a/phpBB/styles/prosilver/theme/content.css
+++ b/phpBB/styles/prosilver/theme/content.css
@@ -466,24 +466,12 @@ blockquote cite {
 	/* Username/source of quoter */
 	font-style: normal;
 	font-weight: bold;
-	margin-left: 0;
 	display: block;
 	font-size: 0.9em;
 }
 
-blockquote cite:before, .uncited:before {
-	/* Font Awesome quote-left */
-	content: '\f10d';
-	font-family: FontAwesome;
-	padding-right: 5px;
-}
-
 blockquote cite cite {
 	font-size: 1em;
-}
-
-blockquote.uncited {
-	padding-top: 0;
 }
 
 blockquote cite > div {
--- a/phpBB/styles/prosilver/theme/content.css
+++ b/phpBB/styles/prosilver/theme/content.css
@@ -483,7 +483,14 @@ blockquote cite cite {
 }
 
 blockquote.uncited {
-	padding-top: 25px;
+	padding-top: 0;
+}
+
+blockquote.uncited:before {
+	/* Font Awesome quote-left */
+	content: '\f10d';
+	font-family: FontAwesome;
+	padding-right: 5px;
 }
 
 blockquote cite > div {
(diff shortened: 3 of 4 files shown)


phpBB/styles/prosilver/theme/bidi.css
```diff
@@ -502,10 +502,18 @@ li.breadcrumbs span:first-child > a {
 
 .rtl blockquote cite {
 	/* Username/source of quoter */
-	margin-right: 20px;
+	margin-right: 0px;
```
vinny: remove `px`

phpBB/styles/prosilver/theme/bidi.css
```diff
@@ -502,10 +502,18 @@ li.breadcrumbs span:first-child > a {
...
 	margin-left: 0;
 }
 
+blockquote cite:before {
```
vinny: `.rtl ` before

phpBB/styles/prosilver/theme/content.css
```diff
@@ -466,11 +466,18 @@ blockquote cite {
 	/* Username/source of quoter */
 	font-style: normal;
 	font-weight: bold;
-	margin-left: 20px;
+	margin-left: 0px;
```
vinny: remove `px`

vinny:
Instead of adding CSS, you should simply use something like: `<i class="icon fa-quote-left fa-fw" aria-hidden="true"></i>`. 

Icon is missing when no user is quoted:


[quote.gif](https://github.com/phpbb/phpbb/blob/master/phpBB/styles/prosilver/theme/images/quote.gif) and [quote_rtl.gif](https://github.com/phpbb/phpbb/blob/master/phpBB/styles/prosilver/theme/images/quote_rtl.gif) should be excluded.

nomind60s:
Thanks for reviewing and offering suggestions, and finding that error. It's late here so I'll give it more thought in the morning. Need to rationalize with comments [here](http://area51.phpbb.com/phpBB/viewtopic.php?p=299166#p299166).

hanakin:
@vinny not sure its possible to handle this via html or at least I could not figure it out based on how the BBCODE engine works or does not work rather!

phpBB/styles/prosilver/theme/content.css
```diff
@@ -466,17 +466,31 @@ blockquote cite {
...
-	padding-top: 25px;
+	padding-top: 0;
+}
+
+blockquote.uncited:before {
+	/* Font Awesome quote-left */
+	content: '\f10d';
+	font-family: FontAwesome;
+	padding-right: 5px;
 }
```
hanakin: pls merge this with the cited version as the code is the same

hanakin: This is also a breaking change if you remove the gifs so keep them for BC until 3.3

phpBB/styles/prosilver/theme/bidi.css
```diff
@@ -502,10 +502,18 @@ li.breadcrumbs span:first-child > a {
...
 	/* Username/source of quoter */
-	margin-right: 20px;
+	margin-right: 0;
 	margin-left: 0;
 }
 
+.rtl blockquote cite:before {
+	/* Font Awesome quote-right */
+	content: '\f10e';
+	font-family: FontAwesome;
```
hanakin: this is redundent as its already included on the base element

phpBB/styles/prosilver/theme/colours.css
```diff
@@ -409,14 +409,9 @@ dl.faq dt {
...
 blockquote {
 	background-color: #EBEADD;
-	background-image: url("./images/quote.gif");
 	border-color:#DBDBCE;
 }
 
-.rtl blockquote {
-	background-image: url("./images/quote_rtl.gif");
-}
-
```
hanakin: This is also a breaking change if you remove the gifs so keep them for BC until 3.3

phpBB/styles/prosilver/theme/bidi.css
nomind60s: The base element specifies quote-left, for the RTL element we need quote-right, so I'm not clear on what's redundant here.

hanakin: the font-family

phpBB/styles/prosilver/theme/bidi.css
```diff
@@ -502,10 +502,17 @@ li.breadcrumbs span:first-child > a {
 
 .rtl blockquote cite {
 	/* Username/source of quoter */
-	margin-right: 20px;
+	margin-right: 0;
 	margin-left: 0;
```
hanakin: just remove the whole property

phpBB/styles/prosilver/theme/content.css
```diff
@@ -466,17 +466,24 @@ blockquote cite {
 	/* Username/source of quoter */
 	font-style: normal;
 	font-weight: bold;
-	margin-left: 20px;
+	margin-left: 0;
```
hanakin: again just remove it completely

phpBB/styles/prosilver/theme/content.css
```diff
@@ -466,17 +466,24 @@ blockquote cite {
...
+}
+
 blockquote cite cite {
 	font-size: 1em;
 }
 
 blockquote.uncited {
-	padding-top: 25px;
+	padding-top: 0;
 }
```
hanakin: same as above

phpBB/styles/prosilver/theme/content.css
```diff
@@ -466,17 +466,24 @@ blockquote cite {
...
 	font-weight: bold;
-	margin-left: 20px;
+	margin-left: 0;
 	display: block;
 	font-size: 0.9em;
 }
 
+blockquote cite:before, .uncited:before {
+	/* Font Awesome quote-left */
+	content: '\f10d';
```
hanakin: add a font-size of 14px to make the icon not look so squashed

hanakin: refactor to the following

```css
blockquote cite:before,
.uncited:before {
	font-family: FontAwesome;
	font-size: 14px;
	padding-right: 5px;
	content: '\f10d'; /* Font Awesome quote-left */
}
```

phpBB/styles/prosilver/theme/bidi.css
```diff
@@ -502,10 +502,17 @@ li.breadcrumbs span:first-child > a {
...
 }
 
+.rtl blockquote cite:before, .uncited:before {
+	/* Font Awesome quote-right */
+	content: '\f10e';
+	padding-right: 0;
+	padding-left: 5px;
+}
+
```
hanakin: refactor to the following

```css
.rtl blockquote cite:before,
.uncited:before {
	padding-right: 0;
	padding-left: 5px;
	content: '\f10e'; /* Font Awesome quote-right */
}
```

phpBB/styles/prosilver/theme/content.css
nomind60s: Appreciate the education. I presume this is to optimize/remove redundant information. Might I ask if you use a tool to determine this or just eye-ball it? I'm using Chrome's Inspect tool for testing.

hanakin: as you should chrome's tool is the best way to work,
I do use it for testing but this was a guess based on the code. 
1. Since we are not applying an img and using `before` it gives us a new box model
2. if you are setting everything to zero this is a clue that it probably can be removed.

phpBB/styles/prosilver/theme/content.css
hanakin: this should probably be moved to the icon.css file so all icons are in the same place.

also we should match the icon class with this for consistency in rendering

```css
blockquote cite:before,
.uncited:before {
    -webkit-font-smoothing: antialiased;
    -moz-osx-font-smoothing: grayscale;
	font: normal normal normal 14px/1 FontAwesome;
   	font-size: inherit;
    text-rendering: auto;
    display: inline-block;
    padding-right: 5px;
	content: "\f10d"; /* Font Awesome quote-left */
}
```

this may require tweaking some of the other css but it will help to ensure all the icons render the same

phpBB/styles/prosilver/theme/bidi.css
hanakin: might make more sense that this be separated out and moved to the icon.css file as well as long as it follows the code for the left
```css
.rtl blockquote cite:before,
.uncited:before {
	content: "\f10e"; /* Font Awesome quote-right */
}
```

nomind60s: In this case I'm inclined to leave it in bidi.css to keep all the RTL coding together rather than keeping all the icon coding together.

phpBB/styles/prosilver/theme/bidi.css
```diff
@@ -502,10 +502,15 @@ li.breadcrumbs span:first-child > a {
...
 .rtl blockquote cite {
 	/* Username/source of quoter */
-	margin-right: 20px;
 	margin-left: 0;
 }
 
+.rtl blockquote cite:before, .uncited:before {
+	padding-right: 0;
+	padding-left: 5px;
+	content: '\f10e'; /* Font Awesome quote-right */
```
hanakin: no move it its the same concept of colors for rtl they are still found in colors.css and not bidi.css(why TF is this not just called rtl.css)

@vinny any issues with this from styles perspective?

nomind60s: Done. My newcomer guess is bidi is short for bidirection.

vinny: @hanakin no.

phpBB/styles/prosilver/theme/icons.css
```diff
@@ -85,3 +85,26 @@
...
+	text-rendering: auto;
+	display: inline-block;
+	padding-right: 5px;
+	content: '\f10d'; /* Font Awesome quote-left */
+}
+
+.rtl blockquote cite:before, .rtl .uncited:before {
+	padding-right: 0;
+	padding-left: 5px;
+	content: '\f10e'; /* Font Awesome quote-right */
```
hanakin: Do not move the whole thing just the content prop

nomind60s: Now I'm getting a bit confused (or lack the understanding) because within icons.css there are already other examples where padding is specified, so why shouldn't it also apply in the case?

hanakin: my intent was not to confuse so sorry. The reason i suggested to move `content` only is for two reasons.

1. because it was split to begin with. The `content` props are replacing the `background-image` props. essentially everything from colours.css should move to icons.css and everything from rtl should remain in rtl as it was since this is for 3.2.x and we want to maintain as much backwards compatibility as possible by keeping changes as small as possible. Moving things around between files is usually something we try to avoid to keep things simple and sane. It might not have any ramifications to me or you but it may for end-users and theme authors who have made changes to the base theme.

2. Because in 3.3 we will be replacing all of the font-awsome stuff with svg icons for better compatibility/flexibility so i am pre-staging for that as icons.css will be completely overhauled.

nomind60s: Thanks for the quick reply and continued mentoring. It's all part of the education process and much appreciated as I learn the ins and outs. I think I  understand the technical details of your point 1 and definitely agree with the philosophical point of keeping changes as small as possible and compatible with other styles and end-users who might have made changes.

nomind60s: I've taken a fresh look using phpBB 3.2.0 as a baseline and followed your point 1. In colours.css the `background-image` property points to a gif of size 15x12. The size of the image is implicit and not specified as a property. In content.css and bidi.css a margin of 20px is specified, creating an implicit padding of 5px. 

In making the change to use Font Awesome icons, we're making all these properties explicit. That's a good thing.

To make sure I fully understand what you're suggesting as changes, icons.css should only have the following added: 
```
blockquote cite:before, .uncited:before {
    content: '\f10d'; /* Font Awesome quote-left */
}

.rtl blockquote cite:before, .rtl .uncited:before {
    content: '\f10e'; /* Font Awesome quote-right */
}
```
and the rest of the properties should be specified in content.css and bidi.css. Is this correct?

hanakin: yes, also instead of retyping everything for the icon coding, just add `blockquote cite:before, .uncited:before` as more selectors in icon.css at the top like so

```css
 .icon, 
.button .icon,
blockquote cite:before, 
.uncited:before  {
 	display: inline-block;
   	font-weight: normal;
   	font-style: normal;
   	font-variant: normal;
   	font-family: FontAwesome;
   	font-size: 14px;
   	line-height: 1;
   	text-rendering: auto; /* optimizelegibility throws things off #1094 */
   	-webkit-font-smoothing: antialiased;
   	-moz-osx-font-smoothing: grayscale;
 }
```

This way the only thing added to content.css is the `padding-right`

hanakin:
#NOTE this will be refactored in 3.3 to use base64 svgs

nomind60s:
I presume with two reviewer approvals that this is ready to be merged.

hanakin:
yes just need to restart the tests

nomind60s:
Someone else will have to restart the tests, I don't have the ability other than closing/re-opening. That's a sloppy work-around IMHO.

hanakin:
agreed  you can always post a link in icq asking for a reset

vinny:
I don't know if I missed something, @hanakin quote.gif and quote_rtl.gif should remain for some compatibility?

hanakin:
@vinny yes they should remain part of the repo otherwise it will break inheritance

phpBB/styles/prosilver/theme/icons.css
```diff
@@ -9,7 +9,7 @@
  * Just change the name of the font after the 14/1 to the name of
  * the font you wish to use.
  */
-.icon, .button .icon  {
+.icon, .button .icon, blockquote cite:before, .uncited:before  {
```
marc1706: Extra space between `:before` and `{`.

phpBB/styles/prosilver/theme/icons.css
```diff
@@ -85,3 +85,13 @@
     line-height: .9;
     height: 12px;
 }
+
+
```
marc1706: Extra new line that should be removed.

phpBB/styles/prosilver/theme/icons.css
nomind60s: Thanks. I thought I had checked for spacing issues. This relatively new contributor continues to learn  and thanks the phpBB team for their assistance, guidance and patience.

PR #4715, opened 2017-03-01
use pathinfo instead of strchr for $phpEx for better usability

naveen17797:
$phpEx hereafter uses PATHINFO_EXTENSION
strchr may fail if there is two extension like file.php.jpg
so pathinfo is used

Checklist:

- [x] Tests pass

### DIFF
--- a/phpBB/app.php
+++ b/phpBB/app.php
@@ -19,7 +19,7 @@
 */
 define('IN_PHPBB', true);
 $phpbb_root_path = (defined('PHPBB_ROOT_PATH')) ? PHPBB_ROOT_PATH : './';
-$phpEx = substr(strrchr(__FILE__, '.'), 1);
+$phpEx = pathinfo(__FILE__, PATHINFO_EXTENSION);
 include($phpbb_root_path . 'common.' . $phpEx);
 
 // Start session management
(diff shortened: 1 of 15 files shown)


Elsensee:
Thank you for your contribution. In its current state, however, we cannot accept it:

1. Your commit doesn't follow our commit message format, please see [this page](https://wiki.phpbb.com/Git#Commit_Messages) for further information.
2. Which webserver would be configured to execute images as php scripts? Also, if I'm correct, your proposed change would still return 'jpg' for your specified example, so this doesn't fix anything.
3. You accidentally commited your `config.php`. We actually ignore this file in our .gitignore, so I don't know, what went wrong there.
4. For future pull requests: Please create a ticket in our bug tracker and then use that ticket id here. For this "bug" we probably don't need a ticket as I don't see a reason to "fix" it. ;)

naveen17797:
hi oliver
im sorry that i didn't follow the rules for contributing.the reason is
that i have been contributing towards my friends projects.They didn't
have a specific set of rules for them.

And i feel that pathinfo is the good way to parse the extension

Hereafter i will contribute according to guidelines

Elsensee:
> And i feel that pathinfo is the good way to parse the extension

There always is more than one solution to a problem. I personally believe that both ways are equal so I don't see a reason to change it.

PR #4736, opened 2017-03-11
[ticket/15123] Check if extension was enabled/disabled before enable or disable

rubencm:
Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15123

### DIFF
--- a/phpBB/language/en/cli.php
+++ b/phpBB/language/en/cli.php
@@ -110,10 +110,10 @@ $lang = array_merge($lang, array(
 
 	'CLI_EXTENSION_DISABLE_FAILURE'		=> 'Could not disable extension %s',
 	'CLI_EXTENSION_DISABLE_SUCCESS'		=> 'Successfully disabled extension %s',
-	'CLI_EXTENSION_ALREADY_DISABLED'	=> 'Extension %s was already disabled',
+	'CLI_EXTENSION_DISABLE_DISABLED'	=> 'Extension %s is not enabled',
 	'CLI_EXTENSION_ENABLE_FAILURE'		=> 'Could not enable extension %s',
 	'CLI_EXTENSION_ENABLE_SUCCESS'		=> 'Successfully enabled extension %s',
-	'CLI_EXTENSION_ALREADY_ENABLED'		=> 'Extension %s was already enabled',
+	'CLI_EXTENSION_ENABLE_ENABLED'		=> 'Extension %s was already enabled',
 	'CLI_EXTENSION_NAME'				=> 'Name of the extension',
 	'CLI_EXTENSION_PURGE_FAILURE'		=> 'Could not purge extension %s',
 	'CLI_EXTENSION_PURGE_SUCCESS'		=> 'Successfully purged extension %s',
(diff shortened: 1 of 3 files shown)


phpBB/language/en/cli.php
```diff
@@ -110,8 +110,10 @@
 
 	'CLI_EXTENSION_DISABLE_FAILURE'		=> 'Could not disable extension %s',
 	'CLI_EXTENSION_DISABLE_SUCCESS'		=> 'Successfully disabled extension %s',
+	'CLI_EXTENSION_ALREADY_DISABLED'	=> 'Extension %s was already disabled',
```
javiexin: This message should probably be "Extension %s is not enabled".  An extension might be neither enabled nor disabled (available, or even, non-existent)...

rubencm: Should I change "Extension %s was already enabled" to "Extension %s is enabled" too?

javiexin: No, that is correct as I see it (but I would prefer present rather than past, but that is just taste).
For me, "Extension %s is already enabled" is ok.

phpBB/language/en/cli.php
```diff
@@ -110,8 +110,10 @@
 
 	'CLI_EXTENSION_DISABLE_FAILURE'		=> 'Could not disable extension %s',
 	'CLI_EXTENSION_DISABLE_SUCCESS'		=> 'Successfully disabled extension %s',
+	'CLI_EXTENSION_DISABLE_DISABLED'	=> 'Extension %s is not enabled',
 	'CLI_EXTENSION_ENABLE_FAILURE'		=> 'Could not enable extension %s',
 	'CLI_EXTENSION_ENABLE_SUCCESS'		=> 'Successfully enabled extension %s',
+	'CLI_EXTENSION_ENABLE_ENABLED'		=> 'Extension %s was already enabled',
```
VSEphpbb: is, not was

VSEphpbb: Just call the keys CLI_EXTENSION_DISABLED and CLI_EXTENSION_ENABLED

phpBB/language/en/cli.php
```diff
@@ -110,8 +110,10 @@
 
 	'CLI_EXTENSION_DISABLE_FAILURE'		=> 'Could not disable extension %s',
 	'CLI_EXTENSION_DISABLE_SUCCESS'		=> 'Successfully disabled extension %s',
+	'CLI_EXTENSION_DISABLE_DISABLED'	=> 'Extension %s is not enabled',
```
VSEphpbb: "is already disabled" (consistent with the other key)

javiexin: No Matt, please take a look at my prior comment: the extension could be available but not disabled, or even not available, so the message would be incorrect stating that it is disabled.  The check is explicitly for `!is_enabled` that is different from `is_disabled`

phpBB/language/en/cli.php
javiexin: Agree with both (as I mentioned in the prior comment, I prefer present as well)

rubencm:
I return 1 when this error ocurrs. Should i return 2, or is reserved for other stuff?

Nicofuma:
Thanks for contributing

PR #4740, opened 2017-03-13
[ticket/14974] Make cookie notice configurable via language files

rubencm:
PHPBB3-14974

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-14974

### DIFF
--- a/phpBB/language/en/common.php
+++ b/phpBB/language/en/common.php
@@ -181,6 +181,7 @@ $lang = array_merge($lang, array(
 	'COOKIE_CONSENT_INFO'	=> 'Learn more',
 	'COOKIE_CONSENT_MSG'	=> 'This website uses cookies to ensure you get the best experience on our website.',
 	'COOKIE_CONSENT_OK'		=> 'Got it!',
+	'COOKIE_CONSENT_HREF'	=> 'http://cookiesandyou.com',
 	'COOKIES_DELETED'		=> 'All board cookies successfully deleted.',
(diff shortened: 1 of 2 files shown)


phpBB/language/en/common.php
```diff
@@ -181,6 +181,7 @@
 	'COOKIE_CONSENT_INFO'	=> 'Learn more',
 	'COOKIE_CONSENT_MSG'	=> 'This website uses cookies to ensure you get the best experience on our website.',
 	'COOKIE_CONSENT_OK'		=> 'Got it!',
+	'COOKIE_CONSENT_HREF'	=> 'http://cookiesandyou.com',
```
CHItA: I would suggest `//` instead of `http://`.

CHItA:
Thanks for contributing.

rubencm:
I dont think thats a good idea since they do not have their ssl certificate properly configured.
Also thats their default option, as you can see here
https://github.com/insites/cookieconsent/blob/master/src/cookieconsent.js#L213

phpBB/language/en/common.php
rubencm: I dont think thats a good idea since they do not have their ssl certificate properly configured.
Also thats their default option, as you can see here
https://github.com/insites/cookieconsent/blob/master/src/cookieconsent.js#L213

CHItA: Well, then that's that.

PR #4752, opened 2017-03-21
[ticket/15137] Global Announcements shouldn't always be never ending

nomind60s:
To be consistent with Sticky and Announcements topic types, Global announcements can 
now stick for a non-zero number of days, i.e. they are no longer never ending.

Before this change, if the topic type was Global and **Stick topic for:** was set to a non-zero 
number of days, the number was silently ignored and a Global announcement was always 
never ending.

PHPBB3-15137

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15137


phpBB/language/en/posting.php
```diff
@@ -241,7 +241,7 @@
 	'SMILIES_ARE_ON'			=> 'Smilies are <em>ON</em>',
 	'STICKY_ANNOUNCE_TIME_LIMIT'=> 'Sticky/Announcement time limit',
 	'STICK_TOPIC_FOR'			=> 'Stick topic for',
-	'STICK_TOPIC_FOR_EXPLAIN'	=> 'Enter 0 for a never ending Sticky/Announcement. Please note that this number is relative to the date of the post.',
+	'STICK_TOPIC_FOR_EXPLAIN'	=> 'Enter 0 for a never ending Sticky/Announcement/Global. Please note that this number is relative to the date of the post.',
```
DavidIQ: I don't think this is necessary. There are only 2 types of posts, not three.

nomind60s: I waffled on it, but there are four choices given, Normal, Sticky, Announcement and Global. I've been around phpBB long enough to know that Global is a special case of Announcement, but I couldn't assume that every board admin would know that, so thought adding it would make it complete and clear.

DavidIQ: You would have to be consistent and also change STICKY_ANNOUNCE_TIME_LIMIT. I'm not sure this change is needed at all however. I'll let the person that would merge this make that decision.

nomind60s: You are correct I missed that one. I'm still on the fence on whether the change is necessary as well and appreciate the feedback. I tend to err on the side of caution and over-explaining. I can live with having whomever merges make the decision also. They've probably got more experience with phpBB code development than I.

Nicofuma: I don't really mind if it is a bit too explicit. Especially because it will makes it clear for existing users used to the current behavior.

Nicofuma:
Thanks for contributing (merged via c9aab0b)

nomind60s:
I'm a bit confused as to why the changes to `includes/functions_posting.php` were not merged. They are part of the fix.

Nicofuma:
certainly an mistake in the merge conflict resolution

nomind60s:
I know what the problem is, a variable name used in the changed line is different between 3.1.x and 3.2.x. Who takes care of the merge conflict resolution, the fixer or the merger?

Nicofuma:
it depends, but in this case I did and I missed it. But it's not a big deal, just send a new PR against 3.2 fixing the issue (if you don't I will do it later). You can reuse the same ticket number.

nomind60s:
I understand the changes to make, but I'm still learning the github/pull request stuff so I'll leave it to you. From what I can see, none of the changes are merged to 3.1.x and they are partially merged to 3.2.x. Too many parts that I'm not comfortable that I understand at this point.

PR #4758, opened 2017-03-23
[ticket/PHPBB3-10424] Added Date-filter to viewtopic_body.html

balabky9:
As per ticket to only display the Join date on forum posts of a user an edit
was done to viewtopic_body.html of styles/prosilver where a twig date formater
was utilized to get only the Day, Date Month-name and Year in this format.

TICKET-PHPBB3-10424

Checklist:

- [ ] Tests pass

Tracker ticket (set the ticket ID to *10424*):

https://tracker.phpbb.com/browse/PHPBB3-10424


hanakin:
why add a date filter to this one occurence? all dates are currently handled consistently via the profile setting in prosilver. While I agree the implementation is poor and should be something the theme author has more control over, consistency is should be maintained.

balabky9:
Hanakin the othe option was changing it in the php file. Marc and I were discussing this on IRC and he favoured that approach more. I was in favour of this approach since it allows theme authors to customize this value and retians the default date data which is being passed to the view. But if that solution seems more consistent I shall make a new PR then.

PR #4763, opened 2017-03-24
[ticket/15146] Date profile field validation incorrect

javiexin:
Fix by changing select input field options to match validation range

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-15146

### DIFF
--- a/phpBB/phpbb/profilefields/type/type_date.php
+++ b/phpBB/phpbb/profilefields/type/type_date.php
@@ -264,7 +264,7 @@ class type_date extends type_base
 		}
 
 		$profile_row['s_year_options'] = '<option value="0"' . ((!$year) ? ' selected="selected"' : '') . '>--</option>';
-		for ($i = $now['year'] - 100; $i <= $now['year'] + 100; $i++)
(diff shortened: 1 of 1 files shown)


marc1706:
Hm, I think having this vary over time actually makes more sense to hard-code 1901 into the code.

javiexin:
Ok, what do you propose?  Today -100 to today+100? Then we fix the other code that has this hardcoded...
Just let me know what.

javiexin:
On second thought, I think it is better to keep a fixed start and move the end as relative.
The reason is that maybe today you create a profile field, and give it a value of the first allowed year.
Then two years down the road, you want to change the value of another profile field, and you will not be able without updating this to the "new" range.
So, I would recommend keeping it as is, but will do whatever you decide is best.

Another alternative would be to make date profile fields have a validation range configurable by the forum admin, through the ACP.  I have done this as an extension, using service replacement, but may be worth including in the core.  Up to you.  For reference: https://github.com/javiexin/valdatepf
This may be converted to a PR in very short time if required/recommended.

PR #4767, opened 2017-03-26
[ticket/10424] Added Date-filter to viewtopic_body.html

balabky9:
As per ticket to only display the Join date on forum posts of a user an edit
was done to viewtopic_body.html of styles/prosilver where a twig date formater
was utilized to get only the Day, Date Month-name and Year in this format.

PHPBB3-10424

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-10424

### DIFF
--- a/phpBB/styles/prosilver/template/viewtopic_body.html
+++ b/phpBB/styles/prosilver/template/viewtopic_body.html
@@ -105,7 +105,7 @@
 			<!-- IF S_CAN_VOTE -->
 				<dl style="border-top: none;" class="poll_vote">
 					<dt>&nbsp;</dt>
-					<dd class="resultbar"><input type="submit" name="update" value="{L_SUBMIT_VOTE}" class="button1 button button-form" /></dd>
+					<dd class="resultbar"><input type="submit" name="update" value="{L_SUBMIT_VOTE}" class="button1" /></dd>
 				</dl>
 			<!-- ENDIF -->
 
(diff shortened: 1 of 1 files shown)


hanakin:
1. why did you recreate the PR (https://github.com/phpbb/phpbb/pull/4758) Close one of them.
2. You need to update your repo as it seems to be behind your PR is removing newly added classes
3. Check comment on old repo

vinny:
As mentioned by bantu in the ticket, I think the line should be removed.

balabky9:
I find the join date valuable. As a forum surfer if a person has been on the forum for a longer time it generally suggest better topic maturity and understanding. So its an important part to display it.

hubaishan:
This PR does not add ability to customize the join date format, the extension in https://www.phpbb.com/customise/db/extension/joined_date_format is more powerful, so I suggest to merge the ext to phpBB and add ability to block joined date as some developers suggest.

DavidIQ:
Yeah no thanks. That's overkill. Stick to the extension if you really want that much control over the display of the join date.

That being said I think removing the join date is something that could be revisited for the new style, but not right now for prosilver.

hubaishan:
Also I am not sure that this PR will handle the timezone of the user how viewing the forum.

balabky9:
It would since its just filtering on the template side. The date variable being passed into the view is not touched at all.

marc1706:
The PR will ignore user's date format settings and only work in the server's default time zone. Also, I don't think directly adding this to the HTML is a good idea.

PR #4773, opened 2017-03-27
[ticket/15154] Load compatibility globals into the CLI installer

VSEphpbb:
PHPBB3-15154

Checklist:

- [x] Tests pass

### DIFF
--- a/phpBB/install/phpbbcli.php
+++ b/phpBB/install/phpbbcli.php
@@ -39,6 +39,9 @@ $input = new ArgvInput();
 // Enable superglobals for cli support
 $phpbb_installer_container->get('request')->enable_super_globals();
 
+// Load compatibility globals to access deprecated globals
+require($phpbb_root_path . 'includes/compatibility_globals.' . $phpEx);
+
 /** @var \phpbb\filesystem\filesystem $phpbb_filesystem */
 $phpbb_filesystem = $phpbb_installer_container->get('filesystem');
 


phpBB/install/phpbbcli.php
```diff
@@ -39,6 +39,9 @@
 // Enable superglobals for cli support
 $phpbb_installer_container->get('request')->enable_super_globals();
 
+// Load compatibility globals to access deprecated globals
+require($phpbb_root_path . 'includes/compatibility_globals.' . $phpEx);
```
Nicofuma: you need to call `register_compatibility_globals()` too.

VSEphpbb: No, Just need the constants which are defined outside of that function.

marc1706: Is there a reason not to call that or is it that you just don't see a point in it right now? If it's the latter I'd recommend just calling it ;)

VSEphpbb: The main reason I had not to was it requires the global `$phpbb_container`. Which does not exist here. The installer here has it's own version of the container `phpbb_installer_container` . I don't know if it's any different. I also don't see the point of calling that function and having it define all those globals it does, when it's not used/needed here. Allz we need are the 3 constants.

VSEphpbb: Confirmed `register_compatibility_globals()` can not be used in the installer CLI because the containers are different, and most of the services in register globals do not exist in the installer's container. So no need to call that method.

PR #4774, opened 2017-03-27
[ticket/15057] Rewording the FIND_USERNAME_EXPLAIN

jajodiaraghav:
PHPBB3-15057

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15057

### DIFF
--- a/phpBB/language/en/memberlist.php
+++ b/phpBB/language/en/memberlist.php
@@ -69,7 +69,7 @@ $lang = array_merge($lang, array(
 	'EMPTY_SUBJECT_EMAIL'	=> 'You must specify a subject for the email.',
 	'EQUAL_TO'				=> 'Equal to',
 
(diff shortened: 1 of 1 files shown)


VSEphpbb:
Thanks for theis PR but this change is not acceptable. phpBB always refers to it's registerants as "members". That should not be changed to users. Also, there are in fact places where multiple members can be selected at once, depending on the form, exactly as described in the original wording (for example in the user control panel on the friends/foes Find a Member form).

screen shot 2017-03-27 at 8 49 45 am

jajodiaraghav:
Then how can we fix the issue PHPBB3-15057. Or is the fix for this issue not required?

DavidIQ:
Basically you should ignore the "members" thing from the ticket. The explanation text could be changed if not coming from ACP so you would have to figure out how to do that. You would end up needing to build the explanation text on the PHP side and dump that out to the template.

VSEphpbb:
Well, actually you'd need to examine everywhere this language is used, and determine if it's a multi or singular selection page, and dole out the intended text for each scenario, which IMO is a little ridiculous considering the original text has it mentioned in parenthesis that some forms allow multiple selects....and this has never been an issue or a complaint or any sort of bug.

PR #4776, opened 2017-03-28
[ticket/15155] Correctly bold the buttons

hanakin:
PHPBB3-15155

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15155

### DIFF
--- a/phpBB/styles/prosilver/template/captcha_default.html
+++ b/phpBB/styles/prosilver/template/captcha_default.html
@@ -12,7 +12,7 @@
 		<dt><label for="confirm_code">{L_CONFIRM_CODE}{L_COLON}</label></dt>
 		<dd class="captcha captcha-image"><img src="{CONFIRM_IMAGE_LINK}" alt="{L_CONFIRM_CODE}" /></dd>
 		<dd><input type="text" name="confirm_code" id="confirm_code" size="8" maxlength="8" tabindex="{$CAPTCHA_TAB_INDEX}" class="inputbox narrow" title="{L_CONFIRM_CODE}" />
-		<!-- IF S_CONFIRM_REFRESH --><input type="submit" name="refresh_vc" id="refresh_vc" class="button1 button button-form-bold" value="{L_VC_REFRESH}" /><!-- ENDIF -->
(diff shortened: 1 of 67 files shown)


hanakin:
!set Styles

hanakin:
!set WIP

hanakin:
should I continue to work this fix or should we just make all buttons consistent?

screenshot 2017-05-13 09 16 26

VSEphpbb:
screen shot 2018-01-28 at 9 29 28 am
These sort of buttons are still way too bold.

hanakin:
@VSEphpbb 
1. that's unrelated to this pr, this pr is dealing with the form buttons being bolded where they should/n't.
2. yes you have to use the google cdn fonts not sure why you are not?

VSEphpbb:
1. If it's unrelated to this PR then why did you ask about continuing on with those buttons (including same screenshot)?

2. It's not about whether I have CDN on or off. It's about the fact that the users may or may not (and the fact that it's off by default). We should still have decent looking button text. The buttons used to look great on their own until they started getting changed very recently. Just the right CSS font and font styling should be simple enough to correct.

hanakin:
1. That was an old post at the time I was referring to whether we wanted to refactor all buttons to the same class/style instead of having two or three different styles, but we have since reverted to maintaining prosilver as is and making all improvements in the new theme to speed up development and simplify the release cycle. The same point applies to fixing the buttons as well. Its easier to fix in a new theme then attempt to overhaul prosilver. This pr as is is not a breaking change or a minor one if at all as it just swaps the classes.

VSEphpbb:
I just do not understand the reluctance to fix these button. In a PR titled "Correctly bold the buttons" these are a button style clearly broken in terms of bolding as of 3.2.2. And all that's needed is to correct their font weight to 400 (or just remove it all together like it is in 3.2.0/3.2.1!!)

Saying it will be fixed in a new theme makes no sense as it was just broken as of the latest release. Literally these specific buttons only need a a class with their font weight corrected, and it should be fixed.

hanakin:
Because this pr is for form buttons making the right buttons bold. We can look at anything else in a separate pr. I would have to look into your problem to find the right fix as 400 is not a bold as 500 is normal default font size so that solves nothing. This pr is ready as is to review

PR #4777, opened 2017-03-28
[ticket/14987] Board closed error message can contain valid HTML

balabky9:
Adds ability to add HTML formatting to board closed error messages by
adding html_entities_decode function
before displaying the board error message.

PHPBB3-14987

Checklist:

- [ ] Tests pass

Tracker ticket (set the ticket ID to **14987**):

https://tracker.phpbb.com/browse/PHPBB3-14987

### DIFF
--- a/phpBB/phpbb/user.php
+++ b/phpBB/phpbb/user.php
@@ -354,7 +354,7 @@ class user extends \phpbb\session
 			{
 				$message = (!empty($config['board_disable_msg'])) ? $config['board_disable_msg'] : 'BOARD_DISABLE';
 			}
-			trigger_error($message);
+			trigger_error(html_entity_decode($message));
 		}
 
 		// Is board disabled and user not an admin or moderator?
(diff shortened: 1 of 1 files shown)


marc1706:
I don't think it's a good idea to allow HTML in something that should be a simple message.

balabky9:
It is helpful in the sense let us say you want to add a link to the message, which goes to a twitter post explaining the issue of an outage, or you might want to bold a part, adjust font and colors etc. It is not just a message but a way of expressing your regret to your user. You can also then probably put in a timer counting down there. Many possibilities by a simple change. Its all about presentation in a forum, so I thought it will be a good idea.

marc1706:
Don't you think it might be better to support BBCodes instead of HTML in those messages? Might actually be a nicer approach then having to enter HTML directly.

balabky9:
I did think so and I'll implement a solution for it.
Both can be supported maybe.
But I don't see a big advantage. If we use bbcode then we would need a
bb-code editor placed over there, so other template changes are needed, but
definitely a nice idea.

marc1706:
Closing as previously noted.

PR #4782, opened 2017-03-29
[ticket/14498] Don't display 'Who is online' members when permission is No

nomind60s:
When the user permission 'Can view profiles, memberlist and online list' is
set to No (or Never), the user should not be able to see which members are
online in the following places, index page, view topic and view forum.

Before this change, guests and bots would see the list of members who are
online, which doesn't match with the user permission and could create privacy
issues where guests or bots could track when a member was online.

PHPBB3-14498

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-14498

### DIFF
--- a/phpBB/styles/prosilver/template/index_body.html
+++ b/phpBB/styles/prosilver/template/index_body.html
@@ -40,8 +40,11 @@
 		<!-- IF U_VIEWONLINE --><h3><a href="{U_VIEWONLINE}">{L_WHO_IS_ONLINE}</a></h3><!-- ELSE --><h3>{L_WHO_IS_ONLINE}</h3><!-- ENDIF -->
 		<p>
 			<!-- EVENT index_body_block_online_prepend -->
-			{TOTAL_USERS_ONLINE} ({L_ONLINE_EXPLAIN})<br />{RECORD_USERS}<br /> <br />{LOGGED_IN_USER_LIST}
-			<!-- IF LEGEND --><br /><em>{L_LEGEND}{L_COLON} {LEGEND}</em><!-- ENDIF -->
+			{TOTAL_USERS_ONLINE} ({L_ONLINE_EXPLAIN})<br />{RECORD_USERS}<br /> 
+			<!-- IF U_VIEWONLINE -->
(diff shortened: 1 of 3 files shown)


DavidIQ:
In my opinion I think we should just axe the online user list for guests at https://github.com/phpbb/phpbb/blob/3.2.x/phpBB/includes/functions.php#L3621 instead of having a template edit in 3 locations. Might also need to look at either an existing permission or a new one.

nomind60s:
There is an existing permission, it just wasn't being checked in all cases. These changes display or hide 'Who is online' per the permission.

DavidIQ:
Yes I understand that. What I'm saying is that rather than editing 3 template files and expecting all style authors to make the change it might be better to change it in one location.

DavidIQ:
Yes I understand that. What I'm saying is that rather than editing 3 template files and expecting all style authors to make the change it might be better to change it in one location.

nomind60s:
I've looked at includes/functions.php and can see it would be possible to return a null LOGGED_IN_USER_LIST when the user permission to view was No. However, I think that would create an awkward or inconsistent output to the user. Look at what's displayed in index_body.html. It includes both the LOGGED_IN_USER_LIST and the LEGEND information. If the LOGGED_IN_USER_LIST was null, the LEGEND would still appear and look out of place. From my viewpoint, the cleaner, more consistent solution is the style changes that I've submitted.

I realize this impacts all style developers, so perhaps the best solution is merge only to 3.2.x, where new events and other style changes are still happening and expected. This would be just one of a number of changes they'd have to deal with.

JimMH:
I would also go for the changes in the template files. It's a relatively minor change for style authors and it would also be easier to chase down for casual board admins that want the "old" behaviour back.

DavidIQ:
Well the other unmentioned part of my suggestion was that with just the template change there are still at least a pair of database queries being done that would not be needed which are the calls to obtain_users_online and obtain_users_online_string. And, on closer inspection, the obtain_users_online_string function probably needs an adjustment anyways since the query being done doesn't even apply for anonymous users (line 3664 should be incorporated into line 3629).

PR #4807, opened 2017-04-23
[ticket/15190] Consolidate validation methods in metadata manager

javiexin:
Use a single method to access all validation options in metadata manager. Before, some validations had their own function, while others shared a single validate method. Now, the validate method acts as a proxy for all others.
Also, with the existing implementation, you should not call validate without calling get_metadata first, as get_metadata was the only way to do the file loading (any call to validate without calling get_metadata first would result in an exception being raised even if the metadata is correct); with this change, the order is irrelevant, and the file is only loaded once.  Some uses of get_metadata that were not using the returned data have been replaced with the corresponding validate call.

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-15190

### DIFF
--- a/phpBB/phpbb/extension/metadata_manager.php
+++ b/phpBB/phpbb/extension/metadata_manager.php
@@ -203,6 +203,7 @@ class metadata_manager
 	*
 	* @return boolean True when passes validation, throws exception if invalid
 	* @throws \phpbb\extension\exception
+	* @internal		Should not be used directly. As of phpBB 4.0, the method will become protected.
 	*/
 	public function validate_authors()
 	{
@@ -227,6 +228,7 @@ class metadata_manager
 	*
 	* @return bool True if validation succeeded, throws an exception if invalid
 	* @throws \phpbb\extension\exception
+	* @internal		Should not be used directly. As of phpBB 4.0, the method will become protected.
 	*/
 	public function validate_enable()
 	{
@@ -239,6 +241,7 @@ class metadata_manager
 	*
 	* @return boolean True when passes validation, throws an exception if invalid
 	* @throws \phpbb\extension\exception
+	* @internal		Should not be used directly. As of phpBB 4.0, the method will become protected.
 	*/
 	public function validate_dir()
 	{
@@ -256,6 +259,7 @@ class metadata_manager
 	*
 	* @return boolean True when passes validation, throws an exception if invalid
 	* @throws \phpbb\extension\exception
+	* @internal		Should not be used directly. As of phpBB 4.0, the method will become protected.
 	*/
 	public function validate_require_phpbb()
 	{
@@ -272,6 +276,7 @@ class metadata_manager
 	*
 	* @return boolean True when passes validation, throws an exception if invalid
 	* @throws \phpbb\extension\exception
+	* @internal		Should not be used directly. As of phpBB 4.0, the method will become protected.
 	*/
 	public function validate_require_php()
 	{
--- a/phpBB/phpbb/extension/metadata_manager.php
+++ b/phpBB/phpbb/extension/metadata_manager.php
@@ -61,11 +61,13 @@ class metadata_manager
 			'enable'		=> ['dir', 'require_php', 'require_phpbb'],
 			'display'		=> ['fields', 'authors'],
 			'fields'		=> ['name', 'type', 'license', 'version'],
+			'version-check'	=> ['version', 'version_check'],
 			// Methods
 			'authors'		=> null,
 			'dir'			=> null,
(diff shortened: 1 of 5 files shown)


javiexin:
!set WIP

javiexin:
!unset WIP

Nicofuma:
All validate_* methods should be marked @internal or @deprecated with a note saying that in 4.0 they won't be public anymore.

About the validate() method, I don't really like how it works now. I think that all should validate everything.

Maybe we can work 2 arrays:
 - An array of validators (Should be a callable, `[$this, 'validate_enable']` per example)
 - A map associating a target to a list of validators
Then when `validate($a)` is called, we check if there is a such target or if there is a such validators and then we call them.

e.g:
```
$validators = [
    'enable' => [$this, 'validate_enable'],
    'dir' => [$this, 'validate_dir'],
];

$map = [
    'all' => ['enable', 'dir'],
    'display' => ['enable'],
];

function validate($target)
{
    $v = isset($map[$target]) ? $map[$target] : (array) $target;

    foreach ($v as $validator) {
        $validators[$validator](); // Plus handle faillure or anything else if necessary
    }
}
```

javiexin:
I like your proposal.  If we are refactoring validate completely, maybe worth adding a validation (and get) for version_check and display_name, quite used in the code.
Will work on this.

While we are at this, what do you think about making metadata completely "opaque", that is, using extension manager as the only accessor to metadata manager, providing proxy methods in extension manager to validate and get metadata?  That would be a separate (but related) PR.  Same thing could be done for extension objects themselves (only `is_enableable` is accessed from outside of the direct control of extension manager).

javiexin:
Ok, I have the code, as follows:
1) Create a single array, as a class property, filled in the constructor (extending the existing one, but moving it to a class property and initializing once)
2) Array key is the validation name, array value is mixed, as follows:
   - `null` for callable methods, of the form `'validate_'$key`
   - string for reg exp validation of regular fields
   - array for combined validation by nesting (self calling)
3) The validate function accesses the array, and determines how to proceed:
   - calling the specific function
   - performing the regular expression check
   - self-calling with the component validations
Is this the right approach?

I could revert back to three arrays, but would be as complex, and more verbose...  Your choice :)

As for the other comment, I will mark all validate_ methods as internal (not deprecated) as they will continue existing, although moving them to protected visibility (as of 4.0, right?).

Nicofuma:
Lets see how it looks.

About merging the extension manager and the metadata manager I think it is a bad idea, lets keep concerns separated

phpBB/phpbb/extension/metadata_manager.php
```diff
@@ -47,6 +53,25 @@ public function __construct($ext_name, $ext_path)
 		$this->ext_name = $ext_name;
 		$this->metadata = array();
 		$this->metadata_file = $ext_path . 'composer.json';
+
+		// Initialize validations
+		$this->validations = array(
```
Nicofuma: should be a static array

javiexin: Agreed.  Didn't do it as I was not sure it was ok for the minimum PHP version.
Will do it tomorrow.

javiexin:
Three commits:
1) New validate method (will change the array initialization to static)
2) Include the "internal" message into the docblocks, please review if ok
3) New method for version_check (maybe it should be a separate PR?)
   Note that this is not yet 100% complete: I want to review the uses of check_version from acp_extension and check CLI command, and may want to adapt to use the new validation option (will work as is, but might be better to update it).

Regarding merging extension manager and metadata manager, I think I miss-explained what I meant.
I do NOT want to merge both.  I just want to make any direct access to metadata manager redundant, as it may be accessed through the extension manager as a proxy for md mgr.  They are always used together, so why not use just one as the single interface?  Also, having the metadata cached (lazy loaded) in extension manager as we have it now, it is clean to access it that way.  Anyhow, definitely, a separate PR would be required.

javiexin:
@Nicofuma completed what we talked about: 
- validations array is now static (accessed through self::, maybe we should use static:: instead?)
- `validate_version_check` is also complete, and cleaned the uses of it in `acp_extensions` and `cli/check`, plus the use in `ext_manager` that was already completed

Anything else you think is worth doing here?

About making the validation a new class... I don't think so: it is heavily mixed with the getter methods for metadata elements.  But maybe we could create a `trait` as a design pattern for validation, that just entering the array as static would validate in any of the three available ways: string pattern, function and combined, or even think of more validation methods...

javiexin:
@Nicofuma Anything else left to do here?  Is this going forward?

PR #4815, opened 2017-04-28
[ticket/15187] Add paths of newly enabled extensions to template

kasimi:
When enabling an extension its paths aren't added immediately to
the template which results in ACP template events not being registered in
the compiled template files.

PHPBB3-15187

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15187

### DIFF
--- a/phpBB/includes/acp/acp_extensions.php
+++ b/phpBB/includes/acp/acp_extensions.php
@@ -214,7 +214,7 @@ class acp_extensions
 					}
 
 					// Update custom style for admin area
-					$template->set_custom_style(array(
+					$this->template->set_custom_style(array(
 						array(
 							'name' 		=> 'adm',
 							'ext_path' 	=> 'adm/style/',
--- a/phpBB/includes/acp/acp_extensions.php
+++ b/phpBB/includes/acp/acp_extensions.php
@@ -212,6 +212,15 @@ class acp_extensions
 							meta_refresh(0, $this->u_action . '&amp;action=enable&amp;ext_name=' . urlencode($ext_name) . '&amp;hash=' . generate_link_hash('enable.' . $ext_name));
 						}
 					}
+
+					// Update custom style for admin area
+					$template->set_custom_style(array(
+						array(
+							'name' 		=> 'adm',
+							'ext_path' 	=> 'adm/style/',
+						),
+					), array($phpbb_root_path . 'adm/style'));
+
 					$this->log->add('admin', $this->user->data['user_id'], $this->user->ip, 'LOG_EXT_ENABLE', time(), array($ext_name));
 				}
 				catch (\phpbb\db\migration\exception $e)
--- a/phpBB/phpbb/extension/manager.php
+++ b/phpBB/phpbb/extension/manager.php
@@ -227,8 +227,6 @@ class manager
 		if ($active)
 		{
 			$this->config->increment('assets_version', 1);
-
-			$this->update_template_paths($name);
 		}
 
 		return !$active;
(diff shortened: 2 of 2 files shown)


kasimi:
I don't know how to fix that timeout error. 😕

DavidIQ:
Restarted it and it passed now 😛

javiexin:
Shouldn't this be done in the ACP instead of in the Extension Manager?  The error is in the ACP, so in my view it should be fixed there, and not in a different place. The code you have proposed would fit also there the same way as here.
Also, is this incrementing the paths, or replacing the existing paths?  If the latter, then other extensions might have issues: you may need to repeat the whole process with all is_enabled extensions...  But I am definitely not sure at all!

kasimi:
I completely agree. The extension manager didn't feel like being the correct place for this from the beginning. I'll see how I can move it to the `acp_extensions` class.

Regarding your second point, it does seem to add paths and not remove existing ones. I duplicated the test extension provided in the ticket, and they installed both fine. If the paths were replaced, I would stop seeing the template code from the first extension when enabling the second extension, unless I'm missing something?

javiexin:
Probably, make the call to the new function here: https://github.com/phpbb/phpbb/blob/3.1.x/phpBB/includes/acp/acp_extensions.php#L215
And most likely you will not need to test for the template object existence, nor use the container, as the template object is available in the class already.
The second point was more a question than a statement: I do not know that part of the code that much...  But if it is incremental, why would you need to add the `adm/style` folder again?  I'll check the code to see if I understand better how it works...

phpBB/phpbb/extension/manager.php
```diff
@@ -589,4 +591,39 @@ public function get_finder($use_all_available = false)
...
+	{
+		if (!$this->container->has('template'))
+		{
+			return;
+		}
+
+		$possible_paths = array(
+			$this->phpbb_root_path . 'ext/' . $name . '/adm/style',
+			$this->phpbb_root_path . 'ext/' . $name . '/styles',
+		);
```
javiexin: Maybe use get_extension_path() instead?

kasimi:
Actually the fix turns out to be much simpler. The paths of all installed extensions are already calculated and added by the `set_custom_style()` method, and they aren't incremental after all. My old code added the newly enabled extension's paths twice, which worked fine but was unnecessary.

What I've added now to the place you suggested is the same that's already done here: https://github.com/phpbb/phpbb/blob/release-3.1.10/phpBB/adm/index.php#L56-L62

I've successfully tested it with 3.1.10 and 3.2.0.

javiexin:
Great!

phpBB/includes/acp/acp_extensions.php
```diff
@@ -212,6 +212,15 @@ function main()
 							meta_refresh(0, $this->u_action . '&amp;action=enable&amp;ext_name=' . urlencode($ext_name) . '&amp;hash=' . generate_link_hash('enable.' . $ext_name));
 						}
 					}
+
+					// Update custom style for admin area
+					$template->set_custom_style(array(
```
javiexin: Please use $this->template

javiexin:
@kasimi Could you please take a look at this? https://github.com/BoardTools/upload/issues/41
It strikes me as related to this issue, but has to do with autoload of PHP classes: may be the autoload cache file is built wrongly after disable/enable same as the compiled templates...

kasimi:
Can someone please tickle Travis again?

javiexin:
@kasimi Maybe a simpler way to get this to work properly would be to move the code https://github.com/phpbb/phpbb/blob/release-3.1.10/phpBB/adm/index.php#L56-L62 to here https://github.com/phpbb/phpbb/blob/release-3.1.10/phpBB/adm/index.php#L82.  The paths are not needed before the page is rendered, and by the time the code execution gets here, all changes would have already happened.  It keeps the code cleaner and clearer as well, without the need for code duplication...  What do you think?

kasimi:
If it works for both 3.1 and 3.2 it would certainly be a better solution.

javiexin:
It works "almost" fine.  Perfect when there is no issue, BUT if there is a call to trigger error from within the module being executed, the page is not loaded correctly (no paths set).  So, I think we should leave it as is, that works fine also with trigger error.

PR #4819, opened 2017-05-02
[ticket/15212] Fix code box double horizontal scrollbars

Louis7777:
Checklist:

- [x] Tests pass

[PHPBB3-15212](https://tracker.phpbb.com/browse/PHPBB3-15212)

### DIFF
--- a/phpBB/styles/prosilver/theme/content.css
+++ b/phpBB/styles/prosilver/theme/content.css
@@ -491,18 +491,17 @@ blockquote cite > div {
 
 /* Code block */
 .codebox {
-	padding: 3px;
 	border: 1px solid transparent;
 	font-size: 1em;
 	margin: 1em 0 1.2em 0;
-	overflow-x: scroll;
 	word-wrap: normal;
 }
 
 .codebox p {
 	text-transform: uppercase;
 	border-bottom: 1px solid transparent;
-	margin-bottom: 3px;
+	margin-bottom: 0;
+	padding: 3px 3px 0 3px;
 	font-size: 0.8em !important;
 	font-weight: bold;
 	display: block;
@@ -517,10 +516,9 @@ blockquote .codebox {
 	display: block;
 	height: auto;
 	max-height: 200px;
-	padding-top: 5px;
+	padding: 8px 3px 5px 3px;
 	font: 0.9em Monaco, "Andale Mono","Courier New", Courier, monospace;
 	line-height: 1.3em;
-	margin: 2px 0;
 }
 
 /* Attachments


phpBB/styles/prosilver/theme/content.css
```diff
@@ -491,18 +491,17 @@ blockquote cite > div {
 
 /* Code block */
 .codebox {
-	padding: 3px;
```
VSEphpbb: Do not remove the padding from here. The codebox needs it to look good.

phpBB/styles/prosilver/theme/content.css
```diff
@@ -491,18 +491,17 @@ blockquote cite > div {
...
-	overflow-x: scroll;
 	word-wrap: normal;
 }
 
 .codebox p {
 	text-transform: uppercase;
 	border-bottom: 1px solid transparent;
-	margin-bottom: 3px;
+	margin-bottom: 0;
+	padding: 3px 3px 0 3px;
```
VSEphpbb: All these padding and margin changes above nd below are unnecessary IMO.

Louis7777: The problem is that I have to remove `.codebox {overflow-x: scroll;}` to get rid of the double scrollbar and keep the overflow for the inner `<code>`, because having an overflow property on the parent would make the "Code: Select all" to scroll as well).

Doing just that fixes the problem, but the scrollbars have margins:

scrollbar-margin

Do we want margins?

However, after the padding and margin changes, it looks like this:

scrollbar-nomargin

Louis7777:
!set Styles

phpBB/styles/prosilver/theme/content.css
VSEphpbb: Wow, scroll bars are ugly on PCs. Scroll bars are invisible on a Mac (until needed, at which point they are extremely thin) so the lack of padding on mac becomes less desirable compared to the way codeboxes have always looked in the past.

Louis7777: Indeed they are ugly.  It would be nice if we could have custom scrollbars (jQuery UI?) in 3.3 for the code box, the notifications dropdown, the topic review, the jumpbox etc. and make them look uniform across different platforms.

(P.S.: I haven't really changed the resulting padding - I just applied it elsewhere to get the same effect.)

Crizz0: Please not custom scrollbars and playing around with JS at basic browser functions.

gooof: Great idea, can you add something like this? 
 ``.codebox code { overflow:hidden; }
.codebox code:hover { overflow: auto; }``

Louis7777: @Crizz0 Just my opinion, but I think that a website's internal scrollbars should look the same across platforms (see Facebook for example) and that they should be part of the default theme and depend on the developers' taste just like the rest of the website. Of course, the main browser scrollbars should never be touched!

@gooof Certainly I can, but I don't think that it would be a desirable change. I for one wouldn't like it that way. There was a debate for something similar in 3.1, about having the post buttons (e.g. "Report this post" and "Reply with quote") hidden by default and make them appear on hover using a transition, and well, it was preferred to have them visible.

hanakin:
to be honest there is a lot of improvements we could easily make to code blocks...

`.codebox`
- loose the overflow-x
- only use top padding

`.codebox code`
- loose the margin, line-height and height as they are doing nothing
- add `white-space: pre-wrap` to prevent unwanted indenting
- change padding to be uniform

Create `.codebox pre`
- add `word-wrap: break-word`
- add `word-break: break-word`

Louis7777:
@hanakin 

For the `.codebox` I have removed the overflow-x and I also removed the padding and applied it to the `<p>` and `<code>` child elements.

For the `.codebox code` I have removed the margin and adjusted the padding to look the same as before. The height doesn't do anything and can be removed, but the line-height does, whether it is removed or changed to "1.75" without units. Adding `white-space: pre-wrap` "kills" the horizontal scrollbar, thus creating lots of vertical scrolling, especially on small screens. Do we want it like that?

Regarding the `.codebox pre`, the `word-wrap: break-word` doesn't seem to do anything and `word-break: break-word` is invalid.

VSEphpbb:
@hanakin The whole point of codebox is to present code as is. There should NOT be any word breaking. It should scroll when code is too long.

The current release of phpBB still use the awful implementation of breaking code, which makes it impossible to read code because of all the inappropriate word breaking and wrapping it does.

We changed it recently so it would preserve the code design and readability, and scroll when the code lines exceed the width of the code box.

VSEphpbb:
@Louis7777 As I mentioned, no wrapping. So no `white-space: pre-wrap`

Please stick to the original ticket, which is removing double scrollbar issues.

Louis7777:
@VSEphpbb @hanakin 

It should be ready for review then.

All I have done is to fix the double scrollbar issue and adjust margins and padding to make it look the same as before. The only minor difference - if you compare the two screenshots above - is that the border below "CODE: SELECT ALL" has now a full width.

phpBB/styles/prosilver/theme/content.css
```diff
@@ -517,10 +516,9 @@ blockquote .codebox {
 	display: block;
 	height: auto;
 	max-height: 200px;
-	padding-top: 5px;
+	padding: 8px 3px 5px 3px;
```
VSEphpbb: Why does the top have to be 8px? Can you just make this `5px 3px`

Louis7777: Because I've removed all the padding including 3px top padding from the `.codebox` and added them to `.codebox p`, but I also removed 3px bottom margin from `.codebox p`, so `.codebox code` needs those 3px as top padding.

VSEphpbb: Sure but inside the actual coebox, there appears to be 3px more padding above the code content than at the bottom of the content because of this being 8px.

Louis7777:
```
<div class="codebox">
<p>...</p>
<pre><code></code></pre>
</div>
```

The codebox initially had 3px padding for all sides, but it was unwanted since it created margins for the scrollbars and so I've removed it.

So the `<code>` then needed 3px padding that was lost. However, I removed 2px bottom and top margin from the `<code>` for the sake of using just padding, so I added those 2px as bottom and top padding -> `padding: 5px 3px 5px 3px;`

Plus 3px top padding (I removed 3px bottom margin from `<p>`) -> `padding: 8px 3px 5px 3px;`

hanakin:
@VSEphpbb completely disagree the prevention of code wrap should be up to the code author following proper stds/formating of 80 char length. scrolling horizontally is not a valid alternative to long code IMO, was there ever any discussion on this anywhere?

DavidIQ:
@hanakin why would we want code box contents to wrap, ever? It isn't done anywhere I've ever seen a code box implemented on any site.

VSEphpbb:
@hanakin we dealt with this already. People post code. You can't help if it doesn't fit in the codebox.

Comsider a code box in responsive, where it's really small.

The whole point is to preserve code, not break and wrap it.

All other code/syntax things preserve code as is, no matter how long it is. Even here on GitHub, e.g.:

```php
{
        $notification = $phpbb_notifications->load_notifications('notification.method.board', array(
	        'notification_id'	=> $mark_notification,
        ));
}
```

The point is readability. The owness is not on the poster to somehow make their code fit within the code box. We fixed this to handle the complete mess of trying to read posted code which has been a disaster for the extension writers forum, for example.

VSEphpbb:
@hanakin 

These examples of how code is currently being broken and wrapped is unacceptable, unreadable and franky, amateur-hour. It's why we finally fixed this issue in phpBB recently (for the next releases)

screen shot 2017-05-02 at 4 44 46 pm
screen shot 2017-05-02 at 4 45 49 pm
screen shot 2017-05-02 at 4 45 28 pm
screen shot 2017-05-02 at 4 45 04 pm

hanakin:
I guess I really never see it most of the code i deal with is html/css/js which is usually 80 lines or less so scrolling is not necessary and wrapping is not much of an issue with exception of possibly on mobile

phpBB/styles/prosilver/theme/content.css
hanakin: leave this for top padding

phpBB/styles/prosilver/theme/content.css
hanakin: none of these changes are nessecary

phpBB/styles/prosilver/theme/content.css
hanakin: this can all be uniform 5px

PR #4822, opened 2017-05-02
[ticket/15212] Fix code box double horizontal scrollbars

Louis7777:
[Old PR](https://github.com/phpbb/phpbb/pull/4819)

Checklist:

- [x] Tests pass

[PHPBB3-15212](https://tracker.phpbb.com/browse/PHPBB3-15212)

### DIFF
--- a/phpBB/styles/prosilver/theme/content.css
+++ b/phpBB/styles/prosilver/theme/content.css
@@ -501,7 +501,7 @@ blockquote cite > div {
 	text-transform: uppercase;
 	border-bottom: 1px solid transparent;
 	margin-bottom: 0;
-	padding: 3px 3px 0 3px;
+	padding: 3px;
 	font-size: 0.8em !important;
 	font-weight: bold;
 	display: block;
--- a/phpBB/styles/prosilver/theme/content.css
+++ b/phpBB/styles/prosilver/theme/content.css
@@ -516,7 +516,7 @@ blockquote .codebox {
 	display: block;
 	height: auto;
 	max-height: 200px;
-	padding: 8px 3px 5px 3px;
+	padding: 8px 3px;
 	font: 0.9em Monaco, "Andale Mono","Courier New", Courier, monospace;
 	line-height: 1.3em;
 }
--- a/phpBB/styles/prosilver/theme/content.css
+++ b/phpBB/styles/prosilver/theme/content.css
@@ -491,18 +491,17 @@ blockquote cite > div {
 
 /* Code block */
 .codebox {
-	padding: 3px;
 	border: 1px solid transparent;
 	font-size: 1em;
 	margin: 1em 0 1.2em 0;
-	overflow-x: scroll;
 	word-wrap: normal;
 }
 
 .codebox p {
 	text-transform: uppercase;
 	border-bottom: 1px solid transparent;
-	margin-bottom: 3px;
+	margin-bottom: 0;
+	padding: 3px 3px 0 3px;
 	font-size: 0.8em !important;
 	font-weight: bold;
 	display: block;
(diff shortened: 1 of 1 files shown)


Louis7777:
@VSEphpbb @DavidIQ @hanakin 

I've reopened the PR, because I screwed up.

Louis7777:
!set Styles

VSEphpbb:
why is it back up to 8px???

Louis7777:
It was the right choice after all. With 8px it looks just right. I took 3px away from the bottom margin of `.codebox p` and gave them to the top padding of `.codebox code`.

But I rushed to change it through the GitHub interface and I think my commit message will fail...

VSEphpbb:
@Louis7777 I think it's not the right choice.

As you can see, the top clearly has more padding than the bottom. They should be equal. It's nit-picking I know, but better to be consistent. (see the yellow padding at top is much thicker than bottom)
screen shot 2017-05-02 at 5 41 03 pm

VSEphpbb:
I'd make top/bottm the same. Don't care though if they are both 8px or 5px.

VSEphpbb:
I also think I'd like to see the overflow hidden until mouse-over makes it auto (but only on non-touch devices, i.e., desktops). It'd be nice to have the scroll bar only visible when you need it. Helps make code even more readable.

Louis7777:
@VSEphpbb Okay, but can you also post a screenshot of your vertical scrollbar?

I had to remove the 3px bottom margin from the `<p>`  in order to make the scrollbar reach the top. This is how it was before:

https://cloud.githubusercontent.com/assets/7398634/25627919/f234f176-2f6c-11e7-968b-b0f866db00cb.png

And this is how it is without the margin:

https://cloud.githubusercontent.com/assets/7398634/25627949/0de4f402-2f6d-11e7-94c9-77b98f1e309d.png

Since the text went up along with the scrollbar, I had to give it + 3px and make it 8px.

Do you still want me to make it 5px?

VSEphpbb:
Here (btw I think you can see how odd the extra padding at top looks here ;)

screen shot 2017-05-02 at 5 56 24 pm


All 5px works for me:
screen shot 2017-05-02 at 5 57 23 pm

Louis7777:
@VSEphpbb Okay, I've changed it again, sorry for the trouble (I'm still trying to get the hang of GitHub).

I made it 5px but kept it 3px for the sides since it was originally like that. What do you think?

As for the "overflow hidden until mouse-over makes it auto" I can change that, but I would simply use media queries for small devices. I'm not sure how I would detect whether it is a touch-enabled device.

VSEphpbb:
Stick to just the original ticket for now.

phpBB/styles/prosilver/theme/content.css
```diff
@@ -491,18 +491,17 @@ blockquote cite > div {
...
-	overflow-x: scroll;
 	word-wrap: normal;
 }
 
 .codebox p {
 	text-transform: uppercase;
 	border-bottom: 1px solid transparent;
-	margin-bottom: 3px;
+	margin-bottom: 0;
+	padding: 3px 3px 0 3px;
```
hanakin: this can be 3px all

Louis7777: The "CODE: SELECT ALL" didn't have any bottom padding before. Plus, I've tried it with 3px for all and I preferred the way it is now. If you also try it out and think it is better with a 3px bottom padding, then I'll proceed to change it.

hanakin: it never had any padding before... the padding was mistakenly added to the box and not the line.

Louis7777:
```
<div class="codebox">
<p>
Code: <a href="#" onclick="selectCode(this); return false;">Select all</a>
</p>
<pre><code>...</code></pre>
</div>
```

The codebox initially had 3px padding for all sides, but it was unwanted since it created margins for the scrollbars and so I removed it. Thus, I had to transfer the padding to the child elements which are `.codebox p` and `.codebox code` (as there's no selector for `<pre>`).

Which is why I didn't give it any bottom padding, but I'll proceed to change it now.

VSEphpbb:
Looks good

PR #4831, opened 2017-05-10
[ticket/15223] Safe extension class loader

javiexin:
Extends the existing class auto loader so that, in case of extension files failing to load,
the offending extension is "invalidated", so that it does not interfere with the normal forum work.

PHPBB-15223

Checklist:

- [ ] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-15223

### DIFF
--- a/phpBB/common.php
+++ b/phpBB/common.php
@@ -110,7 +110,7 @@ else
 	set_error_handler(defined('PHPBB_MSG_HANDLER') ? PHPBB_MSG_HANDLER : 'msg_handler');
 }
 
-$phpbb_class_loader_ext = new \phpbb\class_loader('\\', "{$phpbb_root_path}ext/", $phpEx);
+$phpbb_class_loader_ext = new \phpbb\extension\safe_class_loader('\\', 'ext/', $phpbb_root_path, $phpEx);
 $phpbb_class_loader_ext->register();
 
 // Set up container
--- a/phpBB/phpbb/class_loader.php
+++ b/phpBB/phpbb/class_loader.php
@@ -26,10 +26,10 @@ namespace phpbb;
 */
 class class_loader
 {
-	private $namespace;
-	private $path;
-	private $php_ext;
-	private $cache;
+	protected $namespace;
+	protected $path;
+	protected $php_ext;
+	protected $cache;
 
 	/**
 	* A map of looked up class names to paths relative to $this->path.
--- a/phpBB/phpbb/extension/safe_class_loader.php
+++ b/phpBB/phpbb/extension/safe_class_loader.php
@@ -1,0 +1,163 @@
+<?php
+/**
+*
+* This file is part of the phpBB Forum Software package.
+*
+* @copyright (c) phpBB Limited <https://www.phpbb.com>
+* @license GNU General Public License, version 2 (GPL-2.0)
+*
+* For full copyright and license information, please see
+* the docs/CREDITS.txt file.
+*
+*/
+
+namespace phpbb\extension;
+
+/**
+* The safe extension class loader resolves class names to file system paths
+* and loads them if necessary.
+* In case any file system path is not available, the corresponding extension
+* is disabled, and cache is purged, avoiding further errors due to the same
+* issue.
+*/
+
+class safe_class_loader extends \phpbb\class_loader
+{
+	/** @var string */
+	protected $phpbb_root_path;
+
+	/** @var \phpbb\db\driver\factory */
+	protected $db = null;
+
+	/** @var boolean */
+	protected $is_in_admin;
+
+	/** @var string */
+	protected $extension_table = '';
+
+	/**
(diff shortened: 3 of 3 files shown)


javiexin:
!set WIP

Nicofuma:
A few things: First, do not extend the class_loader, prefer to decorate it. Then I think I would just disable the extension if the container is available and then redirect to a special error page explaining what happenned (and this page will be responsible of the logging). And if the container is not available I won't do anything

javiexin:
Any reason why decoration is preferred here?  Not against it, just trying to understand.  In my view, decorating here would make everything slower, as all method accesses would be transformed into `call_user_func_array` calls (so double calling for each method).
Relative to your other comment, if the container is available, everything is much easier, true.  But then, if nothing is done when it is not present, most/some of the benefit will go away.  Simple way to test it: in a working environment, rename a listener from an enabled extension, and try to fix the problem: it is not so obvious (for the regular admin) as they would have to purge the cache and then figure out a stack dump from the container build phase, and from there, try to identify the right extension.  For me this is too error prone.  Granted this is not a problem for a lot of admins, but phpbb admins are, in the most part, or at least a very relevant fraction, really no experts at all.  There are other cases where the forum will break before the container is created and available, like creating all compatibility globals.  So I'd rather do something in this context, specially given that we have all the information already available.

javiexin:
I have created a Gist with the equivalent to the above, replacing inheritance by a pure service decorator pattern. See https://gist.github.com/javiexin/a40f54ac4f306d572bdcd50ad4e708f3
I have not uploaded the modification to `common.php` to use the decorator.

Nicofuma:
You are decorating an autoloader, so you don't need to decorate anything else than the autoloader entrypoint.

About your point, actually we don't really care because that's not a use case we want to handle: we need to fix issues happening because of errored/incompatible extensions, not issues happening because the admin is messing with the files content.

javiexin:
So, to decorate the autoloader I need to decorate the entry point: `load_class`. Then I need to register and unregister the decorator as an autoloader, so I need `register` and `unregister` as well. And of course, I need `resolve_path`, that is the one I want modified. Ok, I could reuse `set_cache`; wait, no, as I need the cache in the decorator... We have a decorator that implements all and decorates nothing...

And I think that a common source of errors, malfunctions and support requests should be something that we care about... More than errored extensions, as they should be caught during validation.

Nicofuma:
You do not have to decorates anything. What you want to do is an autoloader with some error recovery stuff. For this purpose what you should do is implement an autoloader which uses another object to find and includes the classes but have a special behaviour in case of failure (if after calling the decorated autoloader the class is not defined, then it means there was an error of some kind)

You could even work without any decoration: just register an autoloader function at the end of the autoloader stack. In this case, if your autoloader is called, then it means that the class has not been found by anyone.

> And I think that a common source of errors, malfunctions and support requests should be something that we care about... More than errored extensions, as they should be caught during validation.
That's exactly what I am saying:
`Simple way to test it: in a working environment, rename a listener from an enabled extension, and try to fix the problem:` That's not something an admin should do and it's not a practice we will encourage. If the admin is messing with the files contents it's on his own.

javiexin:
Ok, a new autoloader makes more sense to me as well than decorating.  Although the overhead would be bigger than with the proposed approach.  However, I doubt it will work in most cases: if the file is found by the normal autoloader or is cached from a prior run, it will try to load it via require, and that will throw a non-catchable exception, and not continue with the autoloader chain, as far as I know.  So we will not be able to see it fail.  Only in the case of a new, non-cached, missing file, this approach would work.

> `Simple way to test it: in a working environment, rename a listener from an enabled extension, and try to fix the problem`
> That's not something an admin should do and it's not a practice we will encourage. If the admin is messing with the files contents it's on his own.

Right, I understand, but the situation is the same if the admin is trying to upgrade, and removes the old files at the wrong moment (before disabling extension, for example).  I am not talking about an admin messing around, I am talking about an unexperienced admin that may not do things quite right due to lack of expertise.  The example was just that, an example, so that anyone could see how to easily reproduce one of the issues that would not be caught if we limited the scope of this to the container available time.

javiexin:
Ok, coming back to this again.
The best way I have found to record an "invalid" extension would be in `ext` table itself, with a new field that could store the (serialized) reason for extension not being valid (to be displayed in extension list).  Using the `ext_state` column would be an option, but it might create BC issues with some extensions (that use it for partial migration or `ext` class ongoing work, ie, long running cases).
Another thing that, in my opinion, would be beneficial, and this could be an opportunity to do so, would be to record in the DB the extension VERSION that has been configured/enabled.  That way, you can check the available vs configured version, and eventually run migrations if needed.
Opinions?

CHItA:
I don't see this being useful for most users, as this situation just shouldn't happen at all if one uses the software as intended. If someone removes files from an enabled extension the software supposed to fail.

PR #4834, opened 2017-05-13
[ticket/15056] Update subforum icons with Mark forums read

senky:
PHPBB3-15056

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15056

### DIFF
--- a/phpBB/styles/prosilver/template/ajax.js
+++ b/phpBB/styles/prosilver/template/ajax.js
@@ -26,7 +26,7 @@ phpbb.addAjaxCallback('mark_forums_read', function(res) {
 	});
 
 	// Mark subforums read
-	$('a.subforum[class*="unread"]').removeClass('unread').addClass('read');
+	$('a.subforum[class*="unread"]').removeClass('unread').addClass('read').children('.icon.icon-red').removeClass('icon-red').addClass('icon-blue');
 
 	// Mark topics read if we are watching a category and showing active topics
 	if ($('#active_topics').length) {


hanakin:
This is not the right way to handle this we need to keep things consistent. The problem seems to be with the template bars not the classes... I will look it over tomorrow when I get time.

senky:
@hanakin I try to follow the single-source-of-truth idea whenever possible. What I have discovered here is, that both `a` and it's `i` maintain read status with it's class. That should be avoided. So I removed child's class to keep things consistent.

Another way is to add code to `mark_forums_read` ajax callback. But as I said, I don't like this idea of splitting source of truth.

hanakin:
As I said I will look at it more when I am home I currently in travel mode and do not have the ability to test it, but it seems like there needs to be a cleaner way to handle this then chanining classes which should always be avoided

hanakin:
ok so after looking at this more I do not see what the problem is? why are you thinking this should be avoided? either way if you only wish to use it once then remove it from the link and leave it on the icon as that is where it actually belongs

senky:
The problem is this: when you click Mark forums read, subforum icons remain red, because javascript callback changes `unread` class in `a`, but icon color is determined by `icon-red` class. I think, that class in `a` is unused, so maybe what you suggest is to change javascript part instead of CSS?

hanakin:
yes that was my meaning the class of unread is not used anymore for these icons however it is used still for the imageset...

hanakin:
yeah i like that better for 3.2, in 3.3 we can look at fixing the js in prosilver during the re-write of all the JS for the new theme.

senky:
> also can you pls add the Styles label so that its easier to keep track of changes.

I would love to, but I don't have permission :)

hanakin:
@phpbb/development-team 3.2.1 ?

hanakin:
@senky https://area51.phpbb.com/phpBB/viewtopic.php?f=99&t=46079

senky:
@hanakin I remembered `!set WIP`. Didn't know I can set labels this way, too. Good to know. 🙂

PR #4859, opened 2017-06-25
[ticket/15253] Add storage abstraction

rubencm:
Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15253


senky:
It may be interesting to integrate [Flysystem](https://flysystem.thephpleague.com/) instead of writing custom filesystem abstraction. We used something similar with OAuth and the advantage is, you end up with few filesystems integrated already (AWS, Azure, Dropbox, etc.)

rubencm:
@senky yes, i will integrate flysystem, but probably it will go on a external extension

Nicofuma:
Having our own abstraction gives us more possibilities and flexibility. But there is no reason we can't implement adapters using Flysystem or Gaufrette. And as @rubencm said, it is a part of his GSoC plan, but it is not the priority.

phpBB/config/default/container/services.yml
```diff
@@ -25,6 +25,7 @@ imports:
     - { resource: services_profilefield.yml }
     - { resource: services_report.yml }
     - { resource: services_routing.yml }
+    - { resource: services_storage.yml }
```
Nicofuma: please removes the services, we will commit this part later as it is not really designed for now

phpBB/phpbb/storage/adapter/local.php
```diff
@@ -0,0 +1,151 @@
...
+	 */
+	protected $filesystem;
+
+	/** @var string path */
+	protected $root_path;
+
+	/**
+	 * Constructor
+	 */
+	public function __construct(\phpbb\config\config $config, \phpbb\filesystem\filesystem $filesystem, $phpbb_root_path, $path_key)
```
Nicofuma: please imports the namespaces

phpBB/phpbb/storage/adapter/local.php
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function put_contents($path, $content)
+	{
+		if ($this->exists($path))
+		{
+			throw new exception('', $path); // FILE_EXISTS
```
Nicofuma: I think FILE_EXISTS should be the exception message. But you should use a more precise key: `STORAGE_EXCEPTION_FILE_EXISTS` per example

phpBB/phpbb/storage/adapter/local.php
```diff
@@ -0,0 +1,151 @@
...
+		}
+
+		try
+		{
+			$this->filesystem->dump_file($this->root_path.$path, $content);
+		}
+		catch (filesystem_exception $e)
+		{
+			throw new exception('', $path, array(), $e); // CANNOT_DUMP_FILE
```
Nicofuma: same here

phpBB/phpbb/storage/adapter/local.php
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function get_contents($path)
+	{
+		if (!$this->exists($path))
+		{
+			throw new exception('', $path); // FILE_DONT_EXIST
```
Nicofuma: and here, and everywhere else

phpBB/phpbb/storage/adapter/local.php
```diff
@@ -0,0 +1,151 @@
...
+		}
+
+		if (($content = @file_get_contents($this->root_path.$path)) === false)
```
Nicofuma: we don't really uses this kind of notation here, usually we prefer to split it into 2 lines

phpBB/phpbb/storage/adapter/local.php
```diff
@@ -0,0 +1,151 @@
...
+
+	/**
+	 * Constructor
+	 */
+	public function __construct(\phpbb\config\config $config, \phpbb\filesystem\filesystem $filesystem, $phpbb_root_path, $path_key)
+	{
+		$this->filesystem = $filesystem;
+		$this->root_path = $phpbb_root_path.$config[$path_key];
+
+		if (substr($this->root_path, -1, 1) != DIRECTORY_SEPARATOR)
```
Nicofuma: !==

phpBB/phpbb/storage/adapter/local.php
```diff
@@ -0,0 +1,151 @@
...
+
+		return $content;
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function exists($path)
+	{
+		return $this->filesystem->exists($this->root_path.$path);
```
Nicofuma: $this->root_path . $path

phpBB/phpbb/storage/adapter/local.php
```diff
@@ -0,0 +1,151 @@
...
+ * the docs/CREDITS.txt file.
+ *
+ */
+
+namespace phpbb\storage\adapter;
+
+use phpbb\storage\exception\exception;
+use phpbb\filesystem\exception\filesystem_exception;
+
+class local implements adapter_interface
```
Nicofuma: adds a comment, I think it may be a good idea:
```
/**
 * @internal Experimental
 */
```

phpBB/phpbb/storage/exception/exception.php
```diff
@@ -0,0 +1,42 @@
...
+	 *
+	 * @param string		$message	The Exception message to throw (must be a language variable).
+	 * @param string		$filename	The file that caused the error.
+	 * @param array			$parameters	The parameters to use with the language var.
+	 * @param \Exception	$previous	The previous runtime_exception used for the runtime_exception chaining.
+	 * @param integer		$code		The Exception code.
+	 */
+	public function __construct($message = "", $filename = '', $parameters = array(), \Exception $previous = null, $code = 0)
+	{
+		parent::__construct($message, array_merge(array('filename' => $filename), $parameters), $previous, $code);
```
Nicofuma: Just in case you don't know: you can use the short syntax for the array if you wish

phpBB/phpbb/storage/exception/exception.php
```diff
@@ -0,0 +1,42 @@
...
+	/**
+	 * Constructor
+	 *
+	 * @param string		$message	The Exception message to throw (must be a language variable).
+	 * @param string		$filename	The file that caused the error.
+	 * @param array			$parameters	The parameters to use with the language var.
+	 * @param \Exception	$previous	The previous runtime_exception used for the runtime_exception chaining.
+	 * @param integer		$code		The Exception code.
+	 */
+	public function __construct($message = "", $filename = '', $parameters = array(), \Exception $previous = null, $code = 0)
```
Nicofuma: single quotes: `$message = ''`

phpBB/phpbb/storage/exception/exception.php
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	/**
+	 * Returns the filename that triggered the error
+	 *
+	 * @return string
+	 */
+	public function get_filename()
+	{
+		$parameters = parent::get_parameters();
```
Nicofuma: $this->get_parameters()

phpBB/phpbb/storage/exception/exception.php
```diff
@@ -0,0 +1,42 @@
...
+ * @license GNU General Public License, version 2 (GPL-2.0)
+ *
+ * For full copyright and license information, please see
+ * the docs/CREDITS.txt file.
+ *
+ */
+
+namespace phpbb\storage\exception;
+
+class exception extends \phpbb\exception\runtime_exception
```
Nicofuma: imports the namespace

phpBB/phpbb/storage/exception/not_implemented.php
```diff
@@ -0,0 +1,42 @@
...
+ * @license GNU General Public License, version 2 (GPL-2.0)
+ *
+ * For full copyright and license information, please see
+ * the docs/CREDITS.txt file.
+ *
+ */
+
+namespace phpbb\storage\exception;
+
+class not_implemented extends \phpbb\exception\runtime_exception
```
Nicofuma: couldn't it extends `phpbb\storage\exception` to avoid code duplication ? (and semantically it makes sense too)

phpBB/phpbb/storage/helper.php
```diff
@@ -0,0 +1,365 @@
...
+
+	/**
+	 * Try to resolve real path when PHP's realpath failes to do so
+	 *
+	 * @param string	$path
+	 * @return bool|string
+	 */
+	protected static function phpbb_own_realpath($path)
+	{
+
```
Nicofuma: extra line

phpBB/phpbb/storage/helper.php
```diff
@@ -0,0 +1,365 @@
...
+	 * Given an existing path, convert it to a path relative to a given starting path
+	 *
+	 * @param string $end_path		Absolute path of target
+	 * @param string $start_path	Absolute path where traversal begins
+	 *
+	 * @return string Path of target relative to starting path
+	 */
+	public static function make_path_relative($end_path, $start_path)
+	{
+		$symfony_filesystem = new \Symfony\Component\Filesystem\Filesystem();
```
Nicofuma: imports the namespace

tests/storage/adapter/local_test.php
```diff
@@ -0,0 +1,94 @@
...
+	{
+		file_put_contents('file.txt', 'abc');
+		$this->assertEquals($this->adapter->get_contents('file.txt'), 'abc');
+		unlink('file.txt');
+	}
+
+	public function test_exists()
+	{
+		// Exists with files
+		$this->assertTrue($this->adapter->exists('README.md'));
```
Nicofuma: please uses a dataProvider:

```php
public function data_test_exists()
{
    yield ['README.md', true];
    yield ['nonexistent_file.php', false];
    yield ['phpBB', true];
    yield ['nonexistent_folder', false];
}

/**
 * @dataProvider data_test_exists
 */
public function test_exists($path, $expected)
{
    $this->assertSame($expected, $this->adapter->exists($path));
}
```

tests/storage/adapter/local_test.php
```diff
@@ -0,0 +1,94 @@
...
+		$this->assertFalse($this->adapter->exists('nonexistent_file.php'));
+		// exists with directory
+		$this->assertTrue($this->adapter->exists('phpBB'));
+		$this->assertFalse($this->adapter->exists('nonexistet_folder'));
+	}
+
+	public function test_delete()
```
Nicofuma: please split it into 3 tests: test_delete_file and test_delete_directory

phpBB/phpbb/storage/adapter/local.php
Nicofuma: And adds it to the storage class to

phpBB/phpbb/storage/adapter/local.php
rubencm: Yes, thats why i have left exception messages empty for now

phpBB/phpbb/storage/exception/exception.php
rubencm: Exceptions are a copy/paste of https://github.com/phpbb/phpbb/blob/master/phpBB/phpbb/filesystem/exception/filesystem_exception.php
so there are wrong too

phpBB/phpbb/storage/adapter/local.php
rubencm: The storage class is not an adapter, it has the same methods, but is not exactly the same

Nicofuma: yes I know

phpBB/phpbb/filesystem/filesystem.php
```diff
@@ -162,26 +162,7 @@ public function chown($files, $user, $recursive = false)
 	 */
```
Nicofuma: please add `@deprecated 3.3.0-a1 (To be removed: 4.0.0)`

phpBB/phpbb/filesystem/filesystem.php
```diff
@@ -227,7 +208,7 @@ public function exists($files)
...
 	public function is_absolute_path($path)
```
Nicofuma: please add `@deprecated 3.3.0-a1 (To be removed: 4.0.0)`

phpBB/phpbb/filesystem/filesystem.php
```diff
@@ -305,7 +286,7 @@ public function is_writable($files, $recursive = false)
 	 */
 	public function make_path_relative($end_path, $start_path)
```
Nicofuma: please add `@deprecated 3.3.0-a1 (To be removed: 4.0.0)`

phpBB/phpbb/filesystem/filesystem.php
```diff
@@ -639,6 +620,8 @@ protected function phpbb_is_writable($file)
 	/**
 	 * Try to resolve real path when PHP's realpath failes to do so
 	 *
+	 * @deprecated 3.3.0-a1 (To be removed: 4.0.0)
```
Nicofuma: replace the core of the method by a call to your new helper

phpBB/phpbb/finder.php
```diff
@@ -244,7 +241,7 @@ public function core_directory($core_directory)
 	*/
 	protected function sanitise_directory($directory)
 	{
-		$directory = $this->filesystem->clean_path($directory);
+		$directory = \phpbb\storage\helper::clean_path($directory);
```
Nicofuma: import the namespace

tests/filesystem/clean_path_test.php
```diff
@@ -1,54 +0,0 @@
-<?php
```
Nicofuma: I think you should keep both the old and new tests (even if they look duplicated)

tests/filesystem/is_absolute_test.php
```diff
@@ -1,68 +0,0 @@
-<?php
```
Nicofuma: I think you should keep both the old and new tests (even if they look duplicated)

tests/filesystem/realpath_test.php
```diff
@@ -1,90 +0,0 @@
-<?php
```
Nicofuma: I think you should keep both the old and new tests (even if they look duplicated)

phpBB/phpbb/filesystem/filesystem.php
rubencm: Should i move it to the helper class too?
Also i think i will have to move the helper class to filesystem , it makes no sense in storage

phpBB/phpbb/filesystem/filesystem.php
rubencm: It is already on the interface

phpBB/phpbb/filesystem/filesystem.php
rubencm: It is already on the interface

phpBB/phpbb/filesystem/filesystem.php
rubencm: It is already on the interface

Nicofuma: indeed

phpBB/phpbb/filesystem/helper.php
```diff
@@ -0,0 +1,366 @@
...
+	 * Given an existing path, convert it to a path relative to a given starting path
+	 *
+	 * @param string $end_path		Absolute path of target
+	 * @param string $start_path	Absolute path where traversal begins
+	 *
+	 * @return string Path of target relative to starting path
+	 */
+	public static function make_path_relative($end_path, $start_path)
+	{
+		$symfony_filesystem = new symfony_filesystem();
```
CHItA: it could be static

phpBB/phpbb/filesystem/filesystem.php
```diff
@@ -742,4 +742,19 @@ protected function resolve_path($path, $prefix = '', $absolute = false, $return_
...
 	}
+
+	/**
+	 * Get an instance of symfony's filesystem object.
+	 *
+	 * @return \Symfony\Component\Filesystem\Filesystem	Symfony filesystem
+	 */
+	protected static function get_symfony_filesystem()
+	{
+		if(self::$symfony_filesystem === null)
```
Nicofuma: CS :)

phpBB/phpbb/filesystem/helper.php
```diff
@@ -207,8 +213,7 @@ public static function realpath($path)
 	 */
 	public static function make_path_relative($end_path, $start_path)
 	{
-		$symfony_filesystem = new symfony_filesystem();
-		return $symfony_filesystem->makePathRelative($end_path, $start_path);
+		self::get_symfony_filesystem()->makePathRelative($end_path, $start_path);
```
Nicofuma: you forgot the return

phpBB/includes/functions_compatibility.php
```diff
@@ -96,36 +96,12 @@ function phpbb_check_hash($password, $hash)
...
 {
-	global $phpbb_path_helper, $phpbb_container;
-
-	if (!$phpbb_path_helper && $phpbb_container)
-	{
-		/* @var $phpbb_path_helper \phpbb\path_helper */
-		$phpbb_path_helper = $phpbb_container->get('path_helper');
-	}
-	else if (!$phpbb_path_helper)
+	if (!class_exists('\phpbb\filesystem\helper'))
```
Nicofuma: you can remove this check, rely on the autoloader

phpBB/includes/functions_compatibility.php
```diff
@@ -463,25 +439,31 @@ function phpbb_is_writable($file)
  * @param string $path Path to check absoluteness of
  * @return boolean
  *
- * @deprecated 3.2.0-dev	use \phpbb\filesystem\filesystem::is_absolute_path() instead
+ * @deprecated 3.2.0-dev	use \phpbb\filesystem\helper::is_absolute_path() instead
  */
 function phpbb_is_absolute($path)
 {
-	global $phpbb_filesystem;
+	if (!class_exists('\phpbb\filesystem\helper'))
```
Nicofuma: again, rely on the autoloader

phpBB/includes/functions_compatibility.php
```diff
@@ -463,25 +439,31 @@ function phpbb_is_writable($file)
...
 /**
  * A wrapper for realpath
  *
- * @deprecated 3.2.0-dev	use \phpbb\filesystem\filesystem::realpath() instead
+ * @deprecated 3.2.0-dev	use \phpbb\filesystem\helper::realpath() instead
  */
 function phpbb_realpath($path)
 {
-	global $phpbb_filesystem;
+	if (!class_exists('\phpbb\filesystem\helper'))
```
Nicofuma: again

phpBB/phpbb/filesystem/helper.php
```diff
@@ -0,0 +1,386 @@
...
+ *
+ */
+
+namespace phpbb\filesystem;
+
+use Symfony\Component\Filesystem\Filesystem as symfony_filesystem;
+
+class helper
+{
+
```
Nicofuma: extra line

phpBB/phpbb/filesystem/helper.php
```diff
@@ -0,0 +1,386 @@
...
+	 * 								or false if path cannot be resolved
+	 */
+	protected static function resolve_path($path, $prefix = '', $absolute = false, $return_array = false)
+	{
+		if ($return_array)
+		{
+			$path = str_replace(DIRECTORY_SEPARATOR, '/', $path);
+		}
+
+		trim ($path, '/');
```
Nicofuma: extra space between trim and (

phpBB/phpbb/filesystem/helper.php
```diff
@@ -0,0 +1,386 @@
...
+				if ($absolute)
+				{
+					array_unshift($tmp_array, $prefix);
+				}
+
+				$resolved_path = implode('/', $tmp_array);
+			}
+		}
+
+		return ($return_array) ? $resolved : $resolved_path;
```
Nicofuma: you can remove the parenthesis

tests/filesystem/clean_path_test.php
Nicofuma: do not forget to restore this file

tests/filesystem/is_absolute_test.php
Nicofuma: same here

tests/filesystem/realpath_test.php
Nicofuma: and here

tests/filesystem/clean_path_test.php
rubencm: https://github.com/rubencm/phpbb/blob/ticket/15253/tests/filesystem/helper_clean_path_test.php

Or should i restart the originals too?

Nicofuma: I think we should have both. The new one to test the feature and the old one to guarantee we don't break anything. But it does not really matter here tbh

phpBB/phpbb/filesystem/helper.php
```diff
@@ -0,0 +1,385 @@
...
+		$exploded = explode('/', $path);
+		$filtered = array();
+		foreach ($exploded as $part)
+		{
+			if ($part === '.' && !empty($filtered))
+			{
+				continue;
+			}
+
+			if ($part === '..' && !empty($filtered) && $filtered[sizeof($filtered) - 1] !== '.' && $filtered[sizeof($filtered) - 1] !== '..')
```
rxu: Please use ```count()``` in favour of ```sizeof()``` as the latter is going to be deprecated in PHP 7.2.

CHItA:
I'm not sure if the `filesystem\helper`  is good idea. I haven't thought too much about it, so the only scenario of the top of my head would be the following (I also have not seen the design, so that's that):
Suppose that you have a piece of code which would modify e.g. the permissions of a file (or any other action that is in that class) which is not stored locally. That action doesn't make much sense, but the file could be stored locally, in which case that action would be necessary. Now you have to check in that logic whether the file is stored locally or elsewhere, thus making the abstraction a bit weak IMHO.

Nicofuma:
The helper is for the parts of phpBB that always deal with local files (cache per example)

phpBB/phpbb/filesystem/filesystem.php
Nicofuma: actually I think we can remove this function directly

Nicofuma:
You have a few coding style issues

tests/template/template_test_case.php
```diff
@@ -86,8 +86,12 @@ protected function setup_engine(array $new_config = array())
 			new \phpbb\symfony_request(
 				new phpbb_mock_request()
 			),
+<<<<<<< HEAD
```
Nicofuma: should be removed

PR #4863, opened 2017-06-28
[ticket/15257] Provide extension not enableable messages

javiexin:
Extensions may define the reasons why they cannot be enabled,
so that users are better informed and ready to fix this.
Multiple messages are possible at the same time.
Keeps 100% BC, even with the current "workarounds".

PHPBB3-15257

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-15257

### DIFF
--- a/phpBB/includes/acp/acp_extensions.php
+++ b/phpBB/includes/acp/acp_extensions.php
@@ -167,7 +167,7 @@ class acp_extensions
 				$extension = $this->ext_manager->get_extension($ext_name);
 				if (!$extension->is_enableable())
 				{
-					$message = $extension->get_not_enableable_msg() ?: $this->user->lang['EXTENSION_NOT_ENABLEABLE'];
+					$message = $extension->not_enableable_msg() ?: $this->user->lang['EXTENSION_NOT_ENABLEABLE'];
 					$message = is_array($message) ? implode('<br/>', $message) : $message;
 					trigger_error($message . adm_back_link($this->u_action), E_USER_WARNING);
 				}
@@ -199,7 +199,7 @@ class acp_extensions
 				$extension = $this->ext_manager->get_extension($ext_name);
 				if (!$extension->is_enableable())
 				{
-					$message = $extension->get_not_enableable_msg() ?: $this->user->lang['EXTENSION_NOT_ENABLEABLE'];
+					$message = $extension->not_enableable_msg() ?: $this->user->lang['EXTENSION_NOT_ENABLEABLE'];
 					$message = is_array($message) ? implode('<br/>', $message) : $message;
 					trigger_error($message . adm_back_link($this->u_action), E_USER_WARNING);
 				}
--- a/phpBB/includes/acp/acp_extensions.php
+++ b/phpBB/includes/acp/acp_extensions.php
@@ -129,7 +129,7 @@ class acp_extensions
 				else
 				{
 					$this->config->set('extension_force_unstable', false);
-					trigger_error($this->user->lang['CONFIG_UPDATED'] . adm_back_link($this->u_action));
+					trigger_error($this->user->lang('CONFIG_UPDATED') . adm_back_link($this->u_action));
 				}
 				break;
 
@@ -138,7 +138,7 @@ class acp_extensions
 				if (confirm_box(true))
 				{
 					$this->config->set('extension_force_unstable', true);
-					trigger_error($this->user->lang['CONFIG_UPDATED'] . adm_back_link($this->u_action));
+					trigger_error($this->user->lang('CONFIG_UPDATED') . adm_back_link($this->u_action));
 				}
 
 				$this->list_enabled_exts();
@@ -167,7 +167,7 @@ class acp_extensions
 				$extension = $this->ext_manager->get_extension($ext_name);
 				if (!$extension->is_enableable())
 				{
-					$message = $extension->not_enableable_reason() ?: $this->user->lang['EXTENSION_NOT_ENABLEABLE'];
+					$message = $extension->not_enableable_reason() ?: $this->user->lang('EXTENSION_NOT_ENABLEABLE');
 					$message = is_array($message) ? implode('<br/>', $message) : $message;
 					trigger_error($message . adm_back_link($this->u_action), E_USER_WARNING);
 				}
@@ -199,7 +199,7 @@ class acp_extensions
 				$extension = $this->ext_manager->get_extension($ext_name);
 				if (!$extension->is_enableable())
 				{
-					$message = $extension->not_enableable_reason() ?: $this->user->lang['EXTENSION_NOT_ENABLEABLE'];
+					$message = $extension->not_enableable_reason() ?: $this->user->lang('EXTENSION_NOT_ENABLEABLE');
 					$message = is_array($message) ? implode('<br/>', $message) : $message;
 					trigger_error($message . adm_back_link($this->u_action), E_USER_WARNING);
 				}
--- a/phpBB/includes/acp/acp_extensions.php
+++ b/phpBB/includes/acp/acp_extensions.php
@@ -167,7 +167,7 @@ class acp_extensions
 				$extension = $this->ext_manager->get_extension($ext_name);
 				if (!$extension->is_enableable())
 				{
-					$message = $extension->not_enableable_msg() ?: $this->user->lang['EXTENSION_NOT_ENABLEABLE'];
+					$message = $extension->not_enableable_reason() ?: $this->user->lang['EXTENSION_NOT_ENABLEABLE'];
 					$message = is_array($message) ? implode('<br/>', $message) : $message;
 					trigger_error($message . adm_back_link($this->u_action), E_USER_WARNING);
 				}
@@ -199,7 +199,7 @@ class acp_extensions
 				$extension = $this->ext_manager->get_extension($ext_name);
 				if (!$extension->is_enableable())
 				{
-					$message = $extension->not_enableable_msg() ?: $this->user->lang['EXTENSION_NOT_ENABLEABLE'];
+					$message = $extension->not_enableable_reason() ?: $this->user->lang['EXTENSION_NOT_ENABLEABLE'];
 					$message = is_array($message) ? implode('<br/>', $message) : $message;
 					trigger_error($message . adm_back_link($this->u_action), E_USER_WARNING);
 				}
--- a/phpBB/includes/acp/acp_extensions.php
+++ b/phpBB/includes/acp/acp_extensions.php
@@ -167,7 +167,9 @@ class acp_extensions
 				$extension = $this->ext_manager->get_extension($ext_name);
 				if (!$extension->is_enableable())
 				{
-					trigger_error($this->user->lang['EXTENSION_NOT_ENABLEABLE'] . adm_back_link($this->u_action), E_USER_WARNING);
+					$message = $extension->get_not_enableable_msg() ?: $this->user->lang['EXTENSION_NOT_ENABLEABLE'];
+					$message = is_array($message) ? implode('<br/>', $message) : $message;
+					trigger_error($message . adm_back_link($this->u_action), E_USER_WARNING);
 				}
 
 				if ($this->ext_manager->is_enabled($ext_name))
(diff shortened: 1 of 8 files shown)


phpBB/phpbb/extension/extension_interface.php
```diff
@@ -27,6 +27,13 @@
 	public function is_enableable();
 
 	/**
+	* Indicate why the extension can not be enabled.
+	*
+	* @return string|array with actual processed language string(s) (not language keys), or empty
+	*/
+	public function get_not_enableable_msg();
```
javiexin: I agree with you: the name is terribly ugly... but yours is not much better 😄.  I prefer `not_enableable_msg` if you don't mind, and yes, I will change it.

Regarding the CLI, I was under the impression that this was not considered relevant there.  In fact, the `is_enableable` method is not called (directly) at all in the CLI, it calls `enable` directly (also without going step by step).  If you think this is benefitial, I am happy to add support for it in the CLI, no problem.

javiexin: Done.  Please, review specially the CLI bit, as I am not so experienced there.

rxu: Honestly, I'd go with just ```get_message()``` or similar simple name.

VSEphpbb:
You should add some unit tests for this

phpBB/phpbb/extension/extension_interface.php
3D-I: What about _not_enableable_reason_ ¿

javiexin: I agree, name changed again

javiexin:
@VSEphpbb please, take a look at the added tests, and tell me if ok, or some more are needed (and what do you recommend testing on top of this).

tests/functional/extension_acp_test.php
```diff
@@ -168,6 +168,10 @@ public function test_enable_pre()
 		// Correctly submit the enable form
 		$crawler = self::request('GET', 'adm/index.php?i=acp_extensions&mode=main&action=enable_pre&ext_name=vendor3%2Ffoo&sid=' . $this->sid);
 		$this->assertContainsLang('EXTENSION_NOT_ENABLEABLE', $crawler->filter('.errorbox')->text());
+
+		// Custom error messages returned by not enableable extension
+		$crawler = self::request('GET', 'adm/index.php?i=acp_extensions&mode=main&action=enable_pre&ext_name=vendor4%2Ffoo&sid=' . $this->sid);
+		$this->assertContains('Reason 2', $crawler->filter('.errorbox')->text());
```
VSEphpbb: should also check for `Reason 1`

VSEphpbb:
Tests seem fine to me now.

phpBB/includes/acp/acp_extensions.php
```diff
@@ -197,7 +199,9 @@ function main()
 				$extension = $this->ext_manager->get_extension($ext_name);
 				if (!$extension->is_enableable())
 				{
-					trigger_error($this->user->lang['EXTENSION_NOT_ENABLEABLE'] . adm_back_link($this->u_action), E_USER_WARNING);
+					$message = $extension->not_enableable_reason() ?: $this->user->lang['EXTENSION_NOT_ENABLEABLE'];
+					$message = is_array($message) ? implode('<br/>', $message) : $message;
```
Nicofuma: $message cannot be an array here

phpBB/includes/acp/acp_extensions.php
```diff
@@ -197,7 +199,9 @@ function main()
 				$extension = $this->ext_manager->get_extension($ext_name);
 				if (!$extension->is_enableable())
 				{
-					trigger_error($this->user->lang['EXTENSION_NOT_ENABLEABLE'] . adm_back_link($this->u_action), E_USER_WARNING);
+					$message = $extension->not_enableable_reason() ?: $this->user->lang['EXTENSION_NOT_ENABLEABLE'];
```
Nicofuma: $this->user->lang()

phpBB/includes/acp/acp_extensions.php
Nicofuma: you should call $this->user->lang() for every entry in the array

phpBB/phpbb/extension/extension_interface.php
```diff
@@ -27,6 +27,13 @@
 	public function is_enableable();
 
 	/**
+	* Indicate why the extension can not be enabled.
+	*
+	* @return string|array with actual processed language string(s) (not language keys), or empty
+	*/
+	public function not_enableable_reason();
```
Nicofuma: I think it's better to change the return of is_enableable() to void|string|array

phpBB/includes/acp/acp_extensions.php
javiexin: I just kept the original way it was.  Will change it, no problem.

phpBB/includes/acp/acp_extensions.php
javiexin: Again, I used what was already in the file.  But I will change it, no problem.

phpBB/phpbb/extension/extension_interface.php
javiexin: That would be a tremendous backward compatibility break: a lot of the currently approved extensions would stop working, and a lot of extensions in development, or even abandoned, would fail for this single reason.
It would have been better that way at the very beginning, but not now, in my view.  The intention here was to keep the current interface intact, and that all existing extensions would continue working without any change.  And if some extension wants to change, then it is easy to do it this way, while keeping 100% BC.

VSEphpbb: Yeah is_enableable should only return a Boolean.

Nicofuma: having another method for the reason is a bad design as you will have to implement the logic twice.
Another solution might to to deprecate is_enableable() and add another method which can return a string or an array of string. Then we can deal with both form to maintain BC (as we would have done if we have changed the return type of the existing method)

javiexin: Don't like it too much, but if you want, we might:

1. Deprecate is_enableable
2. Make not_enableable_reason return either string/array like now, or boolean; any "false" return (false, '', array()) would mean the extension is enableable; any "true" value would mean not enableable
3. Have a default implementation of not_enableable_reason as !is_enableable
4. Change the core to use the new function with reverse logic.

Nicofuma: That's exactly what I proposed. But you need a better name than `not_enableable_reason`

javiexin: Open to suggestions. Maybe just `not_enableable`.

Don't agree with the return format: what would be the meaning of [] or ''?
All truish values should mean the same, and all falsish values as well.

Nicofuma: No need for complex parsing, that's what we are already supporting :
 - $language->lang_array('string', array)
 - $language->lang('string')

For the language string not loaded because the extension is not available I see the issue. But we could add in the core a bunch of messages corresponding to the most frequent issue (wrong version of php, wrong version of phpbb, missing php ext, missing dependent ext etc...)

- `''` would just be an empty message, that's not our problem as it won't cause any issue.
- `[]` should use the same default than null

About the truish / falsish values I would ignore that. PHP is a complex language with nuance. The weak typing is a feature but we are not forced to use it everywhere and for everything. It will just make things more complex.

A third solution could be to support and handle exceptions: if an exception cannot be enabled, it just throw an exception in a pre_enable() method per example.

Ref: https://github.com/phpbb/phpbb/blob/master/phpBB/phpbb/language/language.php

javiexin: All of the language methods that you mention take a SINGLE entry and process it.
What we would need is a method that would take either a string, an asoc-array (key->array of params) or an array of keys or array of key+params...  This for me is kind of complex parsing, that of course would use the above services, but needs some significant code.

I don't see that much of an issue with extension messages: the requirement is that the appropriate language file is loaded within the "unnamed new method".  And given that that must be done there (the core cannot have all the possible reasons for an extension not being enableable), for the extension itself would not be complex to process the language strings, using the methods you mention, and without the need of more complex processing in the core.

I understand that we can do what we need with values and types, but I honestly believe that things are much easier to understand and comprehend by the general user when "make sense".  So, having "true" being OK, and having 'some string' being KO does not make much sense to me...

About exceptions, although I think this would be a good solution, it has a significant limitation: single message.

Nicofuma:
```
$result = $ext->get_errrors();

if ($result === true) {
     // continue processing
} else {
    $errs = [];
    if (is_string($result) || is_array($result)) {
        foreach ((array) $result as $message) {
             if (is_array($message)) {
                 $errs[] = $language->lang_array($message[0]; $message[1]);
             } else {
                 $errs[] = $language->lang($message);
             }
        }
    } else {
        $errs[] = $language->lang('GENERIC_ERROR_MESSAGE');
    }
}
```

It is a bit complex because it has to deal with a lot of different cases but you can simplify it if you force the most complex form for the errors and assume you will never have anything else (may be a bad idea) :

```
$result = $ext->get_errrors();

if ($result === true) {
     // continue processing
} else {
    $errs = [];
        foreach ((array) $result as $message) {
             $errs[] = $language->lang_array($message[0]; $message[1]);
        }
}
```

For your other concern, a string does not mean true, not in the real world and it does not mean false either. It is matter of convention and having true for a success or a list of errors in case of error is a common convention in software design.

But as I also said, it's not really the best convention. This kind of error could be solve another way:
 - pre_enable() / pre_disable() / pre_purge() methods
 - these methods are not expected to return anything
 - these methods do nothing by default
 - these methods can throw an exception to stop the installer from going forward 
 - the exception can implement the phpbb exception_interface, and in that case the language var will be handled
 - if we really want to support multiple error message we can add a new class of exception to deal with it

rxu: It seems for me that both points can be satisfied as well as BC retained: if we inject new class property (say, ```$errs = []```), and then just assign it a value(s) inside ```is_enableable()``` in case of ```false```. The ```is_enableable()``` will still return bool, while in case of ```false``` we could get error mesagge with some class method like ```get_errors()``` returning the ```$errs``` propetry value.

javiexin: I clearly prefer rxu's proposal.  Much cleaner, and better BC (perfect, I should say).
But just rename `not_enableable_reason` to `get_errors` and voila!  That's it.
The fact that $errs is in the class is an implementation detail, that might be the norm, but should not be imposed or required.

rxu: @javiexin with theupdated  proposal by @Nicofuma it's actually similar way but more elegant 😉

javiexin: Just for the processing of the language strings, ok.  Otherwise, identical to my proposal.
But again, for that, I would add a method to the language object that would take the "complex" structure and get the clean array of translated messages out (using code similar to the one proposed by Nicofuma).  But, in my view, this is not part of this PR.

Nicofuma: A such method does not belong in the language manager. There is too few usage and to few logic. Actually there is nothing inside, just a loop.

About the solution with 2 methods, as I said many time, I don't like and I won't validate it myself.

javiexin: Ok, single method proposal.

One single method, called `not_enableable`.  All current uses of `is_enableable` should simply be negated, and that would be 100% same behaviour as now.

If the method returns `true` the extension cannot be enabled, without any specified reason, so the default error string should be used.  If the method returns a (non-empty) string, or a (non-empty) array of strings, then this should be interpreted as PROCESSED language strings, and transmitted as such to the end user.  If the method returns false, an empty string or an empty array, the extension may be enabled normally.

The `base` implementation would implement `not_enableable` as `!is_enableable`, therefore making it 100% BC.  For extensions that have implemented a "workaround" for this limitation, they will need to rewrite, or continue using the "workaround".

To help extension developers with the language processing of the returned string/array, a helper function will be included in the `base` ext object, so that the extension developer may simply do `return $this->helper($array_of_key)`, being the array of keys as complex as described above, with all possible variations.

If this proposal is accepted, I will gladly code it and resubmit, but I won't do anything until the concept is approved.  If this proposal is not accepted, please, let me know, so that I can close this PR.

javiexin: @Nicofuma: should I implement this, or still not ok on your side?

javiexin: @Nicofuma if this is not going to progress, better to close it.  Let me know: close it, or implement proposed alternative?

3D-I:
Guys, Javier is not anymore with us, life's sad sometimes. 
Please take care of his contributions here.

Thanks.
https://www.phpbb.com/community/viewtopic.php?p=14938226#p14938226

PR #4864, opened 2017-07-01
[ticket/15259] Fatal error on SQLite/Oracle database update

lavigor:
PHPBB3-15259

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15259

The tools code was changed with some remaining issues that resulted into Fatal error during the update to phpBB 3.2.1-RC1 on SQLite database.

What I've tried to fix in this commit:
- `check_index_name_length` does not take the third parameter into account when calls itself recursively.
- `check_index_name_length` reduces one more symbol from `index_name` when tries to remove `table_prefix` (it is not separated from the remaining `table_suffix` with `_` like `table_name` from `index_name`).
- `check_index_name_length` incorrectly tries to remove `table_name` after it has removed `table_prefix` (instead we need to remove the remaining part of `table_name` - `table_suffix`).
- `sql_index_exists` does not take into account that `$index_name` should be the same on the next iteration of the `while` loop.
- `sql_index_exists` does not take into account the new shortening method for other DBMS (like MySQL).

What I've not fixed (because need some help):
- `sql_unique_index_exists` still needs fixing the same way. It's too complicated for me because I do not have Oracle DBMS for testing its different cases.

Hope that my explanations are informative.

### DIFF
--- a/phpBB/phpbb/db/tools/tools.php
+++ b/phpBB/phpbb/db/tools/tools.php
@@ -941,29 +941,19 @@ class tools implements tools_interface
 				continue;
 			}
 
-			// These DBMS prefix index name with the table name
 			switch ($this->sql_layer)
 			{
+				// These DBMS prefix index name with the table name
 				case 'oracle':
 				case 'sqlite3':
-					$index_name = $this->check_index_name_length($table_name, $table_name . '_' . $index_name, false);
-					$table_prefix = substr(CONFIG_TABLE, 0, -6); // strlen(config)
-
-					if (strpos($index_name , $table_name) === false)
-					{
-						if (strpos($index_name, $table_prefix) !== false)
-						{
-							$row[$col] = substr($row[$col], strlen($table_prefix) + 1);
-						}
-						else
-						{
-							$row[$col] = substr($row[$col], strlen($table_name) + 1);
-						}
-					}
+					$new_index_name = $this->check_index_name_length($table_name, $table_name . '_' . $index_name, false);
+				break;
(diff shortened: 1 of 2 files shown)


rxu:
Do you have an idea why tests on ```sql_index_exists```  pass for the current 3.2.x here and after https://github.com/phpbb/phpbb/blob/3.2.x/tests/dbal/db_tools_test.php#L226?
Care to add some entries for PHPBB3-15259 testing?

DavidIQ:
The fix might be as simple as reverting the changes done to tools.php and going back to this version:
https://github.com/lavigor/phpbb/blob/e974f338afb86c065e9b160363bc2e6156f8566d/phpBB/phpbb/db/tools/tools.php

The reason for the changes might have been mitigated through shortening of the indexes that brought about the need to shorten them to begin with.

lavigor:
@DavidIQ do you suggest shortening the indices instead of fixing that shortening function?

@rxu I've tried adding tests, but the usage of `CONFIG_TABLE` constant stopped me because it requires lots of work for adding a new `$table_prefix` parameter to the tools class to fix it the right way or calculating the length of the base table prefix (that constant cannot be overridden for testing purposes).

lavigor:
@rxu finally implemented a workaround.
It does not pretend to be the fully right solution, but the usage of the constant is not fully right too.

As a result:
- the function and the test work properly on SQLite;
- `sql_unique_index_exists` still needs fixing (and a new test too).

PR #4892, opened 2017-08-07
ticket/15309 Fixing prosilver so that pagination icons don't hit text,

v12mike:
Fixing prosilver so that pagination icons don't hit text,
and text reflows nicely when viewport reduced on veiwforum
and search results pages.

Second try at pulling this change...

Checklist:

- [ ] Tests pass

### DIFF
--- a/phpBB/styles/prosilver/template/search_results.html
+++ b/phpBB/styles/prosilver/template/search_results.html
@@ -107,31 +107,32 @@
 							<!-- EVENT topiclist_row_topic_title_after -->
 
 							<!-- IF not S_IS_BOT -->
-								<div class="responsive-show" style="display: none;">
-									{L_LAST_POST} {L_POST_BY_AUTHOR} {searchresults.LAST_POST_AUTHOR_FULL} &laquo; <a href="{searchresults.U_LAST_POST}" title="{L_GOTO_LAST_POST}">{searchresults.LAST_POST_TIME}</a>
-									<br />{L_POSTED} {L_IN} <a href="{searchresults.U_VIEW_FORUM}">{searchresults.FORUM_TITLE}</a>
-								</div>
-							<!-- IF searchresults.TOPIC_REPLIES --><span class="responsive-show left-box" style="display: none;">{L_REPLIES}{L_COLON} <strong>{searchresults.TOPIC_REPLIES}</strong></span><!-- ENDIF -->
-							<!-- ENDIF -->
+							<div>
+    								<span class="responsive-show" style="display: none;">
+    									{L_LAST_POST} {L_POST_BY_AUTHOR} {searchresults.LAST_POST_AUTHOR_FULL} &laquo; <a href="{searchresults.U_LAST_POST}" title="{L_GOTO_LAST_POST}">{searchresults.LAST_POST_TIME}</a>
+    									<br />{L_POSTED} {L_IN} <a href="{searchresults.U_VIEW_FORUM}">{searchresults.FORUM_TITLE}</a>
+    								</span>
+    							<!-- IF searchresults.TOPIC_REPLIES --><span class="responsive-show left-box" style="display: none;">{L_REPLIES}{L_COLON} <strong>{searchresults.TOPIC_REPLIES}</strong></span><!-- ENDIF -->
+    							<!-- ENDIF -->
 
-							<div class="responsive-hide">
-								<!-- IF searchresults.S_HAS_POLL --><i class="icon fa-bar-chart fa-fw" aria-hidden="true"></i><!-- ENDIF -->
-								<!-- IF searchresults.ATTACH_ICON_IMG --><i class="icon fa-paperclip fa-fw" aria-hidden="true"></i><!-- ENDIF -->
-								{L_POST_BY_AUTHOR} {searchresults.TOPIC_AUTHOR_FULL} &raquo; {searchresults.FIRST_POST_TIME} &raquo; {L_IN} <a href="{searchresults.U_VIEW_FORUM}">{searchresults.FORUM_TITLE}</a>
-							</div>
+								<span class="responsive-hide">
+									<!-- IF searchresults.S_HAS_POLL --><i class="icon fa-bar-chart fa-fw" aria-hidden="true"></i><!-- ENDIF -->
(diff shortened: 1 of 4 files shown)


v12mike:
Sorry, I noticed a minor problem with the code in this PR.  I will fix it and try again

v12mike:
After checking, I decided that the PR as submitted was correct, so re-opening

hanakin:
1. this breaks BC so it will never get merged into 3.2 you would have to work it into 3.3
2. What are you trying to do? pls post screenshots of any style related changes.
3. I am working on all the information for front-end coding stds for 3.3 but do not use spans as blocks please use block level elements. Prosilver was created in the age of divs and can not really be easily broken out of it so just stick with divs.
4. you closing tag for pagination is a div but your opening tag is a span?
5. no-wrap and pagination should not have the same. no-wrap belongs in utilities.css loose the pagination tie and just add no-wrap props to the main pagination class.
6. pls read over the submission policies for PRs as you formatting is incorrect. You are missing the ticket ID and not properly linking to the ticket

v12mike:
I am confused by some of the comments, can you clarify them?

this breaks BC so it will never get merged into 3.2 you would have to work it into 3.3
- can you be more speific, about where it breaks BC?

What are you trying to do? pls post screenshots of any style related changes.
- https://tracker.phpbb.com/browse/PHPBB3-15309 contains a screenshot of the problem of the pagination icons colliding with the text.  There was a patch applied under https://tracker.phpbb.com/browse/PHPBB3-15023 but that really doesn't fix the issue.

I am working on all the information for front-end coding stds for 3.3 but do not use spans as blocks please use block level elements. Prosilver was created in the age of divs and can not really be easily broken out of it so just stick with divs.
- The problem here is the divs.  Having multiple block elements within the td causes unwanted line breaks and poor reflow on narrow viewports.  I don't understand how changing these to spans breaks anything, and I don't kknow how you can eliminate the unwanted line breaks in these tables with divs.

you closing tag for pagination is a div but your opening tag is a span?
- I don't think so.  In viewforum_body.html, the closing </div> on line 219 matches the opening <div> on line 197, and there is no </div> in between them.  Similarly in search_results.html, the closing </div> in line 137 matches the opening <div> on line 110.

no-wrap and pagination should not have the same. no-wrap belongs in utilities.css loose the pagination tie and just add no-wrap props to the main pagination class.
- OK I accept that, and will update the PR

pls read over the submission policies for PRs as you formatting is incorrect. You are missing the ticket ID and not properly linking to the ticket
- OK I will check at fix.

also this is not even really an issue with prosilver... This is an issue with prosilver SE
- I don't understand.  The bug is in prosilver, Can you explain why it would not be fixed in prosilver?

phpBB/styles/prosilver/template/search_results.html
```diff
@@ -107,31 +107,32 @@ <h2 class="searchresults-title"><!-- IF SEARCH_TITLE -->{SEARCH_TITLE}<!-- ELSE
...
+    							<!-- ENDIF -->
+
+								<span class="responsive-hide">
+									<!-- IF searchresults.S_HAS_POLL --><i class="icon fa-bar-chart fa-fw" aria-hidden="true"></i><!-- ENDIF -->
+									<!-- IF searchresults.ATTACH_ICON_IMG --><i class="icon fa-paperclip fa-fw" aria-hidden="true"></i><!-- ENDIF -->
+									<span class="no-wrap">{L_POST_BY_AUTHOR} {searchresults.TOPIC_AUTHOR_FULL}</span> <span class="no-wrap">&raquo; {searchresults.FIRST_POST_TIME}</span> <span class="no-wrap">&raquo; {L_IN} <a href="{searchresults.U_VIEW_FORUM}">{searchresults.FORUM_TITLE}</a></span>
+								</span>
+
+								<!-- IF .searchresults.pagination -->
+								<span class="pagination">
```
hanakin: where is the closing span for this?

phpBB/styles/prosilver/template/search_results.html
```diff
@@ -107,31 +107,32 @@ <h2 class="searchresults-title"><!-- IF SEARCH_TITLE -->{SEARCH_TITLE}<!-- ELSE
...
+									<span><i class="icon fa-clone fa-fw" aria-hidden="true"></i></span>
+									<ul>
```
hanakin: This is a block level element which means you can not wrap it in a span. Ultimately the fix is to remove the wrap altogether and make the ul the pagination class and move the above icon into the ul

aka like the paging here: https://codepen.io/hanakin/pen/dNVWmm?editors=1100

phpBB/styles/prosilver/template/search_results.html
```diff
@@ -107,31 +107,32 @@ <h2 class="searchresults-title"><!-- IF SEARCH_TITLE -->{SEARCH_TITLE}<!-- ELSE
...
-								<ul>
-								<!-- BEGIN pagination -->
-									<!-- IF searchresults.pagination.S_IS_PREV -->
-									<!-- ELSEIF searchresults.pagination.S_IS_CURRENT --><li class="active"><span>{searchresults.pagination.PAGE_NUMBER}</span></li>
-									<!-- ELSEIF searchresults.pagination.S_IS_ELLIPSIS --><li class="ellipsis"><span>{L_ELLIPSIS}</span></li>
-									<!-- ELSEIF searchresults.pagination.S_IS_NEXT -->
-									<!-- ELSE --><li><a class="button" href="{searchresults.pagination.PAGE_URL}">{searchresults.pagination.PAGE_NUMBER}</a></li>
-									<!-- ENDIF -->
-								<!-- END pagination -->
+							<div>
```
hanakin: we do not add not targetable divs. Meaning if this is necessary it needs to have a class. This is also probably the biggest breaking change

hanakin:
1. you are completely changing the structure and css class which means it will break any themes that inherit from prosilver. This is a breaking change. 3.2.x is a maintenance release so breaking changes are not allowed if can be avoided. 3.3 is a minor release which allows for some minor breaking changes if absolutely required due to bugs.

2. Once a PR is started all discussion needs to move to the PR for clarity. Pls post screen shot comparisons of before and after on the PR for review of the PR.

ass for all the coding issues see direct comments

v12mike:
I see the point about other styles dependent on prosilver.

As I am out of my depth in css programming, I will close this PR.  I will leve the ticket open, as the prosilver search results format is still broken.

PR #4893, opened 2017-08-07
[ticket/15305] Add streams for storage

rubencm:
Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15305


phpBB/phpbb/storage/adapter/local.php
```diff
@@ -195,4 +196,39 @@ protected function ensure_directory_exists($path)
...
+		}
+
+		$stream = @fopen($this->root_path . $path, 'w+b');
+
+		if (!$stream)
+		{
+			throw new exception('STORAGE_CANNOT_CREATE_FILE', $path);
+		}
+
+		stream_copy_to_stream($resource, $stream);
```
Nicofuma: you should check the return (false in case of error)

rubencm: Uhm, true, the spanish translation in the php documentation skips that

Crizz0:
Just for me, what does this patch do?

rubencm:
@Crizz0 To be able to use storage with large files without running out of memory

rubencm:
!set WIP

rubencm:
!unset WIP

phpBB/phpbb/storage/stream_interface.php
```diff
@@ -0,0 +1,41 @@
...
+
+interface stream_interface
+{
+	/**
+	 * Reads a file as a stream.
+	 *
+	 * @param string	$path	File to read
+	 *
+	 * @throws \phpbb\storage\exception\exception		When cannot open file
+	 * @throws \phpbb\storage\exception\not_implemented	When the adapter doesnt implement the method
```
Nicofuma: removes this line. If the adapter does not support streams then it does not add the interface

phpBB/phpbb/storage/storage.php
```diff
@@ -133,11 +140,64 @@ public function rename($path_orig, $path_dest)
...
+		return $stream;
+	}
+
+	/**
+	 * Writes a new file using a stream.
+	 *
+	 * @param string	$path		The target file
+	 * @param resource	$resource	The resource
+	 *										When target file cannot be created
+	 * @throws \phpbb\storage\exception\not_implemented	When the adapter doesnt implement the method
```
Nicofuma: should never happen

phpBB/phpbb/storage/storage.php
```diff
@@ -133,11 +140,64 @@ public function rename($path_orig, $path_dest)
 	 * @param string	$path_orig	The original filename
 	 * @param string	$path_dest	The target filename
 	 *
-	 * @throws \phpbb\storage\exception\exception	When target exists
-	 * 												When the file cannot be copied
+	 * @throws \phpbb\storage\exception\exception		When target exists
+	 * 													When the file cannot be copied
+	 * @throws \phpbb\storage\exception\not_implemented	When the adapter doesnt implement the method
```
Nicofuma: actually I don't think we should ever throw a not_implemented exception. These methods are mandatory and can easily be faked (copy = create new file ; rename = create new file + delete old etc...)

phpBB/phpbb/storage/storage.php
```diff
@@ -94,6 +97,8 @@ public function get_contents($path)
 	 *
 	 * @param string	$path	file/directory to check
 	 *
+	 * @throws \phpbb\storage\exception\not_implemented	When the adapter doesnt implement the method
+	 *
 	 * @return bool	Returns true if all files/directories exist, false otherwise
```
Nicofuma: Returns true if the file/directory exist, false otherwise (the type hint is string)

phpBB/phpbb/storage/adapter/local.php
```diff
@@ -195,4 +196,42 @@ protected function ensure_directory_exists($path)
...
+
+		if (!$stream)
+		{
+			throw new exception('STORAGE_CANNOT_CREATE_FILE', $path);
+		}
+
+		if (stream_copy_to_stream($resource, $stream) === false)
+		{
+			throw new exception('STORAGE_CANNOT_COPY_RESOURCE');
+		}
```
Nicofuma: you should close $stream explicitly I think

phpBB/phpbb/storage/storage.php
rubencm: Maybe with some exotic adapter... for example (its an example, probably i am wrong), with gravatar you can read the generated files, but you cant write them directly

phpBB/phpbb/storage/storage.php
rubencm: Ok, i can remove not_implemented exception directly if you want, and they just throws a normal exception like if a generic error ocurred

phpBB/phpbb/storage/storage.php
Nicofuma: but gravatar can't be seen as a storage backend. For me storage should all be read/write. Maybe later we could support read only backend with a few changes (but I think it's not important for now and it should be done specifically otherwise it can be confusing)

phpBB/phpbb/storage/storage.php
Nicofuma: I think we should do that yes

phpBB/phpbb/storage/storage.php
Nicofuma: it may already work (being read only) but I rather avoid to officially support it for now

phpBB/phpbb/storage/storage.php
```diff
@@ -133,11 +134,62 @@ public function rename($path_orig, $path_dest)
...
 	{
 		$this->get_adapter()->copy($path_orig, $path_dest);
 	}
+
+	/**
+	 * Reads a file as a stream.
+	 *
+	 * @param string	$path	File to read
+	 *
+	 * @throws \phpbb\storage\exception\exception		When cannot open file
```
marc1706: When unable to open file

phpBB/phpbb/storage/stream_interface.php
```diff
@@ -0,0 +1,39 @@
...
+namespace phpbb\storage;
+
+interface stream_interface
+{
+	/**
+	 * Reads a file as a stream.
+	 *
+	 * @param string	$path	File to read
+	 *
+	 * @throws \phpbb\storage\exception\exception		When cannot open file
```
marc1706: See above comment

marc1706:
Rest looks good.

PR #4894, opened 2017-08-08
[ticket/15276] Use storage in avatars

rubencm:
Checklist:

- [x] Tests pass

This depends on #4893 

https://tracker.phpbb.com/browse/PHPBB3-15276


phpBB/phpbb/storage/adapter/local.php
```diff
@@ -195,4 +196,57 @@ protected function ensure_directory_exists($path)
...
+	}
+
+	public function file_size($path)
+	{
+		return filesize($this->root_path . $path);
+	}
+
+	public function file_mimetype($path)
+	{
+		return mime_content_type($this->root_path . $path);
```
rubencm: I am looking that this function is obsolete, but i dont find much information about Fileinfo

rubencm:
!set WIP

phpBB/includes/acp/acp_main.php
```diff
@@ -502,26 +502,8 @@ function main($id, $mode)
...
-
-			$avatar_dir_size = get_formatted_filesize($avatar_dir_size);
-		}
-		else
-		{
-			// Couldn't open Avatar dir.
-			$avatar_dir_size = $user->lang['NOT_AVAILABLE'];
-		}
+		// Couldn't open Avatar dir.
+		$avatar_dir_size = $user->lang['NOT_AVAILABLE'];
```
rubencm: I need here another way to get the avatars total filesize

rubencm:
Im pretty sure that i have broken the conversion from phpbb2

phpBB/includes/functions_download.php
```diff
@@ -69,24 +63,23 @@ function send_avatar_to_browser($file, $browser)
...
 
-			if ($fp !== false)
+			while (!feof($fp))
 			{
-				while (!feof($fp))
-				{
-					echo fread($fp, 8192);
-				}
-				fclose($fp);
+				echo fread($fp, 8192);
```
rubencm: Maybe is better to do fwrite('php://output', fread($fp, 8192)); ?

rubencm:
!unset WIP

phpBB/includes/functions_download.php
```diff
@@ -69,24 +76,23 @@ function send_avatar_to_browser($file, $browser)
 			header('Expires: ' . gmdate('D, d M Y H:i:s', time() + 31536000) . ' GMT');
 		}
 
-		$size = @filesize($file_path);
-		if ($size)
-		{
-			header("Content-Length: $size");
+		try {
+			header('Content-Type: ' . $file_info->size);
```
Nicofuma: Content-Length

phpBB/includes/functions_download.php
Nicofuma: actually as we have a stream we can use `stream_copy_to_stream($fp, 'php://output')` http://php.net/manual/fr/function.stream-copy-to-stream.php

phpBB/phpbb/storage/adapter/local.php
```diff
@@ -234,4 +234,19 @@ public function write_stream($path, $resource)
 			throw new exception('STORAGE_CANNOT_COPY_RESOURCE');
 		}
 	}
+
+	public function file_properties($path)
```
Nicofuma: this method should be in the interface

phpBB/includes/acp/acp_main.php
Nicofuma: This is just not possible with most cloud providers

phpBB/phpbb/storage/adapter/local.php
```diff
@@ -234,4 +252,82 @@ public function write_stream($path, $resource)
...
+	/**
+	 * Get image width.
+	 *
+	 * @param string	$path	The file
+	 *
+	 * @return array	Properties
+	 */
+	public function file_image_width($path)
+	{
+		return $this->image_dimensions($path);
```
Nicofuma: return $this->image_dimensions($path)['image_width'];

phpBB/includes/acp/acp_main.php
Nicofuma: Actually it may be possible for some providers and I think it can still be useful (even if it displays the global usage).
You could add a get_total_usage or something like that in the provider which will return an array containing the current total usage and the maximum allowed usage (if it exists). If one information is missing its value being replaced by false.

Example: (all numbers are in bytes)
return ['usage' => 9999, 'max' => 10000] 
return ['usage' => 9999, 'max' => false]
return ['usage' => false, 'max' => false]

phpBB/phpbb/storage/adapter/local.php
rubencm: This if from an old commit, i removed this

phpBB/includes/acp/acp_main.php
rubencm: Yes, but i cannot get the usage easily.
I think i need to create a provider for the storages, with methods to get the list of files, or the total size.
And in the adapter a method to get the total available space.

Nicofuma: well, it depends what we want to do.
If we want to track all files then we need a new table yes (we can have a singe table for all storages) but if we just want the current usage of the cloud storage then we can rely on the adapter directly.

rubencm: I dont know, i still think that is better to do a class for each storage with methods to get the files or somehing like that

Nicofuma: I'm sorry but I don't understand what you have in mind.

rubencm: ok, im going to do a concept, but i have to finish other things first

phpBB/phpbb/storage/adapter/local.php
rubencm: No, this is right, but i think i will remove file_info, i dont like how it works

phpBB/includes/acp/acp_main.php
Nicofuma: please create a dedicated ticket for this issue

Nicofuma:
@rubencm you need to rebase on master I think

rubencm:
done

Nicofuma:
better, thanks

Nicofuma:
@phpbb/development-team it's ready for review and without commits from another PR now

phpBB/includes/functions_download.php
```diff
@@ -25,30 +25,24 @@
...
-	$file_path = $phpbb_root_path . $image_dir . '/' . $prefix . $file;
+		$file_info = $storage->file_info($file_path);
 
-	if ((@file_exists($file_path) && @is_readable($file_path)) && !headers_sent())
-	{
 		header('Cache-Control: public');
 
-		$image_data = @getimagesize($file_path);
-		header('Content-Type: ' . image_type_to_mime_type($image_data[2]));
+		try {
```
marc1706: Braces on new line please (applies to other parts of the code as well).

phpBB/phpbb/storage/file_info.php
```diff
@@ -0,0 +1,71 @@
...
+	 */
+	protected $adapter;
+
+	/**
+	 * @var string
+	 */
+	protected $path;
+
+	/**
+	 * @var array
```
marc1706: Maybe an explanation of what properties these are?

phpBB/phpbb/storage/file_info.php
```diff
@@ -0,0 +1,71 @@
...
+	 */
+	protected $properties;
+
+	public function __construct($adapter, $path)
```
marc1706: Docblock pls

rubencm: This is provisional (how we get metadata from files), if you know a better way to do it tell me

phpBB/develop/adjust_avatars.php
```diff
@@ -30,6 +30,11 @@
 	die('database not up to date');
 }
 
+if (!isset($config['storage\\avatar\\config\\path']) || $config['storage\\avatar\\config\\path'] != 'phpbb\\storage\\provider\\local')
```
CHItA: Use ` !==` if possible.

phpBB/develop/adjust_avatars.php
```diff
@@ -30,6 +30,11 @@
...
+{
+	die('use local provider');
```
CHItA: We should display some translateable error message here i think.

phpBB/includes/functions_user.php
```diff
@@ -2505,22 +2512,29 @@ function group_create(&$group_id, $type, $name, $desc, $group_attributes, $allow
...
 
 	$group_id		= (int) $group_id;
 	$ext 			= substr(strrchr($old_entry, '.'), 1);
 	$old_filename 	= get_avatar_filename($old_entry);
 	$new_filename 	= $config['avatar_salt'] . "_g$group_id.$ext";
 	$new_entry 		= 'g' . $group_id . '_' . substr(time(), -5) . ".$ext";
 
-	$avatar_path = $phpbb_root_path . $config['avatar_path'];
-	if (@rename($avatar_path . '/'. $old_filename, $avatar_path . '/' . $new_filename))
+	try
```
CHItA: This modification is not exactly the same as what the code did before which is fine for now, but it may cause problems if this needs to be extended in the future. (meaning error handling could get difficult if there are multiple statements that could throw exceptions).

phpBB/phpbb/files/types/remote_storage.php
```diff
@@ -0,0 +1,207 @@
...
+
+		$content_length = $response->getBody()->getSize();
+		if ($remote_max_filesize && $content_length > $remote_max_filesize)
+		{
+			$max_filesize = get_formatted_filesize($remote_max_filesize, false);
+
+			return $this->factory->get('filespec')->set_error($this->language->lang($this->upload->error_prefix . 'WRONG_FILESIZE', $max_filesize['value'], $max_filesize['unit']));
+		}
+
+		if ($content_length == 0)
```
CHItA: `===` instead of double

phpBB/develop/adjust_avatars.php
rubencm: I dont know, dont know what is the purpose of this folder, but none of the strings in this file are translatable

phpBB/includes/functions_user.php
rubencm: Uhm, i think it is.
The update query is only executed if rename throws no exception

Nicofuma:
@phpbb/development-team please review it a last time :)

phpBB/develop/adjust_avatars.php
```diff
@@ -30,6 +30,11 @@
 	die('database not up to date');
 }
 
+if (!isset($config['storage\\avatar\\config\\path']) || $config['storage\\avatar\\config\\path'] !== 'phpbb\\storage\\provider\\local')
+{
+	die('use local provider');
```
marc1706: I think this message should be improved as for someone not knowing the exact purpose of this might not be able to understand why it's causing a die().

phpBB/includes/functions_download.php
```diff
@@ -25,30 +25,27 @@
...
 
-		$image_data = @getimagesize($file_path);
-		header('Content-Type: ' . image_type_to_mime_type($image_data[2]));
+		try
+		{
+			header('Content-Type: ' . $file_info->mimetype);
+		}
+		catch (\phpbb\storage\exception\exception $e)
+		{
+			// Just dont send this header
```
marc1706: `// Just don't send this header`

phpBB/includes/functions_download.php
```diff
@@ -69,24 +66,26 @@ function send_avatar_to_browser($file, $browser)
...
 		{
-			header("Content-Length: $size");
+			header('Content-Length: ' . $file_info->size);
 		}
-
-		if (@readfile($file_path) == false)
+		catch (\phpbb\storage\exception\exception $e)
 		{
-			$fp = @fopen($file_path, 'rb');
+			// Just dont send this header
```
marc1706: See above

phpBB/includes/functions_user.php
```diff
@@ -2505,22 +2512,29 @@ function group_create(&$group_id, $type, $name, $desc, $group_attributes, $allow
...
+		$this->storage->rename($old_filename, $new_filename);
+
 		$sql = 'UPDATE ' . GROUPS_TABLE . '
 			SET group_avatar = \'' . $db->sql_escape($new_entry) . "'
 			WHERE group_id = $group_id";
 		$db->sql_query($sql);
 	}
+	catch (\phpbb\storage\exception\exception $e)
+	{
+
```
marc1706: Please add a comment here

phpBB/phpbb/avatar/driver/upload.php
```diff
@@ -316,12 +297,12 @@ public function get_template_name()
...
 	/**
-	* Check if user is able to upload an avatar
+	* Check if user is able to upload an avatar to a temporary folder
 	*
 	* @return bool True if user can upload, false if not
 	*/
 	protected function can_upload()
 	{
-		return ($this->filesystem->exists($this->phpbb_root_path . $this->config['avatar_path']) && $this->filesystem->is_writable($this->phpbb_root_path . $this->config['avatar_path']) && (@ini_get('file_uploads') || strtolower(@ini_get('file_uploads')) == 'on'));
+		return (@ini_get('file_uploads') || strtolower(@ini_get('file_uploads')) == 'on');
```
marc1706: Not sure if we should be using the bantu's ini class? @Nicofuma

phpBB/phpbb/files/types/remote_storage.php
```diff
@@ -0,0 +1,207 @@
...
+		return $this->remote_upload($args[0]);
+	}
+
+	/**
+	 * Remote upload method
+	 * Uploads file from given url
+	 *
+	 * @param string $upload_url URL pointing to file to upload, for example http://www.foobar.com/example.gif
+	 * @return filespec $file Object "filespec" is returned, all further operations can be done with this object
+	 * @access public
```
marc1706: It's both wrong and no longer needed IMHO

phpBB/phpbb/avatar/driver/upload.php
Nicofuma: we could use it yes

phpBB/develop/adjust_avatars.php
```diff
@@ -131,8 +136,8 @@
 function adjust_avatar($old_name, $midfix)
 {
 	global $config, $phpbb_root_path;
-	
-	$avatar_path = $phpbb_root_path . $config['avatar_path'];
+
+	$avatar_path = $phpbb_root_path . $config['storage\\avatar\\config\\path'];
```
Elsensee: I know, there never was a rule about not having (back-)slashes in config names, but without underscores this somehow looks weird to me and breaking the status quo of having configuration names.

phpBB/develop/adjust_avatars.php
rubencm: I think i can rewrite this to use storage, but i dont know under which circunstances this script is used, so i just prevent to use it after changing storage

phpBB/phpbb/files/types/remote_storage.php
rubencm: I don't understand

phpBB/develop/adjust_avatars.php
Nicofuma: Actually I it asked on purpose. For me as there is a status quo of not using \ we can reserved it for internal usage like that and have a guarantee of 0 conflict ever.

phpBB/develop/adjust_avatars.php
Nicofuma: Actually I think we can leave it like that for now. We should start working on the dev doc soon and include a link here

phpBB/phpbb/files/types/remote_storage.php
Nicofuma: just trash the @access line

phpBB/develop/adjust_avatars.php
Elsensee: Internal usage means that you're not having any intention to make the path customizable? Then why not mark it as a parameter in a yaml file?

Otherwise I don't understand the term "internal usage" or have a different opinion.

Nicofuma: the path is customizable from a core page in the acp
but we have a lot of config values related to storage with dynamic names.
Here per example, avatar and path are dynamics the rest is internal. The full name is build internally (including storage, config and \)

marc1706: As personally don't care for this. It's a nice way to prevent conflicts for sure. As long as it doesn't mess up with some dbal

Elsensee: Yeah, okay.. It just looks weird to me but other than that it doesn't do anything harmful, I guess.

phpBB/phpbb/files/types/remote_storage.php
```diff
@@ -0,0 +1,197 @@
...
+		$this->phpbb_root_path = $phpbb_root_path;
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function upload()
+	{
+		$args = func_get_args();
+		return $this->remote_upload($args[0]);
```
CHItA: I don't really understand why this was done this way. What were the original arguments here?

phpBB/phpbb/files/types/remote_storage.php
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	/**
+	 * Get maximum file size for remote uploads
+	 *
+	 * @return int Maximum file size
+	 */
+	protected function get_max_file_size()
+	{
+		$max_file_size = $this->upload->max_filesize;
```
marc1706: I'm wondering if it would have been better to just simply add this as a method to the upload class but oh well

CHItA:
Otherwise it is looking good to me :+1:

phpBB/phpbb/storage/adapter/local.php
marc1706: But now this won't return the actual width but rather just the output of image_dimensions()?

PR #4896, opened 2017-08-08
[ticket/15310] Do not recognize numbers as template names

javiexin:
Currently, something like {923} in a template file is recognized as a potential template name, and replaced by {{ 923 }} for Twig processing.  This also happens with things like {9abc._} or {.a_b.}.
At the very least, things that do not start with a letter should be excluded, and that would rule out most of the "false positives" that it is currently identifying.

PHPBB3-15310

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-15310

### DIFF
--- a/phpBB/phpbb/template/twig/lexer.php
+++ b/phpBB/phpbb/template/twig/lexer.php
@@ -118,7 +118,7 @@ class lexer extends \Twig_Lexer
 
 		// Replace all of our variables, {VARNAME}, with Twig style, {{ VARNAME }}
 		// Appends any filters
-		$code = preg_replace('#{([a-zA-Z][a-zA-Z0-9_\.]*)(\|[^}]+?)?}#', '{{ $1$2 }}', $code);
+		$code = preg_replace('#{([a-zA-Z_][a-zA-Z0-9_\.]*)(\|[^}]+?)?}#', '{{ $1$2 }}', $code);
 
 		return parent::tokenize($code, $filename);
 	}
--- a/phpBB/phpbb/template/twig/lexer.php
+++ b/phpBB/phpbb/template/twig/lexer.php
@@ -118,7 +118,7 @@ class lexer extends \Twig_Lexer
 
 		// Replace all of our variables, {VARNAME}, with Twig style, {{ VARNAME }}
 		// Appends any filters
-		$code = preg_replace('#{([a-zA-Z0-9_\.]+)(\|[^}]+?)?}#', '{{ $1$2 }}', $code);
+		$code = preg_replace('#{([a-zA-Z][a-zA-Z0-9_\.]*)(\|[^}]+?)?}#', '{{ $1$2 }}', $code);
 
 		return parent::tokenize($code, $filename);
 	}


javiexin:
Ok, I see there are some tests that use variables starting with _ (underscore).  I will move it to allow as first character of the template var name.

phpBB/phpbb/template/twig/lexer.php
```diff
@@ -118,7 +118,7 @@ public function tokenize($code, $filename = null)
 
 		// Replace all of our variables, {VARNAME}, with Twig style, {{ VARNAME }}
 		// Appends any filters
-		$code = preg_replace('#{([a-zA-Z0-9_\.]+)(\|[^}]+?)?}#', '{{ $1$2 }}', $code);
+		$code = preg_replace('#{([a-zA-Z_][a-zA-Z0-9_\.]*)(\|[^}]+?)?}#', '{{ $1$2 }}', $code);
```
rxu: I see you want to make varnames starting only with the alpha letters or underscore, but this new regex will match ```test``` in the string ```123test``` for example.
I think you'd need to add the caret anchor to indicate the starting string letter, like ```(^[a-zA-Z_][a-zA-Z0-9_\.]*)```. Thus, it will need adding more specific tests I'd say.

javiexin: No, it won't match that.  Note the opening brace, that fulfills the same function as the caret you were proposing.  Anyhow, the caret would not work, because we are working in the lexer, and therefore the match must trigger anywhere in the string, not only at the beginning.  So, this regex matches `{test123}` but not `{123test}` and, the real issue, it won't match `{123}` either.

DavidIQ:
[3.1 is EOM](https://www.phpbb.com/community/viewtopic.php?f=14&t=2430891). Unless this is a security fix, which it isn't, it should be against 3.2.

javiexin:
It is a bug anyway...  But anyway, up to you really.  The change is the same for 3.1 and 3.2, tested.

DavidIQ:
EOM = End Of Maintenance. This means, for us and pretty much anywhere else, no bug fixes no enhancements, no changes, unless it is a security issue. This is clearly stated in the announcement.

javiexin:
As I said, up to you.  Apply it to 3.2 only.

DavidIQ:
You're missing the point. You need to rebase your pull request against 3.2 and fix any conflicts that may arise from doing that. It's not our job to do that ;)

javiexin:
And you are missing my point.  I already said it works exactly the same for 3.1 and 3.2, so, no conflict resolution.  And you (not David, the team members) change the base merging branch when you see fit, without caring about the author's opinion.  And most probably rightly so, but why sometimes you do it, and other times you keep asking me to do it?  We have spent more time arguing here than it takes to hit three clicks to change the base branch.  Anyhow, done.

Crizz0:
Hi,

Can you give an example in what case a style/extension author would add something like {923}?

Elsensee:
You are changing the regular expressions that we just took from our old template system that is still widely used I presume.
We don't make a difference whether a template file is fully twig or fully old-phpBB style which is also the main reason you can mix both in one file.

The downside to this is, that we have to maintain backwards compatibility with the old syntax and in the old syntax a variable name like `123test` would be totally fine because of that regular expression.

3D-I:
The last two comments don't apply IMHO.
@Crizz0
The issue arised during the development of our extension (fully twigged), I will try to resume the commits from a set of more of 500 in order to give you the exact example.
@Elsensee
There are no conflicts of any sort. Try it and see.

VSEphpbb:
I mentioned in the post referenced by 3D-I how to correctly handle the situation but the OP refused to do it the right way. I don't like the changes I see in this PR, as again it seems like a change for an incredibly rare situation that should never be when using phpBB and writing extensions the eay they are intended to be used.

To use `{123}` in a template, you simply treat each bracket as a separate string:
```twig
{{ '{' ~ '123' ~ '}' }}
```

Nicofuma:
or you can use html entities: https://dev.w3.org/html5/html-author/charref

marc1706:
Sounds to me like there are ways of treating this without causing issues or a BC break. Therefore closing this.

PR #4910, opened 2017-08-19
[ticket/15331] Allow gravatars to be overridden by existing event - B

javiexin:
Option B:
- change the get_avatar function to always trigger the event

PHPBB3-15331

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-15331


javiexin:
This solution modifes the get_avatar() function, so that it does not return prematurely.  Only modifies this function, but would apply to all driver types that use the "custom_html" method. It is also very easy, but a bit less clear.
Advantage: it applies to all avatar types that use the "custom_html" method, and will fix for all.  This would be my preferred choice.
Now, up to you to decide which one to use and merge, if applicable.

CHItA:
Remember to close #4909 if this gets merged.

PR #4913, opened 2017-08-21
[ticket/15333] Call callback when confirm dialog is canceled

kasimi:
PHPBB3-15333

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15333


Nicofuma:
I'm not sure we should call the callback if the dialog is just closed and not validated. What about providing a second callback for the other case?

kasimi:
I thought about that too. It would be anything but elegant though because the signature would look like this:
`function(msg, successCallback, fadedark, cancelCallback)`. 😒 

This PR as it is right now makes the function work exactly as described in the docblock. The core already checks the callback's argument, so no problem there. Extensions that don't check it would have to be updated.

Nicofuma:
ok

PR #4918, opened 2017-08-24
[ticket/15328] Disable checkbox if notification method isn't supported

kasimi:
Stuffing the instances into the `type` and `method` data arrays in `manager.php` doesn't feel like an elegant solution but it's the best I could find that is extensible.

As an alternative, it could all be done in the UCP module, probably with some `instanceof` checks, but that would make it deal with things that it shouldn't deal with, in my opinion. Other suggestions are welcome.

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15328

### DIFF
--- a/phpBB/phpbb/notification/method/email.php
+++ b/phpBB/phpbb/notification/method/email.php
@@ -13,6 +13,8 @@
 
 namespace phpbb\notification\method;
 
+use phpbb\notification\type\type_interface;
+
 /**
 * Email notification method class
 * This class handles sending emails for notifications
@@ -57,11 +59,11 @@ class email extends \phpbb\notification\method\messenger_base
 	* Is this method available for the user?
 	* This is checked on the notifications options
 	*
-	* @param \phpbb\notification\type\type_interface $notification_type An optional instance of a notification type. If provided, this
-	*																	method additionally checks if the type provides an email template.
+	* @param type_interface $notification_type  An optional instance of a notification type. If provided, this
+	*											method additionally checks if the type provides an email template.
 	* @return bool
 	*/
-	public function is_available(\phpbb\notification\type\type_interface $notification_type = null)
+	public function is_available(type_interface $notification_type = null)
 	{
 		return parent::is_available($notification_type) && $this->config['email_enable'] && $this->user->data['user_email'];
 	}
--- a/phpBB/phpbb/notification/method/email.php
+++ b/phpBB/phpbb/notification/method/email.php
@@ -56,10 +56,14 @@ class email extends \phpbb\notification\method\messenger_base
 	/**
 	* Is this method available for the user?
 	* This is checked on the notifications options
+	*
+	* @param \phpbb\notification\type\type_interface $notification_type An optional instance of a notification type. If provided, this
+	*																	method additionally checks if the type provides an email template.
+	* @return bool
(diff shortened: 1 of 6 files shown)


kasimi:
Any idea what Travis is trying to tell me?

phpBB/phpbb/notification/method/email.php
```diff
@@ -56,10 +56,14 @@ public function get_type()
 	/**
 	* Is this method available for the user?
 	* This is checked on the notifications options
+	*
+	* @param \phpbb\notification\type\type_interface $notification_type An optional instance of a notification type. If provided, this
+	*																	method additionally checks if the type provides an email template.
+	* @return bool
 	*/
-	public function is_available()
+	public function is_available(\phpbb\notification\type\type_interface $notification_type = null)
```
CHItA: Add a `use` statement maybe.

phpBB/phpbb/notification/method/jabber.php
```diff
@@ -56,10 +56,14 @@ public function get_type()
 	/**
 	* Is this method available for the user?
 	* This is checked on the notifications options
+	*
+	* @param \phpbb\notification\type\type_interface $notification_type	An optional instance of a notification type. If provided, this
+	*																	method additionally checks if the type provides an email template.
+	* @return bool
 	*/
-	public function is_available()
+	public function is_available(\phpbb\notification\type\type_interface $notification_type = null)
```
CHItA: Same as above.

kasimi:
At the moment it's a front-end fix. Form values aren't checked. Guaranteeing consistent user settings in the database would require a migration to uncheck all existing wrong values, too.

phpBB/styles/prosilver/template/ucp_notifications.html
```diff
@@ -31,7 +31,7 @@
 								<!-- IF notification_types.EXPLAIN --><br />&nbsp; &nbsp;{notification_types.EXPLAIN}<!-- ENDIF -->
 							</td>
 							<!-- BEGIN notification_methods -->
-								<td class="mark"><input type="checkbox" name="{notification_types.TYPE}_{notification_methods.METHOD}"<!-- IF notification_methods.SUBSCRIBED --> checked="checked"<!-- ENDIF --> /></td>
+								<td class="mark"><input type="checkbox" name="{notification_types.TYPE}_{notification_types.notification_methods.METHOD}"<!-- IF notification_types.notification_methods.AVAILABLE and notification_types.notification_methods.SUBSCRIBED --> checked="checked"<!-- ENDIF --><!-- IF not notification_types.notification_methods.AVAILABLE --> disabled="disabled"<!-- ENDIF --> /></td>
```
Elsensee: I believe it must be `notification_methods.AVAILABLE" only? Or was it just wrong before?

kasimi: It used the values of the header row before, which worked coincidentally. `notification_types.notification_methods.METHOD` is the actual row's value. https://wiki.phpbb.com/Template.assign_block_vars#Example_.232_Nested_blocks

Elsensee: Ah, that makes sense!

PR #4919, opened 2017-08-28
[ticket/15309] Fix pagination overlap in prosilver via "left-box"

sdschulze:
This is my attempt at fixing the pagination overlap problem.  Backward-compatibility is mostly preserved, except that non-adapted derived styles will have an extra line for paginated topics even when not needed.  The change to the responsive style is optional (credits to @v12mike for noticing the redundancy).

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15309

### DIFF
--- a/phpBB/styles/prosilver/theme/common.css
+++ b/phpBB/styles/prosilver/theme/common.css
@@ -886,7 +886,6 @@ fieldset.fields1 dl.pmlist dd.recipients {
 /* Pagination in viewforum for multipage topics */
 .row .pagination {
 	display: block;
-	margin-top: -12px;
 }
 
 .row .pagination > ul {
--- a/phpBB/styles/prosilver/theme/common.css
+++ b/phpBB/styles/prosilver/theme/common.css
@@ -886,6 +886,8 @@ fieldset.fields1 dl.pmlist dd.recipients {
 /* Pagination in viewforum for multipage topics */
 .row .pagination {
 	display: block;
+	margin-top: 3px;
+	margin-bottom: 3px;
 }
 
 .row .pagination > ul {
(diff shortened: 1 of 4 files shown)


hanakin:
!set Style

hanakin:
please read over this https://area51.phpbb.com/phpBB/viewtopic.php?f=99&t=46079 in order to add the style label

sdschulze:
!set Styles

phpBB/styles/prosilver/theme/common.css
```diff
@@ -886,7 +886,6 @@ fieldset.fields1 dl.pmlist dd.recipients {
 /* Pagination in viewforum for multipage topics */
 .row .pagination {
 	display: block;
-	margin-top: -12px;
```
hanakin: this should maybe be margin 3px? to avoid collision on mobile

hanakin: perhaps even both?

sdschulze: On mobile, we currently have 4px top and bottom margin, but I would propose to remove the case distinction (see other commit).  Whether or not to have a margin at all is a matter of taste; so far, prosilver has done without one.

hanakin: no the issue is that there is collision with this implementation with both the div above and the bottom of the row so a top and bottom margin is required. but it should be just applied generally instead of on mobile at 3px. try it out an see...

sdschulze: I don't see a collision on mobile:

screenshot-area51

On the other hand, things are admittedly a bit tight, so I wouldn't mind a 3px margin (top and bottom), either:

screenshot-area51-wide

vinny:
Could you check if this fix is also needed on mcp_forum.html, ucp_main_bookmarks.html and ucp_main_subscribed.html?

sdschulze:
Pagination is implemented inconsistently in different places.  That is, in ucp_main_bookmarks.html and ucp_main_subscribed.html, things look OK after my changes, but different from viewforum_topic.html and search_results.html.  In mcp_forum.html, the pagination is not generated at all, even in threads where it should be, which seems like an unrelated bug.

For the 3.2.x branch, I would personally leave those files as they are, but for future releases, there should be some clean-up.

hanakin:
Agreed I also would have preferred a modifier class for the small version my self to require less code and cleaner specificity

PR #4925, opened 2017-09-03
ticket/15345 Add links to icons for forums, topics with no unread posts

v12mike:
A patch that gives icons for forums with no unread posts a link to the forum
(same as existing for forums with unread posts),
and gives icons for topics with unread posts a link to the last post in the topic,
which seems to be the natural thing to do as an extension of the existing behaviour 
for topics with unread posts.

PHPBB3-15345

Checklist:

- [*] Correct branch: master for new features; 3.2.x, 3.1.x for fixes
- [*] Tests pass
- [*] Code follows coding guidelines: [master / 3.2.x](https://area51.phpbb.com/docs/master/coding-guidelines.html), [3.1.x](https://area51.phpbb.com/docs/31x/coding-guidelines.html)
- [*] Commit follows commit message [format](https://wiki.phpbb.com/Git#Commit_Messages)

https://tracker.phpbb.com/browse/PHPBB3-15345

### DIFF
--- a/phpBB/styles/prosilver/template/forumlist_body.html
+++ b/phpBB/styles/prosilver/template/forumlist_body.html
@@ -33,7 +33,7 @@
 			<!-- EVENT forumlist_body_forum_row_prepend -->
 			<dl class="row-item {forumrow.FORUM_IMG_STYLE}">
 				<dt title="{forumrow.FORUM_FOLDER_IMG_ALT}">
-					<!-- IF forumrow.S_UNREAD_FORUM --><a href="{forumrow.U_VIEWFORUM}" class="row-item-link"></a><!-- ENDIF -->
+					<a href="{forumrow.U_VIEWFORUM}" class="row-item-link"></a>
 					<div class="list-inner">
 						<!-- IF S_ENABLE_FEEDS and forumrow.S_FEED_ENABLED -->
 							<!--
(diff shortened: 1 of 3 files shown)


v12mike:
Fixed issues with commit message and base branch

vinny:
This has already been discussed that only unread forums and topics should have the clickable icon. #2178

v12mike:
From #2178 " Updated. Based on IRC discussion, only RED icons indicating unread posts are now clickable."

That is not very informative.
Can you tell me the reasons behind that decision?  It seems bizarre to me.

vinny:
Well, didn't attend the discussion, but I agree  with the decision, which it should keep as is.

@VSEphpbb

v12mike:
Can you tell me why you think the current behaviour is better?  It seems inconsistent to me.

vinny:
the current version was proposed exactly to highlight unread items

How does this make phpBB better? Why is this necessary?

v12mike:
On a small screen, it can be awkward to get to the last post of a multi-page topic, because the pagination icons are rather small and close together.

I tend to use that icon as a link to take me to the most interesting post in a topic, which is either the the first unread post or the last post, which is where I should be if I want to make a reply in the topic.

I am not sure how having links on icons "highlights" anything, and I don't see how not having a link provides any advantage.

vinny:
> On a small screen, it can be awkward to get to the last post of a multi-page topic, because the pagination icons are rather small and close together.
> 
> I tend to use that icon as a link to take me to the most interesting post in a topic, which is either the the first unread post or the last post, which is where I should be if I want to make a reply in the topic.

On small screens you have options to go to the first and last post quickly, not just by pagination

v12mike:
How, from the topics list page do I get to the last post in a topic without using the pagination icons?

CHItA:
@phpbb/development-team opinions?

DavidIQ:
I think @hanakin might have also been involved/instrumental in that change.

hanakin:
the original discussion was what do you link to if read. Still the latest?

PR #4933, opened 2017-09-08
[ticket/15342] Track storage files

rubencm:
Checklist:

- [x] Tests pass

Depends on #4916, also need to be updated after #4899 

https://tracker.phpbb.com/browse/PHPBB3-15342

### DIFF
--- a/phpBB/phpbb/attachment/upload.php
+++ b/phpBB/phpbb/attachment/upload.php
@@ -199,6 +199,7 @@ class upload
 		// Check for attachment quota and free space
 		if (!$this->check_attach_quota() || !$this->check_disk_space())
 		{
+			$this->file->remove($this->storage);
 			return $this->file_data;
 		}
 
@@ -321,8 +322,6 @@ class upload
 				$this->file_data['error'][] = $this->language->lang('ATTACH_QUOTA_REACHED');
 				$this->file_data['post_attach'] = false;
 
-				$this->file->remove($this->storage);
-
 				return false;
 			}
 		}
@@ -352,8 +351,6 @@ class upload
 					$this->file_data['error'][] = $this->language->lang('ATTACH_QUOTA_REACHED');
 				}
 				$this->file_data['post_attach'] = false;
-
-				$this->file->remove($this->storage);
 
 				return false;
 			}
--- a/phpBB/phpbb/attachment/upload.php
+++ b/phpBB/phpbb/attachment/upload.php
@@ -336,28 +336,21 @@ class upload
 	 */
 	protected function check_disk_space()
 	{
-		try
-		{
-			$free_space = $this->storage->free_space();
+		$free_space = $this->storage->free_space();
 
-			if ($free_space <= $this->file->get('filesize'))
+		if ($free_space !== false && $free_space <= $this->file->get('filesize'))
+		{
+			if ($this->auth->acl_get('a_'))
 			{
-				if ($this->auth->acl_get('a_'))
-				{
-					$this->file_data['error'][] = $this->language->lang('ATTACH_DISK_FULL');
-				}
-				else
-				{
-					$this->file_data['error'][] = $this->language->lang('ATTACH_QUOTA_REACHED');
-				}
-				$this->file_data['post_attach'] = false;
-
-				return false;
+				$this->file_data['error'][] = $this->language->lang('ATTACH_DISK_FULL');
 			}
-		}
-		catch (\phpbb\storage\exception\exception $e)
-		{
-			// Nothing
+			else
+			{
+				$this->file_data['error'][] = $this->language->lang('ATTACH_QUOTA_REACHED');
+			}
+			$this->file_data['post_attach'] = false;
+
+			return false;
 		}
 
 		return true;
--- a/phpBB/phpbb/attachment/upload.php
+++ b/phpBB/phpbb/attachment/upload.php
@@ -327,10 +327,36 @@ class upload
 	/**
 	 * Check if there is enough free space available on disk
 	 *
-	 * @return bool True if disk space is available, false if not
+	 * @return bool True if disk space is available or can't get it, false if not
 	 */
 	protected function check_disk_space()
 	{
+		try
+		{
+			$free_space = $this->storage->free_space();
+
+			if ($free_space <= $this->file->get('filesize'))
+			{
+				if ($this->auth->acl_get('a_'))
+				{
+					$this->file_data['error'][] = $this->language->lang('ATTACH_DISK_FULL');
+				}
+				else
+				{
+					$this->file_data['error'][] = $this->language->lang('ATTACH_QUOTA_REACHED');
+				}
+				$this->file_data['post_attach'] = false;
+
+				$this->file->remove($this->storage);
+
+				return false;
+			}
+		}
+		catch (\phpbb\storage\exception\exception $e)
+		{
+			// Nothing
+		}
+
 		return true;
 	}
 
--- a/phpBB/phpbb/attachment/upload.php
+++ b/phpBB/phpbb/attachment/upload.php
@@ -246,8 +246,14 @@ class upload
 			{
 				// Move the thumbnail from temp folder to the storage
 				$fp = fopen($destination, 'rb');
+
 				$this->storage->write_stream($destination_name, $fp);
-				fclose($fp);
+
+				if (is_resource($fp))
+				{
+					fclose($fp);
+				}
+
 				$this->storage->track_file($destination_name);
 			}
 			else
--- a/phpBB/phpbb/attachment/upload.php
+++ b/phpBB/phpbb/attachment/upload.php
@@ -336,21 +336,29 @@ class upload
 	 */
 	protected function check_disk_space()
 	{
-		$free_space = $this->storage->free_space();
-
-		if ($free_space !== false && $free_space <= $this->file->get('filesize'))
+		try
 		{
-			if ($this->auth->acl_get('a_'))
-			{
-				$this->file_data['error'][] = $this->language->lang('ATTACH_DISK_FULL');
-			}
-			else
-			{
-				$this->file_data['error'][] = $this->language->lang('ATTACH_QUOTA_REACHED');
-			}
-			$this->file_data['post_attach'] = false;
+			$free_space = $this->storage->free_space();
 
-			return false;
+			if ($free_space <= $this->file->get('filesize'))
+			{
+				if ($this->auth->acl_get('a_'))
+				{
+					$this->file_data['error'][] = $this->language->lang('ATTACH_DISK_FULL');
+				}
+				else
+				{
+					$this->file_data['error'][] = $this->language->lang('ATTACH_QUOTA_REACHED');
+				}
+
+				$this->file_data['post_attach'] = false;
+
+				return false;
+			}
+		}
+		catch (\phpbb\storage\exception\exception $e)
+		{
+			// Do nothing
 		}
 
 		return true;
--- a/phpBB/phpbb/attachment/upload.php
+++ b/phpBB/phpbb/attachment/upload.php
@@ -248,6 +248,7 @@ class upload
 				$fp = fopen($destination, 'rb');
 				$this->storage->write_stream($destination_name, $fp);
 				fclose($fp);
+				$this->storage->track_file($destination_name);
 			}
 			else
 			{
--- a/phpBB/phpbb/attachment/upload.php
+++ b/phpBB/phpbb/attachment/upload.php
@@ -333,7 +333,7 @@ class upload
 	/**
 	 * Check if there is enough free space available on disk
 	 *
-	 * @return bool True if disk space is available or can't get it, false if not
+	 * @return bool True if disk space is available or not limited, false if not
 	 */
 	protected function check_disk_space()
 	{
--- a/phpBB/phpbb/db/migration/data/v330/storage_track.php
+++ b/phpBB/phpbb/db/migration/data/v330/storage_track.php
@@ -40,4 +40,28 @@ class storage_track extends \phpbb\db\migration\migration
 			),
 		);
 	}
+
+	public function update_data()
+	{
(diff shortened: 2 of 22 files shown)


phpBB/includes/acp/acp_storage.php
```diff
@@ -0,0 +1,267 @@
...
+		$this->config = $phpbb_container->get('config');
+		$this->lang = $phpbb_container->get('language');
+		$this->request = $phpbb_container->get('request');
+		$this->template = $phpbb_container->get('template');
+		$this->user = $phpbb_container->get('user');
+		$this->provider_collection = $phpbb_container->get('storage.provider_collection');
+		$this->storage_collection = $phpbb_container->get('storage.storage_collection');
+
+		// Add necesary language files
+		$this->lang->add_lang(['common', 'acp/storage']);
```
Elsensee: Here, also no common is necessary.

rubencm: This is fixed, i have to rebase

Elsensee: Ah, alright.

rubencm:
!set WIP

Nicofuma:
Hey @rubencm could you rebase this PR please?
Otherwise, do you remember what is its status? (what you have done here and what you are planning to add);
Thanks

rubencm:
Done. Yes, i added 2 checkboxes to remember what i have to do next.

rubencm:
!set WIP :construction:

marc1706:
Can you try to fix the tests for this?

rubencm:
Probably yes, but i left it because database is involved, and i don't know how to do that tests

phpBB/phpbb/db/migration/data/v330/storage_track.php
```diff
@@ -0,0 +1,120 @@
...
+			['custom', [[$this, 'track_avatars']]],
+			['custom', [[$this, 'track_attachments']]],
+			['custom', [[$this, 'track_backups']]],
+		];
+	}
+
+	public function track_avatars()
+	{
+		/** @var storage $storage */
+		$storage = $this->container->get('storage.avatar');
```
rubencm: This services don't exist on installer environment, and i can't add them because they depend of dbal.conn that also is unavailable there

rubencm:
!unset WIP

rubencm:
!unset WIP :construction:

phpBB/phpbb/attachment/upload.php
```diff
@@ -326,10 +327,36 @@ protected function check_attach_quota()
 	/**
 	 * Check if there is enough free space available on disk
 	 *
-	 * @return bool True if disk space is available, false if not
+	 * @return bool True if disk space is available or can't get it, false if not
```
marc1706: What do you mean with "can't get it"?

rubencm: If the available space method is not available.
For example AWS has no limit, so the method won't be available and will always allow to upload files.

rubencm: In other words, if you can't get the free space, the file will be uploaded

rubencm: Here is the same https://github.com/phpbb/phpbb/blob/3.2.x/phpBB/phpbb/attachment/upload.php#L299
if disk_free_space fails, the file will be uploaded

phpBB/phpbb/storage/adapter/adapter_interface.php
```diff
@@ -95,4 +95,13 @@ public function copy($path_orig, $path_dest);
 	 *
 	 */
 	public function get_link($path);
+
+	/*
+	 * Get space available in bytes.
+	 *
+	 * @throws \phpbb\storage\exception\exception		When can't get available space
```
marc1706: `When unable to retrieve available storage space`

phpBB/phpbb/storage/adapter/local.php
```diff
@@ -424,4 +424,23 @@ public function get_link($path)
 	{
 		return generate_board_url() . $this->path . $path;
 	}
+
+	/*
+	 * Get space available in bytes.
+	 *
+	 * @throws \phpbb\storage\exception\exception		When can't get available space
```
marc1706: `When unable to retrieve available storage space`

phpBB/phpbb/storage/storage.php
```diff
@@ -203,6 +235,56 @@ public function write_stream($path, $resource)
 		}
 	}
 
+	public function track_file($path, $update = false)
```
marc1706: Please add a docblock (even if it's just with inheritdoc).

phpBB/phpbb/storage/storage.php
```diff
@@ -203,6 +235,56 @@ public function write_stream($path, $resource)
...
+				SET filesize = ' . $file->size . '
+				WHERE ' . $this->db->sql_build_array('SELECT', $sql_ary);
+			$this->db->sql_query($sql);
+		}
+
+		$this->cache->destroy('_storage_' . $this->get_name() . '_totalsize');
+		$this->cache->destroy('_storage_' . $this->get_name() . '_numfiles');
+	}
+
+	public function untrack_file($path)
```
marc1706: Same here.

phpBB/phpbb/storage/storage.php
```diff
@@ -136,6 +163,8 @@ public function delete($path)
 	public function rename($path_orig, $path_dest)
 	{
 		$this->get_adapter()->rename($path_orig, $path_dest);
+		$this->untrack_file($path_orig);
```
marc1706: Sounds a bit inefficient to me. Could we maybe add another method like `update_file($path_orig, $path_dest)`?

rubencm: Yes

phpBB/phpbb/storage/storage.php
```diff
@@ -203,6 +234,82 @@ public function write_stream($path, $resource)
...
+	 * Rename tracked file.
+	 *
+	 * @param string	$path_orig	The original file/direcotry
+	 * @param string	$path_dest	The target file/directory
+	 */
+	protected function track_rename($path_orig, $path_dest)
+	{
+		$sql = 'UPDATE ' . $this->storage_table . "
+			SET file_path = '" . $path_dest . "'
+			WHERE file_path = '" . $path_orig . "'
```
gooof: $this->db->sql_escape() for both

phpBB/phpbb/storage/adapter/adapter_interface.php
```diff
@@ -99,7 +99,7 @@ public function get_link($path);
 	/*
 	 * Get space available in bytes.
 	 *
-	 * @throws \phpbb\storage\exception\exception		When can't get available space
+	 * @throws \phpbb\storage\exception\exception		When unable to retrieve available storage spac
```
marc1706: space*

phpBB/phpbb/attachment/upload.php
marc1706: Maybe then change it to say something like:
`True if disk space is available or not limited, false if not`

phpBB/phpbb/attachment/upload.php
```diff
@@ -326,10 +333,36 @@ protected function check_attach_quota()
...
+				{
+					$this->file_data['error'][] = $this->language->lang('ATTACH_DISK_FULL');
+				}
+				else
+				{
+					$this->file_data['error'][] = $this->language->lang('ATTACH_QUOTA_REACHED');
+				}
+				$this->file_data['post_attach'] = false;
+
+				$this->file->remove($this->storage);
```
Nicofuma: not sure it should be removed here and not in the calling method

phpBB/phpbb/storage/adapter/adapter_interface.php
```diff
@@ -95,4 +95,13 @@ public function copy($path_orig, $path_dest);
...
 	public function get_link($path);
+
+	/*
+	 * Get space available in bytes.
+	 *
+	 * @throws \phpbb\storage\exception\exception		When unable to retrieve available storage space
+	 *
+	 * @return int	Returns available space
+	 */
+	public function free_space();
```
Nicofuma: I think there is some cases where you cannot estimate the free space (because you have no limit per example).
How should be handle this case? throw an exception ? return PHP_INT_MAX ?

phpBB/phpbb/storage/storage.php
```diff
@@ -229,4 +336,67 @@ public function get_link($path)
...
+				WHERE storage = '" . $this->db->sql_escape($this->get_name()) . "'";
+			$result = $this->db->sql_query($sql);
+
+			$number_files = (int) $this->db->sql_fetchfield('numfiles');
+			$this->cache->put('_storage_' . $this->get_name() . '_numfiles', $number_files);
+
+			$this->db->sql_freeresult($result);
+		}
+
+		return $number_files;
```
Nicofuma: should be casted as int aswell I guess

Nicofuma:
looks pretty good, would be awesome if it could be tested though :)

phpBB/phpbb/storage/adapter/adapter_interface.php
rubencm: With an exception

rubencm: It can be a problem here https://github.com/phpbb/phpbb/pull/5243
Is there any way to catch an exception with twig?

Nicofuma: not really :/

Nicofuma: we could imagine to extend twig for that but I don't think it is a good idea
maybe we could add a parameter to free_space() to not throw but.... don't really like it tbh.
Maybe you will have another idea? or @marc1706 ?

rubencm:
Uhm, maybe in the future xD

phpBB/phpbb/files/filespec_storage.php
```diff
@@ -288,6 +288,10 @@ public function remove($storage)
 		{
 			$storage->delete($this->destination_file);
 		}
+		else
+		{
+			@unlink($this->filename);
```
rubencm: @Nicofuma Check this, i am not sure about it, becuase this wasn't on the original filespec.php

Nicofuma: when will we hit this case?
I don't think it is harmeful

phpBB/phpbb/attachment/upload.php
rubencm: Here

phpBB/phpbb/files/filespec_storage.php
rubencm: Ok, i will remove it, it makes the tests fail

phpBB/phpbb/storage/adapter/adapter_interface.php
marc1706: Yeah, I'd rather not hack exception treatment into twig. If there is no limit I'd rather return PHP_INT_MAX and maybe handle this accordingly before template output. An exception is IMHO not the correct tool to use here as this is not an anomalous condition but rather an expected side effect of the used storage adapter.

rubencm: I prefer it like this, or return -1 or something, because i have to detect if isn't available from twig, if not it will show a few petabytes available.
Also i am looking and php_int_max is only 2gb

rubencm: PHP_INT_MAX normally is 2147483647 (int 32 bits systems, but much bigger in 64 bits systems)

phpBB/phpbb/storage/storage.php
```diff
@@ -183,10 +212,12 @@ public function read_stream($path)
 
 	/**
 	 * Writes a new file using a stream.
+	 * You have to track file after use this method
```
marc1706: Change to: `The file needs to be tracked after using this method`

phpBB/phpbb/storage/storage.php
```diff
@@ -203,6 +234,82 @@ public function write_stream($path, $resource)
 		}
 	}
 
+	/**
+	 * Track file into database.
```
marc1706: change to: `Track file in database`

phpBB/phpbb/storage/storage.php
```diff
@@ -203,6 +234,82 @@ public function write_stream($path, $resource)
...
+				WHERE ' . $this->db->sql_build_array('SELECT', $sql_ary);
+			$this->db->sql_query($sql);
+		}
+
+		$this->cache->destroy('_storage_' . $this->get_name() . '_totalsize');
+		$this->cache->destroy('_storage_' . $this->get_name() . '_numfiles');
+	}
+
+	/**
+	 * Untrack file.
```
marc1706: No need for a full stop

phpBB/phpbb/storage/storage.php
```diff
@@ -203,6 +234,82 @@ public function write_stream($path, $resource)
...
+		$sql = 'DELETE FROM ' . $this->storage_table . '
+			WHERE ' . $this->db->sql_build_array('DELETE', $sql_ary);
+		$this->db->sql_query($sql);
+
+		$this->cache->destroy('_storage_' . $this->get_name() . '_totalsize');
+		$this->cache->destroy('_storage_' . $this->get_name() . '_numfiles');
+	}
+
+	/**
+	 * Rename tracked file.
```
marc1706: Same here, please remove the full stop

phpBB/phpbb/storage/storage.php
```diff
@@ -229,4 +336,66 @@ public function get_link($path)
 	{
 		return $this->get_adapter()->get_link($path);
 	}
+
+	/**
+	 * Get total storage size.
```
marc1706: In other comments as well :)
I don't think we use full stops in other files so using it seems a bit out of place

PR #4943, opened 2017-09-09
[ticket/14933] Make "route not found" more user-friendly

Elsensee:
Checklist:

- [x] Tests pass

Tracker ticket:
https://tracker.phpbb.com/browse/PHPBB3-14933

### DIFF
--- a/phpBB/config/default/container/services_event.yml
+++ b/phpBB/config/default/container/services_event.yml
@@ -9,6 +9,7 @@ services:
         arguments:
             - '@template'
             - '@language'
+            - '%?debug.exceptions'
         tags:
             - { name: kernel.event_subscriber }
 
--- a/phpBB/config/default/container/services_event.yml
+++ b/phpBB/config/default/container/services_event.yml
@@ -9,7 +9,7 @@ services:
         arguments:
             - '@template'
             - '@language'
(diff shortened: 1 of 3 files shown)


phpBB/config/default/container/services_event.yml
```diff
@@ -9,6 +9,7 @@ services:
         arguments:
             - '@template'
             - '@language'
+            - '%?debug.exceptions'
```
Nicofuma: what is this notation?

Elsensee: I used this because I thought `debug.exceptions` would not be available all the time, so to avoid errors when compiling the container I used this notation which should say that the argument is optional (which it is, basically)

Elsensee:
@phpbb/development-team I wonder if this is the correct way to fix the "issue".

I also wonder whether this is the right branch for this.

phpBB/config/default/container/services_event.yml
Nicofuma: I think you forgot a % at the end and it will always be defined, no matter what

Elsensee: Oh, you are correct.

PR #4952, opened 2017-09-10
[ticket/15191] Purge cache directory even on dummy driver

Elsensee:
Checklist:

- [ ] Tests pass

Tracker ticket:
https://tracker.phpbb.com/browse/PHPBB3-15191

### DIFF
--- a/phpBB/phpbb/cache/driver/dummy.php
+++ b/phpBB/phpbb/cache/driver/dummy.php
@@ -76,13 +76,6 @@ class dummy extends \phpbb\cache\driver\base
 	/**
 	* {@inheritDoc}
 	*/
-	function purge()
-	{
-	}
-
-	/**
-	* {@inheritDoc}
-	*/
 	function destroy($var_name, $table = '')
 	{
 	}


ErnadoO:
Hello

This PR doesn't fix the ticket issue, because somes files are never removed:

> else if (strpos($filename, 'container_') === 0 ||
> 				strpos($filename, 'autoload_') === 0 ||
> 				strpos($filename, 'url_matcher') === 0 ||
> 				strpos($filename, 'url_generator') === 0 ||
> 				strpos($filename, 'sql_') === 0  ||
> 				strpos($filename, 'data_') === 0)
> 			{
> 				$this->remove_file($fileInfo->getPathname());
> 			}
in base::purge() :

rxu:
```cache_dir``` is never being initialized for this driver, so there's nothing to purge anyway.

ErnadoO:
Yes but some classes doesn't care about which driver cache is used and write files directly in the cache directory.

I's the case for container_builder():
https://github.com/phpbb/phpbb/blob/b1e37871978702910fbe75acc019268c44bb026b/phpBB/phpbb/di/container_builder.php#L480

router():
https://github.com/phpbb/phpbb/blob/b1e37871978702910fbe75acc019268c44bb026b/phpBB/phpbb/routing/router.php#L213
https://github.com/phpbb/phpbb/blob/b1e37871978702910fbe75acc019268c44bb026b/phpBB/phpbb/routing/router.php#L268


I think these files should never be cached with dummy driver (it's the goal of this driver no ?)

rxu:
> it's the goal of this driver no ?

It should be, as it's previous name is ```null``` 😃

Nicofuma:
Actually these files are not part of the cache managed by these classes.
In phpBB there is 2 different kind of cache:
 - The first one is used to cache data (from the database, an external API, etc. like the list of available styles, the acl...)
 - The second one is used to store generated PHP files for some component that either have a far more important impact on the board performance or just can't work without these files. 

In the future we should better handle that, but for now this PR is acceptable. We just need to ensure that the dummy driver will never fail (even if it can't remove a file)

marc1706:
@rxu is correct. The cache dir is not set so this change will not result in the cache directory being emptied. Therefore I don't see a point in this PR.

PR #4953, opened 2017-09-10
[ticket/15358] Remove resetting the board's start date

Elsensee:
Checklist:

- [x] Tests pass

Tracker ticket:
https://tracker.phpbb.com/browse/PHPBB3-15358

### DIFF
--- a/phpBB/adm/style/acp_main.html
+++ b/phpBB/adm/style/acp_main.html
@@ -214,13 +214,6 @@
 				</dl>
 			</form>
 
-			<form id="action_date_form" method="post" action="{U_ACTION}" data-ajax="true">
-				<dl>
-					<dt><label for="action_date">{L_RESET_DATE}</label><br /><span>&nbsp;</span></dt>
-					<dd><input type="hidden" name="action" value="date" /><input class="button2" type="submit" id="action_date" name="action_date" value="{L_RUN}" /></dd>
-				</dl>
-			</form>
-
 			<form id="action_stats_form" method="post" action="{U_ACTION}">
 				<dl>
 					<dt><label for="action_stats">{L_RESYNC_STATS}</label><br /><span>{L_RESYNC_STATS_EXPLAIN}</span></dt>
--- a/phpBB/includes/acp/acp_main.php
+++ b/phpBB/includes/acp/acp_main.php
@@ -79,10 +79,6 @@ class acp_main
 						$confirm = true;
 						$confirm_lang = 'RESYNC_POSTCOUNTS_CONFIRM';
 					break;
-					case 'date':
-						$confirm = true;
-						$confirm_lang = 'RESET_DATE_CONFIRM';
-					break;
 					case 'db_track':
 						$confirm = true;
 						$confirm_lang = 'RESYNC_POST_MARKING_CONFIRM';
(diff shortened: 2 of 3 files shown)


Louis7777:
I kinda liked that feature...

Elsensee:
For what reason? Please tell me the use cases.

Louis7777:
Sometimes a community may have taken months to go live, the website project may have been abandoned and continued later, so when it finally kicks off and staff is being added etc. I would like them to see an earlier start date rather than two years in the past.

It's not really important and I won't be sad without it, but just saying.

Elsensee:
@Louis7777 I see where you are going but the only thing that is influenced by the board's start date is the posts per day since board start which is displayed only in the ACP.

vinny:
[common.php#L754](https://github.com/phpbb/phpbb/blob/master/phpBB/language/en/acp/common.php#L754)

3D-I:
Nice job. That could be taken care by an extension, which I already coded 1 year ago.

Elsensee:
@vinny To be really honest... I don't know about that. On the one hand, this should be removed as well, on the other hand however, removing this would render the log data weird, if people have already reset their board's start date.
So.. I don't really know.

rxu:
To be honest, it would be better to add the features rather than removing them in common 😉

3D-I:
Bah, at the end of all I wouldn't remove it. Let it be.
There are things more important like this one at the moment IMHO.

marc1706:
In general we should ask ourselves if the features we have are used and how often they are used. More often than not this will not be used, in some cases it might be used once, and in the most rare cases this will be used more than once.
This is IMHO a perfect example of a useless feature as one can simply run a query if one wants to change the board start date. I'd say we should remove it for 3.3 and leave it in 3.2.

Elsensee:
> It might be used once or twice for the lifetime of a board, yes.

Which makes it a perfect candidate for an extension. It's not even that hard to make, actually. Sounds like ~10 lines of actual code and there you go.

> P.S.: I don't get this trend of removing features. Not popular =/= useless. At another PR someone wants to eliminate smilies (whire ARE still very popular) and at another one someone wants to remove IP banning. Plus I don't think that phpBB is bloated. On the contrary, it still lacks a few popular features.

I know, it's kinda weird to *remove* features instead of adding them, especially since I also believe that phpBB is not bloated at all. However, I believe that we need a good base to build new features on, so we go through the legacy part (which is easy to spot, since it's the part *not* in phpbb/) and make the code object-orientated. It's basically a rewrite of the entire code but only bit for bit. (No pun intended!) In that process we might come across features that we feel are not modern anymore and which we'd like to remove, as it saves time rewriting and debugging that part - especially if there is not enough use cases for it. (I don't know about the smilies one, though...)

However, this is about the ACP. If people have a bad experience with it because it's too messy, this will make them not use phpBB. Users don't really care, it's the admins that setup phpBB, so we have to make it as user-friendly as possible. Sure, one setting less won't do much, but somebody has to start somewhere. I guess this will be tackled in full when the module system will be dropped and the ACP will move to a more modern system, but for now..

gooof:
Move it to the maintenance or system tab... admins should never run sql queries or edit the database.

marc1706:
We might re-evaluate this when refactoring the ACP but I think it's best to close this for now.

PR #4962, opened 2017-09-17
[ticket/14972] replace all occurrences of sizeof() with the count()

rubencm:
PHPBB3-14972

Checklist:

- [x] Tests pass

Parent: #4862 

https://tracker.phpbb.com/browse/PHPBB3-14972

### DIFF
--- a/phpBB/phpbb/files/filespec_storage.php
+++ b/phpBB/phpbb/files/filespec_storage.php
@@ -111,7 +111,7 @@ class filespec_storage
 	 */
 	public function set_upload_ary($upload_ary)
 	{
-		if (!isset($upload_ary) || !sizeof($upload_ary))
+		if (!isset($upload_ary) || !count($upload_ary))
 		{
 			return $this;
 		}
@@ -389,7 +389,7 @@ class filespec_storage
 	 */
 	public function move_file($storage, $overwrite = false, $skip_image_check = false)
 	{
-		if (sizeof($this->error))
+		if (count($this->error))
 		{
 			return false;
 		}
@@ -460,7 +460,7 @@ class filespec_storage
 		// Remove temporary filename
 		@unlink($this->filename);
 
-		if (sizeof($this->error))
+		if (count($this->error))
 		{
 			return false;
 		}
--- a/phpBB/phpbb/files/types/form_storage.php
+++ b/phpBB/phpbb/files/types/form_storage.php
@@ -109,7 +109,7 @@ class form_storage extends base
 
 		// PHP Upload file size check
 		$file = $this->check_upload_size($file);
-		if (sizeof($file->error))
+		if (count($file->error))
 		{
 			return $file;
 		}


phpBB/phpbb/files/filespec_storage.php
```diff
@@ -111,7 +111,7 @@ public function __construct(language $language, \bantu\IniGetWrapper\IniGetWrapp
 	 */
 	public function set_upload_ary($upload_ary)
 	{
-		if (!isset($upload_ary) || !sizeof($upload_ary))
+		if (!isset($upload_ary) || !count($upload_ary))
```
Elsensee: Can actually be simplified to just `empty($upload_ary)`...

phpBB/phpbb/files/filespec_storage.php
```diff
@@ -389,7 +389,7 @@ public function check_content($disallowed_content)
 	 */
 	public function move_file($storage, $overwrite = false, $skip_image_check = false)
 	{
-		if (sizeof($this->error))
+		if (count($this->error))
```
Elsensee: Can be simplified to `!empty($this->error)` I believe.

phpBB/phpbb/files/filespec_storage.php
```diff
@@ -460,7 +460,7 @@ public function move_file($storage, $overwrite = false, $skip_image_check = fals
 		// Remove temporary filename
 		@unlink($this->filename);
 
-		if (sizeof($this->error))
+		if (count($this->error))
```
Elsensee: See above.

phpBB/phpbb/files/types/form_storage.php
```diff
@@ -109,7 +109,7 @@ protected function form_upload($form_name)
 
 		// PHP Upload file size check
 		$file = $this->check_upload_size($file);
-		if (sizeof($file->error))
+		if (count($file->error))
```
Elsensee: See above.

phpBB/phpbb/files/filespec_storage.php
rubencm: You should have told this to @rxu in his PR, i am doing here what he did in filespec.php, because this was desynchronized and was causing tests to fail

rubencm: @Elsensee #4862

rxu: The logic is simple. I was replacing ```sizeof()``` with ```empty()``` for non-countable objects only. For countable ones (arrays) it was straight ```sizeof()``` -> ```count()``` replacement.

Elsensee: Why don't you use `empty()` for them too? I think it's faster than `count` or `sizeof` for some arrays?

kasimi: If `$array` is already defined, why use `if (empty($array))`? Why not `if (!$array)`?

rxu: >Why don't you use empty() for them too?

Simply because ```sizeof()``` is an alias of ```count()```, so I just kept the same function call where possible to avoid accidental logic changes as far as possible.

Derky:
@Elsensee agreed to merge it. 🙂 (i.e. not replacing it with empty right now)

PR #4968, opened 2017-09-21
[ticket/15371] Split uploaded files into subdirectories

rubencm:
PHPBB3-15371

Checklist:

- [x] Tests pass

Depends on #4916 #4899 

https://tracker.phpbb.com/browse/PHPBB3-15371

### DIFF
--- a/phpBB/phpbb/storage/adapter/local.php
+++ b/phpBB/phpbb/storage/adapter/local.php
@@ -257,7 +257,7 @@ class local implements adapter_interface, stream_interface
 		$parts = str_split($hash, 2);
 		$parts = array_slice($parts, 0, $this->dir_depth);
 
-		return implode(DIRECTORY_SEPARATOR, $parts) . DIRECTORY_SEPARATOR;
+		return (!empty($parts)) ? implode(DIRECTORY_SEPARATOR, $parts) . DIRECTORY_SEPARATOR : '';
 	}
 
 	/**
--- a/phpBB/phpbb/storage/adapter/local.php
+++ b/phpBB/phpbb/storage/adapter/local.php
@@ -118,7 +118,7 @@ class local implements adapter_interface, stream_interface
 
 		try
 		{
-			$this->filesystem->dump_file($this->root_path . $this->get_filepath($path), $content);
+			$this->filesystem->dump_file($this->root_path . $this->get_path($path) . $this->get_filename($path), $content);
 		}
 		catch (filesystem_exception $e)
 		{
@@ -136,7 +136,7 @@ class local implements adapter_interface, stream_interface
 			throw new exception('STORAGE_FILE_NO_EXIST', $path);
 		}
 
-		$content = @file_get_contents($this->root_path . $this->get_filepath($path));
+		$content = @file_get_contents($this->root_path . $this->get_path($path) . $this->get_filename($path));
 
 		if ($content === false)
 		{
@@ -151,7 +151,7 @@ class local implements adapter_interface, stream_interface
 	 */
 	public function exists($path)
 	{
-		return $this->filesystem->exists($this->root_path . $this->get_filepath($path));
+		return $this->filesystem->exists($this->root_path . $this->get_path($path) . $this->get_filename($path));
 	}
 
 	/**
@@ -161,7 +161,7 @@ class local implements adapter_interface, stream_interface
 	{
 		try
 		{
-			$this->filesystem->remove($this->root_path . $this->get_filepath($path));
+			$this->filesystem->remove($this->root_path . $this->get_path($path) . $this->get_filename($path));
 		}
 		catch (filesystem_exception $e)
 		{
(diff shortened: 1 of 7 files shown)


rubencm:
!set WIP

rubencm:
I still need to know how to handle things like this:
`$storage->create_dir('my/dir');
$storage->put_contents('my/dir/file.txt', '');`

Nicofuma:
imo you should consider that you do whatever you want with these directories.
For me it is something logical, an internal path.

In this specific case and considering how it currently work, I would keep the directories as is and adds the subdirectories inside it (`my/dir/a0/1f/file.txt`)

phpBB/phpbb/storage/adapter/local.php
```diff
@@ -57,6 +57,11 @@ class local implements adapter_interface, stream_interface
 	 */
 	protected $root_path;
 
+	/**
+	 * @var int dir_depth
```
marc1706: Can you add a bit more info on what this member variable does or represents?

phpBB/phpbb/storage/adapter/local.php
```diff
@@ -138,12 +145,14 @@ public function delete($path)
...
 		{
-			$this->filesystem->remove($this->root_path . $path);
+			$this->filesystem->remove($this->root_path . $this->get_filepath($path));
 		}
 		catch (filesystem_exception $e)
 		{
 			throw new exception('STORAGE_CANNOT_DELETE', $path, array(), $e);
 		}
+
+		$this->remove_empty_dirs($path);
```
marc1706: I wonder if it might make sense to have this done via cron. Thoughts? (as in: let cron clean up paths every x hours)

phpBB/phpbb/storage/adapter/local.php
```diff
@@ -215,12 +226,54 @@ protected function ensure_directory_exists($path)
...
+		do
+		{
+			$parts = explode(DIRECTORY_SEPARATOR, $path);
+			$parts = array_slice($parts, 0, -1);
+			$path = implode(DIRECTORY_SEPARATOR, $parts);
+		}
+		while ($path && @rmdir($dirpath . $path));
+	}
+
+	protected function get_filepath($path)
```
marc1706: Please add a docblock for this.

phpBB/phpbb/storage/adapter/local.php
rubencm: No, i don't think so, that can't be optimal when you have lot of files/dirs. It tries do remove empty directories from down to top while they are empty. Example:

rm /ab/cd/ef/file.jpg
rm /ab/cd/ef
rm /ab/cd
rm /ab

At least it should, i haven't tried it.

rubencm:
!unset WIP:construction:

rubencm:
!unset WIP :construction:

phpBB/phpbb/storage/adapter/local.php
```diff
@@ -62,6 +62,22 @@ class local implements adapter_interface, stream_interface
 	 */
 	protected $path;
 
+	/*
+	 * Subdirectories depth
+	 *
+	 * Instead of storing all folders on the same directory, they can be divided
```
marc1706: in the same

phpBB/phpbb/storage/adapter/local.php
```diff
@@ -62,6 +62,22 @@ class local implements adapter_interface, stream_interface
...
+	 * into smaller directories. This variable means the number of subdirectories of
```
marc1706: variable describes the number of

phpBB/phpbb/storage/adapter/local.php
```diff
@@ -62,6 +62,22 @@ class local implements adapter_interface, stream_interface
...
+	 * depth to store the file. For example:
+	 * depth = 0 -> /images/avatars/upload/my_avatar.jpg
+	 * depth = 2 -> /images/avatars/upload/d9/8c/my_avatar.jpg
+	 * This is for those who have problems storing a large number of files on
```
marc1706: of files in a single

phpBB/phpbb/storage/adapter/local.php
```diff
@@ -221,12 +242,68 @@ protected function ensure_directory_exists($path)
...
+
+		$hash = md5(basename($path));
+
+		$parts = str_split($hash, 2);
+		$parts = array_slice($parts, 0, $this->dir_depth);
+
+		// Create path
+		$path = $dirname . DIRECTORY_SEPARATOR;
+
+		if (!empty($parts))
```
gooof: if (isset($parts[0]))

rubencm: is not the same?

gooof: Yes, isset is only a little bit faster. And through the md5 ``$parts[0]`` is never empty if exists.

rubencm: I have test and is almost the same, in fact, empty is a little faster

PR #4969, opened 2017-09-22
[ticket/15245] Fix images in feeds when accessing via app.php

Elsensee:
Checklist:

- [x] Tests pass

Tracker ticket:
https://tracker.phpbb.com/browse/PHPBB3-15245

### DIFF
--- a/phpBB/phpbb/feed/helper.php
+++ b/phpBB/phpbb/feed/helper.php
@@ -27,28 +27,18 @@ class helper
 	/** @var \phpbb\user */
 	protected $user;
 
-	/** @var string */
-	protected $phpbb_root_path;
-
-	/** @var string */
-	protected $phpEx;
-
 	/**
 	 * Constructor
 	 *
-	 * @param	\phpbb\config\config	$config				Config object
-	 * @param	\phpbb\path_helper		$path_helper 		Path helper object
-	 * @param	\phpbb\user				$user				User object
-	 * @param	string					$phpbb_root_path	Root path
-	 * @param	string					$phpEx				PHP file extension
+	 * @param	\phpbb\config\config	$config			Config object
+	 * @param	\phpbb\path_helper		$path_helper 	Path helper object
+	 * @param	\phpbb\user				$user			User object
 	 */
-	public function __construct(\phpbb\config\config $config, \phpbb\path_helper $path_helper, \phpbb\user $user, $phpbb_root_path, $phpEx)
+	public function __construct(\phpbb\config\config $config, \phpbb\path_helper $path_helper, \phpbb\user $user)
 	{
 		$this->config = $config;
 		$this->path_helper = $path_helper;
 		$this->user = $user;
-		$this->phpbb_root_path = $phpbb_root_path;
-		$this->phpEx = $phpEx;
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


phpBB/phpbb/feed/helper.php
```diff
@@ -33,14 +36,16 @@ class helper
...
-	 * @param	string	$phpbb_root_path	Root path
-	 * @param	string	$phpEx				PHP file extension
+	 * @param	\phpbb\config\config	$config				Config object
+	 * @param	\phpbb\path_helper		$path_helper 		Path helper object
+	 * @param	\phpbb\user				$user				User object
+	 * @param	string					$phpbb_root_path	Root path
+	 * @param	string					$phpEx				PHP file extension
 	 */
-	public function __construct(\phpbb\config\config $config, \phpbb\user $user, $phpbb_root_path, $phpEx)
+	public function __construct(\phpbb\config\config $config, \phpbb\path_helper $path_helper, \phpbb\user $user, $phpbb_root_path, $phpEx)
```
Derky: You can completely remove $phpbb_root_path from this file now. 🙂

tests/feed/attachments_base_test.php
```diff
@@ -31,13 +31,22 @@ public function setUp()
 
 		$this->filesystem = new \phpbb\filesystem();
 		$config = new \phpbb\config\config(array());
+		$path_helper = new \phpbb\path_helper(
+			new \phpbb\symfony_request(
+				new phpbb_mock_request()
+			),
+			new \phpbb\filesystem\filesystem(),
```
rubencm: Here you can reuse $this->filesystem instead of instance it again

Elsensee: Good point!

phpBB/phpbb/feed/helper.php
```diff
@@ -105,15 +100,15 @@ public function generate_content($content, $uid, $bitfield, $options, $forum_id,
...
+		$content = preg_replace("#\[quote(=&quot;.*?&quot;)?:$uid\]\s*(.*?)\s*\[/quote:$uid\]#si", "[quote$1:$uid]<br />$2<br />[/quote:$uid]", $content);
 
 		$content = generate_text_for_display($content, $uid, $bitfield, $options);
 
 		// Add newlines
 		$content = str_replace('<br />', '<br />' . "\n", $content);
 
 		// Convert smiley Relative paths to Absolute path, Windows style
-		$content = str_replace($this->phpbb_root_path . $this->config['smilies_path'], $this->get_board_url() . '/' . $this->config['smilies_path'], $content);
+		$content = str_replace($this->path_helper->get_web_root_path() . $this->config['smilies_path'], $this->get_board_url() . '/' . $this->config['smilies_path'], $content);
```
JoshyPHP: I'm ~80% sure that you can change the path in the `text_formatter.renderer` service *before* calling `generate_text_for_display()` a few lines above, rather than replace it with a string function.

```php
$renderer->set_smilies_path($this->get_board_url() . '/' . $this->config['smilies_path']);
```

Elsensee: You are technically right. It just wouldn't work for posts with the old bbcodes, would it?

I wish I could tell the `parse_attachments` function to work with the board url, then I wouldn't need the `path_helper` at all...

JoshyPHP: That's correct, messages posted on 3.1 would have the relative path until the reparser catches up. Since the reparser works backwards starting from the newest posts, that should be a small window of time.

Elsensee: I wonder if we can just assume that the reparser - more specifically cron - works as it should on every board out there or if there are misconfigurations that prevent it from working...

Otherwise I totally agree with you.

JoshyPHP: If you think there's a chance the cron doesn't run you can leave the string replacement in and still set the correct path in the renderer. The other string replacements don't really do anything in 3.2 either. For instance, `"<br />"` are now `"<br>\n"` by default.

Elsensee:
Updated.

@JoshyPHP Mind if you check again? I believe I can't get rid of any other string functions?

marc1706:
Confirmed to be working.

PR #4970, opened 2017-09-22
[ticket/15374] Add core event to modify page title in viewforum.php

rxu:
Checklist:

- [x] Tests pass

<a href="https://tracker.phpbb.com/browse/PHPBB3-15374">PHPBB3-15374</a>.

### DIFF
--- a/phpBB/viewforum.php
+++ b/phpBB/viewforum.php
@@ -161,7 +161,22 @@ $phpbb_content_visibility = $phpbb_container->get('content.visibility');
 $topics_count = $phpbb_content_visibility->get_count('forum_topics', $forum_data, $forum_id);
 $start = $pagination->validate_start($start, $config['topics_per_page'], $topics_count);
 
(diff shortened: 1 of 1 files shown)


rxu:
!set Event

DavidIQ:
`core.page_header` already exists in the `page_header` function which should allow you to do this already. Any reason we need to do it specifically on viewforum.php?

rxu:
Yes, but if you use ```core.page_header```, the listener will be called for any page call and not for ```viewforum.php``` only. Having this event, you can save unnecessary listener calls.
Also, there're similar events in other places, f.e. ```core.index_modify_page_title```,  ```core.viewtopic_modify_page_title```.

rxu:
To add, there're no specific variables in that event, like '''start``` for pagination, ```forum_id``` or```forum_data``` which may be needed to generate the title :)

DavidIQ:
Other than the forum_data, forum_id and start variables would be retrievable from the querystring 😝  If we've already propagated events for such a thing to other areas then I guess a precedent/pattern has already been set so... 🤷‍♂️

rxu:
Well, let's say the event can be also useful not only for the page title but for other purposes too, as currently the 1st "root" event in ```viewforum.php``` (the one isn't wrapped by any conditionals) is only around line 428 😛

PR #4971, opened 2017-09-22
[ticket/15339] Use manual method when adding modules

Elsensee:
PR for master: #5017 

This will fix cases where the module is deleted in later
versions and the migrator thus cannot retrieve the info.

Checklist:

- [x] Tests pass

Tracker ticket:
https://tracker.phpbb.com/browse/PHPBB3-15339

### DIFF
--- a/phpBB/phpbb/db/migration/data/v30x/release_3_0_6_rc1.php
+++ b/phpBB/phpbb/db/migration/data/v30x/release_3_0_6_rc1.php
@@ -156,6 +156,7 @@ class release_3_0_6_rc1 extends \phpbb\db\migration\migration
 					'module_langname'	=> 'ACP_FEED_SETTINGS',
 					'module_mode'		=> 'feed',
 					'module_auth'		=> 'acl_a_board',
+					'after'				=> array('signature', 'ACP_SIGNATURE_SETTINGS'),
 				),
 			)),
 			array('module.add', array(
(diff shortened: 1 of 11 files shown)


Elsensee:
This basically reverts https://tracker.phpbb.com/browse/PHPBB3-14703 to get back to the old intended behaviour from 3.0

VSEphpbb:
@marc1706 @Elsensee 
The changes made here seem to have broken the module exists function. Before if `$parent` was false/0, then the module would still be checked to see if it exists. Now calls to module.exists with a false/0 parent will **always** return false. 

The result of these changes is that a lot of migrations that use `module.exists` and `module.remove` will now fail if they look like:

```
array('module.exists', array('acp', false, 'ACP_MODULE_NAME')),
```

Also the fact it just return the negative of the $lazy boolean at the end doesn't seem to make any sense either.

kasimi:
This might be related: https://tracker.phpbb.com/browse/PHPBB3-15544

VSEphpbb:
LOL, reading the docblock you can see it won't work:
```
* @param int|string|bool $parent The parent module_id|module_langname (0 for no parent).
*	Use false to ignore the parent check and check class wide.
```
```
* @return bool true if module exists in *all* given parents, false if not
```
If the return is going to be false unless a module exists in all parents, then what how can we tell if a module exists when we use false for the parent and check class wide?

PR #4979, opened 2017-10-02
[ticket/15378] Add core.twig_environment_inject_assets_after

senky:
PHPBB3-15378

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15378

### DIFF
--- a/phpBB/phpbb/template/twig/environment.php
+++ b/phpBB/phpbb/template/twig/environment.php
@@ -253,6 +253,20 @@ class environment extends \Twig_Environment
 		$output = str_replace('__STYLESHEETS_' . $placeholder_salt . '__', $this->assets_bag->get_stylesheets_content(), $output);
 		$output = str_replace('__SCRIPTS_' . $placeholder_salt . '__', $this->assets_bag->get_scripts_content(), $output);
 
+		/**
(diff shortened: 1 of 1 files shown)


Elsensee:
Actually, do you need the `placeholder_salt`? If not, you can just use the `kernel.response` event by Symfony..?

senky:
I don't need this. Ticket author asked for it and I just included as much variables as possible to make event more usable. If you think this is a security risk, I will remove it.

Elsensee:
That's not what I'm trying to say.

First, the placeholder_salt will be basically useless *after* everything containing that salt is replaced.
Second, I wanted to say that there already is such an event, just not in phpBB but in Symfony. It's called `kernel.response`.

Nicofuma:
I don't really like this event. The correct event according to the needs expressed in the ticket is kernel.response as said @Elsensee (but it works only for app.php pages I know, an idea could be to fake this event on other pages).

Otherwise, if you want to manipulate the assets included at the end, I think the best way is to override the assets bag but we need to add a way to override it (maybe with an assets_bag_interface and a factory defined as a service : you can override the factory if you want a different implementation of assets_bag_interface)

senky:
Alright, I'm closing this PR then.

PR #4980, opened 2017-10-02
[ticket/15381] Better explain CONTACT_US_ENABLE_EXPLAIN

senky:
PHPBB3-15381

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15381

### DIFF
--- a/phpBB/language/en/acp/board.php
+++ b/phpBB/language/en/acp/board.php
@@ -371,7 +371,7 @@ $lang = array_merge($lang, array(
 	'ACP_CONTACT_SETTINGS_EXPLAIN'		=> 'Here you can enable and disable the contact page and also add a text that is displayed on the page.',
 
 	'CONTACT_US_ENABLE'				=> 'Enable contact page',
-	'CONTACT_US_ENABLE_EXPLAIN'		=> 'This page allows users to send emails to board administrators',
+	'CONTACT_US_ENABLE_EXPLAIN'		=> 'This page allows users to send emails to board administrators. Please note that board-wide emails option must be enabled as well. You can find this option in General > Client Communication > Email settings.',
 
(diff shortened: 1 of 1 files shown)


phpBB/language/en/acp/board.php
```diff
@@ -371,7 +371,7 @@
 	'ACP_CONTACT_SETTINGS_EXPLAIN'		=> 'Here you can enable and disable the contact page and also add a text that is displayed on the page.',
 
 	'CONTACT_US_ENABLE'				=> 'Enable contact page',
-	'CONTACT_US_ENABLE_EXPLAIN'		=> 'This page allows users to send emails to board administrators',
+	'CONTACT_US_ENABLE_EXPLAIN'		=> 'This page allows users to send emails to board administrators. Please note that board-wide emails option must be enabled as well. You can find this option in General > Client Communication > Email settings.',
```
rxu: Instead of kinda "breadcrumbs", probably a link to the respective ACP page could be given?

senky: Eh, not so easy. To generate the link, you need to append SID. So instead of this simple lang key, a new template variable would have to be introduced. No problem for me. I'm just curious if it's worth it.

Elsensee: Instead I would actually just completely remove the hint where to find this setting. I think we're doing that nowhere.

senky: That doesn't mean we shouldn't. phpBB was build with almost no UX in mind and as it is widely discussed at area51, you can see that navigation is really a pain. Why not make it at least a little bit easier for admins?

JimMH: I agree with @senky here. Saves them a trip to our support forums.

PR #4983, opened 2017-10-03
[ticket/15387] Make vertical bars full height on board index forum rows

Sophist-UK:
[ticket/15387] Make vertical bars full height on board index forum rows

Make vertical left border bars for height for Topics and Posts columns
in board index forum rows when these columns have less height than e.g.
the Last Post column.

PHPBB3-15387

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15387

### DIFF
--- a/phpBB/styles/prosilver/theme/content.css
+++ b/phpBB/styles/prosilver/theme/content.css
@@ -18,7 +18,7 @@ ul.topiclist dl {
 }
 
 ul.topiclist li.row dl {
-	padding: 2px 0;
+	margin: 2px 0;
 }
 
 ul.topiclist dt, ul.topiclist dd {
--- a/phpBB/styles/prosilver/theme/content.css
+++ b/phpBB/styles/prosilver/theme/content.css
@@ -64,8 +64,8 @@ ul.topiclist.two-columns dt .list-inner {
 
 ul.topiclist dd {
 	border-left: 1px solid transparent;
-	padding: 4px 0 99999px 0;
-	margin-bottom: -99995px;
+	padding: 4px 0 999px 0;
+	margin-bottom: -995px;
 	-webkit-box-sizing: border-box;
 	-moz-box-sizing: border-box;
 	box-sizing: border-box;
--- a/phpBB/styles/prosilver/theme/content.css
+++ b/phpBB/styles/prosilver/theme/content.css
@@ -64,11 +64,15 @@ ul.topiclist.two-columns dt .list-inner {
 
 ul.topiclist dd {
 	border-left: 1px solid transparent;
-	padding: 4px 0 999px 0;
-	margin-bottom: -995px;
+	padding: 4px 0;
 	-webkit-box-sizing: border-box;
 	-moz-box-sizing: border-box;
 	box-sizing: border-box;
+}
+
+ul.topiclist li.row dd {
+	padding: 4px 0 999px 0;
+	margin-bottom: -995px;
 }
 
 ul.topiclist dfn {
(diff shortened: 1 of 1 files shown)


hanakin:
hmm if you do this, pls add a `margin-bottom: 2px` to `ul.topiclist li.row dl` for consistency

Sophist-UK:
@hanakin Sorry - but I don't understand why you are asking this. The net dimensions of dd have not changed - the net of margin plus padding is still 4px. Indeed I was most careful to ensure that the only impact was on the left borders and that it did not change the size of the box by even 1px.

However, if you would like me to make the selectors more specific i.e. `ul.topiclist.topics li.row dl dd` I would be happy to do so.

Sophist-UK:
Perhaps for consistency I should also switch from using 99999/99995 to 999/995 (like the selector immediately following).

Sophist-UK:
P.S. I am unclear what the standard for specificity is in this situation. Some existing css is `ul li dl` and some `ul dt` without intervening li/dl. And none of it is ul.class > li.class > dl.class > dt/dd which would be meet the performance characteristics in your proposed 3.3 css coding standards.

Ho hum...

hanakin:
1. The 3.2 and 3.3 are completely different beasts. specificity is all over the place as this is a 10yr old theme. and ultimately the correct way to handle it per the 3.3 specs if just .class but all the other specificity would need fix to allow for this which is not practical. We are working on a new default theme which is what the 3.3 specs are really for, but we need to keep prosilver working as much as we can and update what we reasonably can

2. the reason i asked you to make that change was due to the fact that there is a break in the top from the wrap but not the bottom, for consistency there should be, and also yes 999 is probably enough

Sophist-UK:
1. I will leave specificity as-is because there is substantial other css with same specificity. 

2. I am still unclear why you want me to fix some other unconnected issue with a 2px margin in this PR. General opinion is not to mix fixes for separate issues in the same PR.

hanakin:
its not unrelated as you can see the border touches the bottom of the container but not the top. It should not touch either

Sophist-UK:
Yes - I can see that it touches, though unclear why that is an issue since it is the padding which touches.

But it is an issue both before and after this PR, so I am unclear why it is connected to this PR.

hanakin:
it is not an issue before this pr

before
screenshot 2017-10-03 12 29 52

Sophist-UK:
Ah - in that case of course I need to fix it. Can you provide an after screenshot, because on my board it still looks like that.

Yes - I am using a margin hack, but because the padding was 4px and margin 0px before, and after padding plus margin is still 4px, I am unclear why it has any impact.

If it was going to have an impact I would expect it to be c. 995px of impact and immediately obvious.

Sophist-UK:
Here are my two screen shots - similar to those in the  ticket:

Before:
phpbb-bar-before

After:
phpbb-bar-after

As far a I can tell these are pixel identical except for the bar themselves if you look very closely.

Sophist-UK:
Perhap I have now seen what you mean. There is a 2px gap at the top of the white bar but not at the bottom. If this is what you mean, I will fix it.

hanakin:
yes that

Sophist-UK:
It might have helped if I hadn't missed the two previous screen shots and explanation.

Apologies for having doubted you!!!

Sophist-UK:
Please do not merge - MCP is affected by this change and needs a fix.

Sophist-UK:
MCP issue fixed.

marc1706:
Thanks for your pull request. I've taken the liberty to fix the commit message format to fit into our desired one:
https://github.com/phpbb/phpbb/pull/5026

PR #4988, opened 2017-10-04
[ticket/15389] Allow arrays in event dispatcher

Elsensee:
Reasoning is in the tracker.
This change is fully backwards compatible, of course. Nothing in the existing core events or in any extension must be changed.
This change will only allow the short array syntax for the event names and will allow both array syntaxes for multi line event vars instead of just the old syntax.

Checklist:

- [x] Tests pass

Tracker ticket:
https://tracker.phpbb.com/browse/PHPBB3-15389

### DIFF
--- a/phpBB/phpbb/event/dispatcher.php
+++ b/phpBB/phpbb/event/dispatcher.php
@@ -57,7 +57,12 @@ class dispatcher extends ContainerAwareEventDispatcher implements dispatcher_int
 			return $event;
 		}
 
-		return parent::dispatch($eventName, $event);
+		foreach ((array) $eventName as $name)
+		{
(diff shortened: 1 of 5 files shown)


Nicofuma:
could be nice to have a way to mark an event as deprecated too

Elsensee:
@Nicofuma What syntax would you suggest for that?

Nicofuma:
that's a good question and I don't really know right now :/

Elsensee:
Then, how would you mark an event as deprecated? Would you output a trigger_error(E_DEPCREATED) or something completely else?
If yes, then I have an idea.

marc1706:
I think the deprecation can be handled by another ticket and pull request.

PR #4990, opened 2017-10-05
[ticket/15377] Fix INCLUDECSS and INCLUDEJS

rubencm:
PHPBB3-15377

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15377

### DIFF
--- a/phpBB/phpbb/template/twig/loader.php
+++ b/phpBB/phpbb/template/twig/loader.php
@@ -23,6 +23,16 @@ class loader extends \Twig_Loader_Filesystem
 	protected $safe_directories = array();
 
 	/**
+	 * Constructor
+	 *
+	 * @param string|array	$paths
+	 */
+	public function __construct($paths = array())
+	{
+		parent::__construct($paths, filesystem_helper::realpath(dirname(__FILE__)));
+	}
+
+	/**
 	* Set safe directories
 	*
 	* @param array $directories Array of directories that are safe (empty to clear)
--- a/phpBB/phpbb/template/twig/loader.php
+++ b/phpBB/phpbb/template/twig/loader.php
@@ -29,7 +29,7 @@ class loader extends \Twig_Loader_Filesystem
 	 */
 	public function __construct($paths = array())
 	{
-		parent::__construct($paths, filesystem_helper::realpath(dirname(__FILE__)));
+		parent::__construct($paths, __DIR__);
 	}
 
 	/**
--- a/tests/functional/extension_controller_test.php
+++ b/tests/functional/extension_controller_test.php
@@ -90,6 +90,15 @@ class phpbb_functional_extension_controller_test extends phpbb_functional_test_c
 	}
 
 	/**
+	* Check includejs/includecss when the request_uri is a subdirectory
+	*/
+	public function test_controller_template_include_js_css()
+	{
+		$crawler = self::request('GET', 'app.php/help/faq');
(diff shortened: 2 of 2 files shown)


rxu:
Curently, the tests related
https://github.com/phpbb/phpbb/blob/master/tests/template/template_includecss_test.php
https://github.com/phpbb/phpbb/blob/master/tests/template/template_includejs_test.php
don't span this case for some reason, thus we're getting the same bug for the 2nd time.

phpBB/phpbb/template/twig/loader.php
```diff
@@ -23,6 +23,16 @@ class loader extends \Twig_Loader_Filesystem
 	protected $safe_directories = array();
 
 	/**
+	 * Constructor
+	 *
+	 * @param string|array	$paths
+	 */
+	public function __construct($paths = array())
+	{
+		parent::__construct($paths, filesystem_helper::realpath(dirname(__FILE__)));
```
CHItA: Just use `parent::__construct($paths, __DIR__);`?

tests/functional/extension_controller_test.php
```diff
@@ -90,6 +90,15 @@ public function test_controller_with_template()
 	}
 
 	/**
+	* Check includejs/includecss when the request_uri is a subdirectory
+	*/
+	public function test_controller_template_include_js_css()
+	{
+		$crawler = self::request('GET', 'app.php/help/faq');
+		$this->assertContains("./../../assets/javascript/core.js", $crawler->filter('body')->html());
```
CHItA: I think on the long run we should just generate the absolute path.

phpBB/phpbb/template/twig/loader.php
rxu: Looks like working as well. If so, it could be applied to 3.2 too.

rubencm: Yes, is exactly the same.
https://secure.php.net/manual/en/language.constants.predefined.php

CHItA: Exactly my point, what you did is pointless.

rubencm: I haven't done this, i just restore a code that was accidentally removed

rubencm: https://pastebin.com/ANeWhryy

VSEphpbb: `__DIR__` is better because it does not require a function call at execution time, so it's faster.

rubencm: https://tracker.phpbb.com/browse/PHPBB3-15392

phpBB/phpbb/template/twig/loader.php
```diff
@@ -23,6 +23,16 @@ class loader extends \Twig_Loader_Filesystem
 	protected $safe_directories = array();
 
 	/**
+	 * Constructor
+	 *
+	 * @param string|array	$paths
+	 */
+	public function __construct($paths = array())
+	{
+		parent::__construct($paths, filesystem_helper::realpath(__DIR__);
```
CHItA: I still do not understand why would you call `realpath` here.

rubencm: It already was...
Anyway, it was necessary before 5.3.0.
There are many more things that can be improved because old php versions are not supported

PR #4996, opened 2017-10-12
[ticket/15397] Check that table exist before modifying its columns/indexes/keys

blitze:
Given a migration, A, that adds database tables and drops them when reverting,
and a migration, B, that modifies the columns of a table added by migration A,
the migrator runs in the correct order when purging the extension but drops the
first table created by migration A, the starts over with migration B, which
then returns an error because the table no longer exists. This ensures that the
table exists before trying to modify its columns.

PHPBB3-15397

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15397

### DIFF
--- a/phpBB/phpbb/db/tools/tools.php
+++ b/phpBB/phpbb/db/tools/tools.php
@@ -499,6 +499,12 @@ class tools implements tools_interface
 		{
 			foreach ($schema_changes['change_columns'] as $table => $columns)
 			{
+				// only change columns if the table exists
+				if (!$this->sql_table_exists($table))
+				{
+					continue;
+				}
+
 				foreach ($columns as $column_name => $column_data)
 				{
 					// If the column exists we change it, else we add it ;)
@@ -535,6 +541,12 @@ class tools implements tools_interface
 		{
 			foreach ($schema_changes['add_columns'] as $table => $columns)
 			{
+				// only add columns if the table exists
+				if (!$this->sql_table_exists($table))
+				{
+					continue;
+				}
+
 				foreach ($columns as $column_name => $column_data)
 				{
 					// Only add the column if it does not exist yet
@@ -574,6 +586,12 @@ class tools implements tools_interface
 		{
 			foreach ($schema_changes['drop_keys'] as $table => $indexes)
 			{
+				// only remove keys if the table exists
+				if (!$this->sql_table_exists($table))
+				{
+					continue;
+				}
+
 				foreach ($indexes as $index_name)
 				{
 					if (!$this->sql_index_exists($table, $index_name))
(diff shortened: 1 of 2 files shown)


marc1706:
The problem is that with these changes you're simply skipping the changes. Instead, further migration should be stopped and the user be notified of the missing dependency (in this case the table).

kasimi:
Related: https://tracker.phpbb.com/browse/PHPBB3-15396

VSEphpbb:
Re: https://github.com/phpbb/phpbb/pull/4996

We've experienced this problem developing the Ad Management extension. It has some migrations that add keys to existing tables/columns.

When uninstalling, it would fail to revert those migrations, even though the order of migrations made sense, i.e.: it depended on the migration that created the table. It would fal because the table or column had been deleted before it would delete the keys.

We solved it by making the problem migration depend on pratically every single migration that came before it, which somehow forced it to finally work in order. 

But there is a real problem with revert_scjema order because our original depends on should have worked in theory.

I'd rather see the revert order be fixed, rather than this PR, which doesn't solve the problem. Stopping the migration and notifying the user won't just block them from uninstalling an extension in such a situation, and then they'll be stuck.

blitze:
Well, it just made sense to me that we want to check if a table exists before we start modifying it's columns, which fixes this issue.

From my findings, and without fully knowing how the migrator works, the order of the migration is not the problem, at least not initially. Say migration A adds 3 tables and has instructions to revert those 3 tables when purging, and let's say migration B depends on migration A, and modifies the first table added by migration A, and also provides instructions to revert the changes, as it is recommended to do so (BTW I found that this issue only occurs if revert instructions are provided). 

When reverting, the revert_schema method of migration B is called and changes are reverted, then revert_schema of migration A is called and the first table is reverted. Then the migrator starts over with migration B again, and this time, you get a database error since the table was previously dropped. 

It is true that this PR does not address why the migrator starts over but i think it makes sense to ensure that the table exists before we try to mess with its columns/indexes/keys. I'll see if I can figure out why the migrator starts over.

Nicofuma:
I think there is 2 things here:
1) You are right when you say that we should check that the table exists before altering it. But we shouldn't skip the instruction, we should display a nice error, stop here and disable the extension (disable, not remove).
2) It looks like there is a bug in the migrator that needs to be addressed (or at least discribed in a new ticket).

There is also a third thing actually that diserves a ticket imo: before running any migration we should dry-run it and check that all instruction can be executed.

3D-I:
That's a blocker IMO. 

I for one I faced this issue, momentarily solved moving the revert schema into the depending migration.
That's though a bit hackish.

marc1706:
First of, thanks for looking into this. I've taken a deeper look at this issue and to be honest I think we should try to fix the order the revert is executed as this is the underlying issue. If migration B depends on migration A being executed, migration B should be reverted before migration A. Therefore you should never hit the case where the table was already deleted before a column can be removed.

PR #4997, opened 2017-10-13
[ticket/15396] Fix revert_schema() steps not executed in correct order

kasimi:
When applying database changes, [the `try_apply` method _returns_ after applying a migration step from a dependency](https://github.com/phpbb/phpbb/blob/release-3.2.1/phpBB/phpbb/db/migrator.php#L351). When reverting database changes, there is no _return_, which causes the migrator to apply the first step of a depending migration, then the steps of the migration itself, and all remaining steps of the depending migration afterwards.

With these changes, the migrator _returns_ after each `revert_do` of depending migrations and continues at the correct next step, therefore handling all dependencies' steps in correct order.

Deleting data of the test extension from the ticket, @blitze's SiteMaker 3.0.1 and @VSEphpbb's Ad Manangement 1.0.2 works fine.

Some tests coming right up.

PHPBB3-15396

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15396

### DIFF
--- a/phpBB/phpbb/db/migrator.php
+++ b/phpBB/phpbb/db/migrator.php
@@ -503,11 +503,14 @@ class migrator
 			return;
 		}
 
-		foreach ($this->migration_state as $name => $state)
+		foreach ($this->migrations as $name)
 		{
-			if (!empty($state['migration_depends_on']) && in_array($migration, $state['migration_depends_on']))
+			$state = $this->migration_state($name);
+
+			if ($state && in_array($migration, $state['migration_depends_on']) && ($state['migration_schema_done'] || $state['migration_data_done']))
 			{
 				$this->revert_do($name);
+				return;
 			}
 		}
 
--- a/tests/dbal/migration/revert_table.php
+++ b/tests/dbal/migration/revert_table.php
@@ -1,0 +1,39 @@
+<?php
+/**
+*
+* This file is part of the phpBB Forum Software package.
+*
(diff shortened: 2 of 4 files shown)


blitze:
Ah, you beat me to it :). My solution is different from yours though. I noticed that the migrator was not completing all the steps of a migration before moving to the next migration. So if the migration had 3 steps, it would do one, then move on to the next migration and come back and do the next one, etc. My solution was to add the following after [line 617 in migrator.php](https://github.com/phpbb/phpbb/blob/2c01fe67e9313565741f20d207ffb289b1d0b2d3/phpBB/phpbb/db/migrator.php#L617)
`
		if (is_array($result))
		{
			return $this->try_revert($name);
		}
`
That way it completes all the steps for a particular migration before moving on. Migrations depend on other migrations and not their individual steps. So all steps should be completed before moving on. This seems to result in less cycling through the migrations.

kasimi:
@blitze Returning from the `revert_do` method gives control back to the extension manager class which allows more control over timeout handling. That's also why the `update_do` method does it, I think.

Elsensee:
@kasimi Exactly. Unfortunately timeout handling isn't really handled by the extension manager because of that unconditional while-loop. However, there *could* be timeout handling - that's why it was done in `update_do`. (I missed it in `revert_do` because I was working on the updater which has timeout handling but only does updating.. Oops)

blitze:
Ah makes sense. I did not take that into consideration.

PR #5022, opened 2017-10-31
[ticket/15431] Add event core.ucp_register_modify_template_data

kasimi:
I moved the `phpbb_timezone_select()` call below the event so that changes to the `$data['tz']` field in event handlers can take effect.

PHPBB3-15431

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15431

### DIFF
--- a/phpBB/includes/ucp/ucp_register.php
+++ b/phpBB/includes/ucp/ucp_register.php
@@ -528,10 +528,7 @@ class ucp_register
 			break;
 		}
 
-		// Assign template vars for timezone select
-		phpbb_timezone_select($template, $user, $data['tz'], true);
-
-		$template->assign_vars(array(
+		$template_vars = array(
 			'ERROR'				=> (sizeof($error)) ? implode('<br />', $error) : '',
 			'USERNAME'			=> $data['username'],
 			'PASSWORD'			=> $data['new_password'],
@@ -552,7 +549,37 @@ class ucp_register
 
 			'COOKIE_NAME'		=> $config['cookie_name'],
 			'COOKIE_PATH'		=> $config['cookie_path'],
-		));
+		);
+
+		$tpl_name = 'ucp_register';
+		$page_title = 'UCP_REGISTRATION';
+
+		/**
+		* Modify template data on the registration page
+		*
+		* @event core.ucp_register_modify_template_data
+		* @var	array	template_vars		Array with template data
+		* @var	array	data				Array with user data
+		* @var	array	error				Array with errors
+		* @var	array	s_hidden_fields		Array hidden form fields
+		* @var	string	tpl_name			Template name
+		* @var	string	page_title			Page title
+		* @since 3.2.2-RC1
+		*/
+		$vars = array(
+			'template_vars',
+			'data',
+			'error',
+			's_hidden_fields',
+			'tpl_name',
+			'page_title',
+		);
+		extract($phpbb_dispatcher->trigger_event('core.ucp_register_modify_template_data', compact($vars)));
+
+		// Assign template vars for timezone select
+		phpbb_timezone_select($template, $user, $data['tz'], true);
+
+		$template->assign_vars($template_vars);
 
 		//
 		$user->profile_fields = array();
@@ -561,8 +588,8 @@ class ucp_register
 		$cp->generate_profile_fields('register', $user->get_iso_lang_id());
 
 		//
-		$this->tpl_name = 'ucp_register';
-		$this->page_title = 'UCP_REGISTRATION';
+		$this->tpl_name = $tpl_name;
+		$this->page_title = $page_title;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


kasimi:
!set Event

phpBB/includes/ucp/ucp_register.php
```diff
@@ -552,7 +549,37 @@ function main($id, $mode)
...
+		* @var	string	s_hidden_fields		HTML with hidden form field elements
+		* @var	string	tpl_name			Template name
+		* @var	string	page_title			Page title
+		* @since 3.2.2-RC1
+		*/
+		$vars = array(
+			'template_vars',
+			'data',
+			'error',
+			's_hidden_fields',
```
Derky: The vars `$data` (except `$data['tz']`), `$error` and `$s_hidden_fields` were already used to build `$template_vars`. This means overwriting them in an event doesn't have any effect. 

To fix this you could split up creation of `$template_vars` before and after the event.

phpBB/includes/ucp/ucp_register.php
```diff
@@ -561,8 +588,8 @@ function main($id, $mode)
 		$cp->generate_profile_fields('register', $user->get_iso_lang_id());
 
 		//
-		$this->tpl_name = 'ucp_register';
-		$this->page_title = 'UCP_REGISTRATION';
+		$this->tpl_name = $tpl_name;
+		$this->page_title = $page_title;
```
Derky: It turns out that `$this->page_title` isn't used. The language var `UCP_REGISTRATION` doesn't even exist. The page title for registration is set to `REGISTER` in [ucp.php](https://github.com/phpbb/phpbb/blob/3.2.x/phpBB/ucp.php#L71) so trying to overwrite it in this event doesn't work. 

Could you remove `$this->page_title` all together please? (from this file and from the event)

rubencm: I think it is used here:
https://github.com/phpbb/phpbb/blob/master/phpBB/includes/functions_module.php#L968
but is not declared

Derky: @rubencm I tested it and removing the field didn't change the title. (Changing it in ucp.php however worked). If it tries to fall back it also also can't work because the language var `UCP_REGISTRATION` doesn't exist.

rubencm: I see, you use 2 methods to set the titles, you use page_title, but it is overwritten with this:
https://github.com/phpbb/phpbb/blob/master/phpBB/ucp.php#L71
https://github.com/phpbb/phpbb/blob/master/phpBB/includes/functions_module.php#L1012
https://github.com/phpbb/phpbb/blob/master/phpBB/includes/functions_acp.php#L25

Derky: Yes, line 71 of ucp.php is what I linked to in my first reply. 😛
Your third link is incorrect however, it's not going to the function `adm_page_header()` but to `page_header()`.

rubencm: Yes, you are right
I think all modules should use the $page_title method and remove the alternative methods

phpBB/includes/ucp/ucp_register.php
kasimi: Thanks for checking this!

> This means overwriting them in an event doesn't have any effect.

Yes, but reading them can still be useful. Extensions will have to recreate the `$template_vars` value though. If it's split up, extensions don't have a chance to modify the final template value.

Derky: @kasimi Yes they are useful, but should be marked as readonly in the event comments if you can't change them. However when I look at the final template output of `$error` and `$s_hidden_fields` I think these assignments can just be moved after the event so they're editable:

```
'ERROR'				=> (sizeof($error)) ? implode('<br />', $error) : '',
```
```
'S_HIDDEN_FIELDS'	=> $s_hidden_fields,
```

When doing it like this you can easily add or remove errors and you're also able to replace the text of an error without having to implode() and (re)parse to the template again. It's also easy to add an additional hidden field.

The `$data` variable is more tricky. I first thought of just marking that as "readonly" in the event comments, but  `$data['tz']` is being used after the event, meaning the value of it can be edited in the event. (so can't mark it as readonly) When looking at the final template values, the only value that differs from the actual $data value, is this assignment:
```
'S_LANG_OPTIONS'	=> language_select($data['lang']),
```
If we move this assignment to after the event, you can indeed no longer change the final template value of it.

So what we could do is mark `$data` as readonly and create a separate `$tz` variable (from`$data['tz']`) and pass that to the event as being editable. If you have a better way, let me know!

kasimi: * The call to `phpbb_timezone_select()` has been moved above the event. Subscribers will need to clear the `timezone_select` and `timezone_date` template blocks prior to calling this function again, if they intend to.

* The `$event['data']` array  has been marked _read-only_.

* The `$event['error']` array is imploded and added to the `$template_vars` array _below_ the event. This allows subscribers to easily manipulate the array without having to re-assign it to `$event['template_vars']`.

* The `$event['s_hidden_fields']` array is added to the `$template_vars` array _below_ the event for the same reason. Also, the call to `build_hidden_fields()` has been moved below the event which now receives the form elements in an array.

PR #5023, opened 2017-10-31
[ticket/15429] Add function phpbb_general_select() to functions.php

hubaishan:
Add function `phpbb_general_select()` to help extension writers who want to add their config to core ACP configs with combobox

PHPBB3-15429

Checklist:

- [ ] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-15429

### DIFF
--- a/phpBB/includes/functions.php
+++ b/phpBB/includes/functions.php
@@ -457,6 +457,30 @@ function phpbb_get_timezone_identifiers($selected_timezone)
 }
 
 /**
+* Options to pick an element from array
+*
+* @param	array 		$options_array		An array contains the element to include in select options, Array key is the value, Array value is language key
+* @param	string		$default			An element to select
+*
+* @return	string		Returns a string containing the options for the selector.
+*/
+function phpbb_general_select($options_array, $default = '')
+{
+	global $phpbb_container;
+
+	$user = $phpbb_container->get('user');
+	$general_options = '';
+
+	foreach ($options_array as $key => $value)
+	{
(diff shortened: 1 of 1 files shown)


DavidIQ:
This looks so specific to your particular needs that I don't see it helping anyone else much if at all...

If this really is for the ACP then just use `build_cfg_template` which is in `functions_acp.php`.

hubaishan:
> If this really is for the ACP then just use build_cfg_template which is in functions_acp.php.

This new function will be called by build_cfg_template to build 'select' options

rubencm:
Talking about forms..., i think i need something like this #1922

hubaishan:
> Ah yes, forgot about that part of build_cfg_template. That does not, however, change my comment about this only benefiting your specific need.
No it's general. many extension developer may need it, if they want their extension configurations lay beside phpBB core  configurations not in separate module.  they can use 'core.acp_board_config_edit_add' event to add textboxs and radio buttons but not comboboxes because 'build_cfg_template' need a function call to build comboboxes options, this PR will add general function will be useful in many cases.

marc1706:
I'm having a few issues with this pull request. First of, we're trying to get rid of functions that generate HTML inside PHP files. Secondly, your function is missing a phpbb_ prefix to create a more unique name. Thirdly, you're using globals instead of passing the user class object directly.

hanakin:
@rubencm Agree prefer a twig function for this. No HMTL in core...

PR #5025, opened 2017-11-01
[ticket/15433] Fix phpbbcli can enable non-existent extension

hubaishan:
Fix phpbbcli can enable non-existent extension

PHPBB3-15433

Checklist:

- [x] Tests pass

Tracker ticket :

https://tracker.phpbb.com/browse/PHPBB3-15433

### DIFF
--- a/phpBB/language/en/cli.php
+++ b/phpBB/language/en/cli.php
@@ -117,6 +117,7 @@ $lang = array_merge($lang, array(
 	'CLI_EXTENSION_ENABLE_FAILURE'		=> 'Could not enable extension %s',
 	'CLI_EXTENSION_ENABLE_SUCCESS'		=> 'Successfully enabled extension %s',
 	'CLI_EXTENSION_ENABLED'				=> 'Extension %s is already enabled',
+	'CLI_EXTENSION_NOT_EXIST'			=> 'Extension %s does not exist',
 	'CLI_EXTENSION_NAME'				=> 'Name of the extension',
 	'CLI_EXTENSION_PURGE_FAILURE'		=> 'Could not purge extension %s',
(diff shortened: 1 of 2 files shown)


rubencm:
Hello, i think that the ext class is optional, you should validate it better with this:
https://github.com/phpbb/phpbb/blob/master/phpBB/includes/acp/acp_extensions.php#L125-L139

phpBB/language/en/cli.php
```diff
@@ -117,6 +117,7 @@
 	'CLI_EXTENSION_ENABLE_FAILURE'		=> 'Could not enable extension %s',
 	'CLI_EXTENSION_ENABLE_SUCCESS'		=> 'Successfully enabled extension %s',
 	'CLI_EXTENSION_ENABLED'				=> 'Extension %s is already enabled',
+	'CLI_EXTENSION_NOT_EXIST'			=> 'Extension %s is not exist',
```
Crizz0: Does not exist?

hubaishan:
> Hello, i think that the ext class is optional, you should validate it better with this:
https://github.com/phpbb/phpbb/blob/master/phpBB/includes/acp/acp_extensions.php#L125-L139

only check for existence no validate, copied from
https://github.com/phpbb/phpbb/blob/3.2.x/phpBB/phpbb/extension/manager.php#L133-L137
 
`ext` class is not optional, It's preventing enabling something like `phpbb\datetime`

rubencm:
Yes, but i say that you should check the composer.json which is mandatory.
About the ext.php, validate it only if it exists

hubaishan:
> Yes, but i say that you should check the composer.json which is mandatory.
About the ext.php, validate it only if it exist

@rubencm  You are RIGHT, I do not understand you before, please review after last commit.

rubencm:
Now is fine i think  :ok_hand:

PR #5030, opened 2017-11-02
[ticket/15393] Fix wrong direction for text boxs contain code in RTL

hubaishan:
A css class 'always-ltr' was added to adm/style/admin.css 
class 'always-ltr' was added to style files in adm/style
function 'build_cfg_template()' in includes/functions_acp.php was modified to handle sent class
Another files how use 'build_cfg_template()' function was modifed by adding 'always-ltr'

UPDATED:
only changes to admin.css and 2 other template files as vinny and marc directed

Checklist:

- [ ] Tests pass

Tracker ticket 

https://tracker.phpbb.com/browse/PHPBB3-15393

### DIFF
--- a/phpBB/adm/style/admin.css
+++ b/phpBB/adm/style/admin.css
@@ -1126,6 +1126,24 @@ input, textarea {
 	border-bottom: 1px solid #D5D5C8;
 }
 
+.rtl input[type="email"], .rtl input[type="url"],
+.rtl #add_extension, .rtl #ban,
+.rtl #bbcode_match, .rtl #bbcode_tpl,
+.rtl #bot_name, .rtl #bot_agent, .rtl #bot_ip,
+.rtl #forum_image, .rtl #forum_link, .rtl #forum_rules_link,
+.rtl #jab_host, .rtl #jab_username,
+.rtl #email,
+.rtl #upload_path, .rtl #img_imagick,
+.rtl #allowed_schemes_links, .rtl #cookie_domain, .rtl #cookie_name, .rtl #cookie_path,
+.rtl #smilies_path, .rtl #icons_path, .rtl #upload_icons_path, .rtl #ranks_path,
+.rtl #server_protocol, .rtl #server_name, .rtl #script_path,
+.rtl #smtp_host, .rtl #smtp_username,
+.rtl #default_dateformat,
+.rtl #avatar_gallery_path, .rtl #avatar_path,
+.rtl #fulltext_sphinx_data_path, .rtl #fulltext_sphinx_host {
+	direction:ltr;
+}
+
 input:hover, textarea:hover {
 	border-left: 1px solid #AFAEAA;
 	border-top: 1px solid #AFAEAA;
--- a/phpBB/adm/style/acp_bbcodes.html
+++ b/phpBB/adm/style/acp_bbcodes.html
@@ -102,7 +102,7 @@
 	<tbody>
 	<!-- BEGIN bbcodes -->
 		<tr>
-			<td style="text-align: center;">{bbcodes.BBCODE_TAG}</td>
+			<td style="direction: ltr; text-align: center;">{bbcodes.BBCODE_TAG}</td>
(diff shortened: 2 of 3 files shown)


vinny:
@hubaishan a solution using only CSS is not possible?

marc1706:
I'd also very much prefer a solution that requires less core modifications (i.e. employ CSS instead of special classes).

hubaishan:
OK, I will try.

vinny:
[Here](https://github.com/phpbb/phpbb/blob/3.2.x/phpBB/adm/style/admin.css#L1121^L1126) you can add `direction: ltr;`

But you still have to find a way to adjust bbcode.

hubaishan:
Not all `input` and `textarea` must be `ltr`, Only those contains codes, URIs, directories, emails

hubaishan:
So we must use their IDs in css file

hanakin:
I don't not think this is cleanly fixable in 3.2 for all cases really the best way to handle this is with a core fix in 3.3 by factoring out the html. Otherwise we are going to have to factor all of this out in 3.3 as its way hacky. I suggest you split this into to prs one for the ltr utility class fixes and anything that currently requires core edits in another pr.

hubaishan:
!set Styles

phpBB/adm/style/admin.css
```diff
@@ -1126,6 +1126,24 @@ input, textarea {
 	border-bottom: 1px solid #D5D5C8;
 }
 
+.rtl input[type="email"], .rtl input[type="url"],
+.rtl #add_extension, .rtl #ban,
```
hanakin: Why? just add the utility class to the html! We do not target IDs for css as its a bad practice and leads to over specification

marc1706: The previous solution included a special class that was added in a rather weird way by extending the type specifier of acp options. Adding a CSS class that takes care of this was not what bothered me, only the way it was added. Sorry for the confusion.

hubaishan: Editing `build_cfg_template` function is required, An alternate way is adding new non HTML types for ACP options: `path` because most of them are paths, `textltr` for others.

And editing  `build_cfg_template` function to change these new types to `type="text" class="always-ltr"`

hanakin: If it’s doable via ids then it’s doable via two class approach! Just have to factor out the specificity.

hanakin: Most likely I would ha e to look into it further to see what exactly you are trying to accomplish

hubaishan:
@hanakin @marc1706 
Instead of this PR I submitted 2 PRs:
https://github.com/phpbb/phpbb/pull/5092 contains CSS and HTML changes only
https://github.com/phpbb/phpbb/pull/5093 contains core files changes

hanakin:
I’m not in front of a computer right now I’ll look things over tonight if I remember, but did this or fix everything? If so then I fail to see why you would need to split it up? Just have to play with the html some to fix the specificity

hubaishan:
Some ACP modules does not use HTML template files for options they use function `build_cfg_template` to generate HTML for the options these modules need the second PR

hanakin:
So they were not fixed in this or?

hubaishan:
They fixed in this, because I use the IDs. But without IDs , There are need to edit `build_cfg_template`  and its input vars to handle adding class.

hanakin:
so where do those ids come from? that's what needs to be edited we just need to add a class where those ids are...If you can jump on irc and dm me we can discuss the best solution.

hubaishan:
`build_cfg_template` function https://github.com/phpbb/phpbb/blob/3.2.x/phpBB/includes/functions_acp.php#L238^414
`acp_borad.php` prepare configurations in arrays like https://github.com/phpbb/phpbb/blob/master/phpBB/includes/acp/acp_board.php#L382^406
then it uses `build_cfg_template` function to generate HTML https://github.com/phpbb/phpbb/blob/master/phpBB/includes/acp/acp_board.php#L722 in this loop https://github.com/phpbb/phpbb/blob/master/phpBB/includes/acp/acp_board.php#L693^739

hanakin:
ok will comment on approach then in other pr

PR #5034, opened 2017-11-05
[ticket/15437] Update jQuery to latest version

hanakin:
Checklist:

- [x] Tests pass

PHPBB3-15437

https://tracker.phpbb.com/browse/PHPBB3-15437


Louis7777:
Just found the opportunity to say that the wording of the S_ALLOW_CDN variable confuses me.

Altogether I read that line like "If CDN is allowed i.e. if CDN works, then load the local copy of jQuery".

Shouldn't it be the opposite? If CDN doesn't work (so !S_ALLOW_CDN) then load the local copy.

phpBB/phpbb/db/migration/data/v330/jquery_update3.php
```diff
@@ -0,0 +1,37 @@
...
+*
+*/
+
+namespace phpbb\db\migration\data\v330;
+
+class jquery_update3 extends \phpbb\db\migration\migration
+{
+	public function effectively_installed()
+	{
+		return $this->config['load_jquery_url'] !== '//ajax.googleapis.com/ajax/libs/jquery/1.11.0/jquery.min.js';
```
DavidIQ: Is the desire with this that if someone had decided to update to a newer restaurant different version at some point that this migration won’t run? Because that would be the result.

vinny:
why the jquery.min name was changed to jquery-3.2.1.min?

changing the filename, you need to update the following lines:
https://github.com/phpbb/phpbb/blob/master/phpBB/includes/functions.php#L4440
https://github.com/phpbb/phpbb/blob/master/phpBB/adm/style/simple_footer.html#L20
https://github.com/phpbb/phpbb/blob/master/phpBB/adm/style/overall_footer.html#L44
https://github.com/phpbb/phpbb/blob/master/phpBB/styles/prosilver/template/simple_footer.html#L34

hanakin:
sorry forgot the WIP tag. I have not had a chance to test any of this yet still working on it.

hanakin:
so it turns out that not only is the JS horribly for a readability and sanity sake its also coded in a way that its preventing update...

phpBB/phpbb/db/migration/data/v330/jquery_update3.php
VSEphpbb: This should be for version 1.12.4

hanakin: if this is true then its been a problem forever 🤣

hanakin:
@Louis7777 totally agree stupidly reversed

phpBB/phpbb/db/migration/data/v330/jquery_update3.php
VSEphpbb: Not sure what you mean, but as it stands, this migration will not update sh@t unless you change this value to exactly match the value this migration is updating to, so 1.12.4.

hanakin:
```
class jquery_update2 extends \phpbb\db\migration\migration
{
	public function effectively_installed()
	{
		return $this->config['load_jquery_url'] !== '//ajax.googleapis.com/ajax/libs/jquery/1.8.3/jquery.min.js';
	}

	static public function depends_on()
	{
		return array(
			'\phpbb\db\migration\data\v310\jquery_update',
		);
	}

	public function update_data()
	{
		return array(
			array('config.update', array('load_jquery_url', '//ajax.googleapis.com/ajax/libs/jquery/1.11.0/jquery.min.js')),
		);
	}

}
```

VSEphpbb: Yeah I think it should be this, going forward:
```
return $this->config['load_jquery_url'] === '//ajax.googleapis.com/ajax/libs/jquery/1.12.4/jquery.min.js';
```
It's the same logic, but this way the migration author only needs to be concerned with the current version, and not hunt for the exact previous version.

Notice I changed the logic from `!==` to `===` and of course the version number

hanakin: ok

VSEphpbb:
@Louis7777 You are mistaken about the variable `S_ALLOW_CDN`. It does mean what it says. If CDN is enabled/allowed, then use content delivery networks is what it is doing.

You are probably confused by this line: https://github.com/phpbb/phpbb/blob/master/phpBB/styles/prosilver/template/overall_footer.html#L52

In that line, we're falling back to the local copy of jquery in case the CDN failed to deliver jquery.

Louis7777:
@VSEphpbb then why isn't the previous line also wrapped with `<!-- IF S_ALLOW_CDN -->`? If CDN is not enabled/allowed, then the remote jQuery script shouldn't be fetched either, correct?

Thus, since it's missing from there and is present only for the local copy, I thought it should mean "if CDN isn't allowed to load i.e. the remote script didn't work, then proceed to load the local copy".

VSEphpbb:
@Louis7777 The previous line dynamically creates the jquery link based on whther CDN is used or not.  `T_JQUERY_LINK` will either be from the CDN or locally depending on whteher CDN is allowed or not.
https://github.com/phpbb/phpbb/blob/master/phpBB/includes/functions.php#L4440

DavidIQ:
You're overlooking what the script tag is doing in that line. It's basically a fallback for if the CDN is not reachable. Checks if jQuery is loaded from the CDN and if it isn't then load the local one.

VSEphpbb:
@Louis7777 You're initial instinct was correct... `S_ALLOW_CDN` does in fact mean, if CDN is enabled, do this.

If some of the usages of this in the code seem confusing, all I can do is offer you some parmesan cheese with our spaghetti code 😄

Louis7777:
@DavidIQ I get that. My issue was with the `<!-- IF S_ALLOW_CDN -->` condition. I couldn't understand the wording (whether it should be !S_ALLOW_CDN or S_ALLOW_CDN) and why it's there.

@VSEphpbb so if CDN is not allowed, line 52 won't happen, but `T_JQUERY_LINK` will be the local copy so we're good... and if CDN is allowed `T_JQUERY_LINK` will be the CDN's script, so line 52 will happen in order to check whether the remote script was loaded and act as a fallback by fetching the local copy...

I would also like some tomato sauce please. 😛

hanakin:
the whole convo is moot.

1. the only possibly feasible reason I can fathom for an option for CDN is firewall issues which would return a breaking error which is a stupidly configured firewall!
2. The reason for the if allow check is even mooter as regardless what the PHP generated URL was the check would still function properly local would always pass and not return anything anyway.

hence https://github.com/h5bp/html5-boilerplate/blob/master/dist/index.html

VSEphpbb:
We were just getting louis777 to understand. Other than that there is nothing to converse further on as the code is good and now this PR is good.

marc1706:
Correct, this has already been included in phpBB 3.2.5.

PR #5035, opened 2017-11-06
[ticket/15440] Add arguments to core.ucp_register_agreement

kasimi:
PHPBB3-15440

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15440

### DIFF
--- a/phpBB/includes/ucp/ucp_register.php
+++ b/phpBB/includes/ucp/ucp_register.php
@@ -166,7 +166,7 @@ class ucp_register
 					->format($user->lang['DATE_FORMAT'], true);
 				unset($now);
 
-				$template->assign_vars(array(
+				$template_vars = array(
 					'S_LANG_OPTIONS'	=> (sizeof($lang_row) > 1) ? language_select($user_lang) : '',
 					'L_COPPA_NO'		=> sprintf($user->lang['UCP_COPPA_BEFORE'], $coppa_birthday),
 					'L_COPPA_YES'		=> sprintf($user->lang['UCP_COPPA_ON_AFTER'], $coppa_birthday),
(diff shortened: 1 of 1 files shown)


kasimi:
!set Event

phpBB/includes/ucp/ucp_register.php
```diff
@@ -195,22 +195,32 @@ function main($id, $mode)
...
 			* @event core.ucp_register_agreement
+			* @var	string	tpl_name			Template file
+			* @var	array	template_data		Array with data assigned to the template, read only
+			* @var	array	lang_row			Array with available languages, read only
+			* @var	array	s_hidden_fields		Array with hidden form elements, read only
 			* @since 3.1.6-RC1
+			* @changed 3.2.2-RC1 Added tpl_name, template_data, lang_row, s_hidden_fields
 			*/
-			$phpbb_dispatcher->dispatch('core.ucp_register_agreement');
+			$vars = array('tpl_name', 'template_data', 'lang_row', 's_hidden_fields');
```
Derky: `template_data` should be changed to `template_vars`. (Don't forget to update it in the comment as well)

Derky:
@kasimi thanks for fixing that. Currently 3 of the 4 variables you add to this event are read-only. I think it would be nicer to do the same thing as we did in https://github.com/phpbb/phpbb/pull/5022#discussion_r154537327

So move the following code after the event to make `$s_hidden_fields` and  `$template_vars` editable by the event:
```
'S_HIDDEN_FIELDS'	=> build_hidden_fields($s_hidden_fields),
```
```
$template->assign_vars($template_vars);
``` 
Then we only have `$lang_row` as a readonly variable.

kasimi:
Moving `$template->assign_vars()` below the event could break extensions that assign template vars. A somewhat radical solution could be to `@deprecate` this event and add a new one above the `$template->assign_vars()` call, but there doesn't seem to be a procedure in place for deprecating events.

Derky:
Good point. I've discussed the options with the other developers. 
Can you mark this event as deprecated and move your changes into a new event? 😃

kasimi:
Done, please have another look.

Derky:
It's squashed and merged

PR #5038, opened 2017-11-08
[ticket/15444] Merge duplicate BBCodes via a migration

JoshyPHP:
I expect it to be working but I haven't tested it yet. Depends on #5053

PHPBB3-15444

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15444


JoshyPHP:
!set wip

gooof:
It could work that way:
\phpbb\db\migration\data\v320\bbcode_merge_duplicates.php
```php
<?php
/**
 *
 * This file is part of the phpBB Forum Software package.
 *
 * @copyright (c) phpBB Limited <https://www.phpbb.com>
 * @license GNU General Public License, version 2 (GPL-2.0)
 *
 * For full copyright and license information, please see
 * the docs/CREDITS.txt file.
 *
 */

namespace phpbb\db\migration\data\v320;

use phpbb\textreparser\bbcode_merger;

class bbcode_merge_duplicates extends \phpbb\db\migration\migration
{
	static public function depends_on()
	{
		return array(
			'\phpbb\db\migration\data\v320\allowed_schemes_links',
		);
	}

	public function effectively_installed()
	{
		return version_compare($this->config['version'], '3.2.2', '>=');
	}

	public function update_data()
	{
		return array(
			array('custom', array(array($this, 'merge_bbcodes'))),
		);
	}

	public function merge_bbcodes($resume_data)
	{
		$bbcode_merger = $this->container->get('text_formatter.s9e.bbcode_merger');

		$merged = $bbcode_merger->merge_bbcodes(	...	);

		// SQL
		$sql = 'UPDATE ' . $this->table_prefix . "bbcodes
			SET bbcode = {$merged[0]}
			WHERE " . $xy;
		$this->db->sql_query($sql);

		if ($complete)
		{
			return true;
		}
	}
}
```

JoshyPHP:
@marc1706 I'd really like this PR to ship with 3.2.2 because it would halve the number of topics I have to reply to on phpBB.com's Support forums. Can you give me a heads up a few days before your merge window closes? Thanks.

marc1706:
I noticed you still have the WIP label on this which is why I hadn't considered it for a merge yet. Is it ready for review now?

JoshyPHP:
I haven't tested it yet apart from running the migration on a default installation. Is there a way to test migrations automatically or does it have to be done by hand?

Also, note that this PR depends on #5053 which depends on #5046.

rxu:
@JoshyPHP 
As the migration should be fixed anyway... should it merge similar but different case BBCodes? Like ```CODE``` and ```code=``` for example. Currently it doesn't.

JoshyPHP:
@rxu It should. I don't know whether it does but if it doesn't, it would be a good idea to fix it and rerun the migration on next version.

rxu:
Well, it can't wait until the next version because of fastal error https://tracker.phpbb.com/browse/PHPBB3-15502
I'm gonna fix it altogether then 😃 
Should the resulting BBCode be uppercased or lowercased?

JoshyPHP:
No preference as far as I'm concerned. BBCode names are normalized to uppercase in s9e\TextFormatter so the case doesn't matter in the input.

You may want to leave the case as-is and use a case-insensitive comparison.

rxu:
> You may want to leave the case as-is and use a case-insensitive comparison.

Well, as far as one of BBCodes should be deleted, The opper|lower case should be chosen anyway.
Ok, I got the point, thanks.

JoshyPHP:
Just looked at the code, it's [not an a string comparison](https://github.com/phpbb/phpbb/blob/master/phpBB/phpbb/db/migration/data/v32x/merge_duplicate_bbcodes.php#L34). I would probably normalize it to lowercase on the offchance there a dotted I in a Turkish locale that doesn't get case-folded the right way, or something contrived like that.

rxu:
I see. Is ```strtolower()``` enough then or should it be ```utf8_strtolower()``` (which is just a wrapper for ```mb_strtolower()```)?

JoshyPHP:
Come to think of it, the library will only accept A-Z letters for BBCode names so it doesn't matter either way.

rxu:
Oh. Then I rather choose ```strtolower()```.

rxu:
@JoshyPHP https://github.com/phpbb/phpbb/pull/5079/files#diff-59ffeb846a6620e1605d8476afbfe103

JoshyPHP:
@rxu That seems fine.

PR #5070, opened 2018-01-02
[ticket/15467] Fix JS for permissions setting

rxu:
Checklist:

- [x] Tests pass

<a href="https://tracker.phpbb.com/browse/PHPBB3-15467">PHPBB3-15467</a>.

### DIFF
--- a/phpBB/adm/style/permission_mask.html
+++ b/phpBB/adm/style/permission_mask.html
@@ -9,6 +9,8 @@
 
 	var role_options = new Array();
 
+	var no_role_assigned = "{LA_NO_ROLE_ASSIGNED}";
+
 	<!-- IF S_ROLE_JS_ARRAY -->
 		{S_ROLE_JS_ARRAY}
 	<!-- ENDIF -->
--- a/phpBB/adm/style/permissions.js
+++ b/phpBB/adm/style/permissions.js
@@ -279,6 +279,16 @@ function reset_role(id) {
 	}
 
 	t.options[0].selected = true;
+
+	(function($)// Avoid conflicts with other libraries
+	{
+		var parent = $(t).parent();
+		parent.find("span[title=Roles]")[0].innerText = no_role_assigned;
+
+		// Find proper role value
+		var roleInput = parent.find('input[name^=role][data-name]');
+		roleInput.val(0);
+	})(jQuery); // Avoid conflicts with other libraries
 }
 
 /**
--- a/phpBB/adm/style/permissions.js
+++ b/phpBB/adm/style/permissions.js
@@ -280,15 +280,9 @@ function reset_role(id) {
 
 	t.options[0].selected = true;
 
-	(function($)// Avoid conflicts with other libraries
-	{
-		var parent = $(t).parent();
-		parent.find("span[title=Roles]")[0].innerText = no_role_assigned;
-
-		// Find proper role value
-		var roleInput = parent.find('input[name^=role][data-name]');
-		roleInput.val(0);
-	})(jQuery); // Avoid conflicts with other libraries
+	var parent = t.parentNode;
+	parent.querySelector('span.dropdown-trigger').innerText = no_role_assigned;
+	parent.querySelector('input[data-name^=role]').value = '0';
 }
 
 /**


rxu:
!set Styles

phpBB/adm/style/permission_mask.html
```diff
@@ -9,6 +9,8 @@
 
 	var role_options = new Array();
 
+	var no_role_assigned = "{L_NO_ROLE_ASSIGNED}";
```
vinny: `LA_NO_ROLE_ASSIGNED` ?

rxu: Good catch.

phpBB/adm/style/permissions.js
```diff
@@ -279,6 +279,16 @@ function reset_role(id) {
 	}
 
 	t.options[0].selected = true;
+
+	(function($)// Avoid conflicts with other libraries
```
hanakin: why are you using a sub function? just rewrite the whole function for jquery...Also please remove the conflict comments they are redundant

rxu:
@hanakin Sure, the only minor issue is I have no idea how to do that.

phpBB/adm/style/permissions.js
hanakin: 1. change `var t =` line to `var $t = $(id)`
2. `var parent = $(t).parent();` should just be `$t.parent();`
3. everywhere t is used change it to `$t`
4. wrap the whole reset_role function than in the self-invoking function.

rxu: @hanakin Thanks, but that kinda makes no sense for me to try blindly executing this instruction.
That simply leads me to facing errors like
```
permissions.js:271 Uncaught ReferenceError: jQuery is not defined
    at permissions.js:271
(anonymous) @ permissions.js:271
Uncaught ReferenceError: reset_role is not defined
    at HTMLAnchorElement.onclick 
Uncaught ReferenceError: reset_role is not defined
    at HTMLAnchorElement.onclick 
```

Where I have no clue about fixing them.
I guess it's easier to close the PR and leave it as is.

hanakin: hmm not sure why you did not get the error before then. Its saying that jquery is not loaded prior to the loading of permissions.js

hanakin:
ah so yeah the permissions file is loaded in that specific template not sure why this is not in core.js probably should be

rxu:
So what shall I do then? 😃

hanakin:
I’ll investigate it tonight but I would say merge all that into the core.js file

hanakin:
Sorry actually merge it into admin.js and that should solve everything

rxu:
This all will lead to ```reset_role()``` becomes unworkable in ```onclick``` which requires complex changes in template as well as in js itself.
```
index.php?i=acp_permissions&sid=53512b9bf36564527e6ec9e4cdc6f54e&icat=16&mode=setting_group_local:257 Uncaught ReferenceError: reset_role is not defined
```
I'm not sure the trivial bug should trigger such a fundamental changes while it can be fixed by changing a couple of lines in the script. Neither I'm able to implement those, sorry.

hanakin:
I think you misunderstood what I meant. permissions.js should not be included in a template this way. It should actually be merged into admin.js or moved to overaul_header.html via includejs. This is not really a bugs fault this is a global issue with all our js that eventually needs to be fixed. We actually need to rewrite all our js in the future for this and many other reasons, such as the bad practice in the way we are adding global vars in that file which should be converted to the data object but that's another issue. Just noticed you are trying to push this for 3.2 though. This might be outside the scope of 3.2 maybe avoid using jquery for 3.2? @marc1706 @Nicofuma ?

marc1706:
I agree that this is a more broad issue. permissions.js should really be either merged into another file or included via INCLUDEJS. This probably will cause some other issues though. Not sure if it's worth it for 3.2.x, I'd much rather go for 3.3 and see if we can fix this in 3.2 without these major changes.

rxu:
Can we postpone the decision with those weird "major changes" (which I just can't see) and simply fix this specific bug in 3.2/3.3?
The PR is updated.

rxu:
> yes quick fix 3.2

Current PR state is right for that 😃

marc1706:
BTW, will be included in 3.2.8.

PR #5073, opened 2018-01-06
Update functions.php

Zukero:
The 'user_id' field used to fire the event  core.markread_before was always 0, while all calls to mark notifications as read below use $user->data['user_id']
I added a line fetching $user_id from $user if necessary.

Checklist:

- [ ] Tests pass


rxu:
You can do the same right within the event listener.

Zukero:
Sure, and I did, but then what's the point of passing it as an event
parameter?

rxu:
It's a common case when all function arguments are being passed to events. Also, ```$user_id``` can be set for the ```'post'``` mode. The value of ```0``` also indicates that the argument was passed to the function by default, so changing it before the event call changes this as well.

Zukero:
Alright. Thanks for the explanation and never mind this then.

PR #5087, opened 2018-01-13
SMTP server address AND PROTOCOL

chang-zhao:
[master] Saying "SMTP server address and protocol" instead of "SMTP server address".

Many people reading "SMTP server address", think they know what to do
and put "smtp.(...).com" automatically, not reading the explanation
(because there are too many explanations which do not add new meanings,
and maybe people get tired or it's hard for them to read).
So adding this vital part to the title improves accessibility and might
save hours of struggling with errors.

PHPBB3-15514


DavidIQ:
This shows as failing because the commit message doesn't actually follow the format expected mentioned in the pr description above.

chang-zhao:
OK, I think I got it & it's OK now. I hope the check reruns automatically?

DavidIQ:
Commit message still looks the same. Did you push the message fix?

hanakin:
no your commit is not formatted properly. and you are not using the default pr message either which is your guide to what you are doing wrong. https://wiki.phpbb.com/Git#Commit_Messages

chang-zhao:
Not yet. Looking how to do that.

chang-zhao:
Perhaps I can't correct it ("Propose file change" is inactive as I changed only the commit message). So should I discard this PR & create another?

marc1706:
I can change that for you as well. There is no need to recreate this with another PR. I think the more important question is: do you really think this should be targeting master instead of 3.2.x? This looks like something that might make sense to be added to 3.2.x as well.

chang-zhao:
> I can change that for you as well. There is no need to recreate this with another PR. I think the more important question is: do you really think this should be targeting master instead of 3.2.x? This looks like something that might make sense to be added to 3.2.x as well.

Thank you very much. Indeed, it makes sense to add to 3.x.x

Moreover, I did a re-installation and found that during installation only 'SMTP_SERVER' is shown to users, but not 'SMTP_SERVER_EXPLAIN'. So users don't get the information about adding protocol at all, hence errors.

So I think I should add another ticket and another file change - to show 'SMTP_SERVER_EXPLAIN' during install, - and then make those 2 changes in one commit - to 3.2.x, you say?

Is that OK?

marc1706:
Sure, that works as well. I will then go ahead and close this PR for now and you can create a new one with the extra changes included.

PR #5090, opened 2018-01-16
[ticket/15508] Support Twig 2.x

deguif:
Checklist:

- [x] Branch: master
- [x] Tests pass
- [x] Code follows coding guidelines: [master](https://area51.phpbb.com/docs/master/coding-guidelines.html), [3.1.x](https://area51.phpbb.com/docs/31x/coding-guidelines.html)

Tracker ticket: [PHPBB3-15508](https://tracker.phpbb.com/browse/PHPBB3-15508)

Changes introduced by Twig 2.x:
- `Twig_NodeInterface` was removed, `Twig_Node` should be used instead.
- `Twig_Parser::getEnvironment()` was removed
- `Twig_Loader_Filesystem::findTemplate()` has a new argument `$throw = true`
- `Twig_Loader_Filesystem::validateName()` visibility changed from protected to private.

To install Twig 2.x here's the composer command I launched (in order to ignore platform requirements as Twig 2 require php >= 7.0): `php composer.phar update twig/twig --with-dependencies --ignore-platform-reqs`

### DIFF
--- a/phpBB/phpbb/template/twig/loader.php
+++ b/phpBB/phpbb/template/twig/loader.php
@@ -110,7 +110,7 @@ class loader extends \Twig_Loader_Filesystem
 	* Override for Twig_Loader_Filesystem::findTemplate to add support
 	*	for loading from safe directories.
 	*/
-	protected function findTemplate($name)
+	protected function findTemplate($name, $throw = true)
 	{
 		$name = (string) $name;
 
@@ -126,12 +126,12 @@ class loader extends \Twig_Loader_Filesystem
 
 		// First, find the template name. The override above of validateName
 		//	causes the validateName process to be skipped for this call
-		$file = parent::findTemplate($name);
+		$file = parent::findTemplate($name, $throw);
 
 		try
 		{
 			// Try validating the name (which may throw an exception)
-			parent::validateName($name);
+			$this->validateName($name);
 		}
 		catch (\Twig_Error_Loader $e)
 		{
--- a/phpBB/composer.json
+++ b/phpBB/composer.json
@@ -49,7 +49,7 @@
 		"symfony/routing": "~3.1",
 		"symfony/twig-bridge": "~3.1",
 		"symfony/yaml": "~3.1",
-		"twig/twig": "^1.0"
+		"twig/twig": "^1.0 || ^2.0"
 	},
 	"require-dev": {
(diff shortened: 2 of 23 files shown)


phpBB/phpbb/template/twig/loader.php
```diff
@@ -126,12 +126,12 @@ protected function findTemplate($name)
...
 		// First, find the template name. The override above of validateName
 		//	causes the validateName process to be skipped for this call
-		$file = parent::findTemplate($name);
+		$file = parent::findTemplate($name, $throw);
 
 		try
 		{
 			// Try validating the name (which may throw an exception)
-			parent::validateName($name);
+			$this->validateName($name);
```
deguif: I didn't find a decent way to handle this, so that's temporary.
Method `validateName` is now private in `Twig_Loader_Filesystem` so the `parent::validateName()` can't be used anymore, but this breaks the current behavior (validateName was overwritten in this class so that call to `parent::findTemplate()` call this class implementation)
Any help here about how to handle this correctly would be greatly appreciated.

deguif:
The Travis job `PHP 5.6` with env `DB=none;NOTESTS=1`  failed with: `The command "phpenv global 5.6" failed and exited with 1 during .` It seems unrelated to this PR. All others are green.

DavidIQ:
Restarted job. Now the errors are all about your commit messages.

deguif:
I updated commit messages, should I rename branch name? This would require to close this PR and open another.

DavidIQ:
Not really necessary as I don't think there's a check for the branch name. Thanks for fixing.

phpBB/phpbb/template/twig/loader.php
Nicofuma: As validateName does not use any state, you could just duplicate it in our loader... We could also try to rework the whole loader but it's a bit more complicated.

deguif: @Nicofuma yes we can duplicate `validateName` in the loader, but when we call `parent::findTemplate`, it itself calls `validateName()` which is private (for twig 2.x) so this won't call the overriden implementation.

marc1706:
Thanks again for your contribution. We're currently working on using twig 2.x on master (in case that might interest you): https://github.com/phpbb/phpbb/pull/5491

PR #5093, opened 2018-01-16
[ticket/15393] PR2 Fix wrong direction for text boxs contain code in rtl

hubaishan:
This PR require [PR1](https://github.com/phpbb/phpbb/pull/5092).
This PR contains the edits to core files.

This PR specify `path` options type for options with path value,
And specify `textltr` options type for options with value
accept only latin charecters.

function `build_cfg_template` was edited to add class `always-ltr` to
`path` and `textltr` options.

PHPBB3-15393

Checklist:

- [ ] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-15393

### DIFF
--- a/phpBB/includes/acp/acp_attachments.php
+++ b/phpBB/includes/acp/acp_attachments.php
@@ -147,7 +147,7 @@ class acp_attachments
 
 						'allow_attachments'		=> array('lang' => 'ALLOW_ATTACHMENTS',		'validate' => 'bool',	'type' => 'radio:yes_no', 'explain' => false),
 						'allow_pm_attach'		=> array('lang' => 'ALLOW_PM_ATTACHMENTS',	'validate' => 'bool',	'type' => 'radio:yes_no', 'explain' => false),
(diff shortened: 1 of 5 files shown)


hubaishan:
!set Styles

hanakin:
There is not css here so you do not need the styles label

hanakin:
probably cleaner to just add a class var to the arrays and a new param to the build function with a class which equals the array index and you could add the always-ltr to it where needed.

hubaishan:
The pervious PR was adding the class in the in type key `text:20:255:always-ltr` but marc disliked it.

hanakin:
yeah, cause it's not clear or clean which is why I suggested this way, also they need to have classes anyway for the new theme. either way, its probably going to need to be based on 3.3

hanakin:
Approved the other PR, but this one needs rebased against master branch

hanakin:
Looking at this again the best way would be to refactor all of this into a proper twig Form function of some kind to handle them all consistently

PR #5101, opened 2018-01-21
[ticket/15529] Color groups in ACP

senky:
PHPBB3-15529

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15529

### DIFF
--- a/phpBB/adm/style/acp_groups.html
+++ b/phpBB/adm/style/acp_groups.html
@@ -316,7 +316,7 @@
 	<tbody>
 		<!-- ELSE -->
 		<tr>
-			<td><strong>{groups.GROUP_NAME}</strong></td>
+			<td><strong<!-- IF groups.GROUP_COLOR --> style="color: #{groups.GROUP_COLOR}"<!-- ENDIF -->>{groups.GROUP_NAME}</strong></td>
 			<td style="text-align: center;">{groups.TOTAL_MEMBERS}</td>
 			<td style="text-align: center;">{groups.PENDING_MEMBERS}</td>
 			<td style="text-align: center;"><a href="{groups.U_EDIT}">{L_SETTINGS}</a></td>
--- a/phpBB/includes/acp/acp_groups.php
+++ b/phpBB/includes/acp/acp_groups.php
@@ -926,7 +926,7 @@ class acp_groups
 		);
 
 		// Get us all the groups
-		$sql = 'SELECT g.group_id, g.group_name, g.group_type
+		$sql = 'SELECT g.group_id, g.group_name, g.group_type, g.group_colour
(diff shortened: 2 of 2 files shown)


phpBB/includes/acp/acp_groups.php
```diff
@@ -985,6 +985,7 @@ function main($id, $mode)
 					'S_GROUP_SPECIAL'	=> ($row['group_type'] == GROUP_SPECIAL) ? true : false,
 
 					'GROUP_NAME'	=> $group_name,
+					'GROUP_COLOR'	=> $row['group_colour'],
```
Crizz0: We use colour instead of color.

phpBB/adm/style/acp_groups.html
```diff
@@ -316,7 +316,7 @@ <h1>{L_SPECIAL_GROUPS}</h1>
 	<tbody>
 		<!-- ELSE -->
 		<tr>
-			<td><strong>{groups.GROUP_NAME}</strong></td>
+			<td><strong<!-- IF groups.GROUP_COLOR --> style="color: #{groups.GROUP_COLOR}"<!-- ENDIF -->>{groups.GROUP_NAME}</strong></td>
```
Crizz0: Color vs. Colour

hanakin:
@Crizz0 Not for long and that’s actually a bad practice.

Crizz0:
Why? I thought we use British English, so colour is correct, even in the same line two different ways as you can see in the changed code lines.

hanakin:
In the copy yes but British English should not be used in code as all code is based on us English so it’s considered bad practice to mix and match ie style=“color: group-colour” is just plain confusing

phpBB/includes/acp/acp_groups.php
senky: Yeah, taken into account, but as far as I know, when you use this variable to assign CSS' `color` prop, then it should be `COLOR`. But if you really insist, I will rename it.

vinny:
@hanakin `colour` is the term used in the core. Take a look for yourself at files like `includes/acp/acp_groups.php`, `memberlist.php` and `viewtopic.php`

Crizz0:
Yes, I see the problem in the template. On the other hand, we have in the php-files this line:
`'GROUP_COLOR'	=> $row['group_colour'],`
And this looks quite strange as well.

senky:
So what do you want me to do? Assign 2 template variables?
```
'S_GROUP_COLOUR'	=> !empty($row['group_colour']),
'COLOR'			=> $row['group_colour'],
```

Crizz0:
No. Just keep it. I think we need a clean-up here in future, to use consistently only "COLOR" in the code.

PR #5107, opened 2018-01-24
[ticket/15522] Allow multiple color palettes on one page

senky:
PHPBB3-15522

Master PR: https://github.com/phpbb/phpbb/pull/5136

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15522

### DIFF
--- a/phpBB/assets/javascript/core.js
+++ b/phpBB/assets/javascript/core.js
@@ -1654,7 +1654,7 @@ $(function() {
 
 	phpbb.registerPageDropdowns();
 
-	$('#color_palette_placeholder').each(function() {
+	$('[data-orientation]').each(function() {
 		phpbb.registerPalette($(this));
 	});
 


hanakin:
Not sure what this is fixing? As it looks like you just swapped and I’d for a class?

marc1706:
Is there no CSS relating to that ID? Also, I think it would make sense to use the naming conventions mentioned here: https://github.com/phpbb/documentation/pull/128/files#diff-7a53b4a793486ebca256731cfd31015fR1

hanakin:
It’s also js so I would prefer we target a data attribute rather than the css selector

senky:
Problem here is that code in `core.js` looks like it allows multiple color palettes on one page. However, since it selects ID, it always works only for one palette. This PR makes it possible for multiple palettes to occur on one page.

So instead of `class="color_palette_placeholder"` you propose `data-color-palette="true"`?

hanakin:
ok but do not target the class instead target one of the data attributes. I am still working out the framework on the js side but I would rework the js so that data-orientation is changed to data-color-pallette with the same values that are there and then use this as the selector.

senky:
@hanakin if that is the case, then this PR might be irrelevant then. How long till new js framework is out?

hanakin:
long time its part of the work for the new theme, and really by framework I mean philosophy/coding standards for js

senky:
Alright, so I replaced it all with very simple change - `[data-orientation]` is only used in association with color palette, so I hope it's OK.

hanakin:
I would suggest renaming the data property to something more meaningful in the future but looks good

phpBB/assets/javascript/core.js
```diff
@@ -1654,7 +1654,7 @@ $(function() {
 
 	phpbb.registerPageDropdowns();
 
-	$('#color_palette_placeholder').each(function() {
+	$('[data-orientation]').each(function() {
```
marc1706: Use `data-color-palette`?

senky: Currently, there is no such data property. As per hanakin's comments, he would like to see it that way in the future, but this suffices now.

marc1706: Data attributes aren't predefined which is why I think this would make sense. See also: https://www.w3.org/TR/2011/WD-html5-20110525/elements.html#embedding-custom-non-visible-data-with-the-data-attributes

senky: Yeah, I am aware how data attributes work. I just think we don't need to add an extra data attribute just to make this one selector easier to read. User needs to define `data-orientation` anyway, color palette wouldn't work without it.

hanakin: hes not saying add it. what we are saying is rename orientation to color-pallette it would require one extra line of js be change and the html being changed which is how I would like to see it, however as this is 3.2 I'm ok with it being as is since there might be some extensions that use it that would break

senky: So either I leave it as it is or change the base branch to 3.3 and rename it?

hanakin: I would say leave it as is for 3.2 and make a new pr for 3.3 @marc1706 @Nicofuma

senky: https://github.com/phpbb/phpbb/pull/5136

PR #5117, opened 2018-02-01
[ticket/15542] Call Javascript files with assets version

rxu:
Checklist:

- [x] Tests pass

<a href="https://tracker.phpbb.com/browse/PHPBB3-15542">PHPBB3-15542</a>.

### DIFF
--- a/phpBB/adm/style/acp_permission_roles.html
+++ b/phpBB/adm/style/acp_permission_roles.html
@@ -20,7 +20,7 @@
 	// ]]>
 	</script>
 
-	<script type="text/javascript" src="style/permissions.js"></script>
+	<!-- INCLUDEJS permissions.js -->
 
 	<a href="{U_BACK}" style="float: {S_CONTENT_FLOW_END};">&laquo; {L_BACK}</a>
 
--- a/phpBB/adm/style/acp_posting_buttons.html
+++ b/phpBB/adm/style/acp_posting_buttons.html
@@ -29,7 +29,7 @@
 // ]]>
 </script>
 
-<!-- INCLUDEJS {T_ASSETS_PATH}/javascript/editor.js -->
+<!-- INCLUDEJS {T_ASSETS_PATH}/javascript/editor.js?assets_version={T_ASSETS_VERSION} -->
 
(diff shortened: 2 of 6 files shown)


rxu:
!set Styles

marc1706:
Wouldn't it make maybe more sense to inject the assets version directly in INCLUDEJS with twig?

rxu:
Probably yes, but isn't it too big change for 3.2? Because that would require changing for a lot of places in template I think. I'm just not sure.

hanakin:
I am not sure why we need versions? These are not dynamic files nor vendor files? What does this solve? End user modification? We are not using a build process at least not yet ie webpack which is where you would need this sort of thing? And even then the build process handles that sort of thing.

DavidIQ:
@hanakin this addresses phpBB package updates that include changes to JavaScript which then produce support requests because things either don't look or work right after an update. Without the assets version changing the old file in cache is used. Having to tell all your users to clear their browser cache or hard refresh (CTRL+F5) is probably not the best solution.

rxu:
Also, assets version got incremented on every cache purge via ACP.

hanakin:
I get that, what I am asking is why we see the need to do anything for 3.2 as we should not be making any real changes in those specific files until 3.3 and in 3.3 we should be using a build process for our js which will handle that...

hanakin:
Something like https://laravel.com/docs/5.3/elixir

rxu:
This PR is just making js calls consistent with all others. Satying about if we're going to make any real changes in those scripts, I'd say that is possible, example: https://github.com/phpbb/phpbb/pull/5070/files#diff-82491f38daf04277fa9e295dee514eda

marc1706:
My point was to keep the INCLUDEJS calls the same and then add the assets version in the twig token parser.

Nicofuma:
I agree with marc on this

marc1706:
@rxu how about this? https://github.com/phpbb/phpbb/pull/5433

marc1706:
Replaced by https://github.com/phpbb/phpbb/pull/5433

PR #5122, opened 2018-02-12
[ticket/15556] Translate AM/PM in time format

hubaishan:
Taking care in case that `am` may be apart of word

PHPBB3-15556

Checklist:

- [x] Tests pass

Tracker ticket :

https://tracker.phpbb.com/browse/PHPBB3-15556

### DIFF
--- a/phpBB/phpbb/datetime.php
+++ b/phpBB/phpbb/datetime.php
@@ -117,13 +117,23 @@ class datetime extends \DateTime
 					if ($day !== false)
 					{
 						// Format using the short formatting and finally swap out the relative token placeholder with the correct value
-						return str_replace(self::RELATIVE_WRAPPER . self::RELATIVE_WRAPPER, $this->user->lang['datetime'][$day], strtr(parent::format($format['format_short']), $format['lang']));
+						return str_replace(self::RELATIVE_WRAPPER . self::RELATIVE_WRAPPER, $this->user->lang['datetime'][$day], strtr(substr(preg_replace_callback("/(\P{Latin})([AP]M|[ap]m)(\P{Latin})/u",
+								function ($matches)
+								{
+									global $user;
+									return $matches[1].($user->lang($matches[2])).$matches[3];
+						}, parent::format($format['format_short']) . ' '), 0 , -1), $format['lang']));
 					}
 				}
 			}
 		}
 
-		return strtr(parent::format($format['format_long']), $format['lang']);
+		return strtr(substr(preg_replace_callback("/(\P{Latin})([AP]M|[ap]m)(\P{Latin})/u",
+				function ($matches)
+				{
+					global $user;
+					return $matches[1].($user->lang($matches[2])).$matches[3];
+		}, parent::format($format['format_long']) . ' '), 0 , -1), $format['lang']);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


hubaishan:
I also posted lightweight alternative PR in https://github.com/phpbb/phpbb/pull/5124

CHItA:
I would prefer this one over the other solution you posted. @phpbb/development-team thoughts?

hanakin:
@CHItA https://www.phpbb.com/community/viewtopic.php?p=14938071#p14938071

CHItA:
@hanakin that would be better in general I think, however, it is unlikely that we would fix it before 3.3.

Anyhow, what I don't get is whether or not there is a difference between the lower and uppercase versions in any language. If not, couldn't just we use the unicode to uppercase function, and be done with it?

hanakin:
Either way it’s not really a full fix just a minor fix for this specific language. Some languages do there own thing

hubaishan:
Not a specific language, Most of non Latin Languages will by fixed with this.

hanakin:
not true as most languages do not only require a translation but also require reorganizing the entire DTG go here: https://momentjs.com/ and play with the language toggles at the bottom

hubaishan:
After a look to http://momentjs.com it used more then two word for  `AM/PM` for some languages but these languages use 24 hours system not 12 hours system so it is not expected they use `AM/PM` in time format. And also momentjs used long `AM/PM` words for these languages (more than 4 characters).

Solidjeuh:
We don't use that here.
3 AM = 3 hours
3 PM = 15 hours

hanakin:
the best way to handle this is in 3.3 by using something like momentjs possibly as a passthrough twig filter. it also would sole some other issues in that it would allow for custom formats for dates by the style author in every instance

phpBB/phpbb/datetime.php
```diff
@@ -117,13 +117,23 @@ public function format($format = '', $force_absolute = false)
 					if ($day !== false)
 					{
 						// Format using the short formatting and finally swap out the relative token placeholder with the correct value
-						return str_replace(self::RELATIVE_WRAPPER . self::RELATIVE_WRAPPER, $this->user->lang['datetime'][$day], strtr(parent::format($format['format_short']), $format['lang']));
+						return str_replace(self::RELATIVE_WRAPPER . self::RELATIVE_WRAPPER, $this->user->lang['datetime'][$day], strtr(substr(preg_replace_callback("/(\P{Latin})([AP]M|[ap]m)(\P{Latin})/u",
+								function ($matches)
+								{
+									global $user;
```
3D-I, line 123: Why `global $user;` ? Isn't the User object already available in this class? Moreover, it would be better to use the Language object nowadays for this kind of stuffs.

phpBB/phpbb/datetime.php
```diff
@@ -117,13 +117,23 @@ public function format($format = '', $force_absolute = false)
...
 					}
 				}
 			}
 		}
 
-		return strtr(parent::format($format['format_long']), $format['lang']);
+		return strtr(substr(preg_replace_callback("/(\P{Latin})([AP]M|[ap]m)(\P{Latin})/u",
+				function ($matches)
+				{
+					global $user;
```
3D-I, line 134: Why `global $user;` ? Isn't the User object already available in this class? Moreover, it would be better to use the Language object nowadays for this kind of stuffs.

3D-I (changes requested):
See my comments below.

PR #5135, opened 2018-02-18
[ticket/15564] Switch to Invisible reCAPTCHA

senky:
PHPBB3-15564

Invisible reCAPTCHA is great for users but introduces some very ugly changes to the code. If you have something better in mind, please comment.

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15564


phpBB/styles/prosilver/template/captcha_recaptcha.html
```diff
@@ -1,30 +1,48 @@
-<!-- IF S_TYPE == 1 -->
-<div class="panel captcha-panel">
-	<div class="inner">
+<!-- IF S_RECAPTCHA_AVAILABLE -->
+	<noscript>
+		<div>{L_RECAPTCHA_NOSCRIPT}</div>
+	</noscript>
+	<input id="recaptcha_helper" class="hidden">
+	<script src="{RECAPTCHA_SERVER}.js?onload=recaptcha_callback&hl={LA_RECAPTCHA_LANG}" async defer></script>
+	<script>
```
hanakin: is there a reason this is not externalized? This should all be handled via jquery and be located in the core?

phpBB/styles/prosilver/template/captcha_recaptcha.html
```diff
@@ -1,30 +1,48 @@
-<!-- IF S_TYPE == 1 -->
-<div class="panel captcha-panel">
-	<div class="inner">
+<!-- IF S_RECAPTCHA_AVAILABLE -->
+	<noscript>
+		<div>{L_RECAPTCHA_NOSCRIPT}</div>
+	</noscript>
+	<input id="recaptcha_helper" class="hidden">
+	<script src="{RECAPTCHA_SERVER}.js?onload=recaptcha_callback&hl={LA_RECAPTCHA_LANG}" async defer></script>
```
hanakin: is this needed here? can we not just add it to footer with IF S_RECAPTCHA_AVAILABLE

phpBB/assets/javascript/core.js
```diff
@@ -1640,6 +1640,50 @@ phpbb.lazyLoadAvatars = function loadAvatars() {
 	});
 };
 
+// reCAPTCHA-related variables
+var recaptcha_form = $('.g-recaptcha').parents('form'),
+	submit_button = null,
```
hanakin: sorry, I'm working on the guidelines/standards for 3.3 at the moment (using https://github.com/xojs/xo if you're interested) so I'm going to rip this apart lol...
1. always declare vars using keywords pls do not chain
2. might as well use es2015 syntax `recaptcha_form` should be const and the others should be `let`
3. use arrow functions where possible
4. camelCase vars
5. whats with the console logs? I have not tested the code so if its still WIP then ignore...

for simplicity, you can use this.

```
	const recaptchaForm = $('.g-recaptcha').parents('form');
	let submitButton = null;
	let programaticallySubmitted = false;

	phpbb.recaptchaOnLoad = function () {
		console.log('ahoj');
		// Listen to submit buttons in order to know which one was pressed
		$('input[type="submit"]').each(() => {
			$(this).on('click', () => {
				submitButton = this;
			});
		});

		recaptchaForm.on('submit', e => {
			if (!programaticallySubmitted) {
				grecaptcha.execute();
				e.preventDefault();
			}
		});
	}

	phpbb.recaptchaOnSubmit = function () {
		console.log('submit');
		programaticallySubmitted = true;
		// If concrete button was clicked (e.g. preview instead of submit),
		// let's trigger the same action
		if (submitButton) {
			submitButton.click();
		} else {
			// Rename input[name="submit"] so that we can submit the form
			if (typeof recaptchaForm.submit !== 'function') {
				recaptchaForm.submit.name = 'submit_btn';
			}
			recaptchaForm.submit();
		}
	}
```

vinny:
[phpBB/adm/style/captcha_recaptcha.html](https://github.com/phpbb/phpbb/blob/master/phpBB/adm/style/captcha_recaptcha.html#L1~L14) <- It should be updated too.

Also, I believe that some changes in the language file should be made. At the moment there are 3 types of keys:
image
This should be explained by entering the keys in the ACP.

And since the validation is done in the background, this [line](https://github.com/phpbb/phpbb/blob/master/phpBB/language/en/captcha_recaptcha.php#L51) must be updated

phpBB/assets/javascript/core.js
senky: I didn't know we are allowed to use es2015 already. Cool! 😃

senky:
@phpbb/development-team is this my problem why builds are failing?

hanakin:
Looks like there is a up test that may need fixed/removed

senky:
@hanakin it looks like it has nothing to do with my PR (I didn't manipulate permissions), but other PRs doesn't fail on the same test...

hanakin:
It’s somehow related to dropdowns it looks like which is the js code directly before yours

marc1706:
I think the changes are causing some JS side effects which result in the UI tests being unable to select items properly.

hanakin:
@senky most def agree

test.html
```diff
@@ -0,0 +1,1633 @@
+<!DOCTYPE html><html dir="ltr" lang="en-gb"><head>
```
hanakin: did you mean to commit this?

phpBB/assets/javascript/core.js
```diff
@@ -1672,3 +1707,12 @@ $(function() {
 });
 
 })(jQuery); // Avoid conflicts with other libraries
+
+// reCAPTCHA doesn't accept callback functions nested inside objects
+// so we need to make this helper functions here
+function phpbbRecaptchaOnLoad() {
```
hanakin: @senky can we please consider moving these to line 1678 before the window load

phpBB/assets/javascript/core.js
```diff
@@ -1672,3 +1707,12 @@ $(function() {
...
+	phpbb.recaptchaOnLoad();
+}
+function phpbbRecaptchaOnSubmit() {
```
hanakin: how exactly is this being called? its a little confusing what is going on here.

senky: See https://github.com/phpbb/phpbb/pull/5135/files#diff-bbf3aeca1af842b6b834ca6b1e4aca5bR5 and line below. I tried to use `phpbb.recaptchaOnLoad` and `phpbb.recaptchaOnSubmit` but google won't fire those.

senky: I don't get it. What reference? Look, we load a JS file from google servers that will look for `.g-recaptcha[data-callback]` and will fire it's value as JS function. It will do the same with `onload` parameter of said JS file from google servers.

hanakin:
That’s what I was trying to understand cause I do not see the reference to that

senky:
@hanakin still something to change here or are we ok?

hanakin:
I have see no apparent coding issues, I have not been able to test that it works do to needing to spin up a new dev environment. So if someone else can verify then it’s ok from my perspective.

CHItA:
Please rebase instead of merging master into the branch.

marc1706:
Yes, ideally you should be rebasing like this (assuming the main phpBB repo is upstream in your local copy):
`git rebase upstream/master`

marc1706:
I was trying to test this and ended up getting this answer constantly:
`The solution you provided was incorrect`
Thoughts?

CrazyC0d3r:
I just tested this on a demo board and there is an issue when someone tries to login with an incorrect password more than three times...

> You exceeded the maximum allowed number of login attempts. In addition to your username and password you now also have to solve the CAPTCHA below.

There is no CAPTCHA displayed to solve.

senky:
> That user is effectively permanently locked out of their account now.

On the contrary, captcha is present (see the logo in down right corner), you just have to enter correct credentials and hit Login.

CrazyC0d3r:
lol no that's not how it works at all. That is not the captcha, a popup should be displayed where you select all the pictures with a bus or taxi etc...

Entering the correct password results in the same error as the captcha is not displayed.

I can upload a video of how invisible captcha works using phpmyadmin as an example if you need to see it in action?

CrazyC0d3r:
Ok, it's working now. Not sure why it's working now? Maybe js was cached or something...

Screenshot 2019-05-09 at 18 10 54

EA117:
As an update to Vinny's earlier comment (https://github.com/phpbb/phpbb/pull/5135#issuecomment-366865521), note that the way in which Google presents "what kind of reCAPTCHA keys do you want to create" to the customer now is:

recaptchav2

So it is not "v2" versus "Invisible" as it was when this issue was originally opened, and as confirmed in Vinny's screen shot.  "Invisible" is being shown as an option of what you can do with your v2 keys, versus using reCAPTCHA v3 keys (which technically is always "invisible" mode, too).

As such, since we're touching the language files anyway, it seems important (or at least very useful) that we would begin asserting "reCAPTCHA v2" when we talk about this reCAPTCHA plug-in.  Because ever since October 2018 when reCAPTCHA v3 was released, phpBB customers have been asking or attempting to use reCAPTCHA v3 keys with our plug-in, since there in no indication in the ACP "you need to enter v2 keys."

If we update to support "Invisible reCAPTCHA", without also including language "we mean specifically reCAPTCHA v2 Invisible though, and not the reCAPTCHA v3 version of Invisible", this point of confusion will unnecessarily remain and perhaps even be exacerbated since "we did finally update the plug-in after reCAPTCHA v3 was released."

senky:
@EA117 thanks for the notice, I updated the language.

PR #5157, opened 2018-03-09
[ticket/15583] Update session time in AJAX request only once a minute

kasimi:
PHPBB3-15583

The check if the `session_page` needs to be updated was done in two places but they didn't check for the same thing:

* line 1618: `$this->update_session_page && $this->data['session_page'] != $this->page['page'])`
* line 1623: `$this->update_session_page && !$request->is_ajax()`

On AJAX requests, the first condition always passed, so the `if` block was entered and the `session_time` was prepared to be sent to the sessions table, but the second condition never passed, resulting in the `session_page` never actually being updated and only the `session_time` mistakenly being updated.

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15583

### DIFF
--- a/phpBB/phpbb/session.php
+++ b/phpBB/phpbb/session.php
@@ -1614,13 +1614,15 @@ class session
 			return;
 		}
 
+		// Do not update the session page for ajax requests, so the view online still works as intended
+		$page_changed = $this->update_session_page && $this->data['session_page'] != $this->page['page'] && !$request->is_ajax();
+
 		// Only update session DB a minute or so after last update or if page changes
-		if ($this->time_now - ((isset($this->data['session_time'])) ? $this->data['session_time'] : 0) > 60 || ($this->update_session_page && $this->data['session_page'] != $this->page['page']))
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/session.php
```diff
@@ -1614,13 +1614,15 @@ public function update_session_infos()
 			return;
 		}
 
+		// Do not update the session page for ajax requests, so the view online still works as intended
+		$page_changed = $this->update_session_page && $this->data['session_page'] != $this->page['page'] && !$request->is_ajax();
+
 		// Only update session DB a minute or so after last update or if page changes
-		if ($this->time_now - ((isset($this->data['session_time'])) ? $this->data['session_time'] : 0) > 60 || ($this->update_session_page && $this->data['session_page'] != $this->page['page']))
+		if ($this->time_now - ((isset($this->data['session_time'])) ? $this->data['session_time'] : 0) > 60 || $page_changed)
```
gooof: (isset($this->data['session_time'])) 
to 
isset($this->data['session_time'])

PR #5158, opened 2018-03-11
[ticket/15586] Fixed adding module the automatic way without specifying modes

kasimi:
PHPBB3-15586

`$data` isn't used anymore in this method. The code that is deleted in this commit causes an undefined index error when adding a module the automatic way when only specifying the `module_basename` key.

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15586

### DIFF
--- a/phpBB/phpbb/db/migration/tool/module.php
+++ b/phpBB/phpbb/db/migration/tool/module.php
@@ -514,12 +514,6 @@ class module implements \phpbb\db\migration\tool\tool_interface
 		// Allow '' to be sent as 0
 		$parent_id = $parent_id ?: 0;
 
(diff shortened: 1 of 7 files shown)


rxu:
Mind adding specific tests fjr this?

kasimi:
I updated the `test_add()` method to also add modules the automatic way, with a specific mode as well as without the `modes` key, meaning all modes contained in the `_info` file.

However, the test passes with and without the fix for this bug because PHPUnit doesn't seem to fail the test if there's a PHP warning or notice, and the unexpected `null` value of the `$data` variable isn't used anyway.

PR #5175, opened 2018-03-25
[ticket/15592] posting.php changed to add a variable allowed_bbcode and posting_atta…

amalnaeem:
…ch_body.html changed to add a function to hide/display place inline button to resolve [ticket/15592]

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15592

### DIFF
--- a/phpBB/assets/javascript/core.js
+++ b/phpBB/assets/javascript/core.js
@@ -1676,3 +1676,19 @@ $(function() {
 });
 
 })(jQuery); // Avoid conflicts with other libraries
+$(function(){
+
+jQuery( document ).ready(function() {
+var allowed_bbcode = jQuery("#inline").val();
+//From here you can the whaterver you like with you js Value
+if(allowed_bbcode == 0)
+{
+jQuery("#inlineButton").style.visibility='hidden';
+}
+else{
+	jQuery("#inlineButton").style.visibility='visible';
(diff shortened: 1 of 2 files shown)


hanakin:
1. you commits do not follow the proper commit format https://area51.phpbb.com/docs/dev/3.2.x/development/git.html 
2. pls do not inject js scripts into html body, this is a way outdated appraoch. This should be handled via binding and the data object, leverage jQuery

amalnaeem:
Can you elaborate briefly the second point? It will be of much help. Thank you

hanakin:
Yes use jQuery selectors and data objects instead of adding the code to the html? The js itself should be in core.js

phpBB/assets/javascript/core.js
```diff
@@ -1676,3 +1676,19 @@ $(function() {
 });
 
 })(jQuery); // Avoid conflicts with other libraries
+$(function(){
+
+jQuery( document ).ready(function() {
```
hanakin: Do not need all the jQuerys just use $ and wrap it in a self invoked function with jQuery passed into like the example above this in the same file

hanakin: Also why are you double wrapping with document ready?

PR #5198, opened 2018-04-15
[ticket/15637] Extract multi-line PHP event descriptions

kasimi:
PHPBB3-15637

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15637

### DIFF
--- a/phpBB/phpbb/event/php_exporter.php
+++ b/phpBB/phpbb/event/php_exporter.php
@@ -281,7 +281,7 @@ class php_exporter
 						$description_line_num++;
 					}
 
-					$description = trim(implode(' ', $description_lines));
+					$description = trim(implode('<br/>', $description_lines));
 
 					if (isset($this->events[$this->current_event]))
 					{
--- a/phpBB/phpbb/event/php_exporter.php
+++ b/phpBB/phpbb/event/php_exporter.php
@@ -281,7 +281,7 @@ class php_exporter
 						$description_line_num++;
 					}
 
-					$description = implode(' ', $description_lines);
+					$description = trim(implode(' ', $description_lines));
 
 					if (isset($this->events[$this->current_event]))
 					{
--- a/phpBB/phpbb/event/php_exporter.php
+++ b/phpBB/phpbb/event/php_exporter.php
@@ -270,8 +270,8 @@ class php_exporter
 					{
 						$description_line = substr(trim($this->file_lines[$description_line_num]), strlen('* '));
 
-						// Reached end of description if line is empty or a tag
-						if (!strlen($description_line) || $description_line[0] == '@')
+						// Reached end of description if line is a tag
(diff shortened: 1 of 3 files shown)


phpBB/phpbb/event/php_exporter.php
```diff
@@ -264,7 +264,23 @@ public function crawl_php_file($file)
...
 					$description_line_num = $this->find_description();
-					$description = substr(trim($this->file_lines[$description_line_num]), strlen('* '));
+					$description_lines = array();
+
+					while (true)
+					{
+						$description_line = substr(trim($this->file_lines[$description_line_num]), strlen('* '));
+
+						// Reached end of description if line is empty or a tag
+						if (!strlen($description_line) || $description_line[0] == '@')
```
Nicofuma: imo we could take everything between the first line and the first tag (or the end of the comment to be safe in case of mal formed comment)

kasimi: Extracting everything until the first tag is a good idea.

Extracting the entire comment block and then removing all tags should be done in a separate PR, since all the other code (finding tags etc) expects each line in a comment to start with a `*` as well.

kasimi: On second thought, do we actually want to extract everything before the first tag? For example, [this event description](https://github.com/phpbb/phpbb/blob/release-3.2.2/phpBB/common.php#L152-L159) is quite long and would clutter the [events list](https://wiki.phpbb.com/Event_List).

Nicofuma: tbh I think it is better to extract everything because there is little chances that the comment will be read otherwise

Nicofuma: will you have time to work on that?

kasimi: Yes. :smile: I tested extracting the description from a block without tags and from a block without a description and in both cases it threw an exception before the code affected by this PR is run, so I think we can assume that all the information is there.

Nicofuma:
looks like new lines and some trimming are missing (see tests results)

phpBB/phpbb/event/php_exporter.php
```diff
@@ -264,7 +264,24 @@ public function crawl_php_file($file)
...
+						if (strlen($description_line) && $description_line[0] == '@')
+						{
+							break;
+						}
+
+						$description_lines[] = $description_line;
+						$description_line_num++;
+					}
+
+					$description = trim(implode(' ', $description_lines));
```
Nicofuma: implode with space or new line?

kasimi: There's weird font size when using `\n`. I'm voting for `<br/>`.

linebreaks

Nicofuma: ok

PR #5213, opened 2018-04-19
[ticket/15649] Allow to define a custom cache directory

Zoddo:
(Experimented) users should be allowed to easily change the location of the cache directory.

Some examples of use:
* Move the cache directory to a ramdisk filesystem, allowing to reduce writes on drives
* On development boards using different config.php files (to easily switch from a "profile" to another one), it allows to use a different cache directory for each profile

This commit allows to define a constant `PHPBB_CACHE_PATH` (ideally, in `config.php`) which will be used to define the path of the cache directory.

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15649

### DIFF
--- a/phpBB/bin/phpbbcli.php
+++ b/phpBB/bin/phpbbcli.php
@@ -47,6 +47,11 @@ require($phpbb_root_path . 'includes/functions_compatibility.' . $phpEx);
 $phpbb_container_builder = new \phpbb\di\container_builder($phpbb_root_path, $phpEx);
 $phpbb_container = $phpbb_container_builder->with_config($phpbb_config_php_file);
 
+if (defined('PHPBB_CACHE_PATH'))
+{
+	$phpbb_container_builder->with_cache_dir(PHPBB_CACHE_PATH);
+}
+
 $input = new ArgvInput();
 
 if ($input->hasParameterOption(array('--env')))
--- a/phpBB/common.php
+++ b/phpBB/common.php
@@ -112,6 +112,12 @@ $phpbb_class_loader_ext->register();
 try
 {
 	$phpbb_container_builder = new \phpbb\di\container_builder($phpbb_root_path, $phpEx);
+
+	if (defined('PHPBB_CACHE_PATH'))
+	{
+		$phpbb_container_builder->with_cache_dir(PHPBB_CACHE_PATH);
+	}
+
 	$phpbb_container = $phpbb_container_builder->with_config($phpbb_config_php_file)->get_container();
 }
 catch (InvalidArgumentException $e)
--- a/phpBB/download/file.php
+++ b/phpBB/download/file.php
@@ -63,6 +63,12 @@ if (isset($_GET['avatar']))
 
 	// Set up container
 	$phpbb_container_builder = new \phpbb\di\container_builder($phpbb_root_path, $phpEx);
+
+	if (defined('PHPBB_CACHE_PATH'))
+	{
+		$phpbb_container_builder->with_cache_dir(PHPBB_CACHE_PATH);
+	}
+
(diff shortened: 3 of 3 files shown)


Zoddo:
This was discussed with Nicofuma to be included in 3.2.x (even if it can be considered as a new feature)

rubencm:
Why don't create a parameter in the config files with it?

Zoddo:
I think using a constant is more consistant with other values such as `PHPBB_ENVIRONMENT`.

But if you think a variable in `config.php` is better, I can use that instead :)

rubencm:
No, i mean a parameter in the yml files, but i don't know, i think the idea was to move the config.php contants to the yml config files:
https://tracker.phpbb.com/browse/PHPBB3-12623
For example
https://tracker.phpbb.com/browse/PHPBB3-12624

marc1706:
Correct, we should aim to use the container parameters instead of constants.

Zoddo:
Using a container parameter instead of the constant would means users have to modify a phpBB's core file to redefine the cache directory.

Actually, I think this should be configurable without changing any phpBB's files, which mean using the `config.php` file (otherwise, they can just patch the 3 files like I've done in this PR).

If the real problem is the use of a constant, I can change that to use a standard variable (like `$acm_type` or `$phpbb_adm_relative_path`). I used a constant because I though it may helps in some exotic conditions.

senky:
Definitely in `config.php`. Just like `$phpbb_adm_relative_path`, we should have `$phpbb_cache_path` (not relative, it can be out of phpBB root dir).

Zoddo:
@senky Or absolute if it starts with a slash, relative otherwise, defaulted to `$phpbb_cache_path = 'cache/';`

This way, a user using the default value can move his board without having to change this settings.

Zoddo:
!set wip

Zoddo:
!set WIP

Crizz0:
!set WIP

Zoddo:
!set WIP :construction:

rubencm:
I am thinking, and cache folder is a configuration option only for the cache file driver, and here is configured globally, while other cache drivers have their configuration options hardcoded.

Zoddo:
> cache folder is a configuration option only for the cache file driver

Not only. The cache driver is used only for caching data and sql queries. Twig and the container will always use the cache directory, even if another driver is used (like memcache or dummy).

Nicofuma:
The issue when we are talking about cache is that we are using 2 different kind of cache:
 - generated PHP files for Twig, the container and the router. They are always written to the disk
 - object cache used for the SQL queries and the cache service in general. It can use files or many other options like ACP or memcache

For me it should be 2 distinct things with 2 different settings if we want to allow to easily change both:
 - to change the directory of the generated PHP files we need a constant or something like that otherwise it can't be picked up by the container (before we build the container we search it in the cache so the directory can't be defined only in the container itself). But we should also store this path in the container for other generated files (like twig and the router) ; it can be injected the same way as the root_dir or the environment (see the container_builder.php file)
 - to change the object cache dir it should be defined in the config.yml file using semantic configuration

Note: we could also add semantic configuration to define twig and router cache dir with a default value of %core.container_cache_dir% (or any name we will use fot that)

j-angnoe:
Hello,

It is possible to manipulate the cache directory (in the container) via $_SERVER[PHPBB__*], like so:

```php
// in config.php for example:
$_SERVER['PHPBB____core__cache_dir'] = "/path/to/cache";
```

This way cached objects, queries and twig templates will be written to a different cache directory. The autoload and container cache will still be written to phpbb_root_path/cache/<env> due to some inconsistencies inside the container_builder. I'm working on a PR to resolve this, which will arrive shortly.

marc1706:
Obsoleted by solution in https://github.com/phpbb/phpbb/pull/6080

PR #4006, opened 2015-10-25
[feature/sql-bool-builder] New syntax for DBAL query builder for boolean generation

brunoais:
Applying the new syntax as told by Nicofuma in IRC

https://tracker.phpbb.com/browse/PHPBB3-13652

PR #5222, opened 2018-05-11
[ticket/9837] Display unapproved posts to their authors

CHItA:
Checklist:

- [ ] Tests pass
- [ ] Code follows coding guidelines

**TODOs**
- [ ] Review and finalize styling styling
- [ ] Handle anonymous users (exclude them)
- [ ] Optimize query caching
- [ ] Fix displayed values (viewforum shows the topic with -1 replies)
- [ ] Add tests

https://tracker.phpbb.com/browse/PHPBB3-9837

### DIFF
--- a/phpBB/phpbb/content_visibility.php
+++ b/phpBB/phpbb/content_visibility.php
@@ -144,7 +144,12 @@ class content_visibility
 	*/
 	public function is_visible($mode, $forum_id, $data)
 	{
-		$is_visible = $this->auth->acl_get('m_approve', $forum_id) || $data[$mode . '_visibility'] == ITEM_APPROVED;
+		$visibility = $data[$mode . '_visibility'];
+		$poster_key = ($mode === 'topic') ? 'topic_poster' : 'poster_id';
+		$is_visible = $this->auth->acl_get('m_approve', $forum_id) || $visibility == ITEM_APPROVED || (
+			($visibility == ITEM_UNAPPROVED || $visibility == ITEM_REAPPROVE) &&
+			$this->user->data['user_id'] === $data[$poster_key]
+		);
 
 		/**
 		* Allow changing the result of calling is_visible
@@ -216,7 +221,13 @@ class content_visibility
 		}
 		else
 		{
-			$where_sql .= $table_alias . $mode . '_visibility = ' . ITEM_APPROVED;
+			$field_name = ($mode === 'topic') ? 'topic_poster' : 'poster_id';
+			$visibility_query = $table_alias . $mode . '_visibility = ';
+
+			$where_sql .= '(' . $visibility_query . ITEM_APPROVED . ')';
+			$where_sql .= ' OR (';
+			$where_sql .= '(' . $visibility_query . ITEM_UNAPPROVED . ' OR ' . $visibility_query . ITEM_REAPPROVE . ')';
+			$where_sql .= ' AND ' . $table_alias . $field_name . ' = ' . ((int) $this->user->data['user_id']) . ')';
 		}
 
 		return '(' . $where_sql . ')';
--- a/phpBB/phpbb/content_visibility.php
+++ b/phpBB/phpbb/content_visibility.php
@@ -146,9 +146,10 @@ class content_visibility
 	{
 		$visibility = $data[$mode . '_visibility'];
 		$poster_key = ($mode === 'topic') ? 'topic_poster' : 'poster_id';
-		$is_visible = $this->auth->acl_get('m_approve', $forum_id) || $visibility == ITEM_APPROVED || (
-			($visibility == ITEM_UNAPPROVED || $visibility == ITEM_REAPPROVE) &&
-			$this->user->data['user_id'] === $data[$poster_key]
+		$is_visible = $this->auth->acl_get('m_approve', $forum_id) || $visibility == ITEM_APPROVED;
+		$is_visible = $is_visible || (
+			($visibility == ITEM_UNAPPROVED || $visibility == ITEM_REAPPROVE)
+			&& $this->user->data['user_id'] === $data[$poster_key]
 		);
 
 		/**
--- a/phpBB/styles/prosilver/template/viewtopic_body.html
+++ b/phpBB/styles/prosilver/template/viewtopic_body.html
@@ -306,7 +306,7 @@
 				</p>
 			</form>
 			<!-- ELSE -->
-				<p class="information post-notice">
+				<p class="post-notice information">
 					{L_POST_UNAPPROVED}
 				</p>
 			<!-- ENDIF -->
--- a/phpBB/styles/prosilver/template/viewtopic_body.html
+++ b/phpBB/styles/prosilver/template/viewtopic_body.html
@@ -294,6 +294,7 @@
 			<!-- EVENT viewtopic_body_postrow_post_details_after -->
 
 			<!-- IF postrow.S_POST_UNAPPROVED -->
+			<!-- IF postrow.S_CAN_APPROVE -->
 			<form method="post" class="mcp_approve" action="{postrow.U_APPROVE_ACTION}">
 				<p class="post-notice unapproved">
 					<span><i class="icon fa-question icon-red fa-fw" aria-hidden="true"></i></span>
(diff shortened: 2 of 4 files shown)


CHItA:
@hanakin When you have some spare time could you look at the Notice styling?

Also @phpbb/development-team if anyone has any inputs on how this should work and if I'm missing something any pointers would be much appriciated.

phpBB/styles/prosilver/theme/colours.css
```diff
@@ -1151,3 +1151,7 @@ input.disabled {
 	background-color: #d41142;
 	color: #ffffff;
 }
+
+p.information {
```
hanakin: loose the p

phpBB/styles/prosilver/template/viewtopic_body.html
```diff
@@ -304,6 +305,11 @@ <h3 <!-- IF postrow.S_FIRST_ROW -->class="first"<!-- ENDIF -->><!-- IF postrow.P
 					{S_FORM_TOKEN}
 				</p>
 			</form>
+			<!-- ELSE -->
+				<p class="information post-notice">
```
hanakin: post-notice should come first so information can override it.

phpBB/phpbb/content_visibility.php
```diff
@@ -144,7 +144,13 @@ public function get_count($mode, $data, $forum_id)
 	*/
 	public function is_visible($mode, $forum_id, $data)
 	{
-		$is_visible = $this->auth->acl_get('m_approve', $forum_id) || $data[$mode . '_visibility'] == ITEM_APPROVED;
+		$visibility = $data[$mode . '_visibility'];
+		$poster_key = ($mode === 'topic') ? 'topic_poster' : 'poster_id';
+		$is_visible = $this->auth->acl_get('m_approve', $forum_id) || $visibility == ITEM_APPROVED;
+		$is_visible = $is_visible || (
+			($visibility == ITEM_UNAPPROVED || $visibility == ITEM_REAPPROVE)
+			&& $this->user->data['user_id'] === $data[$poster_key]
```
Nicofuma: if I'm not mistaken, this is gonna show unapproved anonymous posts to anonymous users, right?

CHItA: As it is now yes. I added it to the TODO list to handle it. I'm really just trying to mock up what changes are needed first so I can have a better idea on what changes would the database tables need to make this a bit more optimized (as of now all users would have a separate query which doesn't seem optimal).

phpBB/styles/prosilver/template/viewtopic_body.html
lavigor: @hanakin could you please give an article or something about this change please?
Because [this SO answer](https://stackoverflow.com/a/3066365) as well as other ones state clearly that the order of classes defined in HTML is not important for overriding at all.

phpBB/phpbb/content_visibility.php
```diff
@@ -144,7 +144,13 @@ public function get_count($mode, $data, $forum_id)
 	*/
 	public function is_visible($mode, $forum_id, $data)
 	{
-		$is_visible = $this->auth->acl_get('m_approve', $forum_id) || $data[$mode . '_visibility'] == ITEM_APPROVED;
+		$visibility = $data[$mode . '_visibility'];
+		$poster_key = ($mode === 'topic') ? 'topic_poster' : 'poster_id';
+		$is_visible = $this->auth->acl_get('m_approve', $forum_id) || $visibility == ITEM_APPROVED;
```
gooof: to optimize this use
``$visibility == ITEM_APPROVED || $this->auth->acl_get('m_approve', $forum_id)``

phpBB/phpbb/content_visibility.php
gooof: ``$this->user->data['user_id'] === $data[$poster_key]`` first? If is not the user, it not performs the 2 checks.

phpBB/phpbb/content_visibility.php
```diff
@@ -216,7 +222,13 @@ public function get_visibility_sql($mode, $forum_id, $table_alias = '')
 		}
 		else
 		{
-			$where_sql .= $table_alias . $mode . '_visibility = ' . ITEM_APPROVED;
+			$field_name = ($mode === 'topic') ? 'topic_poster' : 'poster_id';
+			$visibility_query = $table_alias . $mode . '_visibility = ';
+
+			$where_sql .= '(' . $visibility_query . ITEM_APPROVED . ')';
+			$where_sql .= ' OR (';
+			$where_sql .= '(' . $visibility_query . ITEM_UNAPPROVED . ' OR ' . $visibility_query . ITEM_REAPPROVE . ')';
```
gooof: This need a switched off, its not good for large databases. There is no optimization possible.
```php
			$visibility_query = $table_alias . $mode . '_visibility = ';

			$where_sql .= $visibility_query . ITEM_APPROVED;
			if ($this->config['show_items_unapproved'])
			{
				$field_name = ($mode === 'topic') ? 'topic_poster' : 'poster_id';
				$where_sql .= ' OR (';
				$where_sql .= $db->sql_in_set($visibility_query, array(ITEM_UNAPPROVED, ITEM_REAPPROVE));
				$where_sql .= ' AND ' . $table_alias . $field_name . ' = ' . ((int) $this->user->data['user_id']) . ')';
			}

File: phpBB/phpbb/content_visibility.php [hunk 6]
```diff
@@ -216,7 +222,13 @@ public function get_visibility_sql($mode, $forum_id, $table_alias = '')
...
+			$where_sql .= ' AND ' . $table_alias . $field_name . ' = ' . ((int) $this->user->data['user_id']) . ')';
```
brunoais: May you please change this to the [query builder](https://github.com/phpbb/phpbb/pull/4006) to allow extensions to more easily change this query for their own needs?

v12mike:
More than a year since the last activity...   Is this still active?  Does it need someone else to pick it up?

CHItA:
Feel free to pick it up, but probably viewtopic and viewforum will move to controllers in 4.0 so this is on the back burner for the moment.

phpBB/phpbb/content_visibility.php
v12mike: This query is already using query builder (in the calling function), here we are just building a WHERE clause for query builder.

phpBB/phpbb/content_visibility.php
v12mike: A config switch will be in my next PR, need to decide whther to default on (my preference) or off.

phpBB/phpbb/content_visibility.php
brunoais: @v12mike query builder: https://github.com/phpbb/phpbb/pull/4006
https://area51.phpbb.com/phpBB/viewtopic.php?t=47306

v12mike: There is an existing event 'core.phpbb_content_visibility_get_visibility_sql_before' in the middle of building the sql, which I think would get broken  if the sql builder was used more extensively here.  Can you see how to do it?

rxu:
What will happen if you set guest posts approval? Will all guests see unapproved guest posts?

v12mike:
> What will happen if you set guest posts approval? Will all guests see unapproved guest posts?

I have updated the code so that guests never see unapproved posts, otherwise it would require session tracking to determine which guest should be able to see it, which is just too difficult.

v12mike:
I should within a few days be able to make a PR with the updated code.  I still need to check the cosmetics and write some tests.

rxu:
> otherwise it would require session or IP address tracking to determine which guest should be able to see it

No need of it, guests just should never see unapproved stuff at all, as well as bots etc.

phpBB/phpbb/content_visibility.php
brunoais: Good point. I failed to see it.
It's sad but...
I rest my case.

brunoais:
There's no reliable way to know if two guests are the same, so I agree that you can just leave them not being able to see the pending posts

phpBB/phpbb/content_visibility.php
v12mike: Would it help sql optimisation if the visibility of unapproved posts to posters had a time limit (only unapproved posts in the last X days)?

brunoais: Even if it does. I believe it would be quite bad for users to have the post disappear. They could think that it had been deleted....

PR #5223, opened 2018-05-14
[ticket/15664] Manage extensions with ajax

senky:
Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15664


phpBB/adm/style/acp_ext_actions.html
```diff
@@ -0,0 +1,3 @@
+{% for action in ext.actions %}
+<a href="{action.U_ACTION}"<!-- IF action.L_ACTION_EXPLAIN --> title="{action.L_ACTION_EXPLAIN}"<!-- ENDIF --><!-- IF action.COLOR -->  style="color: {action.COLOR};"<!-- ENDIF --> data-ajax="{action.ACTION_AJAX}">{action.L_ACTION}</a>{% if not action.S_LAST_ROW %}&nbsp;|&nbsp;{% endif %}
```
3D-I: Any reason why is not fully Twig'ed?
Same for the rest.

lavigor:
I understand that it is highly unlikely that a user does not have JS, but what about accessibility?
This PR seems to drop standard enabling/disabling process in favour of Ajax.
It makes enabling/disabling process impossible if JS is disabled, however it is not stated as a must-have feature in phpBB requirements.

/Unlike Upload Extensions that works fully without JS too./

phpBB/adm/style/acp_ext_list.html
```diff
@@ -42,59 +42,50 @@ <h1>{L_EXTENSIONS_ADMIN}</h1>
...
 	</thead>
 	<tbody>
-		<!-- IF .enabled -->
-		<tr>
+		
+		<tr id="ext_enabled_header"<!-- IF not .enabled --> class="hidden"<!-- ENDIF -->>  
 			<td class="row3" colspan="4"><strong>{L_EXTENSIONS_ENABLED}</strong><!-- EVENT acp_ext_list_enabled_title_after --></td>
 		</tr>
-		<!-- BEGIN enabled -->
+		{% for ext in enabled %}
```
VSEphpbb: I don't think it's a good idea to have mixed template syntaxes. We should probably just stick with old syntax, ugly as it is. Also it looks like you're just using twig to rename the loop array fron `enabled` to `ext` so I don't think this may be necessary.

phpBB/includes/acp/acp_extensions.php
```diff
@@ -1026,7 +1073,11 @@ private function output_actions($block, $actions)
 			}
 
 			$this->template->assign_block_vars($block . '.actions', $vars);
+
+			$vars_ary[] = $vars;
 		}
+
+		return $vars_ary;
```
VSEphpbb: Should be added to doc block

phpBB/includes/acp/acp_extensions.php
```diff
@@ -1006,6 +1051,7 @@ public function list_available_exts(\phpbb\extension\manager $phpbb_extension_ma
 	*/
 	private function output_actions($block, $actions)
 	{
+		$vars_ary = [];
```
VSEphpbb: if using short array syntax, maybe use short array syntax in the entire function for consistency.

phpBB/adm/style/acp_ext_list.html
senky: Yes but that allowed me to use only one loop in `acp_ext_actions.html`. Now it needs 2 loops.

Nicofuma:
seems good. @VSEphpbb ?

PR #5224, opened 2018-05-16
[ticket/15667] Fix first char keys in MemberList

hubaishan:
Fixing memberlist to get first_char keys from `Users` table. So it will support any language, and it will show only used characters.
Edit in last commit:
- Adding first characters to cache to avoid reduce performance.
- Change `includes/functions_user.php` to destroy the cache if new user is added or current user name  is changed.
- Limit first characters to 50 characters for the most 50 characters used in users table.
- When `users first characters` reaches the limit word `other` will be added.
- Change `other` query for more performance and shorter SQL statement.

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15667

### DIFF
--- a/phpBB/includes/functions_user.php
+++ b/phpBB/includes/functions_user.php
@@ -298,7 +298,7 @@ function user_add($user_row, $cp_data = false, $notifications_data = null)
 	$user_id = $db->sql_nextid();
 
 	//	Check users_first_characters
-	$cache = $phpbb_container->get('cache');
+	$cache = $phpbb_container->get('cache.driver');
 	$users_first_characters = $cache->get('users_first_characters');
 	if (($users_first_characters !== false) && (! array_key_exists(mb_substr($username_clean, 0, 1), $users_first_characters)))
 	{
--- a/phpBB/includes/functions_user.php
+++ b/phpBB/includes/functions_user.php
@@ -185,7 +185,7 @@ function user_update_name($old_name, $new_name)
 */
 function user_add($user_row, $cp_data = false, $notifications_data = null)
 {
-	global $db, $config, $cache;
+	global $db, $config;
 	global $phpbb_dispatcher, $phpbb_container;
 
 	if (empty($user_row['username']) || !isset($user_row['group_id']) || !isset($user_row['user_email']) || !isset($user_row['user_type']))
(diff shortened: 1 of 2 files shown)


phpBB/memberlist.php
```diff
@@ -1318,19 +1311,36 @@
 
 		$first_characters = array();
 		$first_characters[''] = $user->lang['ALL'];
-		for ($i = 97; $i < 123; $i++)
+
+		if ($db->get_sql_layer() === 'sqlite3')
+		{
+			$sql = 'SELECT DISTINCT substr(username_clean,1,1) first_char
```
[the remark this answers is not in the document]
hubaishan: `left` is not SQLite or Oracle function.
`substr` work in MySQL, PostgreSQL, Oracle and SQLite, NOT SQL Server.
`substring` is standard work in MySQL, PostgreSQL, Oracle and SQL Server, NOT SQLite.

DavidIQ:
This does not work the way you think it will nor in a way that I think we would like it to work. When I run this query against phpBB.com's database I get back 181 rows:

```sql
SELECT DISTINCT substring(username_clean, 1, 1) first_char
FROM phpbb_users
WHERE user_type IN (0, 1, 3)
ORDER BY first_char;
```

Plus it's pretty inefficient taking almost an entire second just to run this query.

phpBB/includes/functions_user.php
```diff
@@ -288,6 +289,12 @@ function user_add($user_row, $cp_data = false, $notifications_data = null)
 
 	$user_id = $db->sql_nextid();
 
+	//	Check users_first_characters
+	$users_first_characters = $cache->get('users_first_characters');
+	if (($users_first_characters !== false) && (! array_key_exists(mb_substr($username_clean, 0, 1), $users_first_characters)))
+	{
+		$cache->destroy('users_first_characters');
```
VSEphpbb: Instead of adding another global variable for the cache, use the container since it's available, so:
```php
$phpbb_container->get('cache')->destroy('users_first_characters');
```

VSEphpbb: Then just do `$cache = $phpbb_container->get('cache')` instead of `global $cache`

hubaishan: build failed
phpbb_console_user_add_test::test_add_no_dialog
Exception: Could not find service: cache

VSEphpbb: try getting `cache.driver` instead of `cache`

hubaishan:
This PR tested in also in oracle and mssql odbc

PR #5225, opened 2018-05-17
[ticket/13713] User Mentions (GSoC 2018)

lavigor:
This Pull Request is the result of GSoC 2018 Project “User Mentions”.

## The Idea
This project is intended to add the ability to mention specific users in forum topics so that they will receive a notification (just like in many modern messengers and websites).

## Affected modules
This PR is backwards-compatible and works well with phpBB 3.3's BBCode parser (that was newly added in phpBB 3.2).

## How it works
Users can type the `@` symbol in textarea fields on the board to trigger a dropdown with usernames and group names. This list is generated based on several factors (called 'sources'): e.g. friends and topic repliers are shown higher in the list (internally they are given higher priorities).

Besides general user and group sources, built-in name sources include:
* friends,
* topic repliers (open poster receives even higher priority),
* board team members,
* groups that current user is a member of.

Extension developers can add their own sources to give certain users and groups higher priorities based on other conditions.

Special cases taken into consideration:

| Case | Solution |
| --- | --- |
| Users should be able to mention any users/groups regardless of the priorities | Exact matches are always shown on the top the list |
| In most cases users should not mention themselves | Name of the current user is removed from the list |
| Both usernames and group names should have equal chances of appearing in the list | Priorities for usernames and group names are normalized based on the highest priority value |

The name of mentioned user/group becomes a link to its profile highlighed in the colour of that user/group.

## Configuration
Users are able to disable mention notifications in the same way as for other types of notifications (on Edit notification options page in User Control Panel).
Administrators can enable/disable the entire mentions feature and set the maximum number of names shown in the dropdown list as well as the maximum number of names fetched from the server on each request (on Post settings page in Administration Control Panel).

## Background
This feature uses the [At.js](https://github.com/ichord/At.js) library for displaying mention dropdown.
It also involves names caching both on frontend and backend to improve performance and reduce server load.

| Screenshot |
| --- |
| default |

| Unparsed BBCode | Parsed BBCode |
| --- | --- |
| default | default |

## Current Status
User Mentions work. There'll be a PR to development documentation for how to add a new name source in your extension.

-----------------------------------------------------------

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-13713


lavigor:
!set WIP

lavigor:
!set WIP

lavigor:
Seems that the command no longer works...

rubencm:
`!set WIP :construction:`

Persian-phpBB:
well done ...

phpBB/config/default/routing/routing.yml
```diff
@@ -24,5 +24,10 @@ phpbb_help_routing:
     resource: help.yml
     prefix: /help
 
+phpbb_mention_controller:
+    path: /mention
```
gooof: Possible to rename the path like ``mention_json`` or ``mention_api``?

lavigor: I don't think there's a need in introducing a suffix with underscore.

phpBB/styles/prosilver/template/posting_buttons.html
```diff
@@ -5,6 +5,8 @@
 	var text_name = <!-- IF $SIG_EDIT -->'signature'<!-- ELSE -->'message'<!-- ENDIF -->;
 	var load_draft = false;
 	var upload = false;
+	var mention_url = '{UA_MENTION_URL}';
```
hanakin: I would prefer that these were handled via data attributes on a hidden input or other form of html tag to avoid using inline js cause eventually this will all be factored out. Might as well prep for it

rubencm:
You have a problem with phpBB folder in lowercase

senky:
Isn't so much queries adding too much load to the server? Isn't caching the queries a middle way between server load and accuracy of the suggestions (new users/groups will pop up only after cache is past ttl)? Was this discussed somewhere?

lavigor:
@senky caching queries will also be done just a bit later. That part should be one of the most complicated to decide.

@rubencm any idea how to fix this? Probably rebasing can help?

phpBB/styles/prosilver/template/bbcode.html
```diff
@@ -8,6 +8,22 @@
 <!-- BEGIN listitem --><li><!-- END listitem -->
 <!-- BEGIN listitem_close --></li><!-- END listitem_close -->
 
+<!-- BEGIN mention -->
```
hanakin: is this temp code? the xsl stuff needs to be extrapolated out into the core. we should not be shipping any xsl in the bbcode.html file. Joshy needs to remove what he left there...

phpBB/phpbb/textformatter/s9e/mention_helper.php
```diff
@@ -25,18 +32,74 @@ class mention_helper
...
+			$query = $this->db->sql_build_query('SELECT', [
+				'SELECT' => 'u.user_colour, u.user_id',
+				'FROM'   => [
+					USERS_TABLE => 'u',
+				],
+				'WHERE'  => 'u.user_id <> ' . ANONYMOUS . '
+				AND ' . $this->db->sql_in_set('u.user_type', [USER_NORMAL, USER_FOUNDER]) . '
+				AND ' . $this->db->sql_in_set('u.user_id', $user_ids),
+			]);
+			$res = $this->db->sql_query($query);
```
senky: A good habit is to use `$result` as a variable name.

phpBB/phpbb/textformatter/s9e/mention_helper.php
```diff
@@ -25,18 +32,74 @@ class mention_helper
...
+
+			while ($row = $this->db->sql_fetchrow($res))
+			{
+				$this->cached_colors['users'][$row['user_id']] = $row['user_colour'];
+			}
```
senky: `$this->db->sql_freeresult($res);`

phpBB/phpbb/textformatter/s9e/mention_helper.php
```diff
@@ -25,18 +32,74 @@ class mention_helper
...
+					GROUPS_TABLE => 'g',
+				],
+				'WHERE'  => $this->db->sql_in_set('g.group_id', $group_ids),
+			]);
+			$res = $this->db->sql_query($query);
+
+			while ($row = $this->db->sql_fetchrow($res))
+			{
+				$this->cached_colors['groups'][$row['group_id']] = $row['group_colour'];
+			}
```
senky: `$this->db->sql_freeresult($res);`

phpBB/adm/style/acp_posting_buttons.html
```diff
@@ -1,5 +1,7 @@
 <script type="text/javascript">
 // <![CDATA[
+	var mention_url = '{UA_MENTION_URL}';
+	var mention_topic_id = '{S_TOPIC_ID}';
```
senky: `S_` means `switch`, i.e. boolean value. Just use `{TOPIC_ID}` which is maybe already defined.

lavigor: @senky `S_TOPIC_ID` is indeed already defined in `functions.php`.
That is why it is used here.

lavigor: According to guidelines, `S_` stands for `system`, not for `switch`.

phpBB/styles/prosilver/template/posting_buttons.html
lavigor: @hanakin so even very huge scripts such as the one in [plupload.html](https://github.com/phpbb/phpbb/blob/master/phpBB/styles/prosilver/template/plupload.html) will eventually be converted to `data-` attributes or something similar?
Requiring additional code to parse them back to the original JS objects?

I haven't made any performance checks or tests but using a script block seems to me as a rather convenient solution. Probably there should be one inline block at the bottom instead of many other, requiring some additional template features like INCLUDEJS for accumulating script data.

IMHO I'll better leave it as is right now so that it can be changed the same way together with other variables when that change rolls out.

CHItA: The aim would be to get rid of inline js all together.

CHItA: The new theme will probably use some build system for both JS and CSS by default, which will allow us to use some nice features (minification, auto-prefixing etc). This is obviously not usable on inlined JS or CSS.

Probably there will be some backend changes to support some features both for the new theme itself and extensions as well. However, these changes are not yet fully specified so we will probably publish our recommendations for extensions when we release those changes.

lavigor: @CHItA can't wait for those changes to be done :)
Are there any plans for a preprocessor (LESS/SASS) instead of plain CSS?

phpBB/styles/prosilver/template/bbcode.html
lavigor: I wasn't aware of Joshy's plans for that and this place seemed to be the right.
Moved to the core.

There are also many legacy stuff like first and second passes I suppose.
Sometimes it is not obvious to distinguish the legacy, and I suppose that new BBCodes should not be handled by that code.

phpBB/phpbb/textformatter/s9e/mention_helper.php
lavigor: I thought that was common for phpBB but now I realize that this habit probably came from extensions.
Changed to `$result`.

phpBB/phpbb/textformatter/s9e/mention_helper.php
lavigor: Done.

phpBB/phpbb/textformatter/s9e/mention_helper.php
lavigor: Done.

phpBB/adm/style/acp_posting_buttons.html
senky: Ah right, then it's OK.

phpBB/styles/prosilver/template/posting_buttons.html
hanakin: There is never a reason to use inline scripts...just port it to its own file and include it using includejs. Using data attr for any template bar stuff etc... the use of minificatiin in the new theme coupled with the adoption of html2 will mean we can gloving most everything anyway at little cost. 
The new theme will be a completely different beast and will not be compatible with any extensions now nor should as it will be released under 4.0.0 meaning no compatibility. 

I am also looking into vuejs as an additional aspect as working with twig alone is problematic for certain circumstances where vue components just make it effortless and we gain the added benifit if less js for most of the interactivity.

lavigor: @hanakin the only reason I use inline scripts is exactly for template variables that cannot be inserted to the plain JS.
Almost all other JS stuff goes with `INCLUDEJS`.

Looking forward for new changes though.

hanakin: yes but you can instead use data for that and move everything to your js file.

phpBB/includes/constants.php
```diff
@@ -190,6 +190,7 @@
 define('BBCODE_ID_EMAIL', 10);
 define('BBCODE_ID_FLASH', 11);
 define('BBCODE_ID_ATTACH', 12);
+define('BBCODE_ID_MENTION', 13);
```
paul999: How is this going to work out with users having existing custum_bbcodes? Do they start with ID 13? If so, this might cause problems I assume with those. Maybe even a idea to define it completly as custom bbcode instead of hardcoded in core

lavigor: I suppose that existing custom BBCodes don't have the hardcoded constant values.
But probably this constant can be removed entirely because it does not seem that I need to use it anywhere with the new parser. It could be suitable for deprecated stuff that is not necessary with the newly added BBCode.

paul999: The issue is that the first created custom_bbcode is ID 13: 
image
No idea if that can cause issues, but I can see it might happen.

phpBB/assets/javascript/editor.js
```diff
@@ -389,11 +389,32 @@ function getCaretPosition(txtarea) {
...
 (function($) {
+	function handle_mentions(txtarea) {
+		var $mentionParams = $('#mention_params'),
+			mentionURL = $mentionParams.data('mentionUrl'),
+			mentionNamesLimit = $mentionParams.data('mentionNamesLimit'),
+			mentionTopicId = $mentionParams.data('topicId');
+		$(txtarea).atwho({
+			at: "@",
+			displayTpl: function(data) {
+				var avatar = (data.avatar.src) ? "<img src='" + data.avatar.src + "'>" :
```
hanakin: Why not use es6 template structure here https://caniuse.com/#search=Template%20literal

lavigor:
@hanakin what about `admin.css`? It needs the styles to be duplicated/included.
Maybe I can just leave it as is.

As for ES6, PhpStorm is complaining every time about `var` 😊 
I just used the same style as in other places across the file.
Could I maybe refactor the script separately?

P.S. Needed to type this reply the second time, mobile Internet chewed my text 🙄

lavigor:
@hanakin what is the target IE version for 3.3 BTW?
It was IE 8 previously, does your link mean that even IE 11 would not be supported from now on?

hanakin:
No do not split admin.css my bad. We no longer really use our lint for js as it’s antequated and broken. I am working on a fix to switch to xo which is based on raking which is far better. So for now maybe externalize everything into a function call so we can easily refactor

phpBB/adm/style/acp_posting_buttons.html
```diff
@@ -29,6 +29,11 @@
 // ]]>
 </script>
 
+<!-- IF S_ALLOW_MENTIONS -->
+<div id="mention_params" style="display: none;" data-mention-url="{U_MENTION_URL}" data-mention-names-limit="{S_MENTION_NAMES_LIMIT}" data-topic-id="{S_TOPIC_ID}"></div>
```
hanakin: just add the data to the form-buttons div

phpBB/styles/prosilver/template/posting_buttons.html
```diff
@@ -49,6 +49,12 @@
 
 // ]]>
 </script>
+
+<!-- IF S_ALLOW_MENTIONS -->
+<div id="mention_params" style="display: none;" data-mention-url="{U_MENTION_URL}" data-mention-names-limit="{S_MENTION_NAMES_LIMIT}" data-topic-id="{S_TOPIC_ID}"></div>
```
hanakin: same here just add the data to the format buttons div

lavigor:
@hanakin 
Screenshot:
default
Rendered HTML:
```html
<div class="atwho-container">
	<div id="atwho-ground-message">
		<div class="atwho-view" id="at-view-64" style="display: none; top: 356.267px; left: 83.4503px;">
			<ul class="atwho-view-ul">
				<li class="mention-name cur">
					<span class="mention-avatar"><i class="fa fa-user"></i></span><span>test</span>
				</li>
				<li class="mention-name">
					<span class="mention-avatar"><i class="fa fa-group"></i></span><span>Administrators</span>
				</li>
				<li class="mention-name">
					<span class="mention-avatar"><i class="fa fa-group"></i></span><span>Global moderators</span>
				</li>
				<li class="mention-name">
					<span class="mention-avatar"><i class="fa fa-user"></i></span><span>LavIgor</span><span class="mention-rank">Site Admin</span>
				</li>
			</ul>
		</div>
	</div>
</div>
```

>  we are adopting >1% , last 2 browsers

According to caniuse.com, IE 11 complies with this requirement. As we've discussed with @CHItA IE is actually the same as Edge and is no longer supported, but I personally use it sometimes. 😊 

> No do not split admin.css

Should I split regular styles or leave them all as is now?

> maybe externalize everything into a function call

I've done that already, the function is named `handle_mentions`.

hanakin:
@lavigor for prosilver only separate out the colors, put the rest in a new css file called mention.css and import it before colours.css please

phpBB/styles/prosilver/theme/content.css
```diff
@@ -557,6 +557,11 @@ blockquote .codebox {
 	padding: 5px 3px;
 }
 
+/* Mention block */
+.mention {
```
hanakin: where is this used?

hanakin:
also, I have worked up some changes... let me know if you have any issues

https://codepen.io/hanakin/pen/NzwxWa?editors=1100

I kept your class names for the most part. It should be fairly simple to plug and play, however; I changed up the template block for the HTML slightly

phpBB/styles/prosilver/theme/content.css
lavigor: This is used in posts for handling the parsed mention BBCode.

phpBB/styles/prosilver/theme/mentions.css
```diff
@@ -0,0 +1,87 @@
...
+	margin-right: 10px;
+	margin-left: 0;
+}
+
+.mention-media-avatar {
+	font-size: 32px;
+	line-height: 36px;
+	text-align: center;
+	vertical-align: center;
+	width: 36px;
```
hanakin: you should not need anything here but the width and height which should be 40px. Use svgs for the icons, not font-awsome...Were switching to svgs in 3.3 anyway...

phpBB/styles/prosilver/theme/mentions.css
```diff
@@ -0,0 +1,87 @@
...
+	padding: 0;
+	list-style-type: none;
+}
+
+.mention-media {
+	display: inline-flex;
+	flex-shrink: 0;
+	justify-content: center;
+	align-items: center;
+	margin-right: 10px;
```
hanakin: why 10px? instead of 16px?

phpBB/styles/prosilver/theme/mentions.css
```diff
@@ -0,0 +1,87 @@
...
+	font-size: 32px;
+	line-height: 36px;
+	text-align: center;
+	vertical-align: center;
+	width: 36px;
+	height: 36px;
+}
+
+.mention-item {
+	font-size: 14px;
```
hanakin: 16px

phpBB/styles/prosilver/theme/mentions.css
```diff
@@ -0,0 +1,87 @@
...
+	line-height: 1.5;
+	letter-spacing: 0.04em;
+	border-bottom: 1px solid transparent;
+	position: relative;
+	display: flex;
+	overflow: hidden;
+	justify-content: flex-start;
+	align-items: center;
+	max-width: 300px;
+	padding: 10px;
```
hanakin: 16px

phpBB/styles/prosilver/theme/mentions.css
```diff
@@ -0,0 +1,87 @@
...
+.mention-rank {
+	display: block;
+}
+
+.mention-name {
+	line-height: 1.25;
+}
+
+.mention-rank {
+	font-size: 12px;
```
hanakin: 14px

phpBB/styles/prosilver/theme/mentions.css
```diff
@@ -0,0 +1,87 @@
...
+	font-weight: 400;
+	line-height: 1.5;
+	letter-spacing: 0.04em;
+	border-bottom: 1px solid transparent;
+	position: relative;
+	display: flex;
+	overflow: hidden;
+	justify-content: flex-start;
+	align-items: center;
+	max-width: 300px;
```
hanakin: this should not be necessary...it will not overflow the container...

phpBB/adm/style/admin.css
```diff
@@ -1653,6 +1653,107 @@ fieldset.submit-buttons legend {
...
+	list-style-type: none;
+}
+
+.mention-media {
+	color: #757575;
+	display: inline-flex;
+	flex-shrink: 0;
+	justify-content: center;
+	align-items: center;
+	margin-right: 10px;
```
hanakin: 16px

phpBB/adm/style/admin.css
```diff
@@ -1653,6 +1653,107 @@ fieldset.submit-buttons legend {
...
+	text-align: left;
+	background-color: #ffffff;
+	border-radius: 2px;
+	box-shadow:
+		0 3px 1px -2px rgba(0, 0, 0, 0.2),
+		0 2px 2px 0 rgba(0, 0, 0, 0.14),
+		0 1px 5px 0 rgba(0, 0, 0, 0.12);
+	position: absolute;
+	z-index: 999;
+	top: 20px;
```
hanakin: Sorry my bad this may not be necessary as I just added it to give the page some spacing, we can refine it in RC

phpBB/adm/style/admin.css
```diff
@@ -1653,6 +1653,107 @@ fieldset.submit-buttons legend {
...
+	flex-shrink: 0;
+	justify-content: center;
+	align-items: center;
+	margin-right: 10px;
+	margin-left: 0;
+}
+
+.rtl .mention-media {
+	margin-right: 0;
+	margin-left: 10px;
```
hanakin: 16px

phpBB/adm/style/admin.css
```diff
@@ -1653,6 +1653,107 @@ fieldset.submit-buttons legend {
...
+}
+
+.mention-media-avatar {
+	font-size: 32px;
+	line-height: 36px;
+	text-align: center;
+	vertical-align: center;
+	width: 36px;
```
hanakin: only need height and width which should be 40px;

phpBB/adm/style/admin.css
```diff
@@ -1653,6 +1653,107 @@ fieldset.submit-buttons legend {
...
+	height: 36px;
+}
+
+.mention-item {
+	font-size: 14px;
```
hanakin: 16px

phpBB/adm/style/admin.css
```diff
@@ -1653,6 +1653,107 @@ fieldset.submit-buttons legend {
...
+	letter-spacing: 0.04em;
+	border-bottom: 1px solid #dddddd;
+	color: #212121;
+	position: relative;
+	display: flex;
+	overflow: hidden;
+	justify-content: flex-start;
+	align-items: center;
+	max-width: 300px;
+	padding: 10px;
```
hanakin: 16px

phpBB/adm/style/admin.css
```diff
@@ -1653,6 +1653,107 @@ fieldset.submit-buttons legend {
...
+.mention-rank {
+	display: block;
+}
+
+.mention-name {
+	line-height: 1.25;
+}
+
+.mention-rank {
+	font-size: 12px;
```
hanakin: 14px

phpBB/assets/javascript/editor.js
```diff
@@ -389,11 +389,32 @@ function getCaretPosition(txtarea) {
...
+	function handle_mentions(txtarea) {
+		var $mentionDataContainer = $('[data-mention-url]'),
+			mentionURL = $mentionDataContainer.data('mentionUrl'),
+			mentionNamesLimit = $mentionDataContainer.data('mentionNamesLimit'),
+			mentionTopicId = $mentionDataContainer.data('topicId');
+		$(txtarea).atwho({
+			at: "@",
+			displayTpl: function(data) {
+				var avatar = (data.avatar.src) ? "<img src='" + data.avatar.src + "' class='mention-media-avatar'>" :
+					"<i class='mention-media-avatar fa fa-" + data.avatar.type + "'></i>",
```
hanakin: use the svg code for the icons not font-awsome.

phpBB/styles/prosilver/theme/bidi.css
```diff
@@ -365,6 +365,24 @@
...
+
+/* Mention dropdown
+---------------------------------------- */
+.rtl .atwho-view { /* mention-container */
+	text-align: right;
+}
+
+.rtl .mention-media {
+	margin-right: 0;
+	margin-left: 10px;
```
hanakin: 16px

phpBB/phpbb/mention/source/base_group.php
```diff
@@ -0,0 +1,148 @@
...
+		{
+			$group_ids[] = $row['group_id'];
+		}
+
+		$this->db->sql_freeresult($result);
+
+		// Grab group data
+		$groups = $this->get_groups();
+
+		$matches = preg_grep('/^' . $keyword . '.*/i', $groups['names']);
```
gooof: preg_quote for $keyword?

lavigor:
@hanakin do you mean that I must do it **exactly** as it is on Codepen?
Yes, I've changed the font size because the dropdown looked too huge for me.
It doesn't look so huge on Codepen because there's nothing else to compare it with.

Side note: `svg` class goes to `normalize.css`, am I right?
I also googled for `viewBox` attribute, it seems to be camelCase, but actually works as lowercased too.
Should we leave it lowercased?

Screenshots? OK:
Version from Codepen:
default

My version:
default

And I've fixed a bug when the name is too long by introducing `max-width`.
(I don't think that the block should become 1200+ pixels in width for really long names.)

Also I've fixed a bug with the lack of styling for selected element (`.cur` class of atwho).

You code on Codepen doesn't comply with your Stylelint rules BTW.
And I wonder why there's no autofix for them in PhpStorm (red lines are loaded after ~20-30 seconds which is also irritating).

FontAwesome also has support for SVGs (and a JS framework for converting icons to SVGs too) - are there any plans for using it?

phpBB/phpbb/mention/source/base_group.php
lavigor: Thank you.

lavigor:
> Don't forget to update 

@gooof `[mention]` BBCode won't be displayed on a page, so there's no need in adding it to that list.

gooof:
@lavigor The line is only to make sure that nobody can add a own [mention] bbcode. Because he already exists hardcoded.

lavigor:
@gooof I thought about another file, sorry.
Thank you again. 😉

phpBB/phpbb/mention/source/team.php
```diff
@@ -0,0 +1,38 @@
...
+			'FROM'      => [
+				USERS_TABLE => 'u',
+				USER_GROUP_TABLE => 'ug',
+				TEAMPAGE_TABLE => 't',
+			],
+			'WHERE'     => 'ug.group_id = t.group_id AND ug.user_id = u.user_id AND ug.user_pending = 0
+				AND u.user_id <> ' . ANONYMOUS . '
+				AND ' . $this->db->sql_in_set('u.user_type', [USER_NORMAL, USER_FOUNDER]) . '
+				AND u.username_clean ' . $this->db->sql_like_expression($keyword . $this->db->get_any_char()),
+			'ORDER_BY'  => 'u.user_lastvisit DESC'
```
CHItA: Same as above.

phpBB/phpbb/mention/source/topic.php
```diff
@@ -0,0 +1,41 @@
...
+			],
+			'LEFT_JOIN' => [
+				[
+					'FROM' => [POSTS_TABLE => 'p'],
+					'ON'   => 'u.user_id = p.poster_id'
+				]
+			],
+			'WHERE'     => 'p.topic_id = ' . $topic_id . ' AND u.user_id <> ' . ANONYMOUS . '
+				AND ' . $this->db->sql_in_set('u.user_type', [USER_NORMAL, USER_FOUNDER]) . '
+				AND u.username_clean ' . $this->db->sql_like_expression($keyword . $this->db->get_any_char()),
```
CHItA: Same as above.

phpBB/phpbb/mention/source/user.php
```diff
@@ -0,0 +1,35 @@
...
+	protected function query($keyword, $topic_id)
+	{
+		$query = $this->db->sql_build_query('SELECT', [
+			'SELECT'    => 'u.username, u.user_id',
+			'FROM'      => [
+				USERS_TABLE => 'u',
+			],
+			'WHERE'     => 'u.user_id <> ' . ANONYMOUS . '
+				AND ' . $this->db->sql_in_set('u.user_type', [USER_NORMAL, USER_FOUNDER]) . '
+				AND u.username_clean ' . $this->db->sql_like_expression($keyword . $this->db->get_any_char()),
```
CHItA: Same as above.

phpBB/phpbb/textformatter/s9e/factory.php
```diff
@@ -84,6 +84,12 @@ class factory implements \phpbb\textformatter\cache_interface
 		'img'   => '[IMG src={IMAGEURL;useContent}]',
 		'list'  => '[LIST type={HASHMAP=1:decimal,a:lower-alpha,A:upper-alpha,i:lower-roman,I:upper-roman;optional;postFilter=#simpletext} #createChild=LI]{TEXT}[/LIST]',
 		'li'    => '[* $tagName=LI]{TEXT}[/*]',
+		'mention' =>
+			"[MENTION
+				group_id={UINT;optional}
```
CHItA: It might be better to merge all `*_id` tags into one and have that information in the `id` tag. E.g. instead of this: `group_id=foobar` it could be `id=group-foobar`.

phpBB/phpbb/textformatter/s9e/mention_helper.php
```diff
@@ -0,0 +1,178 @@
...
+		$this->cached_colours['groups'] = [];
+
+		if (!empty($user_ids))
+		{
+			$query = $this->db->sql_build_query('SELECT', [
+				'SELECT' => 'u.user_colour, u.user_id',
+				'FROM'   => [
+					USERS_TABLE => 'u',
+				],
+				'WHERE'  => 'u.user_id <> ' . ANONYMOUS . '
```
CHItA: Just filter out `ANONYMOUS` from `$user_ids`.

phpBB/assets/javascript/editor.js
```diff
@@ -391,21 +391,24 @@ function getCaretPosition(txtarea) {
 
 (function($) {
 	function Mentions() {
+		let $mentionDataContainer = $('[data-mention-url]:first'), cachedNames = null, filteredNamesCount = 0;
```
hanakin: all vars should be explicitly defined, please break them up on their own line

phpBB/assets/javascript/editor.js
```diff
@@ -391,21 +391,24 @@ function getCaretPosition(txtarea) {
...
 		this.isEnabled = function() {
-			return $('[data-mention-url]').length;
+			return $mentionDataContainer.length;
 		};
 
 		this.handle = function(txtarea) {
-			let $mentionDataContainer = $('[data-mention-url]'),
-				mentionURL = $mentionDataContainer.data('mentionUrl'),
+			let mentionURL = $mentionDataContainer.data('mentionUrl'),
 				mentionNamesLimit = $mentionDataContainer.data('mentionNamesLimit'),
```
hanakin: all vars should be explicitly defined do not group them with commas

phpBB/assets/javascript/editor.js
```diff
@@ -415,11 +418,66 @@ function getCaretPosition(txtarea) {
...
+						}
 						let params = {keyword: query, topic_id: mentionTopicId, _referer: location.href};
 						$.getJSON(mentionURL, params, function (data) {
-							callback(data)
+							cachedNames = data;
+							callback(data);
+						});
+					},
+					sorter: function(query, items, searchKey) {
+						let _unsorted, _results, _exactMatch, i, item, len, highestPriority = 0;
```
hanakin: all vars should be explicitly declared please break them up on their own line.

phpBB/assets/javascript/editor.js
```diff
@@ -389,13 +389,166 @@ function getCaretPosition(txtarea) {
...
+		let cachedNames = null;
+		let cachedFor = null;
+		let cachedSearchKey = 'name';
+
+		function defaultAvatar(type) {
+			return (type === 'group') ? '<svg class="mention-media-avatar" xmlns="http://www.w3.org/2000/svg" viewbox="0 0 24 24"><path fill-rule="evenodd" d="M16 11c1.66 0 2.99-1.34 2.99-3S17.66 5 16 5c-1.66 0-3 1.34-3 3s1.34 3 3 3zm-8 0c1.66 0 2.99-1.34 2.99-3S9.66 5 8 5C6.34 5 5 6.34 5 8s1.34 3 3 3zm0 2c-2.33 0-7 1.17-7 3.5V19h14v-2.5c0-2.33-4.67-3.5-7-3.5zm8 0c-.29 0-.62.02-.97.05 1.16.84 1.97 1.97 1.97 3.45V19h6v-2.5c0-2.33-4.67-3.5-7-3.5z"/></svg>' : '<svg class="mention-media-avatar" xmlns="http://www.w3.org/2000/svg" viewbox="0 0 24 24"><path fill-rule="evenodd" d="M12,19.2C9.5,19.2 7.29,17.92 6,16C6.03,14 10,12.9 12,12.9C14,12.9 17.97,14 18,16C16.71,17.92 14.5,19.2 12,19.2M12,5A3,3 0 0,1 15,8A3,3 0 0,1 12,11A3,3 0 0,1 9,8A3,3 0 0,1 12,5M12,2A10,10 0 0,0 2,12A10,10 0 0,0 12,22A10,10 0 0,0 22,12C22,6.47 17.5,2 12,2Z"/></svg>';
+		}
+
+		function getMatchedNames(query, items, searchKey) {
+			let i, len;
```
hanakin: missed one heree

phpBB/assets/javascript/editor.js
```diff
@@ -389,13 +389,166 @@ function getCaretPosition(txtarea) {
...
+						let params = {keyword: query, topic_id: mentionTopicId, _referer: location.href};
+						$.getJSON(mentionURL, params, function (data) {
+							cachedNames = data;
+							cachedFor = query;
+							queryInProgress = null;
+							callback(data);
+						});
+					},
+					sorter: function(query, items, searchKey) {
+						let i, len;
```
hanakin: missed another one here

phpBB/assets/javascript/editor.js
```diff
@@ -389,13 +389,166 @@ function getCaretPosition(txtarea) {
...
+					}
+				}
+			});
+		};
+	}
+	phpbb.mentions = new Mentions();
+
 	$(document).ready(function() {
-		var doc, textarea;
+		let doc, textarea;
```
hanakin: another here

phpBB/includes/functions.php
```diff
@@ -4392,6 +4395,11 @@ function page_header($page_title = '', $display_online_list = false, $item_id =
 		'U_RESTORE_PERMISSIONS'	=> ($user->data['user_perm_from'] && $auth->acl_get('a_switchperm')) ? append_sid("{$phpbb_root_path}ucp.$phpEx", 'mode=restore_perm') : '',
 		'U_FEED'				=> $controller_helper->route('phpbb_feed_index'),
 
+		'S_ALLOW_MENTIONS'		=> ($config['allow_mentions'] && $auth->acl_get('u_mention') && (empty($forum_id) || $auth->acl_get('f_mention', $forum_id))) ? true : false,
+		'S_MENTION_BATCH_SIZE'	=> 100, // TODO: do not hardcode the value
```
marc1706: Yep, makes sense to add this to the load settings

phpBB/language/en/ucp.php
```diff
@@ -335,6 +335,7 @@
 	'NOTIFICATION_TYPE_BOOKMARK'						=> 'Someone replies to a topic you have bookmarked',
 	'NOTIFICATION_TYPE_GROUP_REQUEST'					=> 'Someone requests to join a group you lead',
 	'NOTIFICATION_TYPE_IN_MODERATION_QUEUE'				=> 'A post or topic needs approval',
+	'NOTIFICATION_TYPE_MENTION'							=> 'Someone mentiones you in a post',
```
marc1706: mentiones --> mentions

lavigor: Oh, missed that, thank you

lavigor:
@hanakin I'd like some help with avatars (see TODO in this PR).
Your solution on Codepen is only for images with 1:1 ratio.
But currently phpBB accepts any avatars, not only square/circle but rectangle too.
How should I handle them?

Regarding your previous reply, PhpStorm incorporates WebStorm built specifically for front-end development. Maybe it's not `front-end editor` because it doesn't have a WYSIWYG editor, it's an IDE and has lots of other useful features.

As for `max-width`, I was thinking about really long usernames (2k+ characters), but maybe that should never happen in production (however still possible).

As for CSS formatting, I tried to execute the built-in Gulp file, but it changed all CSS files for prosilver.
I thought that something is wrong with it or it's not used at the moment, and I just shouldn't fix every single CSS that is out of the scope of this PR.

> We have our own method we are working which will be far superior and simpler to maintain.

I wonder what that will be. As for Font Awesome, were you talking about the latest release (now it is 5.2)?
These guys gathered over $1m on Kickstarter, SVG support was one of the features. However not sure whether that is included in the free plan.

hanakin:
Long answers coming haha... I will get to this tonight when I get off work

phpBB/styles/prosilver/theme/mentions.css
```diff
@@ -0,0 +1,86 @@
...
+	margin-right: 16px;
+	margin-left: 0;
+}
+
+.mention-media-avatar {
+	width: 40px;
+	height: 40px;
+}
+
+svg.mention-media-avatar { /* TODO: remove it after general normalization */
```
hanakin: not sure why you are class typing? This is a global for all svg...leave the comment for me to clean up but remove the class it should just be `svg {`

phpBB/adm/style/admin.css
```diff
@@ -1653,6 +1653,106 @@ fieldset.submit-buttons legend {
...
+	margin-right: 0;
+	margin-left: 16px;
+}
+
+.mention-media-avatar {
+	width: 40px;
+	height: 40px;
+}
+
+svg.mention-media-avatar { /* TODO: remove it after general normalization */
```
hanakin: should only be `svg {` see comment in mentions.css for theme.

phpBB/adm/style/admin.css
```diff
@@ -1653,6 +1653,106 @@ fieldset.submit-buttons legend {
...
+
+.mention-item:hover,
+.mention-item.cur {
+	text-decoration: none;
+	background-color: #eeeeee;
+	color: #2d80d2;
+}
+
+.mention-item:hover .mention-media-avatar,
+.mention-item.cur .mention-media-avatar {
```
hanakin: what is this for? why are you changing the avatar color on hover?

phpBB/styles/prosilver/theme/mentions.css
lavigor: Added that not to make a global impact in the scope of this PR.
Ok, I can make it global if you say so.

phpBB/adm/style/admin.css
lavigor: That is for default avatars only.
You have this on your Codepen too.

phpBB/styles/prosilver/theme/mentions.css
hanakin: Yes it’s a global it needs to be in the reset

lavigor:
>If they do not use a 1:1 avatar it will look like half an image. Not our problem!

It's indeed our problem: most phones and cameras make rectangle photos.
Most websites with square/circle avatars have a JS tool for selecting a square/circle area by the user.
Is any similar tool planned for inclusion into phpBB?

In fact original photo can be stored as a profile picture besides the cutted avatar (e.g. just like in social webs), IMO that's more convenient.
However leaving only the cutted result is also OK (for saving disk space).

But I'm sure that the user should be able to select the area for cutting, that should not be done automatically.

lavigor:
@hanakin could you please be more specific about what part of my reply you disagree with.
Regarding this PR, I can do exactly that (scaling and centering with JS).

About the `concept of the avatar`: if they are intended to be in 1:1 ratio, why is it difficult to simplify creating them by allowing to select the square/circle area when a user uploads an avatar?
A JS plugin of that kind won't affect website performance.

hanakin:
With the fact that it’s our problem here. Exactly as you said it’s a problem during creation which needs solved not something for here. Our avatar creation is out dated and needs updated to match something like Facebook

lavigor:
Ah, OK, thank you.
It's not the problem of this PR, of course.

CHItA:
@phpbb/development-team reviews would be much appriciated.

phpBB/styles/prosilver/theme/colours.css
```diff
@@ -314,6 +314,41 @@ p.post-notice {
 	background-image: none;
 }
 
+/*	colours and backgrounds for mentions.css */
```
hanakin: Remove the tab should be a space

phpBB/styles/prosilver/theme/mentions.css
```diff
@@ -0,0 +1,86 @@
+/* -------------------------------------------------------------- /*
+	$Mentions
```
hanakin: Tab should be a space

phpBB/styles/prosilver/theme/colours.css
lavigor: I've made that the same as for other modules, changed that.
The same for your comment below.

lavigor:
-------------
### GSoC code ends here.
-------------

CHItA:
@JoshyPHP could you have a look at the text-formater related code if we missed anything?

phpBB/phpbb/textformatter/s9e/factory.php
```diff
@@ -84,6 +84,10 @@ class factory implements \phpbb\textformatter\cache_interface
 		'img'   => '[IMG src={IMAGEURL;useContent}]',
 		'list'  => '[LIST type={HASHMAP=1:decimal,a:lower-alpha,A:upper-alpha,i:lower-roman,I:upper-roman;optional;postFilter=#simpletext} #createChild=LI]{TEXT}[/LIST]',
 		'li'    => '[* $tagName=LI]{TEXT}[/*]',
+		'mention' =>
+			"[MENTION={PARSE=/^(?<type>[ug]):(?<id>\d+)$/}
+				profile_url={URL;optional;postFilter=#false}
```
JoshyPHP: It would be a good idea to define the `color` attribute here (e.g. `color={COLOR}`) otherwise the template is potentially unsafe since it's used as-is in CSS.

JoshyPHP:
@CHItA I took a look at `mention_helper::inject_metadata()` and it seems alright. I don't think the library is used anywhere else.

I added a comment about declaring the `color` attribute when creating the BBCode.

Unrelated to markup handling, I'd be careful to measure the impact of the new feature on performance and resource usage. I think that `mention_helper::cache_colours()` is going to cache enough data to make a measurable difference on big boards.

lavigor:
@JoshyPHP @gooof if there's a way to make a single DB query to fetch all necessary info for the whole topic page at once, I'll use your suggestions.
I've done it this way so that we won't have a new DB query for every single post parsed for current request.

As for the `color` attribute, I'll add it as internal, thank you (in fact it's not something that the user can fill in).

CHItA:
On parsing it is fine to have a few queries, what needs to be fast is rendering.

lavigor:
@CHItA sorry I meant rendering because colors aren't stored in the DB due to possible changes.
(And if we did store them, the colors would be messed up after post edits).

lavigor:
!unset WIP 🚧

lavigor:
!unset WIP

phpBB/phpbb/mention/controller/mention.php
```diff
@@ -0,0 +1,67 @@
...
+	{
+		if (!$this->request->is_ajax())
+		{
+			return new RedirectResponse(append_sid($this->phpbb_root_path . 'index.' . $this->php_ext));
+		}
+
+		$keyword = $this->request->variable('keyword', '', true);
+		$topic_id = $this->request->variable('topic_id', 0);
+		$names = [];
+		$hasNamesRemaining = false;
```
Nicofuma: phpBB use snake_case instead of camelCase

phpBB/phpbb/mention/source/base_group.php
```diff
@@ -0,0 +1,172 @@
...
+		{
+			$group_ids[] = $row['group_id'];
+		}
+
+		$this->db->sql_freeresult($result);
+
+		// Grab group data
+		$groups = $this->get_groups();
+
+		$matches = preg_grep('/^' . preg_quote($keyword) . '.*/i', $groups['names']);
```
Nicofuma: I don't really understand why you are doing that, couldn't just lookup the returned group ID in the groups array and do that directly in the loop parsing the query results? (something like that:

```
		// Grab group data
		$groups = $this->get_groups();

		// Grab all group IDs, cache for 5 minutes
		$result = $this->db->sql_query($this->query($keyword, $topic_id), 300);
		$group_ids = [];
		while ($row = $this->db->sql_fetchrow($result))
		{
			if (isset($this->groups[$row['group_id']])) {
				$group_ids[] = $row['group_id'];
			}
		}
```

phpBB/phpbb/mention/source/base_user.php
```diff
@@ -0,0 +1,143 @@
...
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function get(array &$names, $keyword, $topic_id)
+	{
+		$fetched_all = false;
+		$keyword = utf8_clean_string($keyword);
+
+		// Do not query all possible users (just a moderate amount), cache results for 5 minutes
```
Nicofuma: you do not limit the size of the batch yourself, it is up to the query itself. So this comment should be moved to the query() method imo

phpBB/phpbb/mention/source/friend.php
```diff
@@ -0,0 +1,58 @@
...
+	 */
+	protected function query($keyword, $topic_id)
+	{
+		/*
+		 * For optimization purposes all friends are returned regardless of the keyword
+		 * Names filtering is done on the frontend
+		 * Results will be cached on a per-user basis
+		 */
+		$query = $this->db->sql_build_query('SELECT', [
+			'SELECT'    => 'u.username_clean, u.username, u.user_id',
```
Nicofuma: you shouldn't return u.username as you don't use it

phpBB/phpbb/mention/source/friend.php
```diff
@@ -0,0 +1,58 @@
...
+			'FROM'      => [
+				USERS_TABLE => 'u',
```
Nicofuma: better using the table names coming from the container

phpBB/phpbb/mention/source/base_group.php
Nicofuma: just noticed the filtering is not done in the SQL query

phpBB/phpbb/mention/controller/mention.php
lavigor: Thanks, that came after fixing JS, so...

phpBB/phpbb/textformatter/s9e/mention_helper.php
```diff
@@ -0,0 +1,277 @@
...
+			return;
+		}
+
+		$types = ['u', 'g'];
+
+		foreach ($types as $type)
+		{
+			$this->cached_colours[$type] = [];
+
+			$query = $this->db->sql_build_query('SELECT', $this->get_colours_sql($type));
```
Nicofuma: this is gonna return a lot of data

phpBB/phpbb/mention/source/base_user.php
lavigor: This comment is no longer valid, all user IDs are queried and cached.
Fixed that.

phpBB/phpbb/mention/source/friend.php
lavigor: It's used [here](https://github.com/phpbb/phpbb/pull/5225/files#diff-a9071ee38a42033caed608cc816e37e2R129).

lavigor: But I can get rid of this in queries and use the data from user_loader as it's being fetched anyway.
Fixed.

phpBB/phpbb/textformatter/s9e/mention_helper.php
CHItA: yeah, probably we just shouldn't render the user color, just do a simple CSS styling as gooof suggested above.

lavigor: Yes, but the database is queried only once in 5 minutes.
Do you have suggestions for changing that?

phpBB/phpbb/textformatter/s9e/factory.php
lavigor: Added color attribute.
@JoshyPHP BTW could you please explain the meaning/necessity of `postFilter=#false`?
I've taken that from the `quote` BBCode.

JoshyPHP: [It unconditionally invalidates the value](https://s9etextformatter.readthedocs.io/Filters/Built-in_filters/). That prevents the user from supplying a value for the profile URL.

phpBB/phpbb/textformatter/s9e/mention_helper.php
Nicofuma: I agree with @CHItA and @gooof lets not use the user color.
The issue is not that the database won't like it but that migth be a huge amount of data to load and put in cache and it can be noticed and having one crying user every 5 minutes is not really acceptable. But even if it was having that much data in cache (and so in memory) and not that good.

phpBB/phpbb/mention/source/user.php
```diff
@@ -0,0 +1,47 @@
...
+	 */
+	protected function query($keyword, $topic_id)
+	{
+		$query = $this->db->sql_build_query('SELECT', [
+			'SELECT'    => 'u.username_clean, u.user_id',
+			'FROM'      => [
+				USERS_TABLE => 'u',
+			],
+			'WHERE'     => $this->db->sql_in_set('u.user_type', [USER_NORMAL, USER_FOUNDER]),
+			'ORDER_BY'  => 'u.user_lastvisit DESC'
```
Nicofuma: you can't select all users like that, you need to add some limit.
I don't think it is acceptable to load the entire user table and to cache it like that.
@phpbb/development-team what do you think

phpBB/phpbb/mention/source/topic.php
```diff
@@ -0,0 +1,69 @@
...
+				[
+					'FROM' => [POSTS_TABLE => 'p'],
+					'ON'   => 'u.user_id = p.poster_id'
+				],
+				[
+					'FROM' => [TOPICS_TABLE => 't'],
+					'ON'   => 't.topic_id = p.topic_id'
+				],
+			],
+			'WHERE'     => 'p.topic_id = ' . $topic_id . '
```
Derky: Please cast `$topic_id` with `(int)`

phpBB/phpbb/mention/source/user.php
Derky: Yeah I agree this will get way too big. What is this used for actually since `$keyword` is not even used in the query?

lavigor: @Derky I tried to find the right balance between the DB performance and the amount of cache files being saved. So that we do not have a cache file for every keyword/user/etc as well as we don't query the DB too many times during the topic display process.

Hmm, actually I stopped using the keyword in any query methods, the parameter could probably be removed then.
Do you have any suggestions for making it in a different way?

CHItA: well, you could add a where on the keyword and a limit statement. One thing to note here is that the more users we pull from the DB at once will reduce the number of requests to the API endpoint, so probably you want to balance that.

Derky: @lavigor are you only using this to display the username in the post or also to create the list of options while typing?

If you only use it to display the username in the post, I advise to look at how we do that with quoting since we show the username as well. (I think it's just hardcoded in the post text)

phpBB/phpbb/textformatter/s9e/factory.php
lavigor: OK, so my guesses were right, thank you.

phpBB/phpbb/mention/source/user.php
lavigor: @Derky quite the opposite - this is only used to generate the list of options.

@CHItA how many cache files are supposed to be created? I can't imagine the exact amount for every typed letters and their combinations, but this will likely take more space then a single request to fetch them all.

phpBB/adm/style/admin.css
```diff
@@ -1704,6 +1704,11 @@ fieldset.submit-buttons legend {
 	height: 40px;
 }
 
+.mention-media-avatar img { /* TODO: reconsider for the future design */
```
Nicofuma: you should say why

lavigor: Added lines stating `with 1:1 ratio images`.

lavigor:
@marc1706 sad to hear that, the suggested replacement seems to have less features.
But I can test it then.

marc1706:
Yeah, would be great if you could check it out. I've just rebased your PR on top of latest master and added some jsdoc docblocks.

marc1706:
And I broke the builds. Let me fix that. ;)

hanakin:
@marc1706 @lavigor whats the status of this? Would be nice to see this on area51 for proper testing sometime this year

By-oussef:
+1

luzpaz:
Haven't seen any updates to this code, does that mean it's stable ?

JoshyPHP:
I would advise against using the same attributes for two different type of values. Here, user mentions and group mentions use the same `id` attribute plus a `type` attribute to determine what the value of `id` really means. For instance, here are two mentions taken from the test cases:

```xml
<r><MENTION id="3" type="u"><s>[mention=u:3]</s>test<e>[/mention]</e></MENTION></r>
```
```xml
<r><MENTION id="3" type="g"><s>[mention=g:3]</s>test<e>[/mention]</e></MENTION></r>
```

I think there's an argument for having two different BBCodes, one for user mentions and another for group mentions. Regardless, if the same BBCode is used for both, it would be a good idea to use two different attributes to hold the different types of values. The BBCode declaration could look like this:
```
[MENTION={PARSE=/^(?:g:(?<group_id>\\d+)|u:(?<user_id>\\d+))$/}
    group_id={UINT?}
    user_id={UINT?}
    profile_url={URL?;postFilter=#false}]{TEXT}[/MENTION]
```
The XML output would look like this:
```xml
<r><MENTION user_id="3"><s>[mention=u:3]</s>test<e>[/mention]</e></MENTION>
<MENTION group_id="3"><s>[mention=g:3]</s>test<e>[/mention]</e></MENTION></r>
```

hanakin:
@lavigor I am going to add the abandoned by author tag to this unless you can make the necessary changes listed above by JoshyPHP

lavigor:
@hanakin sorry for the delay, I'll look into it soon.
Should I also replace the aforementioned plugin?

hanakin:
not sure which plugin you are referring too?

lavigor:
I mean the At.js library.

paul999:
It is no longer maintained, see https://github.com/ichord/At.js. I think it wouldn't be to hard to replace it with tribute, which is the suggested replacement

JoshyPHP:
Wrt having different BBCodes for different types of mentions, as a point of comparison on XenForo when somebody uses a mention such as `@user` it's automatically converted into `[USER=123]@user[/USER]`. Even if both types of mentions are displayed the same now, having two different BBCodes would make it easier to present or handle differently in the future.

lavigor:
@JoshyPHP, so should I split the BBCode or should we put this into a new discussion?

JoshyPHP:
I don't know. It's not strictly better to have different BBCodes either and it somewhat increases the risk of name collision with existing custom BBCodes, so I don't have a strong opinion about it. You're free to ignore it if you don't see an upside to having different BBCodes.

Similarly, Flarum has post mentions in the form `@user#123` where `123` is the ID of the post. I don't think it's relevant to phpBB because quoting is the primary mechanism to refer to other posts but I think it's worth mentioning in case somebody has a particularly good idea for it.

Dode123:
Can't wait for this to be a part of phpBB core =)

marc1706:
Rebased on latest master

lavigor:
@marc1706 thanks
Very sorry for such a delay

marc1706:
No worries. Wanted to see how cleanly it applies and I guess based on tests there might be some incompatibilities we need to resolve. :)

marc1706:
I've worked a bit on switching from the no longer developed atwho.js to zurb tribute:
https://github.com/phpbb/phpbb/compare/master...marc1706:ticket/13713

@lavigor would you have some time to maybe check if the behavior is the same as previously expected?

marc1706:
I have now replaced this PR with https://github.com/phpbb/phpbb/pull/6206

PR #5237, opened 2018-06-05
[ticket/15680] Point to 3.2 documentation

avijc:
The old link pointed to 3.1 documentation

https://tracker.phpbb.com/browse/PHPBB3-15680

### DIFF
--- a/phpBB/docs/INSTALL.html
+++ b/phpBB/docs/INSTALL.html
@@ -44,7 +44,7 @@
 	<p>
 		A basic overview of running phpBB can be found in the accompanying <a href="README.html">README</a> file.
 		Please ensure you read that document in addition to this! For more detailed information on using, installing,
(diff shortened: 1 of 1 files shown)


Crizz0:
Why do you close the PR? 

You can fix the error within this PR, no need for a new one.

avijc:
It was unclear why exactly Travis rejected the pull request (all I got was "Unexpected EOF encountered"), so I wanted to take a fresh start. Had I known the exact reason for the failure at that stage I might have decided to do something differently. In addition, as the change was just a single character, I did not bother with cloning the repository on my own computer at all, but instead did all the work directly on Github's website which has its own text editor, among other tools. It would have worked OK that way as well, apart from the fact that Github seems to use CR/LF line breaks in commit messages by default, which phpBB's pull request validation scripts will apparently reject. This is unfortunate.

PR #5239, opened 2018-06-05
[ticket/15683] Better error message when commit message has CRLF

avijc:
When someone tries to do a pull request with incorrect line breaks in the commit message, the error message as given by Travis is "Unexpected EOF encountered". The error message should be clearer that CR/LF was the problem. This change adds a separate check for CRLF line endings.

PHPBB3-15683

### DIFF
--- a/git-tools/hooks/commit-msg
+++ b/git-tools/hooks/commit-msg
@@ -147,6 +147,15 @@ then
 	quit $ERR_LENGTH;
 fi
 
+# Check for CR/LF line breaks
+if grep -q $'\r$' "$1"
+then
+	complain "The commit message uses CR/LF line breaks, which are not permitted." >&2
+	complain >&2
+
+	quit $ERR_EOF;
+fi
+
 lines=$(wc -l "$1" | awk '{ print $1; }');
 expecting=header;
 in_description=0;


marc1706:
Any chance you could rebase this onto 3.2.x? I think it would be fitting to also add this to 3.2.x. If you're unsure how to rebase it onto a different branch I can assist you with that as well.

avijc:
Sure, I thought this would be more fitting in master, because the changes are not really related to 3.2.x.

git-tools/hooks/commit-msg
```diff
@@ -147,6 +147,15 @@ then
 	quit $ERR_LENGTH;
 fi
 
+# Check for CR/LF line breaks
+if grep -q $'\r$' "$1"
```
marc1706: Right now this would only detect \r at the end of a line, which is the classic windows line ending. CRLF should however be \r\n which means there probably needs to be an optional addition for \n at the end of the line between the \r and the $.

avijc: That's what one would think, but consider this:

    $ echo test > unix
    $ echo test > dos
    $ unix2dos dos
    unix2dos: converting file dos to DOS format ...
    $ od -a unix
    0000000   t   e   s   t  nl
    0000005
    $ od -a dos
    0000000   t   e   s   t  cr  nl
    0000006
    $ if grep -q $'\r$' unix; then echo crlf; fi
    $ if grep -q $'\r$' dos; then echo crlf; fi
    crlf
    $ if grep -q $'\r\n$' unix; then echo crlf; fi
    crlf
    $ if grep -q $'\r\n$' dos; then echo crlf; fi
    crlf

Based on this, adding \n would be harmful.

Nicofuma:
They are not related to 3.2 yes but it does not harm to have it here, on the contrry, it is helpfull :)

git-tools/hooks/commit-msg
Nicofuma: that's because grep works line by line and never consider \n :)

PR #5257, opened 2018-06-18
[ticket/15651] Remove unnecessary condition

senky:
PHPBB3-15651

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15651

### DIFF
--- a/phpBB/phpbb/db/migration/data/v320/add_help_phpbb.php
+++ b/phpBB/phpbb/db/migration/data/v320/add_help_phpbb.php
@@ -32,10 +32,7 @@ class add_help_phpbb extends \phpbb\db\migration\migration
 		return array(
 			array('config.add', array('help_send_statistics', true)),
 			array('config.add', array('help_send_statistics_time', 0)),
-			array('if', array(
-				array('module.exists', array('acp', false, 'ACP_SEND_STATISTICS')),
-				array('module.remove', array('acp', false, 'ACP_SEND_STATISTICS')),
-			)),
+			array('module.remove', array('acp', false, 'ACP_SEND_STATISTICS')),
 			array('module.add', array(
 				'acp',
 				'ACP_SERVER_CONFIGURATION',


rubencm:
https://tracker.phpbb.com/browse/PHPBB3-15090

senky:
Funny thing is, this condition is always false so module was never removed... I guess we need to use explicit calls as [VSE used](https://github.com/VSEphpbb/abbc3/commit/108132bd0662eda01ed740fce1e5730422bfc250).

rubencm:
If that always return false, is a bug in module.exists, so that is what has to be fixed i think

senky:
`module.exists` works just like expected, but maybe `if` should be able to accept input other than false.

kasimi:
`module.remove` doesn't need an `if` check in the migration at all because it checks if the module exists internally: https://github.com/phpbb/phpbb/blob/release-3.2.2/phpBB/phpbb/db/migration/tool/module.php#L386

I agree that the `if` should support calling other tools, and it should be fairly simple: if `$condition` is an array [here](https://github.com/phpbb/phpbb/blob/release-3.2.2/phpBB/phpbb/db/migrator.php#L763), use something like `$this->run_step($this->get_callable_from_step($condition), $last_result, $reverse)`, as the condition, otherwise use `$condition` itself.

senky:
@kasimi does your comment mean this ticket should cover the changes?

kasimi:
No, you're right, the 2nd part of my reply above should be done in a new ticket. But the first part is relevant to this one.

senky:
https://github.com/phpbb/phpbb/pull/5260

Nicofuma:
is it still relevant?

senky:
I think so. PR enabling other tools being used in conditions isn't merged yet. Better safe than sorry.

Nicofuma:
well the other PR will be merged before this one.. so ? (and it's not an issue to try to remove a non existing module?)

senky:
>> and it's not an issue to try to remove a non existing module?

No, because `module.remove` internally checks if module exists. So it's secure.

Nicofuma:
ok

PR #5263, opened 2018-06-18
[ticket/12591] Improve breadcrumb naming and extend it's usage

senky:
PHPBB3-12591

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-12591

### DIFF
--- a/phpBB/styles/prosilver/template/navbar_header.html
+++ b/phpBB/styles/prosilver/template/navbar_header.html
@@ -187,7 +187,7 @@
 
 			<!-- BEGIN navlinks -->
 				<!-- EVENT overall_header_navlink_prepend -->
-				<span class="crumb" {$MICRODATA}<!-- IF navlinks.MICRODATA --> {navlinks.MICRODATA}<!-- ENDIF -->><a href="{navlinks.U_VIEW_FORUM}" itemprop="url"><span itemprop="title">{navlinks.FORUM_NAME}</span></a></span>
+				<span class="crumb" {$MICRODATA}<!-- IF navlinks.MICRODATA --> {navlinks.MICRODATA}<!-- ENDIF -->><a href="{navlinks.U_BREADCRUMB}" itemprop="url"><span itemprop="title">{navlinks.BREADCRUMB_NAME}</span></a></span>
 				<!-- EVENT overall_header_navlink_append -->
 			<!-- END navlinks -->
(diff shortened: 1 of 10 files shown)


VSEphpbb:
This will have backwards compatibility breaking consequences for any extensions with breadcrumbs. Extension like **Ideas**, **Pages**, **Board Rules** all use breadcrumbs and rely on `FORUM_NAME` and `U_VIEW_FORUM` existing in the core Prosilver templates.

Personally I don't think changing the name just for symantec reasons is worth the B.C. breakage. But if this is so desired then it should have some sort of BC fallback, i.e. set and use a template variable based on the existence of the new or the old variable names:
```
{{ set BC_NAME = BREADCRUMB_NAME || FORUM_NAME }}
{{ set BC_LINK = U_BREADCRUMB || U_VIEW_FORUM }}
```

senky:
Funny that you advocate for original naming because I used your comment in the ticket to name variables. 😅

I did consider extensions, but this will fall into 3.3 release which can break compatibility. Your solution, however, is quite clever so I'm going for it. But we should, in any case, make a list of BC changes when releasing 3.3.0 and this should be mentioned.

phpBB/styles/prosilver/template/navbar_header.html
```diff
@@ -186,8 +186,10 @@
 				<span class="crumb" {$MICRODATA}><a href="{U_INDEX}" itemprop="url" accesskey="h" data-navbar-reference="index"><!-- IF not U_SITE_HOME --><i class="icon fa-home fa-fw"></i><!-- ENDIF --><span itemprop="title">{L_INDEX}</span></a></span>
 
 			<!-- BEGIN navlinks -->
+				{{ set NAVINK_NAME = navlinks.BREADCRUMB_NAME || navlinks.FORUM_NAME }}
+				{{ set NAVINK_LINK = navlinks.U_BREADCRUMB || navlinks.U_VIEW_FORUM }}
```
3D-I: May I ask why `NAVINK_*` isn't `NAVLINK_*`? Out of curiosity...

senky: Typo. 😅

Nicofuma:
If we can easily keep compatibility we should (and say that the older way is deprecated of course)

senky:
It keeps compatibility.

PR #5264, opened 2018-06-18
[ticket/12578] Lazy-load template service

senky:
PHPBB3-12578

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-12578

### DIFF
--- a/phpBB/config/default/container/services_twig.yml
+++ b/phpBB/config/default/container/services_twig.yml
@@ -62,6 +62,7 @@ services:
             - '@user'
             - '@template.twig.extensions.collection'
             - '@ext.manager'
+        lazy:  true
 
     template_context:
         class: phpbb\template\context


senky:
Btw I think pretty much every service should be lazy-loaded. Even `@user` isn't used on every request. What do you think, @phpbb/development-team?

Elsensee:
Well, are there downsides to this?

Nicofuma:
complexity and complex bugs, a long time ago I think I tried to make everything lazy and it was a bad idea. But it was a long time ago.

Also it is not always worth it (especially when the constructor does nothing)

Elsensee:
Yeah, it probably isn't worth it. Also, the proxy service that is instantiated there adds something to the memory footprint so it's rather bad than good?

Except of course we instantiate objects in the constructor of an service that we really don't need.. Then it might be worth it but I'm not sure.

senky:
Generally, it's a good idea. Most of the requests don't use even half of the services that are initiated. But as @Nicofuma pointed out, it's somehow buggy and the proxy doesn't behave like intended. Or rather it doesn't work at all, it won't expose the required service upon call. So it's a good idea that would reduce the server resources but for the cost of the huge changes to the core.

PR #5265, opened 2018-06-18
[ticket/12587] Wrap all globals in proxy

senky:
PHPBB3-12587

Note that `@user` can't be lazy-loaded due to limitations of indirect modifications of `$user->data` property.

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-12587

### DIFF
--- a/phpBB/config/default/container/services.yml
+++ b/phpBB/config/default/container/services.yml
@@ -43,6 +43,7 @@ services:
              - '@dbal.conn'
              - '%core.root_path%'
              - '%core.php_ext%'
+        lazy: true
 
     cache.driver:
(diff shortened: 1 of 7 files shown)


senky:
Damn, this is tricky. Tests won't pass because literally every lazy-loaded service doesn't work as expected. Any idea why?

Nicofuma:
We rarely need a proxy. Basically we need one only when many (expensive) things are done in the constructor. Otherwise the performance gain is rarely significant enough to be worse it.

About your problem it's hard to tell, but lazy loaded services using a proxy are tricky

PR #5266, opened 2018-06-18
[ticket/12624] Add debug.load_time parameter

senky:
PHPBB3-12624

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-12624

### DIFF
--- a/phpBB/phpbb/db/driver/mssql_odbc.php
+++ b/phpBB/phpbb/db/driver/mssql_odbc.php
@@ -156,7 +156,7 @@ class mssql_odbc extends \phpbb\db\driver\mssql_base
 			{
 				$this->sql_report('start', $query);
 			}
-			else if (defined('PHPBB_DISPLAY_LOAD_TIME'))
+			else if ($this->debug_load_time)
 			{
 				$this->curtime = microtime(true);
 			}
@@ -176,7 +176,7 @@ class mssql_odbc extends \phpbb\db\driver\mssql_base
 				{
 					$this->sql_report('stop', $query);
 				}
-				else if (defined('PHPBB_DISPLAY_LOAD_TIME'))
+				else if ($this->debug_load_time)
 				{
 					$this->sql_time += microtime(true) - $this->curtime;
 				}
--- a/phpBB/phpbb/db/driver/mssqlnative.php
+++ b/phpBB/phpbb/db/driver/mssqlnative.php
@@ -128,7 +128,7 @@ class mssqlnative extends \phpbb\db\driver\mssql_base
 			{
 				$this->sql_report('start', $query);
 			}
-			else if (defined('PHPBB_DISPLAY_LOAD_TIME'))
+			else if ($this->debug_load_time)
 			{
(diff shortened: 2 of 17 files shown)


rubencm:
Remove this too https://github.com/phpbb/phpbb/blob/master/vagrant/after.sh#L41

phpBB/phpbb/db/driver/mssql_odbc.php
```diff
@@ -156,7 +158,7 @@ function sql_query($query = '', $cache_ttl = 0)
 			{
 				$this->sql_report('start', $query);
 			}
-			else if (defined('PHPBB_DISPLAY_LOAD_TIME'))
+			else if ($phpbb_container->getParameter('debug.load_time'))
```
Nicofuma: like for your other PR. I think it could be a parameter that we set during a boostrap phase once the container is built

senky: But how? How do we receive the value which we want to be stored inside yml file?

Nicofuma: a setter in the class and an after boot step that : 
1) retrieve the parameter value from the container
2) retrieve the service from the container
3) call the service setters with the parameter value

senky: Yeah true. But the problem is more fundamental. SQL query is called two times before container is built (without cache). `sql_query()`, however, uses `debug.load_time`. So first 2 queries wouldn't have SQL time calculated.

senky: @Nicofuma like this?

phpBB/phpbb/db/driver/mssql_odbc.php
```diff
@@ -147,6 +147,8 @@ function _sql_transaction($status = 'begin')
 	*/
 	function sql_query($query = '', $cache_ttl = 0)
 	{
+		global $phpbb_container;
```
Nicofuma: you can remove this

phpBB/phpbb/db/driver/mssqlnative.php
```diff
@@ -119,6 +119,8 @@ function _sql_transaction($status = 'begin')
 	*/
 	function sql_query($query = '', $cache_ttl = 0)
 	{
+		global $phpbb_container;
```
Nicofuma: and here

phpBB/phpbb/db/driver/mysql.php
```diff
@@ -167,6 +167,8 @@ function _sql_transaction($status = 'begin')
 	*/
 	function sql_query($query = '', $cache_ttl = 0)
 	{
+		global $phpbb_container;
```
Nicofuma: and here

phpBB/phpbb/db/driver/mssql_odbc.php
Nicofuma: something like that yes

PR #5267, opened 2018-06-18
[ticket/12627] Add debug.sql_explain parameter

senky:
PHPBB3-12627

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-12627


rubencm:
I think you can inject the parameters into the controllers in config files instead of using globals

senky:
DB classes aren't loaded by a service, thus it's not that easy.

senky:
This has a major flaw in the logic. DB is needed to construct the container, but only the container holds the information about debugging.

So we either test whether the container is already constructed (`!empty($phpbb_container) && $phpbb_container->getParameter('debug.sql_explain')` or we abandon the idea altogether...

Nicofuma:
I think it could be a parameter of the classes that we set at bootstrap once the container is built

senky:
@Nicofuma all the same as here: https://github.com/phpbb/phpbb/pull/5266#discussion_r200448131

PR #5269, opened 2018-06-23
[ticket/15705] fix phpbbcli language vs PHP

3D-I:
PHPBB3-15705

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15705

### DIFF
--- a/phpBB/bin/phpbbcli.php
+++ b/phpBB/bin/phpbbcli.php
@@ -71,9 +71,12 @@ require($phpbb_root_path . 'includes/compatibility_globals.' . $phpEx);
 
 register_compatibility_globals();
 
+/** @var \phpbb\config\config $config */
+$config = $phpbb_container->get('config');
+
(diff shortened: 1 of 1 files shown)


3D-I:
bug_language_config

rubencm:
Is valid since php 5.4 (which is the lowest php version supported for phpbb), so the syntax is valid.

https://secure.php.net/manual/en/language.types.array.php#example-62

3D-I:
I know, as a matter of fact though the logic isn't consistent IMO.
First, it releases an error message instead of a valid and meaningful message that can be easily interpreted by the average user. Which is harsh somehow. No error handling here.

Second, the user has tried with PHP 5.5.30 as can be seen from this post, with failure.
https://www.phpbb.com/customise/db/extension/mathjax_2/support/topic/193236?p=639351#p639351
I don't have 5.5.30 but 5.5.38.

PR #5271, opened 2018-06-27
[ticket/15286] Download remote attachments properly

rubencm:
PHPBB3-15286

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15286

### DIFF
--- a/phpBB/includes/functions_download.php
+++ b/phpBB/includes/functions_download.php
@@ -248,30 +248,8 @@ function send_file_to_browser($attachment, $category)
 
 		if ($fp !== false)
 		{
-			// Deliver file partially if requested
-			if ($range = phpbb_http_byte_range($size))
-			{
-				fseek($fp, $range['byte_pos_start']);
-
-				send_status_line(206, 'Partial Content');
-				header('Content-Range: bytes ' . $range['byte_pos_start'] . '-' . $range['byte_pos_end'] . '/' . $range['bytes_total']);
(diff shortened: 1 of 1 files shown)


phpBB/includes/functions_download.php
```diff
@@ -248,30 +248,8 @@ function send_file_to_browser($attachment, $category)
 
 		if ($fp !== false)
 		{
-			// Deliver file partially if requested
-			if ($range = phpbb_http_byte_range($size))
```
Nicofuma: are you removing the partial download on purpose?

rubencm: Yes, that probably don't work with remote files, only with local files, that is why i said that maybe i should use BinaryFileResponse for local storage only, and stream_copy_to_stream for the others

Nicofuma: ok
actually for remote file it could work but it's more complicated (something like "read fast-forward until you reach the asked bits, then stream them") not sure it is worse it.

rubencm: Yes, that should work, but better for #5274 pr, because is where i moved this code.

rubencm: @Nicofuma If i don't understand wrong, this is a very bad idea, because if a browser request a file by ranges, for each range request you have to read the file.
This can be useful only with local adapter

Nicofuma: yep, that's why I said I'm not sure it is worse it

PR #5291, opened 2018-07-12
[ticket/15722] Allow forum topicrow pagination to use topicrow values

toxyy:
Changes topicrow pagination to use values from topicrow

This allows pagination to reflect values from core.viewforum_modify_row

PHPBB3-15722

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15722

### DIFF
--- a/phpBB/viewforum.php
+++ b/phpBB/viewforum.php
@@ -1027,7 +1027,7 @@ if (count($topic_list))
 
 		$template->assign_block_vars('topicrow', $topic_row);
 
-		$pagination->generate_template_pagination($view_topic_url, 'topicrow.pagination', 'start', $replies + 1, $config['posts_per_page'], 1, true, true);
+		$pagination->generate_template_pagination($topic_row['U_VIEW_TOPIC'], 'topicrow.pagination', 'start', $topic_row['REPLIES'] + 1, $config['posts_per_page'], 1, true, true);
 
 		$s_type_switch = ($row['topic_type'] == POST_ANNOUNCE || $row['topic_type'] == POST_GLOBAL) ? 1 : 0;
 


toxyy:
If anyone disagrees, it's either adding $view_topic_url and $replies to the core.viewforum_modify_row event, or adding theme here.  This is the last place that those two are called, so it wouldn't hurt to use their values from topicrow so the event can modify them as it is now.

3D-I:
I don't see the reason to fiddle with the pagination this way.
`$view_topic_url` and `$replies` are those that generate those 2 topicrow template vars, used in the template, hence it's a nonsense IMO.

https://github.com/toxyy/phpbb/blob/48927104feb03c108607e8c038485a803db0376d/phpBB/viewforum.php#L917

https://github.com/toxyy/phpbb/blob/48927104feb03c108607e8c038485a803db0376d/phpBB/viewforum.php#L951

The event can already modify them as it is now, BTW.

If you feel you need to modify the $pagination just add `pagination` to 
`core.viewforum_topic_row_after` instead.

toxyy:
I figured since it's the last time those two variables are used it wouldn't hurt to just replace them in pagination.  But that would also be a solution, though I'm not really sure how

`pagination->generate_template_pagination($view_topic_url, 'topicrow.pagination', 'start', $replies + 1, $config['posts_per_page'], 1, true, true);`

could be added to an event.  It looks like more like a function to me.  If I instantiate it to a $topicrow_pagination variable, would modify that object work?  Or would `pagination->generate_template_pagination` have to be run again?  And if it's the latter case, then perhaps it's better to add `$view_topic_url` and `$replies` to `core.viewforum_modify_row`.

3D-I:
Is not clear what you want to do and the reason why, has been also not discussed previously.

Pagination has to remain this way IMO, if you are to add something to an event to fit your needs you should specifically discuss (not here) what you are after and all of the rest, **at area51.com**, then eventually make a PR.

In any case it's better to add just `$view_topic_url_params` (which serves $view_topic_url) to core.viewforum_modify_topicrow. I guess that's what you are after.

My previous pagination's suggestion does not apply anymore by my side.

DavidIQ:
This seems like a reasonable request and adjustment. We aren't strictly requiring a discussion be started anymore for every core event addition or adjustment, just that a ticket be created.

One of our developers might come along and say that the pagination variables should be added to the event instead, but I would just wait for one of them to suggest that first.

3D-I:
> We aren't strictly requiring a discussion be started anymore for every core event addition or adjustment, just that a ticket be created.
Good news here, I am still old fashion somehow..

The reasoning beyond this is I know of an extension toxyy managed, we discussed that at writers forum, which adds something to the url.. hence the suggestion about add just `$view_topic_url_params` (which serves _$view_topic_url_) to `core.viewforum_modify_topicrow`.

toxyy:
I want members to see soft deleted posts and to reflect that in the topic's post count and pagination in viewtopic and viewforum.  I was able to do this `viewtopic_assign_template_vars_before` changing `$start` and `$total_posts`, but the only way to access those per topicrow in viewforum is `$replies`.  I included `$view_topic_url` for good measure.

hanakin:
@toxyy this will need rebased on to 3.3 or master

hanakin:
This needs rebased on 3.3.x

toxyy:
@hanakin Rebased to 3.3.x

toxyy:
!unset Abandoned by Author 🏚️

marc1706 (approved):

rmcgirr83:
The event should have been modified. Now one of my extensions is borked. Thank you very much.

PR #5320, opened 2018-08-12
[ticket/15755] Fix event vars in /phpbb/attachment/delete.php

rxu:
For the ```3.2.x``` and ```master``` branches.

Checklist:

- [ ] Tests pass

<a href="https://tracker.phpbb.com/browse/PHPBB3-15755">PHPBB3-15755</a>.

### DIFF
--- a/phpBB/phpbb/attachment/delete.php
+++ b/phpBB/phpbb/attachment/delete.php
@@ -104,6 +104,8 @@ class delete
 
 		$this->set_sql_constraints($mode);
 
+		$sql_id = $this->sql_id;
+
 		/**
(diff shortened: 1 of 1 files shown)


Nicofuma:
@marc1706

marc1706:
I'm not a fan of adding parameters to methods just to be able to pass the mode or ids over to the events in a function that does the actual deletion. Instead, users should be able to use the event in the main method and handle things accordingly there. Therefore, variables that don't exist in the protected methods should just be dropped.

rxu:
If we drop variables in events that would break BC then. It is already broken as no variables passed to the event at all.

PR #5173, opened 2018-03-24
[ticket/15300] Maximum attachment file size shown in error alert

artur99:
Checklist:

- [ ] Correct branch: master for new features; 3.2.x for fixes
- [x] Tests pass
- [ ] Code follows coding guidelines: [master](https://area51.phpbb.com/docs/dev/master/development/coding_guidelines.html) and [3.2.x](https://area51.phpbb.com/docs/dev/3.2.x/development/coding_guidelines.html)
- [ ] Commit follows commit message [format](https://area51.phpbb.com/docs/dev/3.2.x/development/git.html)

Tracker ticket:
https://tracker.phpbb.com/browse/PHPBB3-12345

PR #5327, opened 2018-08-30
[ticket/15300] Show maximum attachment filesize

rubencm:
PHPBB3-15300

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15300

### DIFF
--- a/phpBB/includes/functions_posting.php
+++ b/phpBB/includes/functions_posting.php
@@ -730,7 +730,7 @@ function posting_gen_attachment_entry($attachment_data, &$filename_data, $show_a
 		'S_HAS_ATTACHMENTS'				=> count($attachment_data),
 		'FILESIZE'						=> $config['max_filesize'],
 		'FILE_COMMENT'					=> (isset($filename_data['filecomment'])) ? $filename_data['filecomment'] : '',
-		'MAX_ATTACHMENT_FILESIZE'		=> $user->lang('MAX_ATTACHMENT_FILESIZE', get_formatted_filesize($config['max_filesize'])),
+		'MAX_ATTACHMENT_FILESIZE'		=> $config['max_filesize'] > 0 ? $user->lang('MAX_ATTACHMENT_FILESIZE', get_formatted_filesize($config['max_filesize'])) : '',
 	));
 
 	if (count($attachment_data))
--- a/phpBB/includes/functions_posting.php
+++ b/phpBB/includes/functions_posting.php
@@ -726,10 +726,11 @@ function posting_gen_attachment_entry($attachment_data, &$filename_data, $show_a
 
 	// Some default template variables
 	$template->assign_vars(array(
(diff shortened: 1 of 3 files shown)


rubencm:
Another version of #5173

hanakin:
@rubencm perhaps just rebase on 3.3.x and we can merge this

rubencm:
@hanakin done

3D-I:
why just 3.3?

phpBB/includes/functions_posting.php
```diff
@@ -726,10 +726,11 @@ function posting_gen_attachment_entry($attachment_data, &$filename_data, $show_a
...
 	$template->assign_vars(array(
-		'S_SHOW_ATTACH_BOX'	=> $show_attach_box,
-		'S_HAS_ATTACHMENTS'	=> count($attachment_data),
-		'FILESIZE'			=> $config['max_filesize'],
-		'FILE_COMMENT'		=> (isset($filename_data['filecomment'])) ? $filename_data['filecomment'] : '',
+		'S_SHOW_ATTACH_BOX'				=> $show_attach_box,
+		'S_HAS_ATTACHMENTS'				=> count($attachment_data),
+		'FILESIZE'						=> $config['max_filesize'],
+		'FILE_COMMENT'					=> (isset($filename_data['filecomment'])) ? $filename_data['filecomment'] : '',
+		'MAX_ATTACHMENT_FILESIZE'		=> $user->lang('MAX_ATTACHMENT_FILESIZE', get_formatted_filesize($config['max_filesize'])),
```
3D-I, line 733: This will show `0 Bytes` instead of `Unlimited`, in case,  if I am not wrong?

hanakin:
its an improvement which technically should only go into master by definition, but its very minimal so I think we can do it in a 3.3.x update, but 3.2.x is frozen to bug/security fixes only which this is not.

phpBB/includes/functions_posting.php
marc1706: Adjusted to not display the info if the limit is set to 0.

3D-I:
@marc1706 
this deserves more attention, I have reread the ticket and reread this PR. Everything is fine but not great, this does not satisfy the request, it is a panacea. It certainly does not finalize the ticket in its entirety, which you can get IMHO.

phpBB/includes/functions_posting.php
3D-I: I do not tend to agree with it though. Sorry.

marc1706:
What part of the request does this not satisfy and what do you not agree with in the current design? Please be more specific, thanks.

3D-I:
@marc1706 

I'm sorry I didn't answer you right away but I don't know how I missed this notification, the last few weeks have been quite difficult here. I'll come back to the subject as soon as possible or I'll see if I can create a new PR with my suggestions, in the end it's nothing serious and we can move on if you like.

Thanks.

PR #5333, opened 2018-08-31
[ticket/12631] Add finder.not_use_cache

rubencm:
PHPBB3-12631

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-12631


kasimi:
I'd suggest renaming it to `finder.use_cache` and changing the logic accordingly, it's less confusing.

phpBB/phpbb/finder.php
```diff
@@ -421,10 +421,12 @@ public function find_from_extension($extension_name, $extension_path, $cache = t
 	*/
 	public function find_from_paths($extensions, $cache = true, $is_dir = false)
 	{
+		global $phpbb_container;
```
Nicofuma: I don't like this, I think it should be a parameter of the constructor but as a new finder is built each time, ww should create a finder_factory service that can be injected anywhere we need a finder.

phpBB/phpbb/hook/finder.php
```diff
@@ -55,7 +55,9 @@ public function __construct($phpbb_root_path, $php_ext, \phpbb\cache\driver\driv
 	*/
 	public function find($cache = true)
 	{
-		if (!defined('DEBUG') && $cache && $this->cache)
+		global $phpbb_container;
+
+		if (!$phpbb_container->getParameter('finder.not_use_cache') && $cache && $this->cache)
```
Nicofuma: just remove the DEBUG usage inn this file

phpBB/phpbb/finder.php
rubencm: Yes, that is why i asked on irc

rubencm: But... i dont like this finder that do a lot of things, i think there should be a finder, and other finder that extends from this one or something in extensions folder

Nicofuma: don't think it's a good idea to have multiple classes. The whole point of the finder is to be configurable.

phpBB/phpbb/hook/finder.php
rubencm: Can the hooks be removed? i remember it was like events but from phpBB 3.0/3.1 or something like that

phpBB/config/development/config.yml
```diff
@@ -17,3 +17,6 @@ core:
     extensions:
         composer_debug: true
         composer_verbose: true
+
+    finder:
+        not_use_cache: true
```
Nicofuma: and call it use_cache as suggested it is easier to read and understand :)

rubencm: Ah ok

tests/test_framework/phpbb_ui_test_case.php
```diff
@@ -458,7 +459,7 @@ protected function get_extension_manager()
 			$container,
 			$db,
 			$config,
-			$user,
```
rubencm: This argument is incorrect, but it wasn't giving any errors, maybe this method is never executed

PR #5337, opened 2018-09-04
[ticket/14754] Only one email notification per topic

senky:
I created a new table to track email notifications. Row is deleted when user reads the notification (e.g. visits a topic).
_______________________

PHPBB3-14754

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-14754

### DIFF
--- a/phpBB/phpbb/notification/method/email.php
+++ b/phpBB/phpbb/notification/method/email.php
@@ -32,19 +32,7 @@ class email extends \phpbb\notification\method\messenger_base
 	protected $db;
 
 	/** @var string */
-	protected $topics_watch_table;
-
-	/** @var string */
-	protected $topics_track_table;
-
-	/** @var string */
-	protected $posts_table;
-
-	/** @var string */
-	protected $forums_watch_table;
-
-	/** @var string */
-	protected $forums_track_table;
+	protected $email_notifications_table;
 
 	/**
 	 * Notification Method email Constructor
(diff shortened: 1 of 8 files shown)


phpBB/phpbb/notification/method/email.php
```diff
@@ -68,6 +98,53 @@ public function is_available(type_interface $notification_type = null)
...
 	}
 
+	/**
+	* {@inheritdoc}
+	*/
+	public function get_notified_users($notification_type_id, array $options, $notification_type_name)
+	{
+		$notified_users = array();
+
+		if ($notification_type_name == 'notification.type.post' && !empty($options['item_parent_id']))
```
Nicofuma: you don't need the name: you already have the notification ID and you should use that to retrieve the name if you need it

phpBB/phpbb/notification/method/email.php
```diff
@@ -68,6 +98,53 @@ public function is_available(type_interface $notification_type = null)
...
+			// Topics watch
+			$sql = 'SELECT tw.user_id
+					FROM ' . $this->topics_watch_table . ' tw
+					LEFT JOIN ' . $this->topics_track_table . ' tt
+						ON (tt.user_id = tw.user_id AND tt.topic_id = tw.topic_id)
+					LEFT JOIN ' . $this->posts_table . ' p
+						ON (p.topic_id = tw.topic_id)
+					WHERE tw.topic_id = ' . (int) $options['item_parent_id'] . '
+						AND p.post_time > tt.mark_time
+					HAVING COUNT(p.post_id) > 1';
```
Nicofuma: HAVING without GROUP BY is not allowed in all MySQL configuration

phpBB/phpbb/notification/method/email.php
senky: I didn't want to add yet another query to the process since the data are already available.

Nicofuma: it won't add more queries, we have helpers for that which are cachin the results (and the cache will already be loaded here)

phpBB/phpbb/notification/method/email.php
DavidIQ: Not allowed in SQL Server at all.

Nicofuma:
I'm not very a fan of this solution. Imo we should replicate what has been done for board notifications: track the notification status in a dedicated table and use that to know if it has already been sent. Maybe we could use the same table or maybe we shoud just duplicate it. I don't know.

What do you think @phpbb/development-team

phpBB/phpbb/notification/method/email.php
senky: Can you direct me to the method that converts ID to name? Because in the manager, we only have a name-to-id method, not the other way around...

phpBB/phpbb/notification/method/email.php
senky: 😭

Elsensee:
I agree with @Nicofuma. We can just use the information in the tracking table to see if the user has already seen that post and not send another notification until they have.

Also, you've removed files from the cache folder. (Also, why aren't they in .gitignore?)

joelst:
This PR has been sitting for a long time without any movement. Are you at an impasse or do you think that this can be resolved and merged?

Eldenroot:
bump, is there anyone who can finish this?

PR #5235, opened 2018-06-04
[ticket/15682] Revamp PM to chat based (GSoC 2018)

akbarhashmi:
Checklist:

- [x] Correct branch
- [ ] Tests pass
- [x] Code follows coding guidelines: 
- [x] Commit follows commit message 

**Screenshots:**
screen shot 2018-08-14 at 7 41 22 am

PHPBB3-15682
Tracker ticket 
https://tracker.phpbb.com/browse/PHPBB3-15682

PR #5351, opened 2018-09-13
[ticket/15682] Revamp PM

senky:
This is a continuation of https://github.com/phpbb/phpbb/pull/5235

Removals:
- smileys are removed (you can use emojis now, there is no reason to keep it in 2018)
- icons are removed (almost not used)
- reply subject is removed (hidden actually)

Changes:
- you can mark conversation important (previously you could mark every message separately)

------------------------------

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15682


Louis7777:
>     * smileys are removed (you can use emojis now, there is no reason to keep it in 2018)

Emojis are inconsistent across devices whereas with smileys you know exactly what you get and what others will see, plus you can have your own, the way you want them to be. They were and will be always superior, and relevant. In 2018 messenger apps (e.g. WhatsApp, Viber), Facebook, and all forum software still have smileys. Imo, it will be a bad move to remove a feature that has followed phpBB since the begin and that is still loved.

marc1706:
Looking at the PR, folders were partially removed. Is this supposed to be removed entirely?

hanakin:
in relation to the smiles removal. Add it back in as a dropdown I am working on a new emoji picker that will include them as well. It will work like the one on discord.

Indianos:
> Changes:
> * you can mark conversation important (previously you could mark every message separately)

I'm sorry for not digging myself in the code as it is currently not possible.
Will there be important conversations for each side separately or will it be important for everyone involved when somebody marks it?

PR #5361, opened 2018-09-15
[ticket/15790] Allow dynamic loading services files for extensions

rxu:
Checklist:

- [x] Tests pass

<a href="https://tracker.phpbb.com/browse/PHPBB3-15790">PHPBB3-15790</a>.

### DIFF
--- a/phpBB/phpbb/extension/di/extension_base.php
+++ b/phpBB/phpbb/extension/di/extension_base.php
@@ -155,7 +155,7 @@ class extension_base extends Extension
 	 */
 	public function getServicesFilenames($services_directory, $services_file_prefix = '')
 	{
-		$servises_files = array();
+		$services_files = array();
 		$finder = new \Symfony\Component\Finder\Finder();
 		$finder
 			->name("{$services_file_prefix}*.yml")
@@ -165,9 +165,9 @@ class extension_base extends Extension
 
 		foreach ($finder as $file)
 		{
-			$servises_files[] = $file->getBasename();
+			$services_files[] = $file->getBasename();
 		}
 
-		return $servises_files;
+		return $services_files;
 	}
 }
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/extension/di/extension_base.php
```diff
@@ -135,4 +143,30 @@ public function getAlias()
...
+	 * Gets the array of the Yaml filenames.
+	 *
+	 * @param string $services_directory   Directory in the extension folder containing services files
+	 * @param string $services_file_prefix Services files prefix to look for
+	 *
+	 * @return array The services filenames array
+	 */
+	public function getServicesFilenames($services_directory, $services_file_prefix = '')
+	{
+		$servises_files = array();
```
Derky: Typo in var name

Derky:
This way you can't inherit and overwrite settings from other files right? And it doesn't seem to check if it already tried to load the file right?

Elsensee:
In general, I don't get this change. The services are cached in the Container and it will only be rebuilt if the cache folder is cleared or extensions are enabled/disabled.

rxu:
> This way you can't inherit and overwrite settings from other files right?

As far as the files loading queue is not ordered, than rather right. Is it really needed? And if yes, any ideas about how to implement that?

> And it doesn't seem to check if it already tried to load the file right?

Didn't get the question. All files are unique, so why do we need to check that?

> The services are cached in the Container and it will only be rebuilt if the cache folder is cleared or extensions are enabled/disabled.

Sure. You disable the extension, drop in files, then enable the extension. Voila.

Elsensee:
> Sure. You disable the extension, drop in files, then enable the extension. Voila.

Yeah, but you could also make another extension that extends it, can't you?

rxu:
With this change you don't need to create new extension every time you need to make an addition to another extension.

rxu:
What I can say additionally is that with this change extension developer is free to decide if to use this feature or not.
That's easy: if you use `services_` prefix for your `.yml` config file, it's gonna be autoloaded. And vice versa.
If anyone is wanting the old behavior, just name files without `services_` prefix.

rxu:
@marc1706 Can we probably make a decision on this PR?

PR #5365, opened 2018-09-19
[ticket/15794] Convert prosilver to Twig Syntax

meis2m:
Convert prosilver to Twig Syntax

Checklist:

- [ ] Correct branch: master for new features; 4.0.x for fixes
- [ ] Tests pass
- [x] Code follows coding guidelines: [master](https://area51.phpbb.com/docs/dev/master/development/coding_guidelines.html) 

https://tracker.phpbb.com/browse/PHPBB3-15794


DavidIQ:
No X 1000000

meis2m:
> 
> 
> No X 1000000

why not David ?

DavidIQ:
1. 3.2.x is feature frozen.
2. This is basically a new feature when you consider you've modified 120 files.
3. It will confuse users at this point.
4. Seems like a waste of time when you consider prosilver will be replaced in 3.3 making this just some capricious thing that isn't needed that will cause more problems while solving none.

meis2m:
1. Yes you right not in 3.2
2. Yes agian.
3. Confuse?
4. i thought my work is a part of new theme for future. yes prosilver will be not in 3.3 or 4 i don't know. but new theme must should twig syntax right?
my work is reply to requesting from Front-end Development Lead.

DavidIQ:
So you're doing this at @hanakin's request? I was under the impression that prosilver is not going to be used at all. I guess he'll need to chime in.

rubencm:
Maybe for 3.3, but you will have to modify the tests too

hanakin:
@DavidIQ Ok so this was discussed in Krakow for those who were not there. I asked him to do this and its also been discussed on are51. This will go along way to help with the new themes development. It's not going to going to be in master. Its a starting point to work from and also we will still have to release a prosilver theme that we intend to provide minimal support for going forward for the transition to the new theme. It needs to be closely related to the same code.

@meis2m although I did say that you would not be able to do it in one pr. We will probably want to break it up into 4-5 files at a time to make it easier to review. @rubencm is correct so that is another reason that this will need to be done in smaller prs.

phpBB/styles/prosilver/template/index_body.html
```diff
@@ -1,79 +1,79 @@
...
 
-<!-- EVENT index_body_forumlist_body_after -->
+{% EVENT index_body_forumlist_body_after %}
 
-<!-- IF not S_USER_LOGGED_IN and not S_IS_BOT -->
-	<form method="post" action="{S_LOGIN_ACTION}" class="headerspace">
-	<h3><a href="{U_LOGIN_LOGOUT}">{L_LOGIN_LOGOUT}</a><!-- IF S_REGISTER_ENABLED -->&nbsp; &bull; &nbsp;<a href="{U_REGISTER}">{L_REGISTER}</a><!-- ENDIF --></h3>
+{% if not S_USER_LOGGED_IN and not S_IS_BOT %}
+	<form method="post" action="{{ S_LOGIN_ACTION }}" class="headerspace">
+	<h3><a href="{{ U_LOGIN_LOGOUT }}">{{ lang('LOGIN_LOGOUT') }}</a>{% if S_REGISTER_ENABLED %}&nbsp; &bull; &nbsp;<a href="{{ U_REGISTER }}">{{ lang('REGISTER') }}</a>{% endif %}</h3>
```
mrgoldy: `L_LOGIN_LOGOUT` is a php generated language string.

meis2m: Hi
it can not replace with
{ lang('LOGIN_LOGOUT') }
?

3D-I: No.

`{{ L_LOGIN_LOGOUT }}`

DavidIQ: To be clear:
https://github.com/phpbb/phpbb/blob/master/phpBB/language/en/common.php

A `LOGIN_LOGOUT` key does NOT exist so the `lang` helper function wouldn't be able to retrieve it. It is a template key built through code, as pointed out by @3D-I.

meis2m:
please see this:
https://www.phpbb.com/community/viewtopic.php?t=2424606
in "Language keys" section said:
 phpBB syntax is:
{L_FOO_BAR}

 Twig syntax is:
{{ lang('FOO_BAR') }}

DavidIQ:
Did you actually test these changes?

meis2m:
> 
> 
> Did you actually test these changes?

yes and it was working very well. you can test it.

DavidIQ:
It doesn't really seem to be correct usage of the `lang` function because `L_LOGIN_LOGOUT` isn't really a language key in the sense that it doesn't exist in a language file. It is actually a variable as per the topic you posted. However the `lang` function does find it because of how it searches for it within the `root_ref`. I guess it's maybe less confusing leaving the `lang` function than doing `{{ L_LOGIN_LOGOUT}}`? I'll let whoever is merging this decide.

DavidIQ:
Build fails anyway so you'll need to fix those issues first.

mrgoldy:
This raises a couple of question in regards to coding style.
There are a lot of possibilities with twig, so not sure what the phpBB standards are / should be.

## `include`
> As of Twig 1.12, it is recommended to use the include function instead as it provides the same features with a bit more flexibility.

Currently `{% INCLUDE TEMPLATE_FILE %}` file is bein used, which is phpBB's added syntax. Why not use `{{ include(TEMPLATE_FILE) }}`

## Colon
Currently colons are not concatenated: `{{ lang('SOMETHING') }}{{ lang('COLON') }}` while it looks better in my opinion if just writing `{{ lang('SOMETHING') ~ lang('COLON') }}`. Replaces `}}{{` with `~` is more overseeable.

## Variable concatenation
Current there are strings like these:
```html
{% if S_TOPIC_MODE %}{{ DELETE_TOPIC_PERMANENTLY_EXPLAIN }}{% else %}{{ DELETE_POST_PERMANENTLY_EXPLAIN }}{% endif %}
```
Which can be reduced to 
```html
{{ S_TOPIC_MODE ? DELETE_TOPIC_PERMANENTLY_EXPLAIN : DELETE_POST_PERMANENTLY_EXPLAIN }}
```

Or if we look at this line:
```html
{% if LOGIN_EXPLAIN %}{{ LOGIN_EXPLAIN }}{% else %}{{ lang('LOGIN') }}{% endif %}
```
This can be reduced to
```html
{{ LOGIN_EXPLAIN ?: lang('LOGIN') }}
```

## `S_ROW_COUNT`
This is often used to style background rows, zebra style.
`usernotes.S_ROW_COUNT is even` can be replaced with `loop.index is even` or `loop.index0 is even` is it has to be zero-based.

DavidIQ:
I'm not sure Twig's `include` function will look in the parent template or in the "all" folder for the template files. Probably won't work very well in extensions either.

The rest of what you said is on point and should probably be done.

hanakin:
Row count is not even nessecary code anymore can be completely removed for css. 💯 no to short ifs it’s not readable. Possibly ok with concatenation. Same question as David with includes

mrgoldy:
As far as the include calls go.
I can use `{{ include('pagination.html') }}` from within an extension.
This will work on prosilver and and on a custom style that has not event got a `template` folder.
So that means that inheritance is working properly.

As for including my own extension's files, both options work:
`{{ include('custom_body.html') }}`
`{{ include('@vendor_extension/custom_body.html') }}`

VSEphpbb:
I prefer to stick with the include as we've been using it all these years ... it works ... won't create confusion about which way we're supposed to use it down the road.

PR #5381, opened 2018-09-25
[ticket/15616] Remove jumpbox from login_forum.html

senky:
PHPBB3-15616

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15616

### DIFF
--- a/phpBB/includes/functions.php
+++ b/phpBB/includes/functions.php
@@ -2539,7 +2539,7 @@ function login_box($redirect = '', $l_explain = '', $l_success = '', $admin = fa
 */
 function login_forum_box($forum_data)
 {
-	global $db, $phpbb_container, $request, $template, $user, $phpbb_dispatcher;
+	global $db, $phpbb_container, $request, $template, $user, $phpbb_dispatcher, $phpbb_root_path, $phpEx;
 
 	$password = $request->variable('password', '', true);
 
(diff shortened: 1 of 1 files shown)


hanakin:
@Elsensee The question is really do we need a jumpbox from the login form? login auto redirects to index anyway...Do guests need to jump around?

Elsensee:
@hanakin Clearly not. I don't even know if the jumpbox is used that often in general...
But to me, this one is about consistency.

senky:
@hanakin but this isn't general login box. This is the page where you type a password for password-protected forum. You are redirected to the forum afterwards. So maybe it makes sense.

hanakin:
Wait this sounds like a larger stupid issue then. Essentially jump box is not callable outside of the individual pages which is just stupid. This should be a 100% agnostic capability. It’s the same issue we currently face with the ucp and mcp not being accessible outside itself

Elsensee:
@senky Can you reword the commit message, then this is good to go.

@hanakin I see why this was done this way. In earlier version we had our shitty own template engine which was not very capable of doing stuff, so in order to save resources and avoid querying all accessible forums on every page, there had to be a special function call.
Now, with twig, we could technically just remove that function call. We could just do `{{ jumpbox() }}` and then the calculation would run right there and no extra call in the code is necessary.

PR #5385, opened 2018-09-26
[ticket/15043] Rewrite get_context()

rubencm:
PHPBB3-15043

Checklist:

- [x] Tests pass
- [x] Handle case when there are no coincidences
- [x] If first fragment dont start at the beginning, add '... '

https://tracker.phpbb.com/browse/PHPBB3-15043

### DIFF
--- a/phpBB/includes/functions_content.php
+++ b/phpBB/includes/functions_content.php
@@ -325,83 +325,101 @@ function bump_topic_allowed($forum_id, $topic_bumped, $last_post_time, $topic_po
 *
 * @return	string			Context of the specified words separated by "..."
 */
-function get_context($text, $words, $length = 400)
+function get_context(string $text, array $words, int $length = 400): string
 {
-	$text_length = utf8_strlen($text);
-
-	// Replace all spaces/invisible characters with single spaces
-	$text = preg_replace("/\s+/", ' ', $text);
+	if ($length <= 0)
+	{
+		return '...';
+	}
 
 	// we need to turn the entities back into their original form, to not cut the message in between them
 	$text = html_entity_decode($text);
 
+	// Replace all spaces/invisible characters with single spaces
+	$text = preg_replace("/\s+/u", ' ', $text);
+
+	$text_length = utf8_strlen($text);
+
 	// Get first occurrence of each word
-	$word_indizes = [];
+	$word_indexes = [];
 	foreach ($words as $word)
 	{
 		$pos = utf8_stripos($text, $word);
 
 		if ($pos !== false)
 		{
-			$word_indizes[$pos] = $word;
+			$word_indexes[$pos] = $word;
 		}
 	}
 
-	// If there are coincidences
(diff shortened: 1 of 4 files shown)


phpBB/includes/functions_content.php
```diff
@@ -362,26 +362,28 @@ function get_context($text, $words, $length = 400)
 			$start = $end = 0;
 			foreach ($word_indizes as $indize => $word)
```
3D-I: `indize` is more a German word, may I ask the reason why?

rubencm: I kept some variable names from the original code while rewriting it

3D-I: I see :+1:

phpBB/includes/functions_content.php
```diff
@@ -327,117 +327,82 @@ function bump_topic_allowed($forum_id, $topic_bumped, $last_post_time, $topic_po
...
-	// first replace all whitespaces with single spaces
-	$text = preg_replace('/ +/', ' ', strtr($text, "\t\n\r\x0C ", '     '));
+	$output = '';
+	$text_length = utf8_strlen($text);
+
+	// Replace all spaces/invisible characters with single spaces
+	$text = preg_replace("/\s+/", ' ', $text);
 
 	// we need to turn the entities back into their original form, to not cut the message in between them
-	$entities = array('&lt;', '&gt;', '&#91;', '&#93;', '&#46;', '&#58;', '&#058;');
```
marc1706: Is there a reason why you're changing this to be broader and cover all entities? I think there was a point to only decode the characters that need to be decoded.

rubencm: Are you sure?

rubencm: That was written 10 years ago, this is the commit
https://github.com/phpbb/phpbb/commit/433f03107de1162744a1d0b8c9fa3c4ba1b3ed21

I think htmlentities and html_entity_decode do exactly that, and is more complete

rubencm: Maybe I should use the any of the flags and utf-8, but i'm not sure

marc1706: Ah ok, then there was no special point to it. Then this should be fine as it's just used for cutting at a specific length.

marc1706:
Since there are no tests for this function yet, I think it's a good idea to add some now while refactoring it. I think this is similar to what Joas did a while back in so these test cases might be of help with getting started: https://github.com/nickv-phpbb/phpbb-tool-trimmessage/blob/master/src/Nickvergessen/TrimMessage/Tests/TrimTest.php

rubencm:
Probably @JoshyPHP can do this much better without breaking bbcodes, I don't understand much that part.
I made this after runing strip_tags, so all tags are removed, but there are better ways to do it.

rubencm:
@marc1706

PR #5386, opened 2018-09-28
[ticket/15809] Allow twig syntax for template events

senky:
PHPBB3-15809

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15809

### DIFF
--- a/phpBB/phpbb/event/md_exporter.php
+++ b/phpBB/phpbb/event/md_exporter.php
@@ -439,16 +439,9 @@ class md_exporter
 		$event_list = array();
 		$file_content = file_get_contents($this->path . $file);
 
-		$events = explode('<!-- EVENT ', $file_content);
-		// Remove the code before the first event
-		array_shift($events);
-		foreach ($events as $event)
-		{
-			$event = explode(' -->', $event, 2);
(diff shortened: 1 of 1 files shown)


rubencm:
```
preg_match_all('/<!-- EVENT (.*) -->/U', $file_content, $event_list);
$event_list = $event_list[1];
```

Elsensee:
Still can't believe this is done without regular expressions and just with explode()'s...
We can actually just use on regex for both styles then:
```
preg_match_all('/(?:{%|<!--) EVENT (.*) (?:%}|-->)/U', $file_content, $event_list);
$event_list = $event_list[1];
```

rubencm:
Yes, that's better, but is better to have the preg_match_all twice, because one of them will have to be removed, so will be easier.

Elsensee:
You mean when we get rid of the old syntax? We could also just change the regular expression then...

rubencm:
Well, to put the `@deprecated` comment somewhere so its easier to find in the future, but ok

marc1706:
I was thinking about writing a PHP extension to extract the events but until that happens we should probably go with the regex and change it if needed (so just one please).

PR #5387, opened 2018-09-28
[ticket/15593] Do not allow print view with direct URL

senky:
PHPBB3-15593

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15593

### DIFF
--- a/phpBB/includes/ucp/ucp_pm.php
+++ b/phpBB/includes/ucp/ucp_pm.php
@@ -170,6 +170,12 @@ class ucp_pm
 					trigger_error('NO_AUTH_READ_MESSAGE');
 				}
 
+				if ($view == 'print' && (!$config['print_pm'] || !$auth->acl_get('u_pm_printpm')))
+				{
+					send_status_line(403, 'Forbidden');
+					trigger_error('NO_AUTH_PRINT_MESSAGE');
+				}
+
 				// Do not allow hold messages to be seen
 				if ($folder_id == PRIVMSGS_HOLD_BOX)
 				{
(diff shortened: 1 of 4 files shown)


Elsensee:
*prints the page without print view*

Elsensee:
Let me elaborate: I think it's stupid that you can disallow printing in the ACP because, well, you can't.
Administrators get a sense of false security if they choose to do so.

hanakin:
The current prosilver print file is fairly deprecated we have not maintained it. I am honestly looking not looking to include this in the new theme as it’s not really something I think very many use

senky:
I created a ticket to remove print view in 3.3 so we don't forget about it when the new style is up:
https://tracker.phpbb.com/browse/PHPBB3-15810

Louis7777:
Shame, I liked that one... I once had a forum with recipes and I know a few people would "print view" before printing (the first post with the recipe).

DavidIQ:
>           Shame, I liked that one... I once had a forum with recipes and I know a few people would "print view" before printing (the first post with the recipe).

A "print view" is possible with just CSS and does not require an actual feature for it (I agree that having it as a permissions-controlled feature is dumb). This is apparently what @hanakin plans to do in the new style. 😃

hanakin:
Tbh it’s a separate theme really so it will not be something that is part of the new theme. It should be something in the all folder

marc1706:
I think this fix is fine as it stands as it enforces what should already be enforced. Having a different type of print view is something for 3.3 and is somewhat unrelated to this change.

PR #5389, opened 2018-10-01
[ticket/15820] Fix `mssqlnative` Can`t store unicode

hubaishan:
SQL server treats any strings as ANSI until putting `N` before 
 quoted unicode text 

PHPBB3-15820

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15820


hubaishan:
This PR is not Enough  to fix this bug after merging this PR any `"'" . $db->sql_escape($x) . "'"` must be replaced with `$db->sql_quote($x)`

hanakin:
@hubaishan why not just handle in this pr?

hubaishan:
!set WIP

hubaishan:
!unset WIP

3D-I:
AFAIK the `$db->sql_quote()` has been added recently as per https://github.com/phpbb/phpbb/blame/864321ecf69a694a385ec1a623ac685c4eedd4e0/phpBB/phpbb/db/driver/factory.php#L463 (7 months ago) for 3.3.3-rc1.
Maybe I'm missing something but how do you use a feature that was not yet implemented in 2018 / 2019 🤔 
I've explored the whole lot of files included in this PR BTW.

phpBB/develop/adjust_avatars.php
```diff
@@ -51,13 +51,13 @@
 	if ($new_avatar_name !== false)
 	{
 		$sql = 'UPDATE ' . USERS_TABLE . "
-			SET user_avatar = '" . $db->sql_escape($new_avatar_name) . "'
-			WHERE user_avatar = '" . $db->sql_escape($row['group_avatar']) . "' 
+			SET user_avatar = " . $db->sql_quote($new_avatar_name) . "
+			WHERE user_avatar = " . $db->sql_quote($row['group_avatar']) . "
 			AND user_avatar_type = " . AVATAR_UPLOAD;
```
rxu, line 56: Here and in other places, phpBB core performs string escaping and then quotes it. Removing `sql_escape()` is not correct as it leads to SQL injections, so in this case it should be `$db->sql_quote($db->sql_escape)`, which doesn't look beautiful though.

3D-I: But `sql_quote()` did not exist at the time when this PR was made.
 I'm not sure where that came from, maybe the author forgot to push the necessary change as well?

hubaishan: see changes of ` phpBB/phpbb/db/driver/driver.php` my function is differ than rxu function

rxu: Strings have to be escaped, not quoted.

hubaishan: really!!
How can SQL identify  it as string without quote!!???

PR #5413, opened 2018-10-14
[ticket/15822] Create twig functions for jumpbox

rubencm:
PHPBB3-15822

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15822

### DIFF
--- a/phpBB/config/default/container/services_twig_extensions.yml
+++ b/phpBB/config/default/container/services_twig_extensions.yml
@@ -1,6 +1,8 @@
+# Twig extensions don't needed by the installer
+
 services:
-    jumpbox.twig.extension:
-        class: phpbb\jumpbox\twig\extension
+    template.twig.extensions.jumpbox:
+        class: phpbb\template\twig\extension\jumpbox
         arguments:
             - '@auth'
             - '@dbal.conn'
--- a/phpBB/phpbb/template/twig/extension/jumpbox.php
+++ b/phpBB/phpbb/template/twig/extension/jumpbox.php
@@ -120,7 +120,11 @@ class jumpbox extends \Twig_Extension
 			$tpl_ary = array(
 				'FORUM_ID'		=> $row['forum_id'],
 				'FORUM_NAME'	=> $row['forum_name'],
+				'SELECTED'		=> ($row['forum_id'] == $forum_id) ? ' selected="selected"' : '',
+				'S_FORUM_COUNT'	=> $iteration,
 				'S_IS_CAT'		=> ($row['forum_type'] == FORUM_CAT) ? true : false,
+				'S_IS_LINK'		=> ($row['forum_type'] == FORUM_LINK) ? true : false,
+				'S_IS_POST'		=> ($row['forum_type'] == FORUM_POST) ? true : false,
 				'LINK'			=> $this->path_helper->append_url_params("viewforum.php", array('f' => $row['forum_id'])),
 			);
 
--- a/phpBB/phpbb/template/twig/extension/jumpbox.php
+++ b/phpBB/phpbb/template/twig/extension/jumpbox.php
@@ -11,14 +11,14 @@
 *
 */
 
-namespace phpbb\jumpbox\twig;
+namespace phpbb\template\twig\extension;
 
 use phpbb\auth\auth;
 use phpbb\db\driver\driver_interface;
 use phpbb\event\dispatcher_interface;
 use phpbb\path_helper;
 
-class extension extends \Twig_Extension
+class jumpbox extends \Twig_Extension
 {
 	/**
 	 * Constructor
--- a/phpBB/phpbb/template/twig/extension/jumpbox.php
+++ b/phpBB/phpbb/template/twig/extension/jumpbox.php
@@ -54,7 +54,7 @@ class jumpbox extends \Twig_Extension
 	public function getFunctions()
 	{
 		return array(
-			new \Twig_SimpleFunction('jumpbox', [$this, 'jumpbox'], array('needs_environment' => true)),
+			new \Twig_SimpleFunction('JumpBox', [$this, 'jumpbox'], array('needs_environment' => true)),
 		);
 	}
 
--- a/phpBB/styles/prosilver/template/faq_body.html
+++ b/phpBB/styles/prosilver/template/faq_body.html
@@ -47,7 +47,7 @@
 <!-- END faq_block -->
 
 <div class="action-bar actions-jump">
-	{{ jumpbox() }}
+	{{ JumpBox() }}
 </div>
 
 <!-- INCLUDE overall_footer.html -->
--- a/phpBB/styles/prosilver/template/faq_body.html
+++ b/phpBB/styles/prosilver/template/faq_body.html
@@ -46,5 +46,8 @@
 	</div>
 <!-- END faq_block -->
 
-<!-- INCLUDE jumpbox.html -->
+<div class="action-bar actions-jump">
+	{{ jumpbox() }}
+</div>
+
 <!-- INCLUDE overall_footer.html -->
--- a/phpBB/styles/prosilver/template/login_forum.html
+++ b/phpBB/styles/prosilver/template/login_forum.html
@@ -38,7 +38,7 @@
 </form>
 
 <div class="action-bar actions-jump">
-	{{ jumpbox() }}
+	{{ JumpBox() }}
 </div>
 
 <!-- INCLUDE overall_footer.html -->
--- a/phpBB/styles/prosilver/template/login_forum.html
+++ b/phpBB/styles/prosilver/template/login_forum.html
@@ -37,5 +37,8 @@
 
 </form>
 
-<!-- INCLUDE jumpbox.html -->
+<div class="action-bar actions-jump">
+	{{ jumpbox() }}
+</div>
+
 <!-- INCLUDE overall_footer.html -->
--- a/phpBB/styles/prosilver/template/mcp_forum.html
+++ b/phpBB/styles/prosilver/template/mcp_forum.html
@@ -1,7 +1,7 @@
 <!-- INCLUDE mcp_header.html -->
 
 <!-- DEFINE $CUSTOM_FIELDSET_CLASS = 'forum-selection2' -->
-<!-- INCLUDE jumpbox.html -->
+<!-- INCLUDE mcp_jumpbox.html -->
 
 <h2><a href="{U_VIEW_FORUM}">{L_FORUM}{L_COLON} {FORUM_NAME}</a></h2>
 
--- a/phpBB/styles/prosilver/template/mcp_jumpbox.html
+++ b/phpBB/styles/prosilver/template/mcp_jumpbox.html
@@ -1,0 +1,18 @@
+<div class="action-bar actions-jump">
+	<div class="jumpbox dropdown-container dropdown-container-right<!-- IF not S_IN_MCP --> dropdown-up<!-- ENDIF --> dropdown-{S_CONTENT_FLOW_BEGIN} dropdown-button-control" id="jumpbox">
+			<span title="<!-- IF S_IN_MCP and S_MERGE_SELECT -->{L_SELECT_TOPICS_FROM}<!-- ELSEIF S_IN_MCP -->{L_MODERATE_FORUM}<!-- ELSE -->{L_JUMP_TO}<!-- ENDIF -->" class="button button-secondary dropdown-trigger dropdown-select">
+				<span><!-- IF S_IN_MCP and S_MERGE_SELECT -->{L_SELECT_TOPICS_FROM}<!-- ELSEIF S_IN_MCP -->{L_MODERATE_FORUM}<!-- ELSE -->{L_JUMP_TO}<!-- ENDIF --></span>
+				<span class="caret"><i class="icon fa-sort-down fa-fw" aria-hidden="true"></i></span>
+			</span>
+		<div class="dropdown">
+			<div class="pointer"><div class="pointer-inner"></div></div>
+			<ul class="dropdown-contents">
+				<!-- BEGIN jumpbox_forums -->
+				<!-- IF jumpbox_forums.FORUM_ID neq -1 -->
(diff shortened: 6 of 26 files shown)


rubencm:
!set WIP :construction:

rubencm:
Weird error, it only fails on mssql, i haven't changed the code, only moved it

Elsensee:
Why not have this function under \phpbb\template\twig\extension. I think it's weird to create a new jumpbox namespace..

Elsensee:
Well, the route function is also not part of the template engine and it's still there, so I'd definitely vote for putting it there; it just makes more sense. (The jumpbox also isn't a separate function like the captcha's for example, so its own namespace doesn't make sense to me)

Seems like MSSQL doesn't like the query you passed it. Query looks good to me, though...

rubencm:
@Elsensee I think it was better outside, so it's outside the installer environment where some of the services it uses don't exist

mrgoldy:
You're removing some variable from the template array, such as `S_IS_LINK`, `S_SELECTED`, etc.
Which aren't used by prosilver to theme anything differently, but might be by other style authors.
I see no down side in assigning those variables again?

rubencm:
Yes, probably is a good idea to add them again, I just tried to clean the function a little with unused/legacy code

phpBB/config/default/container/services_twig_extensions.yml
```diff
@@ -0,0 +1,12 @@
+# Twig extensions don't needed by the installer
+
+services:
+    template.twig.extensions.jumpbox:
+        class: phpbb\template\twig\extension\jumpbox
```
hanakin: Tlaking to @marc1706 we are going to be enforceing a naming convention for all of our twig funcitns that return template code using Pascal Case so please use JumpBox for the function name to avoid clshing with css class names

phpBB/phpbb/template/twig/extension/jumpbox.php
```diff
@@ -0,0 +1,178 @@
...
+*/
+
+namespace phpbb\template\twig\extension;
+
+use phpbb\auth\auth;
+use phpbb\db\driver\driver_interface;
+use phpbb\event\dispatcher_interface;
+use phpbb\path_helper;
+
+class jumpbox extends \Twig_Extension
```
hanakin: JumpBox

phpBB/styles/prosilver/template/faq_body.html
```diff
@@ -46,5 +46,8 @@ <h2 class="faq-title">{faq_block.BLOCK_TITLE}</h2>
 	</div>
 <!-- END faq_block -->
 
-<!-- INCLUDE jumpbox.html -->
+<div class="action-bar actions-jump">
+	{{ jumpbox() }}
```
hanakin: JumpBox()

phpBB/styles/prosilver/template/login_forum.html
```diff
@@ -37,5 +37,8 @@ <h2 class="login-title">{L_LOGIN}</h2>
 
 </form>
 
-<!-- INCLUDE jumpbox.html -->
+<div class="action-bar actions-jump">
+	{{ jumpbox() }}
```
hanakin: JumpBox()

phpBB/styles/prosilver/template/memberlist_body.html
```diff
@@ -159,6 +159,8 @@ <h2 class="solo">{PAGE_TITLE}<!-- IF SEARCH_WORDS -->{L_COLON} <a href="{U_SEARC
 <!-- IF S_IN_SEARCH_POPUP -->
 	<!-- INCLUDE simple_footer.html -->
 <!-- ELSE -->
-	<!-- INCLUDE jumpbox.html -->
+	<div class="action-bar actions-jump">
+		{{ jumpbox() }}
```
hanakin: JumpBox()

phpBB/styles/prosilver/template/memberlist_team.html
```diff
@@ -43,5 +43,8 @@ <h2 class="solo">{PAGE_TITLE}</h2>
 
 </form>
 
-<!-- INCLUDE jumpbox.html -->
+<div class="action-bar actions-jump">
+	{{ jumpbox() }}
```
hanakin: JumpBox()

phpBB/styles/prosilver/template/memberlist_view.html
```diff
@@ -139,6 +139,8 @@ <h3>{L_SIGNATURE}</h3>
 
 <!-- EVENT memberlist_view_content_append -->
 
-<!-- INCLUDE jumpbox.html -->
+<div class="action-bar actions-jump">
+	{{ jumpbox() }}
```
hanakin: JumpBox()

phpBB/styles/prosilver/template/search_results.html
```diff
@@ -237,6 +237,23 @@ <h3><a href="{searchresults.U_VIEW_POST}">{searchresults.POST_SUBJECT}</a></h3>
...
+	</p>
+	<!-- ELSEIF S_SEARCH_ACTION -->
+	<p class="jumpbox-return">
+		<a class="left-box arrow-{S_CONTENT_FLOW_BEGIN}" href="{U_SEARCH}" title="{L_SEARCH_ADV}" accesskey="r">
+			<i class="icon fa-angle-{S_CONTENT_FLOW_BEGIN} fa-fw icon-black" aria-hidden="true"></i><span>{L_GO_TO_SEARCH_ADV}</span>
+		</a>
+	</p>
+	<!-- ENDIF -->
+
+	{{ jumpbox() }}
```
hanakin: JumpBox()

phpBB/styles/prosilver/template/viewforum_body.html
```diff
@@ -293,7 +293,16 @@ <h3><a href="{U_LOGIN_LOGOUT}">{L_LOGIN_LOGOUT}</a><!-- IF S_REGISTER_ENABLED --
...
 
-<!-- INCLUDE jumpbox.html -->
+<div class="action-bar actions-jump">
+	<p class="jumpbox-return">
+		<a href="{U_INDEX}" class="left-box arrow-{S_CONTENT_FLOW_BEGIN}" accesskey="r">
+			<i class="icon fa-angle-{S_CONTENT_FLOW_BEGIN} fa-fw icon-black" aria-hidden="true"></i><span>{L_RETURN_TO_INDEX}</span>
+		</a>
+	</p>
+
+	{{ jumpbox() }}
```
hanakin: JumpBox()

phpBB/styles/prosilver/template/viewonline_body.html
```diff
@@ -59,5 +59,8 @@ <h2 class="viewonline-title">{TOTAL_REGISTERED_USERS_ONLINE}</h2>
 	</div>
 </div>
 
-<!-- INCLUDE jumpbox.html -->
+<div class="action-bar actions-jump">
+	{{ jumpbox() }}
```
hanakin: JumpBox()

phpBB/styles/prosilver/template/viewtopic_body.html
```diff
@@ -440,7 +440,17 @@ <h3 <!-- IF postrow.S_FIRST_ROW -->class="first"<!-- ENDIF -->><!-- IF postrow.P
...
-<!-- INCLUDE jumpbox.html -->
+
+<div class="action-bar actions-jump">
+	<p class="jumpbox-return">
+		<a href="{U_VIEW_FORUM}" class="left-box arrow-{S_CONTENT_FLOW_BEGIN}" accesskey="r">
+			<i class="icon fa-angle-{S_CONTENT_FLOW_BEGIN} fa-fw icon-black" aria-hidden="true"></i><span>{L_RETURN_TO_FORUM}</span>
+		</a>
+	</p>
+
+	{{ jumpbox() }}
```
hanakin: JumpBox()

phpBB/styles/prosilver/template/mcp_jumpbox.html
```diff
@@ -0,0 +1,18 @@
+<div class="action-bar actions-jump">
+	<div class="jumpbox dropdown-container dropdown-container-right<!-- IF not S_IN_MCP --> dropdown-up<!-- ENDIF --> dropdown-{S_CONTENT_FLOW_BEGIN} dropdown-button-control" id="jumpbox">
+			<span title="<!-- IF S_IN_MCP and S_MERGE_SELECT -->{L_SELECT_TOPICS_FROM}<!-- ELSEIF S_IN_MCP -->{L_MODERATE_FORUM}<!-- ELSE -->{L_JUMP_TO}<!-- ENDIF -->" class="button button-secondary dropdown-trigger dropdown-select">
+				<span><!-- IF S_IN_MCP and S_MERGE_SELECT -->{L_SELECT_TOPICS_FROM}<!-- ELSEIF S_IN_MCP -->{L_MODERATE_FORUM}<!-- ELSE -->{L_JUMP_TO}<!-- ENDIF --></span>
+				<span class="caret"><i class="icon fa-sort-down fa-fw" aria-hidden="true"></i></span>
+			</span>
+		<div class="dropdown">
+			<div class="pointer"><div class="pointer-inner"></div></div>
+			<ul class="dropdown-contents">
+				<!-- BEGIN jumpbox_forums -->
```
hanakin: why are we not using the twig function for this? Or have you just not gotten to it yet?

rubencm: I think is better to use another thing for this, because we had two diferent jumpbox in one single function.
They are similar, they can use the same forum_provider or whatever (when its done), but don't use both in the same function.
Anyway, i think this pr was just a concept.

hanakin: sure im just not sure why they need to be different. I have not gotten there yet with the new theme . Granted im not using the db scheme from prosilver for my testsfor chameleon, but the way I am handling it is far simpler https://github.com/phpbb/chameleon/blob/master/all/views/includes/jumpto.twig

rubencm: That looks much better than in prosilver.
But the purpose of this was to propose a way to remove the make_jumpbox function from all controllers.

hanakin: ok got it, but we still need to maintain consistency across the theme in the way that we are creating the jumpbox. This is probably something that needs flagged as a bug then for the fron-end to be handle first. As again not sure why we are doing this differently in different places possibly an oversight when jumpbox.html was created

rubencm: Just close this if you are going to do it in another way.

phpBB/styles/prosilver/template/mcp_forum.html
```diff
@@ -1,7 +1,7 @@
 <!-- INCLUDE mcp_header.html -->
 
 <!-- DEFINE $CUSTOM_FIELDSET_CLASS = 'forum-selection2' -->
-<!-- INCLUDE jumpbox.html -->
+<!-- INCLUDE mcp_jumpbox.html -->
```
hanakin: eitherway I would still do the action bar div external to the file
`
<div class="action-bar actions-jump">

</div>
`

phpBB/styles/prosilver/template/mcp_jumpbox.html
hanakin: wait Im still confused sorry I do not have access to a code editor so I m stuck with only github at my disposal. What exactly id the differrence between jumpbox.html and mcp_jumpbox.html besides the wrapping div you left in mcp_jumpbox.html

rubencm: In jumpbox.html i removed all the code related to the mcp.
The diference is that one is the normal jumpbox to jump between forums, and the other is to jump between moderate forums in mcp (one displays the forums the user have permission to see, and the other display the forums the user have permission to moderate). I think both jumpbox should share code/dependencies, but they shouldn't be the same, because the code becomes complex (but i can be wrong)
Also there is another jumpbox that isn't a jumpbox, that is the one used in mcp.

Normal jumpbox
https://i.imgur.com/KIRdigj.png

Mcp jumpbox:
https://i.imgur.com/U9oZNPK.png

Select that isn't a jumpbox, but do the same:
https://i.imgur.com/2oICfKB.png

hanakin: ok I gotcha yeah its not really overly complex looks like the only differencs is the title and the css classes which can both be passed in as paramaters to the function.

`JumpBox(css_classes, title = lang('JUMP_TO'), )`

`lang('DIRECTION') == 'ltr' ? 'left' : 'right'` is not needed just use {{ S_CONTENT_FLOW_BEGIN }}

something like

```
<div class="jumpbox dropdown-container dropdown-container-righ dropdown-{{ S_CONTENT_FLOW_BEGIN }}} dropdown-button-control{% css_classes ? css_classes : '' %}" id="jumpbox">
			<span title="{{ title }}" class="button button-secondary dropdown-trigger dropdown-select">
				<span{{ title }}</span>
				<span class="caret"><i class="icon fa-sort-down fa-fw" aria-hidden="true"></i></span>
			</span>
		<div class="dropdown">
			<div class="pointer"><div class="pointer-inner"></div></div>
			<ul class="dropdown-contents">
				{% for forum in jumpbox_forums %}
				{% if forum.FORUM_ID != -1 %}
				<li>
					<a href="{{ forum.LINK }}" class="{% if forum.level %}jumpbox-sub-link{% elseif forum.S_IS_CAT %}jumpbox-cat-link{% else %}jumpbox-forum-link{% endif %}">
						{% for level in forum.level %}<span class="spacer"></span>{% endfor %}
						<span>
							{% if forum.level %}{% if lang('DIRECTION') == 'rtl' %}&#8626;{% else %}&#8627;{% endif %} &nbsp;{% endif %}
							{{ forum.FORUM_NAME }}
						</span>
				</li>
				{% endif %}
				{% endfor %}
			</ul>
		</div>
	</div>
```

rubencm: Yes, the template probably can be the same, i was saying the php function.
So finally you like this or is going to be done in other way?

hanakin: as long as you can handle it with the same template for all then yes

rubencm: Ah ok, the idea for now is wait for the index refactor and use its forum_provider (or whatever) here.
I will update the name tomorrow

hanakin: ok no worries sounds good

phpBB/phpbb/template/twig/extension/jumpbox.php
rubencm: That should be in line 57 I think, php classes should use the phpbb code style

PR #5446, opened 2018-11-14
[Ticket/14771]  Support playing audio attachments

hubaishan:
replaces https://github.com/phpbb/phpbb/pull/4439
Nearly all work is from https://github.com/danchr
I rebased and resolved conflicts.

I am obliged to remove his commit because of error in commit title

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-14771


3D-I:
The `[audio-support]` header on the first old commit isn't allowed..

3D-I:
A TABs issue https://travis-ci.org/phpbb/phpbb/jobs/455419314#L470-L479

phpBB/includes/acp/acp_attachments.php
```diff
@@ -630,6 +630,7 @@ function main($id, $mode)
 				$cat_lang = array(
 					ATTACHMENT_CATEGORY_NONE		=> $user->lang['NO_FILE_CAT'],
 					ATTACHMENT_CATEGORY_IMAGE		=> $user->lang['CAT_IMAGES'],
+					ATTACHMENT_CATEGORY_AUDIO		=> $user->lang['CAT_AUDIO_FILES'],
```
rubencm: You should use language component better, and also don't access the property directly

hubaishan: Change in whole file or only in this line.
In ACP language component used only in acp_extensions.php all other files uses user->lang

rubencm: Uhm, only that line

hubaishan: You mean using `$user->lang('CAT_AUDIO_FILES')` only because using $language require adding to global, right?

DavidIQ: Using it as an array is deprecated and Ruben has a task to replace all instances of the old with the new. Adding more of the old deprecated method will not be allowed.

phpBB/install/schemas/schema_data.sql
```diff
@@ -795,12 +796,15 @@ INSERT INTO phpbb_extensions (group_id, extension) VALUES (4, 'ods');
...
-INSERT INTO phpbb_extensions (group_id, extension) VALUES (5, 'mp3');
 INSERT INTO phpbb_extensions (group_id, extension) VALUES (5, 'mpeg');
 INSERT INTO phpbb_extensions (group_id, extension) VALUES (5, 'mpg');
-INSERT INTO phpbb_extensions (group_id, extension) VALUES (5, 'ogg');
 INSERT INTO phpbb_extensions (group_id, extension) VALUES (5, 'ogm');
 
+INSERT INTO phpbb_extensions (group_id, extension) VALUES (6, 'mp3');
+INSERT INTO phpbb_extensions (group_id, extension) VALUES (6, 'wav');
+INSERT INTO phpbb_extensions (group_id, extension) VALUES (6, 'm4a');
+INSERT INTO phpbb_extensions (group_id, extension) VALUES (6, 'ogg');
```
senky: `webm` should be here as well.

hubaishan: `webm` is used mostly for video. can be used for  audio only but how we can determine it is audio only.

senky: Well, mimetype is either audio/webm or video/webm.

hubaishan:
I can`t locate the error in travis check. very long report

rubencm:
I have removed the \` character in the commit message, you will have to update that

phpBB/includes/functions_content.php
```diff
@@ -1280,6 +1281,19 @@ function parse_attachments($forum_id, &$message, &$attachments, &$update_count_a
...
 				break;
 
+				// Audio files
+				case ATTACHMENT_CATEGORY_AUDIO:
+					$inline_link = append_sid("{$phpbb_root_path}download/file.$phpEx", 'id=' . $attachment['attach_id']);
+					$block_array += array(
+						'S_AUDIO_FILE'			=> true,
+						'U_FORUM'			=> generate_board_url(),
+						'ATTACH_ID'			=> $attachment['attach_id'],
+					);
```
rxu, line 1291: Short array syntax `[]` can be used here (actually this is what being done for all new core code applying arrays).

phpBB/install/schemas/schema_data.sql
```diff
@@ -835,12 +836,16 @@ INSERT INTO phpbb_extensions (group_id, extension) VALUES (4, 'ods');
...
+INSERT INTO phpbb_extensions (group_id, extension) VALUES (6, 'webm');
```
rxu, line 847: For this and above changes in this file - if you change it for new installations, then migration for update existing ones is required.

phpBB/styles/prosilver/template/attachment.html
```diff
@@ -29,6 +29,19 @@
...
+			<!-- IF _file.COMMENT --><dd><em>{_file.COMMENT}</em></dd><!-- ENDIF -->
+			<dd>{_file.DOWNLOAD_NAME} ({_file.FILESIZE} {_file.SIZE_LANG}) {_file.L_DOWNLOAD_COUNT}</dd>
+			<dd>
+				<audio controls="controls" preload="none" style="max-width:100%;width:360px">
+					<source src="{_file.U_DOWNLOAD_LINK}" type="{_file.MIMETYPE}">
+				</audio>
+			</dd>
+		</dl>
+		<!-- ENDIF -->
+
```
rxu, line 44: For new template code, Twig is to be used.

phpBB/install/schemas/schema_data.sql
hubaishan: No migration was done for `phpbb_extensions` before, Users are able to change this values, So we can't expect what is in user table

rxu: The fact extensions table wasn't changed previously doesn't mean it shouldn't be changed :) 
Without migration, updated boards will be simply missing the feature.

marc1706:
Rebased the PR and cleaned up a bit, e.g. with CSS classes and re-arranged items. Audio playback looks like this now:
image

phpBB/styles/prosilver/theme/content.css
```diff
@@ -682,6 +682,10 @@ dl.thumbnail dt a:hover img {
 	border: 1px solid transparent;
 }
 
+audio.file-audio-controls {
```
hanakin, line 685: we do not allow over-qualifying selectors in 4.0. Please change this to `.c-file-audio-controls`

phpBB/styles/prosilver/template/attachment.html
```diff
@@ -29,6 +29,14 @@
 		</dl>
 		<!-- ENDIF -->
 
+		{% if _file.S_AUDIO_FILE %}
+		<dl class="file">
+			<dt><audio class="file-audio-controls" controls="controls" preload="metadata"><source src="{{ _file.U_DOWNLOAD_LINK }}" type="{{ _file.MIMETYPE }}"></audio></dt>
```
hanakin, line 34: Every element needs a class in 4.0, also every element should be on its own line this line should be
```
		<dt class="c-file-header">
			<audio class="c-file-audio-controls" controls="controls" preload="metadata">
				<source class="c-file-audio-source" src="{{ _file.U_DOWNLOAD_LINK }}" type="{{ _file.MIMETYPE }}">
			</audio>
		</dt>
```

phpBB/styles/prosilver/template/attachment.html
```diff
@@ -29,6 +29,14 @@
...
+			{% if _file.COMMENT %}<dd><em>{{ _file.COMMENT }}</em></dd>{% endif %}
```
hanakin, line 35: again please fix it like above give the dd a class and do not use the em tag but instead style it using the class.
```
		{% if _file.COMMENT %}
			<dd class="c-file-comment">{{ _file.COMMENT }}</dd>
		{% endif %}
```

phpBB/styles/prosilver/template/attachment.html
```diff
@@ -29,6 +29,14 @@
...
+			<dd>{% if _file.UPLOAD_ICON %}{{ _file.UPLOAD_ICON }} {% endif %}<a class="postlink" href="{{ _file.U_DOWNLOAD_LINK }}" download>{{ _file.DOWNLOAD_NAME }}</a> ({{ _file.FILESIZE }} {{ _file.SIZE_LANG }}) {{ _file.L_DOWNLOAD_COUNT }}</dd>
```
hanakin, line 36: yet again please fix the spacing here, also give the link a component class as well like so
```
		<dd class="c-file-actions">
			{% if _file.UPLOAD_ICON %}
				<span class="o-icon c-file-icon">{{ _file.UPLOAD_ICON }}</span>
			{% endif %}
			<a class="c-file-actions-link postlink" href="{{ _file.U_DOWNLOAD_LINK }}" download>{{ _file.DOWNLOAD_NAME }}</a>
			<span class="c-file-actions-meta"> ( {{ _file.FILESIZE }} {{ _file.SIZE_LANG }}) {{ _file.L_DOWNLOAD_COUNT }} )</span>
		</dd>
```
I am not sure about the icon I will need to look deeper into how this icon is working? and what its purpose is.. Why is this coming from the database and not just using the normal icon?

phpBB/styles/prosilver/template/attachment.html
```diff
@@ -29,6 +29,14 @@
 		</dl>
 		<!-- ENDIF -->
 
+		{% if _file.S_AUDIO_FILE %}
+		<dl class="file">
```
hanakin, line 33: this needs indented

hanakin (changes requested):

phpBB/phpbb/db/migration/data/v400/add_audio_files_attachment_group.php
```diff
@@ -0,0 +1,86 @@
...
+			WHERE ' . $this->db->sql_build_array('SELECT', ['group_name' => 'AUDIO_FILES']);
+		$result = $this->db->sql_query($sql);
+		$audio_group_id = $this->db->sql_fetchfield('group_id');
+		$this->db->sql_freeresult($result);
+
+		if ($audio_group_id === false)
+		{
+			$sql = 'INSERT INTO ' . $this->table_prefix . 'extension_groups ' . $this->db->sql_build_array('INSERT', [
+				'group_name'		=> 'AUDIO_FILES',
+				'cat_id'			=> 7,
```
marc1706, line 43: Should use actual constants

phpBB/phpbb/db/migration/data/v400/add_audio_files_attachment_group.php
```diff
@@ -0,0 +1,86 @@
...
+				'upload_icon'		=> '',
+				'max_filesize'		=> 0,
+				'allowed_forums'	=> '',
+			]);
+			$this->db->sql_query($sql);
+			$audio_group_id = $this->db->sql_nextid();
+		}
+		else
+		{
+			$sql = 'UPDATE ' . $this->table_prefix . 'extension_groups SET cat_id = 7
```
marc1706, line 54: Should use actual constant

marc1706:
I've also created a ticket that should add the constants as constants in the attachment manager class.

rubencm (approved):

phpBB/styles/prosilver/template/attachment.html
```diff
@@ -29,6 +29,26 @@
...
+					<audio class="c-file-audio-controls" controls="controls" preload="metadata">
+						<source class="c-file-audio-source" src="{{ _file.U_DOWNLOAD_LINK }}" type="{{ _file.MIMETYPE }}">
+					</audio>
+				</dt>
+				{% if _file.COMMENT %}
+					<dd class="c-file-comment">{{ _file.COMMENT }}</dd>
+				{% endif %}
+				<dd class="c-file-actions">
+					{% if _file.UPLOAD_ICON %}
+						<span class="o-icon c-file-icon">{{ _file.UPLOAD_ICON }}</span>
```
hanakin, line 44: again not sure where this icon is coming from...does it need to be a DB icon? Is it user specified in some way? Eitherway it needs an svg implementation using the icon twig function The goal is all icon/image assets in 4.0 use this to include eventually smiles and rank imgs. May be outside of the scope here but if I do not ask I will forget down the road...

hanakin (approved):

phpBB/styles/prosilver/template/attachment.html
marc1706: This is from the database, correct. I think we have similar things to change for the topic icons and should probably also rethink these types of icons then. I do also agree that it is outside the scope of this PR.

phpBB/styles/prosilver/template/attachment.html
```diff
@@ -29,6 +29,26 @@
 		</dl>
 		<!-- ENDIF -->
 
+		{% if _file.S_AUDIO_FILE %}
+			<dl class="file">
+				<dt class="c-file-header">
+					<audio class="c-file-audio-controls" controls="controls" preload="metadata">
+						<source class="c-file-audio-source" src="{{ _file.U_DOWNLOAD_LINK }}" type="{{ _file.MIMETYPE }}">
```
3D-I, line 36: The `class="c-file-audio-source"` is not defined in the CSS.

marc1706: Which is fine, there is currently no need to style it but with this being declared one could style it if one would want to.

PR #5452, opened 2018-11-22
[ticket/15879] Add core.posting_modify_uninit_data

toxyy:
Allows the modification of default variables put into post_data.
Also allows you to unset variables in post_data to inherit the default ones.

PHPBB3-15879

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15879

### DIFF
--- a/phpBB/posting.php
+++ b/phpBB/posting.php
@@ -597,6 +597,20 @@ if (isset($post_data['post_text']))
 // Set some default variables
 $uninit = array('post_attachment' => 0, 'poster_id' => $user->data['user_id'], 'enable_magic_url' => 0, 'topic_status' => 0, 'topic_type' => POST_NORMAL, 'post_subject' => '', 'topic_title' => '', 'post_time' => 0, 'post_edit_reason' => '', 'notify_set' => 0);
 
+/**
+* This event allows you to modify the default variables for post_data, and unset them in post_data if needed
+*
+* @event core.posting_modify_uninit_data
+* @var	array	post_data	Array with post data
+* @var	array	uninit		Array with default vars to put into post_data, if they aren't there
+* @since 3.2.5-RC1
+*/
+$vars = array(
+	'post_data',
+	'uninit',
+);
+extract($phpbb_dispatcher->trigger_event('core.posting_modify_uninit_data', compact($vars)));
+
 foreach ($uninit as $var_name => $default_value)
 {
 	if (!isset($post_data[$var_name]))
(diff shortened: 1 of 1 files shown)


toxyy:
!set Event

phpBB/posting.php
```diff
@@ -606,9 +606,24 @@
...
+*/
+$vars = array(
+	'post_data',
+	'attachment_user_id',
+);
+extract($phpbb_dispatcher->trigger_event('core.posting_modify_attachment_user_id', compact($vars)));
+
 // Always check if the submitted attachment data is valid and belongs to the user.
 // Further down (especially in submit_post()) we do not check this again.
-$message_parser->get_submitted_attachment_data($post_data['poster_id']);
```
3D-I: I'd move the event underneath this line and add the `message_parser` to the array.

Then undo the changes to the `message_parser` 's line of code.

3D-I:
Wrong header for the commit `add $post_data to event ` - should be like the first one.

phpBB/posting.php
toxyy: Unfortunately if `get_submitted_attachment_data` is called at all with an unmodified poster_id in my situation, users would still get an error.  I attempted to move it and add the message parser and re execute it under my conditions, it doesn't bypass the error.

toxyy:
I'll change it once the review you made is resolved.

phpBB/posting.php
3D-I: Without testing I'd go that way, more or less.

```
$message_parser->get_submitted_attachment_data($post_data['poster_id']);

/**
* This event allows you to modify the post data etc..
*
* @event core.posting_modify_post_data
* @var	array	post_data		Array with post data
* @var	@todo	message_parser	The message parser etc..
* @since 3.2.5-RC1
*/
$vars = array(
	'post_data',
	'message_parser',
);
extract($phpbb_dispatcher->trigger_event('core.posting_modify_post_data', compact($vars)));

```

and in the event
```

$message_parser = event['message_parser'];
$poster_id = event['post_data']['poster_id'];

$my_id = do something to define your user ID;
$poster_id = (int) $my_id;

$message_parser->get_submitted_attachment_data($poster_id);

event['message_parser'] = event['message_parser'];

```

Just to throw something on the table though.

toxyy: That's pretty much exactly what I did after your initial suggestion.  Since `$message_parser->get_submitted_attachment_data($post_data['poster_id'])` is ran with the wrong poster_id initially, it throws the error and doesn't give the message_parser a second chance to run.

toxyy: image

See the error, it gets automatically called at line 614 (the first call to get_submitted_attachment_data), and nothing in the event can be executed.

3D-I: Reading the code a bit more, the `poster_id` is being set here (` $user->data['user_id']` )

https://github.com/phpbb/phpbb/blob/838b5159acf3a10168b28c23226ff229b7960c90/phpBB/posting.php#L597

Why don't you try to use that array  `uninit` with a new event?

toxyy: I'll test doing that tomorrow, as I'm unsure of the affect of letting the remainder of posting.php process the post with a modified poster_id.  I'd like to keep it ANONYMOUS as long as I can, until I might change it in other events.  Nevertheless I'll attempt this tomorrow and see what happens.

toxyy: Seems to be working just fine, and I suppose this event is more useful than the one I made anyways.  Thanks for your suggestions and help.

3D-I:
The problem here will be you should rebase (and possibly cherry-pick) the PR/commits since there is always the wrong header for the prior commit `add $post_data to event` 

If you are not confortable with that I suggest you to close this PR and open a new one.
It is faster, nonetheless.

3D-I:
In fact
```

Error: heading should be a sentence beginning with a letter. 
You entered: 
add $post_data to event 
Syntax error on line 1: 
>> add $post_data to event 
Expecting: comment, header, 
Aborting commit. 
```

PR #5453, opened 2018-11-22
[ticket/15880] Add core event display_user_activity_modify_sql

toxyy:
Allows the sql queries that fetch most active topic and forum to be modified.
These queries are only run when viewing a user's profile page.
You can exclude posts with certain parameters this way.

PHPBB3-15880

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15880

### DIFF
--- a/phpBB/includes/functions_display.php
+++ b/phpBB/includes/functions_display.php
@@ -1196,15 +1196,52 @@ function display_user_activity(&$userdata_ary)
 		$phpbb_content_visibility = $phpbb_container->get('content.visibility');
 
 		// Obtain active forum
-		$sql = 'SELECT forum_id, COUNT(post_id) AS num_posts
-			FROM ' . POSTS_TABLE . '
-			WHERE poster_id = ' . $userdata_ary['user_id'] . '
-				AND post_postcount = 1
-				AND ' . $phpbb_content_visibility->get_forums_visibility_sql('post', $forum_ary) . '
-			GROUP BY forum_id
-			ORDER BY num_posts DESC';
-		$result = $db->sql_query_limit($sql, 1);
+		$forum_sql_ary = array(
+			'SELECT'	=> 'forum_id, COUNT(post_id) AS num_posts',
+			'FROM'		=> array(
+				POSTS_TABLE	=> 'p',
+			),
+			'WHERE'		=> array(
+				'poster_id = ' . $userdata_ary['user_id'],
+				'AND post_postcount = 1',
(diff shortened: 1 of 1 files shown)


toxyy:
!set Event

toxyy:
Not sure if

```
$result = $db->sql_query_limit($db->sql_build_query('SELECT', $forum_sql_ary), 1);
$active_f_row = $db->sql_fetchrow($result);
$db->sql_freeresult($result);

$result = $db->sql_query_limit($db->sql_build_query('SELECT', $topic_sql_ary), 1);
$active_t_row = $db->sql_fetchrow($result);
$db->sql_freeresult($result);
```

is the best way to reformat this, I just copied the existing code and placed it adjacent to one another.

3D-I:
It's ok the reformatting.
But the code placement is not IMO.

Those two "fetches" should go exactly each one underneath its related query and the 2 sub-queries should go back to where they came from.

At this point you can dispatch those two arrays as you did but not into a new event, simply adding those two to the `core.display_user_activity_modify_actives` below. Hence get rid of your event and add those two arrays to it modifying the existing event.

At a first glance.

phpBB/includes/functions_display.php
```diff
@@ -1196,17 +1196,54 @@ function display_user_activity(&$userdata_ary)
...
+		*
+		* @event core.display_user_activity_modify_sql
+		* @var	array	forum_sql_ary		SQL array to get the user's most active forum
+		* @var	array	topic_sql_ary		SQL array to get the user's most active topic
+		* @since 3.2.5-RC1
+		*/
+		$vars = array('forum_sql_ary', 'topic_sql_ary');
+		extract($phpbb_dispatcher->trigger_event('core.display_user_activity_modify_sql', compact($vars)));
+
+		$result = $db->sql_query_limit($db->sql_build_query('SELECT', $forum_sql_ary), 1);
```
3D-I: It's ok the reformatting.
But the code placement is not IMO.

Those two "fetches" should go exactly each one underneath its related query and the 2 sub-queries should go back to where they came from.

At this point you can dispatch those two arrays as you did but not into a new event, simply adding those two to the `core.display_user_activity_modify_actives` below. Hence get rid of your event and add those two arrays to it modifying the existing event.

At a first glance.

marc1706:
I've pondered a bit about whether this we should merge this. I understand that you wanted to prevent duplicating the two queries in your extension, however I think this might actually be the better solution than this event both from a maintenance viewpoint in phpBB and how much use this event would otherwise be.

Therefore I'm going to close this PR.

PR #5457, opened 2018-11-26
[ticket/15883] New error message for when users try adding invalid usernames to a group

battye:
Update the ACP and the UCP so that when bulk adding users to a group, if invalid usernames are submitted alongside valid usernames then a message will be displayed to inform the user what the invalid usernames are.

I'm using the old `$user->lang[]` format for now so I don't interfere with PR #5447.

PHPBB3-15883

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-15883

### DIFF
--- a/phpBB/includes/acp/acp_groups.php
+++ b/phpBB/includes/acp/acp_groups.php
@@ -29,6 +29,9 @@ class acp_groups
 		global $phpbb_root_path, $phpbb_admin_path, $phpEx;
 		global $request, $phpbb_container, $phpbb_dispatcher;
 
+		/** @var \phpbb\language\language $language Language object */
+		$language = $phpbb_container->get('language');
+
 		$user->add_lang('acp/groups');
 		$this->tpl_name = 'acp_groups';
 		$this->page_title = 'ACP_GROUPS_MANAGE';
@@ -293,7 +296,7 @@ class acp_groups
 				// Add user/s to group
 				if ($error = group_user_add($group_id, false, $name_ary, $group_name, $default, $leader, 0, $group_row))
 				{
-					$display_message = $user->lang[$error];
+					$display_message = $language->lang($error);
 
 					if ($error == 'GROUP_USERS_INVALID')
 					{
@@ -302,7 +305,7 @@ class acp_groups
 						$actual_user_id_ary = false;
 						user_get_id_name($actual_user_id_ary, $actual_name_ary, false, true);
 
-						$display_message = sprintf($user->lang['GROUP_USERS_INVALID'], implode($user->lang['COMMA_SEPARATOR'], array_diff($name_ary, $actual_name_ary)));
+						$display_message = $language->lang('GROUP_USERS_INVALID', implode($language->lang('COMMA_SEPARATOR'), array_udiff($name_ary, $actual_name_ary, 'strcasecmp')));
 					}
 
 					trigger_error($display_message . adm_back_link($this->u_action . '&amp;action=list&amp;g=' . $group_id), E_USER_WARNING);
--- a/phpBB/includes/acp/acp_groups.php
+++ b/phpBB/includes/acp/acp_groups.php
@@ -302,7 +302,7 @@ class acp_groups
 					{
 						// Find which users don't exist
 						$actual_name_ary = $name_ary;
-						$actual_user_id_ary = false;
+						$actual_user_id_ary = [];
 						user_get_id_name($actual_user_id_ary, $actual_name_ary, false, true);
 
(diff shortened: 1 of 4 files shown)


rubencm:
Use the new format better, my pr wont be accepted for now

battye:
Ah okay, I will change it then - thank you @rubencm

Of these two techniques:

1/ using the container like described [here](https://area51.phpbb.com/docs/dev/3.2.x/language/usage.html):


    /** @var \phpbb\language\language $language */
    $language = $phpbb_container->get('language');
    $language->lang('...');

2/ or, the user class like you used in your PR:

    $user->lang('...');

Which is the preferred way to access language variables moving forward?

rubencm:
The idea is to change all to the first one.

battye:
How does that look now @rubencm? :smiley: 

I also changed:
* Using `array_udiff` now so that the comparison of names to be displayed in the error message is case insensitive
* The return to previous page link wasn't taking users back to the previous page after any sort of add to usergroup error occurred (some pre-existing bug whereby it was directing them back to the manage group list, instead of the specific group they were just managing)

phpBB/includes/acp/acp_groups.php
```diff
@@ -293,7 +296,19 @@ function main($id, $mode)
...
 				if ($error = group_user_add($group_id, false, $name_ary, $group_name, $default, $leader, 0, $group_row))
 				{
-					trigger_error($user->lang[$error] . adm_back_link($this->u_action . '&amp;action=list&amp;g=' . $group_id), E_USER_WARNING);
+					$display_message = $language->lang($error);
+
+					if ($error == 'GROUP_USERS_INVALID')
+					{
+						// Find which users don't exist
+						$actual_name_ary = $name_ary;
+						$actual_user_id_ary = false;
```
marc1706: Why are you setting this to a boolean value? user_get_id_name expects an array.

phpBB/includes/functions_user.php
```diff
@@ -26,8 +26,9 @@
 * @param array &$user_id_ary The user ids to check or empty if usernames used
 * @param array &$username_ary The usernames to check or empty if user ids used
 * @param mixed $user_type Array of user types to check, false if not restricting by user type
+* @param bool $update_references If false, the supplied array is unset and appears unchanged from where it was called
 */
```
marc1706: Can you also add a `@return null` to the doc block while you're at it?

phpBB/includes/ucp/ucp_groups.php
```diff
@@ -1054,13 +1057,27 @@ function main($id, $mode)
...
 							if ($error = group_user_add($group_id, false, $name_ary, $group_name, $default, 0, 0, $group_row))
 							{
-								trigger_error($user->lang[$error] . $return_page);
+								$display_message = $language->lang($error);
+
+								if ($error == 'GROUP_USERS_INVALID')
+								{
+									// Find which users don't exist
+									$actual_name_ary = $name_ary;
+									$actual_user_id_ary = false;
```
marc1706: Same as above.

phpBB/includes/acp/acp_groups.php
battye: @marc1706 There's lots of places in the code where we already treat it as a boolean. Like when [`group_user_attributes()`](https://github.com/phpbb/phpbb/search?q=group_user_attributes&unscoped_q=group_user_attributes), [`group_user_del()`](https://github.com/phpbb/phpbb/search?q=group_user_del&unscoped_q=group_user_del) or [`group_user_add()`](https://github.com/phpbb/phpbb/search?q=group_user_add&unscoped_q=group_user_add) get called. 

Would you like me to change mine to `$actual_user_id_ary = array();` or change the `@param` type to mixed in the function seeing we already use booleans in quite a few areas?

phpBB/includes/functions_user.php
battye: Not a problem, will do.

battye:
The changes are applied now

battye:
@marc1706 Actually what if we did `@return mixed Returns false on success, error string on failure` in the doc block instead? 

Even though `user_get_id_name()` is called directly most of the time, there are a few times where the result is actually checked.

marc1706:
How about maybe `@return boolean|string Returns false on success, error string on failure`?

battye:
Done :thumbsup:

phpBB/includes/functions_user.php
battye: I noticed this yesterday, scroll down a bit further to 
https://github.com/phpbb/phpbb/pull/5457#issuecomment-451495389 😃

mrgoldy: Oh my bad 😵

PR #5459, opened 2018-11-29
[ticket/15886] Group helper functions

mrgoldy:
PHPBB3-15886

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15886


mrgoldy:
!set WIP :construction:

mrgoldy:
I am running `$ phpBB/vendor/bin/phpunit -d memory_limit=2048M` but that is not picking up these errors. Anyway to catch these errors locally aswell?
Sorry for the failed travis jobs.

mrgoldy:
!unset WIP :construction:

3D-I:
`!unset WIP :construction: `

3D-I:
One thing we have discussed, which still leaves me perplexed, is the "a priori" inclusion of events in this code. I think they should be submitted as separate PRs and linked to here, but I am not so sure. Let us ask @marc1706 about everything.

phpBB/phpbb/group/helper.php
```diff
@@ -15,17 +15,58 @@
...
+	protected $phpbb_root_path;
+
+	/** @var string PHP file extension */
+	protected $php_ext;
+
 	/**
 	 * Constructor
 	 *
-	 * @param \phpbb\language\language $language	Language object
+	 * @param \phpbb\auth\auth $auth Authentication object
```
marc1706: Can you please change these to not always use fully qualified class names? (see http://php.net/manual/en/language.namespaces.importing.php)

phpBB/phpbb/group/helper.php
```diff
@@ -15,17 +15,58 @@
...
+	 * @param \phpbb\cache\service $cache Cache object
```
marc1706: Are you sure it's the cache service and not the driver_interface?

phpBB/phpbb/group/helper.php
```diff
@@ -37,4 +78,198 @@ public function get_name($group_name)
 	{
 		return $this->language->is_set('G_' . utf8_strtoupper($group_name)) ? $this->language->lang('G_' . utf8_strtoupper($group_name)) : $group_name;
 	}
+
+	/**
+	 * Get group name details for placing into templates.
+	 *
+	 * @param string	$mode				Can be profile (for getting an url to the profile), group_name (for obtaining the group name), colour (for obtaining the group colour), full (for obtaining a html string representing a coloured link to the group's profile) or no_profile (the same as full but forcing no profile link)
```
marc1706: Please split this explanation onto multiple lines. Ideally your line should be no longer than 120 characters

phpBB/phpbb/group/helper.php
```diff
@@ -37,4 +78,198 @@ public function get_name($group_name)
...
+		 *
+		 * @event core.modify_group_name_string
+		 * @var string	mode				profile|group_name|colour|full|no_profile
+		 * @var int		group_id			The group identifier
+		 * @var string	group_name			The group name
+		 * @var string	group_colour		The group colour
+		 * @var string	custom_profile_url	Optional parameter to specify a profile url.
+		 * @var string	group_name_string	The string that has been generated
+		 * @var array	_profile_cache		Array of original return templates
+		 * @since 3.2.5-RC1
```
marc1706: Should be 3.2.6-RC1 now

phpBB/phpbb/group/helper.php
```diff
@@ -37,4 +78,198 @@ public function get_name($group_name)
...
+			'img'		=> null,
+			'img_src'	=> null,
+		);
+
+		/**
+		 * Preparing a group's rank before displaying
+		 *
+		 * @event core.get_group_rank_before
+		 * @var	array	group_data		Array with group's data
+		 * @since 3.2.5-RC1
```
marc1706:
```suggestion
		 * @since 3.2.6-RC1
```

phpBB/phpbb/group/helper.php
marc1706:
```suggestion
		 * @since 3.2.6-RC1
```

phpBB/phpbb/group/helper.php
```diff
@@ -37,4 +78,198 @@ public function get_name($group_name)
...
+	 * @param int		$group_id			The group id
+	 * @param string	$group_name			The group name
+	 * @param string	$group_colour		The group colour
+	 * @param mixed		$custom_profile_url	optional parameter to specify a profile url. The group id gets appended to this url as &amp;g={group_id}
+	 *
+	 * @return string A string consisting of what is wanted based on $mode.
+	 */
+	public function get_name_string($mode, $group_id, $group_name, $group_colour = '', $custom_profile_url = false)
+	{
+		$_profile_cache = array(
```
marc1706: I see this used to be a static variable. Instead of redeclaring this every every time you can just add it as a class member variable.

phpBB/phpbb/group/helper.php
marc1706: I checked and you're in fact using the cache service.

phpBB/phpbb/group/helper.php
mrgoldy: Adjusted

phpBB/phpbb/group/helper.php
mrgoldy: Done

phpBB/phpbb/group/helper.php
mrgoldy: Done

phpBB/phpbb/group/helper.php
mrgoldy: Done

phpBB/phpbb/group/helper.php
mrgoldy: Done
Just for the record, I tried using `use phpbb\cache` and then use `cache\service`, but Travis doesn't like that. See: https://travis-ci.org/phpbb/phpbb/jobs/473299146#L498https://travis-ci.org/phpbb/phpbb/jobs/473299146#L498
Otherwise I had to use `service` which is not very descriptive. So changed it to `use phpbb\cache\service as cache`.

phpBB/phpbb/group/helper.php
mrgoldy: Yeah I have to `obtain_ranks()` which is within the `cache\service`.

mrgoldy:
Okay, some questions still remain:
- I have not written a test for the `get_avatar()` function, as it is just a wrapper function. Is that okay or should I copy/create a test for that aswell?
- In the `group_helper.php` `get_rank()` I use _(just as in `phpbb_get_user_rank()`)_ the [`$path_helper->update_web_root_path()`](https://github.com/phpbb/phpbb/pull/5459/files#diff-cf13160eb2109d6e5c7abbfa09da68a8R236). I do not know how to create/mimic [that function in the tests](https://github.com/phpbb/phpbb/pull/5459/files#diff-b6079a3272d92a8ceae4646094de71b7R27).
- I would like to change the _wrong_ (`$user_row` instead of `$group_row`) variable names in [`phpbb_get_group_avatar()`](https://github.com/phpbb/phpbb/blob/master/phpBB/includes/functions.php#L4064) aswell. Is that allowed in this PR aswell, or should that be done in a separate PR?
- This PR in it self does not fix anything, but will be used for [Group rank not showing memberlist_body](https://tracker.phpbb.com/browse/PHPBB3-15885). Should I include that fix in this PR, or also in a separate PR?

marc1706:
- If it's just a wrapper function then I think not having a test for it is ok.
- You can mock the return of that method to make sure it's returning what you're expecting. See https://github.com/phpbb/phpbb/blob/master/tests/template/asset_test.php#L36
- You can change it in this PR, just make sure it's a separate commit
- Make it a separate PR and note that that PR depends on this one in the pull request message.

mrgoldy:
Okay, think I am done here 👍 
Created one additional check to make sure no profile urls are created for the bots group.
#5494 is currently failing travis tests as this PR first needs to be merged.

Should I create a PR where the helper functions are used in various places, such as: https://github.com/phpbb/phpbb/blob/master/phpBB/index.php#L109-L119 
Or should I leave the current code as is? I guess I mean: should I implement this everywhere or not?

phpBB/phpbb/group/helper.php
```diff
@@ -13,19 +13,73 @@
...
 		$this->language = $language;
+		$this->dispatcher = $dispatcher;
+		$this->path_helper = $path_helper;
+		$this->user = $user;
+
+		$this->phpbb_root_path = $path_helper->get_phpbb_root_path();
+
+		$this->name_strings = array(
+			'base_url'				=> append_sid("{$path_helper->get_phpbb_root_path()}memberlist.{$path_helper->get_php_ext()}", 'mode=group&amp;g={GROUP_ID}'),
+			'tpl_noprofile'			=> '<span class="username">{GROUP_NAME}</span>',
```
marc1706: @hanakin I guess this should ideally be refactored out?

phpBB/phpbb/group/helper.php
```diff
@@ -13,19 +13,73 @@
...
+		$this->config = $config;
 		$this->language = $language;
+		$this->dispatcher = $dispatcher;
+		$this->path_helper = $path_helper;
+		$this->user = $user;
+
+		$this->phpbb_root_path = $path_helper->get_phpbb_root_path();
+
+		$this->name_strings = array(
+			'base_url'				=> append_sid("{$path_helper->get_phpbb_root_path()}memberlist.{$path_helper->get_php_ext()}", 'mode=group&amp;g={GROUP_ID}'),
```
marc1706: Please don't call this in the constructor but actually use append_sid() when you need it. The container will automatically share services and therefore might end up sharing this URL with the SID across users.

mrgoldy: Changed it so assignment is still as a class member, but `append_sid()` is only called when required.

phpBB/phpbb/group/helper.php
hanakin: @marc1706 well ideally none of these would be there, but that means we would have to replace all the uses with the correct format in the templates which will be tedious and would have huge bc concerns most likely. Probably outside the scope of this pr and definitely 3.2.x. Also, I have to double check the CSS for username vs username-colored not sure what exactly the difference between those two are other than for style creators which will lead to customization issues? Might be easier to leave it and just document it clearly so we can easily find and change the html if need be. There are probably a few cases that it may make sense to keep it in the backend like this, but we need to have a clear documented list of the uses. For now, I guess maybe @mrgoldy can add an @html tag comment in the doc block that captures them.

phpBB/phpbb/group/helper.php
```diff
@@ -37,4 +91,198 @@ public function get_name($group_name)
...
+
+			if (isset($ranks['special'][$group_data['group_rank']]))
+			{
+				$rank = $ranks['special'][$group_data['group_rank']];
+
+				$group_rank_data['title'] = $rank['rank_title'];
+
+				$group_rank_data['img_src'] = (!empty($rank['rank_image'])) ? $this->path_helper->update_web_root_path($this->phpbb_root_path . $this->config['ranks_path'] . '/' . $rank['rank_image']) : '';
+
+				$group_rank_data['img'] = (!empty($rank['rank_image'])) ? '<img src="' . $group_rank_data['img_src'] . '" alt="' . $rank['rank_title'] . '" title="' . $rank['rank_title'] . '" />' : '';
```
hanakin: @marc1706 I would be more in favor of factoring this type out as it prevents us from using svgs

phpBB/phpbb/group/helper.php
mrgoldy: What exactly do you want me to put in the DocBlock? If you can let me know, I'll add it.
Do you also want me to add it to the `get_username_string` occurences of these strings ([here](https://github.com/phpbb/phpbb/blob/master/phpBB/includes/functions_content.php#L1485))?

hanakin: Just add @html and a desc of what html is return. For every function that returns html so we can easily find it to factor out later if need be.

phpBB/phpbb/group/helper.php
hanakin: This is html as well

tests/group/helper_get_name_string_test.php
```diff
@@ -0,0 +1,115 @@
...
+	public function test_get_name_string_colour($group_id, $group_name, $group_colour, $custom_profile_url, $expected)
+	{
+		$this->assertEquals($expected, $this->group_helper->get_name_string('colour', $group_id, $group_name, $group_colour, $custom_profile_url));
+	}
+
+	public function get_name_string_full_data()
+	{
+		global $phpbb_root_path, $phpEx;
+
+		return array(
```
hanakin: Html

tests/group/helper_get_name_string_test.php
```diff
@@ -0,0 +1,115 @@
...
+	 * @dataProvider get_name_string_full_data
+	 */
+	public function test_get_name_string_full($group_id, $group_name, $group_colour, $custom_profile_url, $expected)
+	{
+		$this->assertEquals($expected, $this->group_helper->get_name_string('full', $group_id, $group_name, $group_colour, $custom_profile_url));
+	}
+
+	public function get_name_string_no_profile_data()
+	{
+		return array(
```
hanakin: Html

tests/group/helper_get_rank_test.php
```diff
@@ -0,0 +1,43 @@
...
+{
+	public function get_rank_data()
+	{
+		global $phpbb_root_path;
+
+		return array(
+			array(
+				array('group_id' => 0, 'group_rank' => 1),
+				array(
+					'title' => 'Site admin',
```
hanakin: Html

phpBB/phpbb/group/helper.php
hanakin: Well possibly reword it to be “group name and username spans and links” also would consider adding to the doc block as well since the function needs replaced or rewritten essentially.

phpBB/phpbb/group/helper.php
mrgoldy: Done

phpBB/phpbb/group/helper.php
mrgoldy: Done

tests/group/helper_get_name_string_test.php
mrgoldy: These are tests.

tests/group/helper_get_name_string_test.php
mrgoldy: These are tests.

tests/group/helper_get_rank_test.php
mrgoldy: These are tests.

3D-I:
A bit confused now, where this PR should be merged?

hanakin:
3.3 according to Marc’s last change. 3.2.x branch will be branched into a new 3.3 branch when Marc gets a chance possibly after .6 release

3D-I:
This is the last of 5 PRS we made for an extension to be released with 3.2.6. 😕
4 of those have been already merged for 3.2.6.

hanakin:
@marc then 😜

3D-I:
Another PR stated for 3.2.6 depend on this one BTW ;)

https://github.com/phpbb/phpbb/pull/5494

phpBB/phpbb/group/helper.php
```diff
@@ -37,4 +92,203 @@ public function get_name($group_name)
...
 		return $this->language->is_set('G_' . utf8_strtoupper($group_name)) ? $this->language->lang('G_' . utf8_strtoupper($group_name)) : $group_name;
 	}
+
+	/**
+	 * Get group name details for placing into templates.
+	 *
+	 * @html Group name spans and links
+	 *
+	 * @param string	$mode				profile (for getting an url to the profile),
+	 *                                        group_name (for obtaining the group name),
```
marc1706: Alignment indentation should be done with tabs and not spaces.

phpBB/phpbb/group/helper.php
```diff
@@ -37,4 +92,203 @@ public function get_name($group_name)
...
+		 *
+		 * @event core.modify_group_name_string
+		 * @var string	mode				profile|group_name|colour|full|no_profile
+		 * @var int		group_id			The group identifier
+		 * @var string	group_name			The group name
+		 * @var string	group_colour		The group colour
+		 * @var string	custom_profile_url	Optional parameter to specify a profile url.
+		 * @var string	group_name_string	The string that has been generated
+		 * @var array	name_strings		Array of original return templates
+		 * @since 3.2.6-RC1
```
marc1706: Needs to be updated to 3.2.8-RC1

phpBB/phpbb/group/helper.php
```diff
@@ -37,4 +92,203 @@ public function get_name($group_name)
...
+
+		return $group_name_string;
+	}
+
+	/**
+	 * Get group rank title and image
+	 *
+	 * @html Group rank image element
+	 *
+	 * @param array		$group_data		the current stored group data
```
marc1706:
```suggestion
	 * @param array		$group_data		The current stored group data
```

PR #5477, opened 2018-12-04
[ticket/15900] Converting install/index.html to UTF-8 and fixing tag

orynider:
PHPBB3-15900

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15900

### DIFF
--- a/phpBB/install/index.html
+++ b/phpBB/install/index.html
@@ -1,7 +1,7 @@
 <html>
 <head>
 <title></title>
-<meta http-equiv="Content-Type" content="text/html; charset=iso-8859-1">
+<meta http-equiv="Content-Type" content="text/html" charset="UTF-8" />
 <meta http-equiv="refresh" content="0; url=./app.php" />
 </head>
 


mrgoldy:
You still have to end your commit messages with an empty line and `PHPBB3-15900`.
```
[ticket/15900] Converting install/index.html to UTF8 and fixing tag

PHPBB3-15900
```

3D-I:
That file, however, is only there to make sure that the contents of the directory is not seen by pointing it directly. There's no reason to change it, it's just like dozens of others in the package. Practically this PR doesn't make sense as the file only displays a blank page with no text in it. Like the .htaccess colleague.

Examples:

https://github.com/phpbb/phpbb/blob/master/phpBB/files/index.htm
https://github.com/phpbb/phpbb/blob/master/phpBB/download/index.htm
https://github.com/phpbb/phpbb/blob/master/phpBB/cache/index.htm

etc.. etc..

marc1706:
Can you elaborate a little bit on what you're trying to achieve / change with this PR before we talk about specifics like commit message format? The file you're changing is HTML 4.01, you seem to be trying to change it to HTML5 syntax. What I don't understand is why exactly since this file is only used as a placeholder as @3D-I already mentioned.

orynider:
It's about the tag above that line that probably is HTML 4, and there are files like that with open tags at end that we never use but we are redirected to this one and from this one to app.php every time we do a fresh install.
 <meta http-equiv="Content-Type" content="text/html; charset=iso-8859-1">
. 
 <meta http-equiv="Content-Type" content="text/html" charset="utf-8" />
Replacing this also require a conversion in UTF-8 in Notepad++, or if is just converted to HTML6 standards we can just put the comas correct:
 <meta http-equiv="Content-Type" content="text/html" charset="iso-8859-1" />

orynider:
It's about the tag above that line that probably is HTML 4, and there are files like that with open tags at end that we never use but we are redirected to this one and from this one to app.php every time we do a fresh install.

.

Replacing this also require a conversion in UTF-8 in Notepad++, or if is just converted to HTML6 standards we can just put the comas correct:
<meta http-equiv="Content-Type" content="text/html" charset="iso-8859-1" />
or
<meta http-equiv="Content-Type" content="text/html" "charset=utf-8" />
instead of this line that breaks syntax highlighting in Notepad++:
<meta http-equiv="Content-Type" content="text/html; charset=iso-8859-1">

PR #5487, opened 2018-12-24
[ticket/15918] Fix addslashes in mcp_ban.html

3D-I:
Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15918


3D-I:
Lang keys used in javascript contexts and prefixed with `A_` are automatically escaped by the template engine. This fixes the issue as per the tracker and the regression issued by https://github.com/phpbb/phpbb/commit/4a3c00d1c84284c030e7a7a37efc3ada57228726#diff-b21a294b8b5400fdea6e473cd77f03c6 IMHO.

kasimi:
The current solution removes the fix for line breaks that cause a JS error that was added in #5170.

Please apply the Twig escape filter to both `bans.REASON` and `bans.GIVE_REASON` respectively (not the addslashed `A_` contents). This should fix the backslash being double-escaped while not breaking reasons that contain line breaks.

3D-I:
Do you mean `{{ bans.GIVE_REASON | e('js') }}` for example?

kasimi:
Yes. :+1:

3D-I:
Gotcha, thanks @kasimi :+1:

kasimi:
In general I don't see why not, but better not do it in this PR.

hanakin:
We actually would like to convert everything to twig for 3.3 so anything you want to push a pr for go for it ;) also consider if you are making any js fixes for 3.3 to factor out the js completely from the templates and into the js files. For template cars assign as data attributes to the respective elements if you can. Both will go towards making it easier for implementation in the new theme

PR #5488, opened 2018-12-26
[ticket/15921] Update textformattter to 1.3.2

marc1706:
Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15921


phpBB/phpbb/textformatter/s9e/factory.php
```diff
@@ -354,6 +354,7 @@ function ($m)
 
 		// Load the Emoji plugin and modify its tag's template to obey viewsmilies
 		$tag = $configurator->Emoji->getTag();
+		$tag->template = '<img alt="{.}" class="emoji" draggable="false" src="//twemoji.maxcdn.com/2/svg/{@tseq}.svg"/>';
```
JoshyPHP: Old content won't have a value for `@tseq` (which was added in 1.3.0) and the URL won't be valid unless the old content is reparsed. You can test for the presence of the attribute inside of the template and use a different URL or markup accordingly.
```suggestion
		$tag->template = '<xsl:choose>
			<xsl:when test="@tseq">
				<img alt="{.}" class="emoji" draggable="false" src="//twemoji.maxcdn.com/2/svg/{@tseq}.svg"/>
			</xsl:when>
			<xsl:otherwise>
				' . $tag->template . '
			</xsl:otherwise>
		</xsl:choose>';
```

The downside of this method is that you end up with two different sets of emoji. Alternatively, if there's no `@tseq` you may optimistically use `@seq` with the Twemoji URL and rely on an `onerror` handler on the image (client-side, then) to fall back to an EmojiOne URL. `@seq` and `@tseq` are only different if the emoji contains a ZWJ or VS16 in the middle of a sequence.

Lastly, I have a Twemoji fork that uses the same naming scheme as EmojiOne, which is accessible via jsDelivr.
```suggestion
		$tag->template = '<xsl:choose>
			<xsl:when test="@tseq">
				<img alt="{.}" class="emoji" draggable="false" src="//twemoji.maxcdn.com/2/svg/{@tseq}.svg"/>
			</xsl:when>
			<xsl:otherwise>
				<img alt="{.}" class="emoji" draggable="false" src="https://cdn.jsdelivr.net/gh/s9e/emoji-assets-twemoji@11.2/dist/svgz/{@seq}.svgz"/>
			</xsl:otherwise>
		</xsl:choose>';
```

marc1706: I think I'd actually prefer the last one as this seems to come without any real side effects.

PR #4765, opened 2017-03-26
[ticket/15148] Covert contact menu to svg (Retina) 

hanakin:
Checklist:

- [x] Correct branch: master for new features; 3.2.x, 3.1.x for fixes
- [x] Tests pass
- [x] Code follows coding guidelines: [master]
- [x] Commit follows commit message [format](https://wiki.phpbb.com/Git#Commit_Messages)

Ticket/15148

https://tracker.phpbb.com/browse/PHPBB3-15148

Examples

screenshot 2017-03-25 21 09 22
screenshot 2017-03-25 21 09 02

This is a test case for the full icon conversion to svg. all icons will be broken down into sets(html or twig files) of relatable icons and each file file be include at the top of the page. You can see the sets I have in mind here: http://codepen.io/hanakin/pen/XMBaaV?editors=1100

> **NOTE** that the actual icons themselves may be different but the groupings should be the same

PR #5489, opened 2018-12-27
[ticket/15538] Icon usage twig function

mrgoldy:
PHPBB3-15538

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15538

---
This is for the Chameleon Theme, to help out @hanakin.
No PNG icons will be shipped initially with the theme, but SVG's will be.
This is not to say that other style authors do not want to use PNG's, so there is support for png aswell.
The assumption is that icons will be located in
- `theme/icons/png/` `some_icon.png`
- `theme/icons/svg/` `some_icon.svg`

In regards to loading times, I do not know what is faster:
- hard coding html strings for the various icons and caching those
- rendering the `icon.html` through the environment

This also goes for the svg's `<path d="">`. Cause the svg's will mostlikely be _'stand-alone'_ versions of the icon (`<svg>..</svg>`) and we only want to use the the `d=""` part from the `<path>`. Anyways, I am not sure what the fastest way is to get the contents of the svg file:
- Through the environment `render()`
- Through the filesystem?
- ... ?

Moreover the svg `<path d="">`, I couldn't quite find it online yet but what are the exact allowed characters in that specific attribute? I am currently using `'/d="([a-zA-Z0-9., ]+)"/'` as a regex. Anything that needs to be included/excluded?

The current `icon.html` file needs work, was just a quick mock up to get something to be displayed.


mrgoldy:
!set WIP :construction:

phpBB/styles/prosilver/template/icon.html
```diff
@@ -0,0 +1,10 @@
+{% if TYPE == 'png' %}
+<img class="icon png-{{ ICON ~ S_CLASSES }}" src="{{ T_PATH }}"{% if TITLE %} alt="{{ TITLE }}"{% endif %} style="background: black;">
```
hanakin: whats with the hardcoded style?

phpBB/styles/prosilver/template/icon.html
```diff
@@ -0,0 +1,10 @@
...
+{% elseif TYPE == 'font' %}
+<i class="icon fa-{{ ICON ~ S_CLASSES }}" aria-hidden="true"{{ S_ATTRIBUTES }}></i>{% if TITLE %}<span{% if S_HIDDEN %} class="sr-only"{% endif %}>{{ TITLE }}</span>{% endif %}
```
hanakin: aria hidden should only be there if title else its should be false

phpBB/styles/prosilver/template/icon.html
mrgoldy: Oh yeah my bad, I downloaded the white png 😵 couldn't see it, like I said, the entire html file needs work as it's just a mock. Decided that that will be for you.
First want to make sure the PHP and loading times are correct.

phpBB/styles/prosilver/template/icon.html
```diff
@@ -0,0 +1,10 @@
...
+{% elseif TYPE == 'svg' %}
+<svg class="icon svg-{{ ICON ~ S_CLASSES }}"{% if TITLE %} aria-hidden="true" aria-labeled-by="{{ TITLE_CLEAN }}"{% endif %} xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1"  width="24" height="24" viewBox="0 0 24 24">
```
hanakin: we do not need all of the extra stuff here you can remove
```
xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1"  width="24" height="24"
```

hanakin:
@CHItA @Nicofuma you insights for the best way to implement this would probably help as well. IIRC one of you had an idea for handling this

hanakin:
have a look at some of the code from this old pr https://github.com/phpbb/phpbb/pull/4765/files might help with testing and what not

hanakin:
@mrgoldy where did we leave of with this? should probably create some tests. We will need to test it in all locations/possible uses that is to say I want to use it from icons, imagesets, contact icons, forum icons, smiles

the last one is a maybe

but this means we need to be able to specify in the case of forum icons a different path as they are in the files folder same with the smiles.

PR #5493, opened 2018-12-31
[ticket/15924] Switch to trusty build environment

marc1706:
PHPBB3-15924

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15924


marc1706:
The builds will still be executed based on the `.travis.yml` configuration that the extensions are using. However, extensions depending on the `setup-mariadb.sh` script will probably start failing if they don't update to trusty as well.
I think it makes sense to create a sticky in Extensions Writers Discussion mentioning this. Do you want to create that or should I?

VSEphpbb:
You could since you know what needs changing

marc1706:
@VSEphpbb so, with the latest change this should no longer be necessary and the builds should continue work as they did before even when an extension is still using the precise builds. :champagne:

PR #5498, opened 2019-01-05
[ticket/15926] Deny 3.2.x installation if PHP >= 7.3-dev

3D-I:
PHPBB3-15926

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15926


phpBB/install/app.php
```diff
@@ -20,9 +20,9 @@
 $phpbb_root_path = '../';
 $phpEx = substr(strrchr(__FILE__, '.'), 1);
 
-if (version_compare(PHP_VERSION, '5.4') < 0)
+if ( version_compare(PHP_VERSION, '5.4.7', '<') || version_compare(PHP_VERSION, '7.3-dev', '>=') )
```
marc1706: Should be this according to coding guidelines:
```suggestion
if (version_compare(PHP_VERSION, '5.4.7', '<') || version_compare(PHP_VERSION, '7.3-dev', '>='))
```

3D-I: Ok, though I think that the code's readability gains here.

phpBB/install/app.php
```diff
@@ -20,9 +20,9 @@
 $phpbb_root_path = '../';
 $phpEx = substr(strrchr(__FILE__, '.'), 1);
 
-if (version_compare(PHP_VERSION, '5.4') < 0)
+if (version_compare(PHP_VERSION, '5.4.7', '<') || version_compare(PHP_VERSION, '7.3-dev', '>='))
 {
-	die('You are running an unsupported PHP version. Please upgrade to PHP 5.4.0 or higher before trying to install or update to phpBB 3.2');
+	die('You are running an unsupported PHP version. Please upgrade to PHP equal or greater than 5.4.7 but less than 7.3-dev in order to install or update to phpBB 3.2');
```
VSEphpbb: should be more like: 
equal to or greater than 5.4.7

Solidjeuh:
Why hardcoded language? Should be added in the language files...

DavidIQ:
Because loading the language would mean loading phpBB, which is what this change attempts to do for PHP 7.3+

VSEphpbb:
Because it has to be installed to load the language sustem, but this error comes before it is installed.

PR #5504, opened 2019-01-10
[ticket/13889] Refactor viewtopic code into a controller

CHItA:
PHPBB3-13889

Checklist:

- [ ] Tests pass


CHItA:
This basically a non-functioning refactoring attempt of `viewtopic.php`. I only posted this PR as i think it sort of demonstrates that the current data flow is pretty much unmaintainable (see references in method arguments). Also the data retrieving, rendering, various moderation functionality, poll management and session handling (in which forum the user is) follow each other in a seemingly random order.

This leads to a problem, where basically whatever sane way one would prefer to take with this will break events and thus extension BC, which I would assume would be pretty painful for extension authors. However, it also seems clear to me that the current data flow has to go, so if we are breaking BC, at least we should come up with some design that will be flexible enough for all our needs, and possibly work for a long time.

So if anyone has any suggestions, they are most definitely welcome.

PR #5505, opened 2019-01-12
[ticket/15925] Add core events for sync and mcp functions

toxyy:
Add core events to modify topic and forum info when syncing
Also add core events to account for missing variables when using mcp functions
Custom post, topic, or forum columns use these events to account for their data

PHPBB3-15925

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15925

### DIFF
--- a/phpBB/includes/functions_admin.php
+++ b/phpBB/includes/functions_admin.php
@@ -2118,6 +2118,22 @@ function sync($mode, $where_type = '', $where_ids = '', $resync_parents = false,
 					$topic_data[$topic_id]['last_poster_name'] = ($row['poster_id'] == ANONYMOUS) ? $row['post_username'] : $row['username'];
 					$topic_data[$topic_id]['last_poster_colour'] = $row['user_colour'];
 				}
+
+				/**
+				* Event to modify the topic_data when syncing topics
+				*
+				* @event core.sync_modify_topic_data
+				* @var	array	topic_data		Array with the topics' data we are syncing
+				* @var	array	row				Array with some of the current user and post data
+				* @var	int		topic_id		The current topic_id of $row
+				* @since 3.2.6-RC1
+				*/
+				$vars = array(
+					'topic_data',
+					'row',
+					'topic_id',
+				);
+				extract($phpbb_dispatcher->trigger_event('core.sync_modify_topic_data', compact($vars)));
 			}
 			$db->sql_freeresult($result);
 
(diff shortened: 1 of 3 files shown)


toxyy:
!set Event

toxyy:
First build error fixed in #5506 

Second one is facebook curl :thumbsdown: fixed by restarting too many times, thanks Ruben

toxyy:
Alternatively, with the last two sync_before event commits, the sync functions could be moved after `core.mcp_topics_merge_posts_after` and `core.mcp_forum_merge_topics_after` for the same affect.

phpBB/includes/functions_admin.php
```diff
@@ -1872,7 +1887,6 @@ function sync($mode, $where_type = '', $where_ids = '', $resync_parents = false,
 						}
 					}
 				}
-				unset($post_info);
```
marc1706: This shouldn't be removed. The current placement ensures that the post info array is both created and deleted inside step 5 and therefore stays encapsulated. If `$post_ids` is empty the compact in your event will throw a PHP notice with `Notice: Undefined variable: post_info`

phpBB/includes/functions_posting.php
```diff
@@ -258,14 +258,35 @@ function update_post_information($type, $ids, $return_update_sql = false)
...
+		* @since 3.2.6-RC1
+		*/
+		$vars = array(
+			'type',
+			'sql_ary',
+		);
+		extract($phpbb_dispatcher->trigger_event('core.update_post_info_modify_posts_sql', compact($vars)));
+		$result = $db->sql_query($db->sql_build_query('SELECT', $sql_ary));
+
+		$rowset = array();
```
marc1706: Should be unset after the event.

phpBB/includes/functions_admin.php
toxyy: @marc1706 I'm unsure how to resolve this one.  [In the event within my extension here,](https://github.com/toxyy/anonymousposts/blob/master/event/listener.php#L547) I'm appending data to` forum_data` from `post_info` which requires a new fieldname as well.  I could separate this into two events, one to append to `forum_data` and one to append to the fieldnames, but since they are both related, it seemed better to do it in one event as it's meaningless to add a key onto fieldnames without corresponding data from `forum_data` or vice versa.  Would it be better to make a separate variable for this, or add a line
```php
if( !isset($post_info)
{
    $post_info = array()
}
```
or `$post_info = isset($post_info) ?: array();`

before the `sync_modify_forum_data` event.

Which would be the better solution?

phpBB/includes/functions_posting.php
toxyy: I believe this has been resolved? Thanks for that.

marc1706 (approved):

marc1706:
I was planning to already include this in 3.3.4-RC1, sorry for letting this slip for that long.

toxyy:
No problem, glad to finally have it in.  Thanks Mark!

PR #5506, opened 2019-01-12
[ticket/15935] Don't install APCu if it's already installed

toxyy:
Travis fails when the APCu install fails due to it already being installed
A check is added here to make sure that it does nothing in that case

PHPBB3-15935

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15935

### DIFF
--- a/travis/setup-php-extensions.sh
+++ b/travis/setup-php-extensions.sh
@@ -49,10 +49,13 @@ echo 'opcache.enable=0' >> "$php_ini_file"
 # APCu
 if [ `php -r "echo (int) (version_compare(PHP_VERSION, '7.0.0-dev', '>=') && version_compare(PHP_VERSION, '7.3.0-dev', '<'));"` == "1" ]
 then
-	echo 'Enabling APCu PHP extension'
-	printf "\n" | pecl install apcu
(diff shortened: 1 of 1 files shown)


travis/setup-php-extensions.sh
```diff
@@ -49,10 +49,13 @@ echo 'opcache.enable=0' >> "$php_ini_file"
 # APCu
 if [ `php -r "echo (int) (version_compare(PHP_VERSION, '7.0.0-dev', '>=') && version_compare(PHP_VERSION, '7.3.0-dev', '<'));"` == "1" ]
```
rubencm: @marc1706 Why between this two versions? Can this condition be replaced by the new one?

toxyy: That's what was in place, which version did you have in mind again?

3D-I: Perhaps starting from 7.1, not sure though.

marc1706: The APCu extension was not available via pecl in 7.3 yet. Not sure if this has changed yet.

rubencm: Then the check should be check if is available or something and is not installed

PR #5515, opened 2019-01-17
[ticket/15943] Update core.posting_modify_default_variables

toxyy:
Update event to take advantage of get_submitted_attachment_data's parameter
This is useful if we want to not check the attachment poster's id
If the new variable is at all set by this event, then it is used instead

PHPBB3-15943

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15943

### DIFF
--- a/phpBB/posting.php
+++ b/phpBB/posting.php
@@ -596,18 +596,22 @@ if (isset($post_data['post_text']))
 
 // Set some default variables
 $uninit = array('post_attachment' => 0, 'poster_id' => $user->data['user_id'], 'enable_magic_url' => 0, 'topic_status' => 0, 'topic_type' => POST_NORMAL, 'post_subject' => '', 'topic_title' => '', 'post_time' => 0, 'post_edit_reason' => '', 'notify_set' => 0);
+$attachment_poster_id = '';
 
 /**
 * This event allows you to modify the default variables for post_data, and unset them in post_data if needed
 *
 * @event core.posting_modify_default_variables
-* @var	array	post_data	Array with post data
-* @var	array	uninit		Array with default vars to put into post_data, if they aren't there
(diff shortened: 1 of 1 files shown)


toxyy:
!set Event

toxyy:
Might not be necessary, will close for now.

phpBB/posting.php
```diff
@@ -620,9 +624,10 @@
 }
 unset($uninit);
 
+($attachment_poster_id !== '') ?: $attachment_poster_id = $post_data['poster_id'];
```
kasimi: Maybe you could initialize `$attachment_poster_id = $post_data['poster_id'];` and remove the above line? It would make the code a bit easier to follow and event handlers would be able to use the `attachment_poster_id` value for read-only purposes, too.

toxyy: I remember why I changed it to that now.  At first I did do what you suggested, though I realized that `$message_parser->get_submitted_attachment_data()` is calling the poster_id made after [the foreach on this line](https://github.com/phpbb/phpbb/blob/179fc6e7bc35c77c0b00ead381e0834e302316eb/phpBB/posting.php#L618).  This means that with your suggestion, if the event is unused, it will be checking it against the $post_data['poster_id'] variable before it is possibly changed.  This may lead to unintended behavior as it was originally coded to use the poster_id after all of that business is done.

I tried a few ways of going about this, but I settled on checking if the new attachment_poster_id has been used at all; if it has then we use that variable, if it hasn't then we use the $post_data['poster_id'] value that has potentially been updated after that foreach I linked.

marc1706:
I think the current design is just confusing. I think it would be better to handle this in the get_submitted_attachment_data() method.

marc1706:
Closing this as previously discussed.

PR #5523, opened 2019-01-24
[ticket/15951] Add core events to mcp

toxyy:
Allow modification of permissions for normal users to access the mcp
Also allows modification of the hidden form fields when deleting topics
This lets normal users delete their own topics with replies

PHPBB3-15951

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15951

### DIFF
--- a/phpBB/includes/mcp/mcp_main.php
+++ b/phpBB/includes/mcp/mcp_main.php
@@ -895,7 +895,7 @@ function mcp_delete_topic($topic_ids, $is_soft = false, $soft_delete_reason = ''
 	* @var	string	soft_delete_reason		The reason we're soft deleting
 	* @var	string	action					The current delete action
 	* @var	array	check_permission		The array with a permission to check for, can be set to false to not check them
-	* @since 3.3.2-RC1
+	* @since 3.3.3-RC1
 	*/
 	$vars = array(
 		'topic_ids',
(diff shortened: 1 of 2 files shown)


phpBB/includes/mcp/mcp_main.php
```diff
@@ -821,16 +821,38 @@ function mcp_restore_topic($topic_ids)
...
+		'topic_ids',
+		'forum_id',
+		'is_soft',
+		'soft_delete_reason',
+		'action',
+		'check_permission',
+	);
+	extract($phpbb_dispatcher->trigger_event('core.mcp_delete_topic_modify_permissions', compact($vars)));
+
+	if (!phpbb_check_ids($topic_ids, TOPICS_TABLE, 'topic_id', $check_permission))
```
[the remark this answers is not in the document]
toxyy: The reason that I changed that up, is because it was calling `phpbb_check_ids` like this:

`phpbb_check_ids($topic_ids, TOPICS_TABLE, 'topic_id', array($check_permission))`

In `phpbb_check_ids`, I need to set the 4th parameter, `$acl_list`, to false.  `array(false)` is NOT false and it does not behave the same way.  So, I had to change `$check_permission` from a string to an array with one element, and now within the event you are able to set it to false and it will register in `phpbb_check_ids`.

toxyy:
!set Event

toxyy:
Error said there was whitespace at the end of line 118, there was none to be found... closing and reopening.

phpBB/includes/mcp/mcp_main.php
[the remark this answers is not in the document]
marc1706: I think this is fine as it's still passed as an array.

phpBB/mcp.php
```diff
@@ -116,15 +116,35 @@
 	);
 
 	$allow_user = false;
+	$topic_info = phpbb_get_topic_data(array($topic_id));
```
marc1706: You're changing this from being only executed if needed in an edge case to always being executed. This will probably increase the overall server load and shouldn't be done. Instead, you should probably also pass `$quickmod` to the event and then execute this if needed

toxyy: Removed these changes and added `$quickmod` to the event instead.

toxyy:
Since version updated with the fixes.

toxyy:
Unsure why travis failed...

toxyy:
!set '3.3 (Proteus)

toxyy:
!set 3.3 (Proteus)

phpBB/includes/mcp/mcp_main.php
```diff
@@ -1002,6 +1024,28 @@ function mcp_delete_topic($topic_ids, $is_soft = false, $soft_delete_reason = ''
 			$s_hidden_fields['delete_permanent'] = '1';
 		}
 
+		/**
+		* This event allows you to modify the hidden form fields when deleting topics
+		*
+		* @event core.mcp_delete_topic_modify_hidden_fields
+		* @var	string	l_confirm				The mode we are deleting in (DELETE_TOPIC(S), DELETE_TOPIC(S)_PERMANENTLY)
```
marc1706, line 1031: This is actually the confirmation text language variable

toxyy: I added the DELETE_TOPIC(S) etc for usage context if looking up the event. Should I change the whole thing or just the text before it to reflect that it's the confirmation text language variable?

marc1706: Changing just the text before the possible values is fine

marc1706 (approved):

PR #5525, opened 2019-01-26
[ticket/15954] Add safeguards to include() calls

kasimi:
All functions and classes that are being checked for are actually the ones being used, except for the check in ucp_groups.php: https://github.com/kasimi/phpbb/commit/75007697ae55632200eb7f2c6579026e23fe3d7d#diff-ec446c980dec86d6991a822a459adfe6R404 I couldn't find out which of the functions in `includes/functions_display.php` is needed so I picked a random one for the `function_exists()` check.

PHPBB3-15954

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15954

### DIFF
--- a/phpBB/includes/mcp/mcp_main.php
+++ b/phpBB/includes/mcp/mcp_main.php
@@ -176,7 +176,7 @@ class mcp_main
 			case 'front':
 				if (!function_exists('mcp_front_view'))
 				{
-					mcp_front_view($id, $mode, $action);
+					include($phpbb_root_path . 'includes/mcp/mcp_front.' . $phpEx);
 				}
 
 				$user->add_lang('acp/common');
--- a/phpBB/includes/mcp/mcp_main.php
+++ b/phpBB/includes/mcp/mcp_main.php
@@ -174,7 +174,10 @@ class mcp_main
 		switch ($mode)
 		{
 			case 'front':
-				include($phpbb_root_path . 'includes/mcp/mcp_front.' . $phpEx);
+				if (!function_exists('mcp_front_view'))
+				{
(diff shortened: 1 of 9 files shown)


phpBB/includes/mcp/mcp_main.php
```diff
@@ -174,7 +174,10 @@ function main($id, $mode)
 		switch ($mode)
 		{
 			case 'front':
-				include($phpbb_root_path . 'includes/mcp/mcp_front.' . $phpEx);
+				if (!function_exists('mcp_front_view'))
+				{
+					mcp_front_view($id, $mode, $action);
```
battye: Did you mean for this line to be `include($phpbb_root_path . 'includes/mcp/mcp_front.' . $phpEx);` ❓

kasimi: Indeed, thank you!

rubencm:
Why don't use include_once or even better, require_once?

kasimi:
Sounds good to me. Any other opinions before I go ahead and change to `require_once()`

rubencm:
Get rid of $phpEx

DavidIQ:
That might require some commenting from @marc1706 and/or @paul999 as there is a reason that exists.

paul999:
The phpEx comes from back in phpBB2 when it was really common to use different extensions for different php versions. In the current times it is not that usefull anymore.

However, in my personal opinion, especially in the old code base like this part of phpBB, it is really common to use both include with a function check and $phpEx, and as such I think that should be the way to go.

3D-I:
I agree with @paul999 FWIW

PR #5527, opened 2019-02-01
[ticket/15959] Fix invalid CNAME record

ajquick:
PHPBB3-15959

https://tracker.phpbb.com/browse/PHPBB3-15959


ajquick:
Of course also not passing due to timeout issues related to another issue?

rubencm:
@ajquick Thank you very much for the fix

But please, can you change it to any subdomain we own with CNAME record like area51.phpbb.com?

ajquick:
> But please, can you change it to any subdomain we own with CNAME record like area51.phpbb.com?

Yes, that makes more sense!

marc1706:
Does this only affect master or also 3.2.x?

ajquick:
It affects all branches for development. It is likely that no Travis builds will pass from now on without this change on 3.1.x, 3.2.x, master and 3.3.x.

PR #5534, opened 2019-02-13
[ticket/15917] Unapproved topics message for forums with no other posts

battye:
Fix for a minor bug whereby if a forum had no posts and someone added a topic which required approval, for privileged users it would say "No posts" in the forumlist while in the topic/post count it would have a number that indicated there were posts.

This change makes it so in this rare case (which has actually happened at .com a few times recently), instead of "No posts" it will either say "Topic awaiting approval" or "Topics awaiting approval".

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15917

### DIFF
--- a/phpBB/styles/prosilver/template/forumlist_body.html
+++ b/phpBB/styles/prosilver/template/forumlist_body.html
@@ -104,9 +104,9 @@
 								<br />{forumrow.LAST_POST_TIME}
 							<!-- ELSE -->
 							{% if forumrow.U_UNAPPROVED_TOPICS %}
-								{{ lang('TOPIC_UNAPPROVED_FORUM', forumrow.TOPICS) }}<br />
+								{{ lang('TOPIC_UNAPPROVED_FORUM', forumrow.TOPICS) }}
 							{% else %}
-								{{ lang('NO_POSTS') }}<br />
+								{{ lang('NO_POSTS') }}
 							{% endif %}
 							<!-- ENDIF -->
 						</span>
--- a/phpBB/styles/prosilver/template/forumlist_body.html
+++ b/phpBB/styles/prosilver/template/forumlist_body.html
@@ -103,7 +103,11 @@
 								<!-- ENDIF -->
 								<br />{forumrow.LAST_POST_TIME}
 							<!-- ELSE -->
-								{L_NO_POSTS}<br />&nbsp;
+							{% if forumrow.U_UNAPPROVED_TOPICS %}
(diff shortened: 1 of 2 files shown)


3D-I:
immagine

and

immagine

phpBB/styles/prosilver/template/forumlist_body.html
```diff
@@ -103,7 +103,11 @@
 								<!-- ENDIF -->
 								<br />{forumrow.LAST_POST_TIME}
 							<!-- ELSE -->
-								{L_NO_POSTS}<br />&nbsp;
+							{% if forumrow.U_UNAPPROVED_TOPICS %}
+								{{ lang('TOPIC_UNAPPROVED_FORUM', forumrow.TOPICS) }}<br />
```
3D-I: `<a href="{{ forumrow.U_UNAPPROVED_TOPICS }}">{{ lang('TOPIC_UNAPPROVED_FORUM', forumrow.TOPICS) }}</a><br />`

In case you want to directly link the authed user to the MCP without entering the forum.

hanakin: acctually we do not allow br tags anyway these will need to be factored out.

hanakin: This also makes changes to the templates which is a breaking change and should instead be pushed to 3.3 or 4.0

DavidIQ: This isn't a breaking change. If a style author does not apply this change it will not break their style. We routinely make style changes in 3.2 so this isn't any different.

DavidIQ: And unless we're going to remove all of the `<br />` tags in this file I'd just leave it there for this line. If there is a desire to change or remove these tags it should be done in a separate pr.

hanakin: Generally we like to prevent them if possilbe except for additions if absolutly required. Which I guess is what this is just did not look close enough ;)

battye:
Are you able to resolve this request @hanakin as per the conversation above? Thanks :)

hanakin:
@battye remove the <br>s they are useless and we need to stop using them as they are not styleable

battye:
@hanakin 

Done (https://github.com/phpbb/phpbb/pull/5534/commits/d3564a4a07299d2c9e6e77fa451b22cf62d077e2)

ch13

PR #5545, opened 2019-03-08
[ticket/15538] Twig icon function

mrgoldy:
PHPBB3-15538

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15538

---

### For usage in the **Chameleon** style.
Reference PR's: #5489 & #4765 

The only thing to be added from this PR is the twig extension (declaration and class), the rest is for testing purposes and the icon templates should go in the Chameleon style.


mrgoldy:
Couple of things to discuss:

Currently using the `DOMDocument` approach to manipulate the SVG file, as this is often considered a better approach than using `preg_replace()`. However, `DOMDocument` does not like HTML5 and SVG's, as `loadHTML()` will throw: 
```
An exception has been thrown during the rendering of a template ("Warning: DOMDocument::loadHTML(): Tag svg invalid in Entity, line: 1")
```
Therefore suppression is needed, currently simply done by `@`, but not sure what the prefered method is? `@` or `libxml_use_internal_errors(true)` and `libxml_clear_errors()`. 

Moreover, eventhough I am currently using the `DOMDocument` approach, I would like to point out that the `preg_replace` approach is a lot faster. I did some timing tests, with a 1000 iterations _(time in seconds)_:
- `DOMDocument`: 
  - Total time: 0.12599396705627
  - Average time: 0.00012599396705627
- `preg_replace`:
  - Total time: 0.0043911933898926
  - Average time: 4.3911933898926E-6

And lastly, currently there are two instances of where I catch `\Twig_Error` inside the `icon()` function. Should those errors be logged in the admin/error log?

mrgoldy:
!set WIP :construction:

mrgoldy:
!set Chameleon :unicorn:

phpBB/styles/prosilver/theme/icons.css
```diff
@@ -27,6 +32,13 @@ blockquote cite:before,
 	text-rendering: auto; /* optimizelegibility throws things off #1094 */
 }
 
+.icon[class*="png-"],
+.icon[class*="svg-"] {
```
hanakin: do not do this check comment above. These should all have there own class `.svg` `.png` `.fa`. Furthermore do not worry about any changes to the styling that is outside the scope of this pr and will be handled when i convert everything in the templates over in a seperate pr as The whole file will be replaced. The svg style above would not really go in this file but at the top of common.css with the other resets hereh: https://github.com/phpbb/phpbb/blob/master/phpBB/styles/prosilver/theme/common.css#L150

phpBB/styles/prosilver/template/icons/png.html
```diff
@@ -0,0 +1,8 @@
+{% spaceless %}
+<img
+	class="icon png-{{ ICON ~ (CLASSES ? ' ' ~ CLASSES) }}"
```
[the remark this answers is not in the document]
mrgoldy: Not sure if I understood correctly:
`class="o-icon png png-SOME_ICON"` or
`class="o-icon png-SOME_ICON"`?

phpBB/styles/prosilver/template/icons/svg.html
```diff
@@ -0,0 +1,21 @@
...
+<svg
+	class="icon svg-{{ ICON ~ (CLASSES ? ' ' ~ CLASSES) }}"
+	viewBox="0 0 24 24"
+	{% if TITLE %}
+		{% if S_HIDDEN %}aria-hidden="true"{% endif %}
+		aria-labelledby="{{ TITLE_ID }}"
+	{% endif %}
+	role="img"
+	{{ ATTRIBUTES }}
+>
```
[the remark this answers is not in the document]
mrgoldy: Okay, I've added the `xmlns="http://www.w3.org/2000/svg"`.
This does however make my editor's inspections say that the attributes `aria-*` and `role="img"` are not allowed here.

phpBB/styles/prosilver/theme/icons.css
mrgoldy: Ah right, I understand.
I'll just make those small changes to just the `icons.css` as it's rather small file anyways and allows me/us to properly test things and make sure they are all working as intended.

mrgoldy:
Then I still need to know what elements and attributes to remove / replace for SVG's?
Currently removing any `<title>` element, along with all its children.
Removing any `<svg>` element, but preserving its children.
Removing any `fill=""` attribute, exept if its value is `none`.

Also, in an other comment that you would also like to use this for forum icons, I'm guessing you mean "New Post", "Forum locked", etc with that. Why would they go in a different directory than the theme? It sounds like they are very closely tight to the actual theme and currently a lot of them already use different ones. Instead, perhaps you can just add a class to the icon like `o-forum-icon` or w/e.

phpBB/styles/prosilver/template/icons/png.html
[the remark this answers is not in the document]
hanakin: The first one

phpBB/styles/prosilver/theme/icons.css
hanakin: You can make them locally just do not merge them to the pr

hanakin:
Any and all fill rules are ok. You not not need to sanitize any of the content other than the svg tag and the title needs to be dynamic as you have done. The rest of the contents just pass through.

This will be used for everything potentially which includes:
Icons = anything currently handled by font-awesome
Forum icons = the icons added when you create a forum currently found in the files folder
Imagesets = the images in front of the forum and topic rows.’
Forum Images = an optional image added in the app to each forum that appears next to the imageset 
Contact icons = the icons in the menu that appears when clicking the contact bubble icon in the profile of a post.

phpBB/styles/prosilver/template/icons/svg.html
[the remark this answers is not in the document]
hanakin: Possibly because you have them on different lines

phpBB/styles/prosilver/theme/icons.css
mrgoldy: Okay, I was going for doing it now. Test everything and then revert the PR in such a state so it only has the necessary stuff (twig extension and icon templates). This way other people can easily "test/see/fiddle" with it too.

phpBB/styles/prosilver/template/index_body.html
```diff
@@ -33,14 +33,44 @@ <h3><a href="{U_LOGIN_LOGOUT}">{L_LOGIN_LOGOUT}</a><!-- IF S_REGISTER_ENABLED --
 	</form>
 <!-- ENDIF -->
 
+<div class="panel">
```
hanakin: Is this a test file? If so like the css changes uncommit this as well

hanakin:
So talking it over as I mentioned on your other PR this would need to be `Icon()`

cyberalien:
Don't add fill rule to css!

1. Not every SVG has fill. Some rely on stroke. This rule will cause problems to icons that are using stroke.

2. It won't work for external images anyway, so its useless.

3. Its better to just specify fill="currentColor" on shapes that use fill and stroke="currentColor" on shapes that use stroke.

As for attributes removal: title attribute must be removed because it is used as tooltip in some browsers. It overrides title from parent element, so for example if you have this:
```
<a href="#" title="Link title!"><svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="1.27em" height="1em" preserveAspectRatio="xMidYMid meet" viewBox="0 0 1664 1312" style="-ms-transform: rotate(360deg); -webkit-transform: rotate(360deg); transform: rotate(360deg);"><title>SVG title!</title><path d="M1408 768v480q0 26-19 45t-45 19H960V928H704v384H320q-26 0-45-19t-19-45V768q0-1 .5-3t.5-3l575-474 575 474q1 2 1 6zm223-69l-62 74q-8 9-21 11h-3q-13 0-21-7L832 200 140 777q-12 8-24 7-13-2-21-11l-62-74q-8-10-7-23.5T37 654L756 55q32-26 76-26t76 26l244 204V64q0-14 9-23t23-9h192q14 0 23 9t9 23v408l219 182q10 8 11 21.5t-7 23.5z" fill="currentColor"/></svg></a>
```
tooltip will show "SVG title!" instead of "Link title!" as tooltip.

Best option for cleaning up is to run SVGO with all options to clear everything.

hanakin:
@cyberalien 
- There is no bandwidth lost they are server generated embeded svgs. 
- We do not use  parents unless links as wrapping an icon is very poor semantics and leads to bad selector performance which means using the icons toolip is prefered.
- We are using svgo on our icons but the problem is svgo as far as I know is only available as a js library that requires node to run which does not work for custom icons.

cyberalien:
SVGO can be ran before uploading icons. There is also web interface for it: https://jakearchibald.github.io/svgomg/

As for bandwidth, I mean including icons on every page load. If icons are used for all decorations instead of fonts, that could be a lot of icons per page. Its one of reasons why glyph fonts became so popular - fonts use minimal markup, reducing page loading times compared to embedding SVG.

hanakin:
🤔 calling 💩 on that fonts were introduced because of a lack of support of svg. Massive html pages load way faster than dynamic pages as it’s all cached and griped. There is a negative return actually. Let’s say the entire page was 2mb which is massive for an html page. That’s only on initial load subsequent loads would be like 10kb tops with gzip and cache. And there is no way it’s anywhere near that size unless you are trying to load every emoji on a page.

cyberalien:
Please read what I posted again. I said "one of reasons". Lack of support for SVG was definitely main reason though.

Lets do some math. FontAwesome 5 SVG size is between 300 for most basic icon and 1k for most complex icon. So 650b on average. Those are very simple icons. EmojiOne icons are mostly several kb in size up to 11kb. Those are shiny icons used for decorations, such as forum icons. 11kb is extreme icon, on average its about 2-3kb, so 2.5kb.

phpBB.com forum index is currently 66.5kb. Lets calculate replacing icons with SVG embedded in document:
- Header: 3 feed icons + search placeholder
- Navigation: 5 in quick links + 3-6 icons depending on permissions + 3 in user menu
- Forums list: 19 big forum icons, 8 sub-forum icons, 16 last post icons, 6 minimise (and 6 hidden maximize) icons.
- Footer: 5 icons

Big forum icons should be shiny, so that's similar to EmojiOne icons. Other icons are basic, so similar to FontAwesome 5 icons. That's on average 650b for basic icon, 2.5kb for forum icon. Let's assume minimum permissions, so in navigation only 3 icons.

Header: 4 * 0.65 = 2.6kb
Navigation: (4 + 5 + 3 + 3) * 0.65 = 9.75kb
Forums list: 19 * 2.5 + (8 + 16 + 12) * 0.65 = 70.9kb
Footer: 5 * 0.65 = 3.25kb

That's extra 86.5kb for page that's currently 66.5kb. Download size is more than doubled. Its added to every page load. In forum view there are even more icons. In posts there are lots of buttons in every post.

Today page load is very important. It affects SEO. For me that's a massive no for usage.

hanakin:
🤔 Well prosilver is an irrelavent exp, will have to build some test I guess inthe new theme to figure everythingout.

cyberalien:
Cool. Please post results, it would be interesting to know how those calculations work for similar size forum with new layout. 🙂

hanakin:
Ok so still have a lot more tests to run but will update this post once its all done.

Tests run through Chrome Audit 'Sim Fast 3G, 4x CPU Throttle - Mobile'

todo: 
- the test with iconify that will take some time converting the current version of the index and forums pages over
- we do not use the larger icons you speak of with the new theme. I will have to test that still. As all of our icons are very well optimized and monochrome even our imagesets.

I did convert over to use fully inline svgs, using the themes current icon pack 
> would be better once integrated into the app with actual assets instead of the mock ones I'm using which are all very un-performant. This is also with loading all css and js files via jsdeliver cdn directly from the repo as well. 

Results:

New Theme (express static gzip)
index.html: 98kb 2.1secs (99%)
forum.html: 198kb 2.5secs (98%)

phpbb.com (local static tests loaded assets from live sites)
community index: 79kb 6.3secs (56%) [express static gzip] 3.2secs (89%) - probably due to static rendering and gzip
community support forum: 97kb 5.8secs (66%)  [express static gzip] 3.5secs (87%) - probably due to static rendering and gzip

area51 (local static tests loaded assets from live sites)
index: 40kb 3.9secs (83%) [express static gzip .com] 2.8secs (98%)
3.0 disscussion forum: 68kb 3.9secs (83%) [express static gzip] 2.9secs (93%)

hanakin:
ok so finished and updated the above post with the results. sort of as I expected there is not really much tangible difference with static files. would really need to test with phpbb page renders. But the numbers are essentially vacuumed meaning that from a performance gain either way is under 2 sec loads with 10ms interactivity times. Compare this to current static versions of .com and area51 at 3+ secs vs live versions and you see a consistent extra sec for area51 and 2-3 secs for .com

I would not mind entertaining a 3.3 prosilver convert to iconify over font-awesome given these results.

I also put up some tests for that as well

prosilver integration: https://codepen.io/hanakin/pen/oVyvBp?editors=1100
icon list: https://codepen.io/hanakin/pen/GPZvev?editors=1100

cyberalien:
That is not real forum test, it is missing what forums actually have - content. You didn't replace forum icon, which is biggest icon and you've made only 2 rows. Those are conditions of a dead forum. Real forums use 20 rows, not 2. That's 10 times more replication of big icons. Try replacing forum icon with svg, which will happen in new theme and add more rows to see conditions of actual forum.

hanakin:
That was not the test that is a planning environment I put up

cyberalien:
Oh, ok. Now it makes sense ☺️

hanakin:
@mrgoldy check out this 
https://stackoverflow.com/questions/52855116/twig-auto-import-macros-by-environment
Thks to @Nicofuma 

Seems macros are still able to be used

mrgoldy:
I am not sure if this approach will work.. Will investigate some more.
```
Fatal error: Uncaught Twig\Error\RuntimeError: Accessing \Twig\Template attributes is forbidden
```

phpBB/phpbb/template/twig/extension/icon.php
```diff
@@ -0,0 +1,150 @@
...
+	 * @param string		$icon			The icon name
+	 * @return string
+	 */
+	public function png_path(environment $environment, $icon)
+	{
+		$board_url	= defined('PHPBB_USE_BOARD_URL_PATH') && PHPBB_USE_BOARD_URL_PATH;
+		$web_path	= $board_url ? generate_board_url() . '/' : $environment->get_web_root_path();
+		$style_path	= $this->user->style['style_path'];
+
+		return "{$web_path}styles/{$style_path}/template/icons/png/{$icon}.png";
```
hanakin: This should be in theme folder

phpBB/phpbb/template/twig/extension/macro.php
```diff
@@ -0,0 +1,64 @@
...
+	 * Returns a list of global variables to add to the existing list.
+	 *
+	 * @return array An array of global variables
+	 */
+	public function getGlobals()
+	{
+		$macros = null;
+
+		try
+		{
```
hanakin: We may want to consider keeping it as a twig file to differentiate macros from normal template files?

phpBB/styles/all/template/macros.html
```diff
@@ -0,0 +1,5 @@
+{# Wrapper function #}
```
hanakin: This should be in a sub folder called macros

phpBB/styles/prosilver/template/index_body.html
```diff
@@ -11,6 +11,8 @@
 <!-- ENDIF -->
 <!-- EVENT index_body_markforums_after -->
 
+Test: {{ macros.Icon('fa', 'phone') }}
```
hanakin: Hmm what’s with the “macros.”

phpBB/phpbb/template/twig/extension/macro.php
mrgoldy: You mean `.twig` extension, or what do you mean?

phpBB/styles/prosilver/template/index_body.html
mrgoldy: As per [this comment](https://github.com/phpbb/phpbb/pull/5545#issuecomment-474015052), it would mean adding a loaded template file to a 'super global'. Which is assigned in the `phpbb/template/twig/extension/macro.php`. Meaning, that all macros defined in the `macros.html` file will be available under `macros.XYZ()`. It is not possible to add them without a 'prefix', eg `{{ Icon() }}`. That would either require still importing them in every file, or creating them as twig functions in php.

phpBB/phpbb/template/twig/extension/icon.php
mrgoldy: Okay, btw this will not handle 'style inheritance'. Is that correct behaviour?

phpBB/styles/prosilver/template/index_body.html
hanakin: I see not sure if we want to call it macros then but we can change that later

hanakin: I see not sure if we want to call it macros then but we can change that later

phpBB/phpbb/template/twig/extension/icon.php
hanakin: What do you mean?

phpBB/phpbb/template/twig/extension/macro.php
hanakin: Yes

phpBB/phpbb/template/twig/extension/icon.php
mrgoldy: If the icon is not present in the user's style it will display the 'No image'.
It will not inherit from any parent style (eg. Chameleon).

phpBB/config/default/container/services_twig.yml
```diff
@@ -54,6 +54,13 @@ services:
         tags:
             - { name: twig.extension }
 
+    template.twig.extensions.icon:
```
rubencm: Why you don't add this services in another file better?
This file is included in the installer environment, and probably can cause problems with some dependencies

mrgoldy: The dependency is `user`, that should be available, right?

But I know some other icon extensions have the similar problem, so perhaps it's a good idea to start using a separate config file, something like `services_twig_extensions.yml` so we can only include those in the production environment?

mrgoldy:
I tried to rebase, resolve conflicts and force push.
But somehow managed to double my own commits, yet the master commit message are still in there as Travis is failing on them.
Any chance somebody could help me out here?

phpBB/phpbb/template/twig/extension/icon.php
```diff
@@ -0,0 +1,183 @@
...
+	 * @param string		$title			Icon title
+	 * @param bool			$hidden			Hide the icon title from view
+	 * @param string		$classes		Additional classes (eg. "fa-fw")
+	 * @param array			$attributes		Additional attributes for the icon, where the key is the attribute.
+	 *                      				{'data-ajax': 'mark_forums'} results in ' data-ajax="mark_forums"'
+	 * @return string
+	 */
+	public function icon(environment $environment, $type, $icon, $title = '', $hidden = false, $classes = '', array $attributes = [])
+	{
+		$type = utf8_strtolower($type);
```
JoshyPHP: Since the only relevant input is ASCII, you can use `strtolower()` here so you don't depend on the mbstring polyfill.

phpBB/phpbb/template/twig/extension/icon.php
```diff
@@ -0,0 +1,183 @@
...
+
+				foreach ($element->childNodes as $node)
+				{
+					$children[] = $node;
+				}
+
+				foreach ($children as $child)
+				{
+					$element->parentNode->insertBefore($child, $element);
+				}
```
JoshyPHP: I believe you can replace the content of that block with a single loop:
```php
				while (isset($element->firstChild))
				{
					$element->parentNode->insertBefore($element->firstChild, $element);
				}
```

mrgoldy: Tested and you are absolutely correct 👍

mrgoldy:
Okay, tests added aswell.

Requesting another review by @hanakin 
Especially verify / double check the HTML files ([font](https://github.com/phpbb/phpbb/pull/5545/files#diff-7349591346d45c1f27954d7530a19647), [png](https://github.com/phpbb/phpbb/pull/5545/files#diff-a41f5d894fb3d5756452262def600dc9) and [svg](https://github.com/phpbb/phpbb/pull/5545/files#diff-a41f5d894fb3d5756452262def600dc9))
And if the expected outcomes in [the tests](https://github.com/phpbb/phpbb/pull/5545/files#diff-857dac0d1e818fe89027f9bb8c958aef) are correct, as in, what you want them to be.

Then the only thing still be discussed is that SVG has 'style inheritance', where it will look in parent styles directories for the image, but PNG does not have that. It will only look for the image in the user's default style. Do I have to look into a way that PNG also has some sort of style inheritance?

hanakin:
@mrgoldy thanks I have a look over everything in more detail but there are a couple changes that need implemented. yes there needs to be inheritance for the pngs... also we need to add a fourth icon type for iconify. look at the implementation in the current chameleon for reference https://github.com/hanakin/chameleon/blob/master/gulpfile.js#L61. This will require including iconify.js in the footer.

hanakin:
@mrgoldy to speed things up you can just duplicate the font template for iconify and change the first two classes to `iconify o-icon` in that order and add a switch for `iconify` to the twig function.

phpBB/styles/prosilver/template/icons/iconify.html
```diff
@@ -0,0 +1,4 @@
+{% spaceless %}
+	<i class="iconify o-icon {{ ICON ~ (CLASSES ? ' ' ~ CLASSES) }}"{% if S_HIDDEN %}{% if TITLE %} title="{{ lang(TITLE) }}"{% endif %} aria-hidden="true"{% endif %}{{ ATTRIBUTES }}></i>
```
hanakin: hmm now that I see it this will not work this way. It should be a span tag. iconify will actually replace everything with an svg so the next line will be removed. we may also need an inline false flag not sure will have to test it this weekend. Also forgot iconify does not use classess for the icons but instead a data attribute.

`<span class="iconify" title="android" data-icon="mdi-android" data-inline="false"></span>`

lastly since its completely replaced any custom attributes will not work we may need a wrapper around them. 

Let me do some testing...

hanakin: ok so we just need to tweak this line to

`<i class="iconify o-icon {{ CLASSES ? ' ' ~ CLASSES }}" {% if S_HIDDEN %}{% if TITLE %} title="{{ lang(TITLE) }}"{% endif %} aria-hidden="true"{% endif %} data-icon="fa:power-off" data-inline="true" {{ ATTRIBUTES }}></i>`

tests/template/extension_test.php
```diff
@@ -253,8 +259,197 @@ public function data_template_extensions()
...
+					'hidden'		=> true,
+					'classes'		=> 'icon-lg',
+					'attributes'	=> [
+						'data-swap'		=> 'Swap text',
+					],
+				],
+				[
+					'ICON_PENCIL'	=> 'Pencil icon',
+				],
+				'<i class="iconify o-icon pencil icon-lg" title="Pencil icon" aria-hidden="true" data-swap="Swap text"></i>
```
hanakin: update to `'<i class="iconify o-icon icon-lg" title="Pencil icon" aria-hidden="true" data-icon="fa:pencil" data-inline="true" data-swap="Swap text"></i>`

tests/template/extension_test.php
```diff
@@ -253,8 +259,197 @@ public function data_template_extensions()
...
+				[
+					'type'			=> 'iconify',
+					'icon'			=> 'phone',
+					'title'			=> '',
+					'hidden'		=> false,
+					'classes'		=> '',
+					'attributes'	=> [],
+				],
+				[],
+				'<i class="iconify o-icon phone"></i>',
```
hanakin: change to

`<i class="iconify o-icon" data-icon="fa:phone" data-inline="true"></i>`

phpBB/styles/prosilver/template/icons/iconify.html
```diff
@@ -0,0 +1,4 @@
+{% spaceless %}
+	<i class="iconify o-icon{{ CLASSES ? ' ' ~ CLASSES }}"{% if S_HIDDEN %}{% if TITLE %} title="{{ lang(TITLE) }}"{% endif %} aria-hidden="true"{% endif %} data-icon="fa:{{ ICON }}" data-inline="true"{{ ATTRIBUTES }}></i>
```
hanakin: `data-icon="fa:{{ ICON }}"` is incorrect the fa: is part of the `ICON` variable since you can use any icon from any library you could have fa:home, mdi:home, is:home, dash:home, mdl:home, etc...

tests/template/extension_test.php
```diff
@@ -253,8 +259,197 @@ public function data_template_extensions()
...
+				[
+					'type'			=> 'svg',
+					'icon'			=> 'dirty',
+					'title'			=> '',
+					'hidden'		=> false,
+					'classes'		=> '',
+					'attributes'	=> [],
+				],
+				[],
+				'<svg class="o-icon svg svg-dirty" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" role="img">
```
hanakin: what is svg-dirty?

hanakin:
pending the test looks good

tests/template/extension_test.php
mrgoldy: Most of the provided SVG files provided are 'clean' already. As in, don't contain any unnecessary or unwanted tags, such as `DOCTYPE`, `<?xml ?>`, `<title>`, etc..
The 'dirty' one does, so it test if the svg gets cleaned properly.
And if you mean the `svg-dirty` class, that's the svg file(/icon) name that gets added to to `svg-`, like with all the others, such as `svg-pencil` and `svg-phone`.

hanakin: o your adding a class for svgs? gotcha guess it does not hurt...

mrgoldy: I mean, if it is not necessary I will remove it. You're the master 😛 
I thought it was desired, to perhaps style specific svg's.
Custom classes are also added, as per the test above.

hanakin: it does not hurt anything, I would not recommend using it to style. Its better to style based on where something is used not what it is usually which custom classes handle

hanakin:
ill work on the icon rework for prosilver pr based off this tomorrow probably so we can merge both together.

hanakin:
@mrgoldy just noticed that you have all the templates in the template folder. This is not right the icons folder needs moved to all/templates/macros

phpBB/styles/prosilver/template/icons/svg.html
```diff
@@ -0,0 +1,9 @@
+{% spaceless %}
+	{% set TITLE_ID = TITLE ? TITLE|lower|replace({' ': '_'}) ~ '-' ~ random() %}
+
+	<svg class="o-icon svg svg-{{ ICON ~ (CLASSES ? ' ' ~ CLASSES) }}" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"{% if TITLE %}{% if S_HIDDEN %} aria-hidden="true"{% endif %} aria-labelledby="{{ TITLE_ID }}"{% endif %} role="img"{{ ATTRIBUTES }}>
```
hanakin: change the `svg` class to `o-icon-svg`

phpBB/styles/prosilver/template/icons/png.html
```diff
@@ -0,0 +1,3 @@
+{% spaceless %}
+	<img class="o-icon png png-{{ ICON ~ (CLASSES ? ' ' ~ CLASSES) }}" src="{{ SOURCE }}" alt="{{ lang(TITLE) }}"{{ ATTRIBUTES }} />
```
hanakin: change the `png` class to `o-icon-png`

phpBB/styles/prosilver/template/icons/font.html
```diff
@@ -0,0 +1,4 @@
+{% spaceless %}
+	<i class="o-icon font fa-{{ ICON ~ (CLASSES ? ' ' ~ CLASSES) }}"{% if S_HIDDEN %}{% if TITLE %} title="{{ lang(TITLE) }}"{% endif %} aria-hidden="true"{% endif %}{{ ATTRIBUTES }}></i>
```
hanakin: change the `font` class to `o-icon-font`

phpBB/styles/prosilver/template/icons/iconify.html
```diff
@@ -0,0 +1,4 @@
+{% spaceless %}
+	<i class="iconify o-icon{{ CLASSES ? ' ' ~ CLASSES }}"{% if S_HIDDEN %}{% if TITLE %} title="{{ lang(TITLE) }}"{% endif %} aria-hidden="true"{% endif %} data-icon="{{ ICON }}" data-inline="true"{{ ATTRIBUTES }}></i>
```
hanakin: change the `iconify` class to `iconify o-icon-svg` since this will return an svg icon.

phpBB/phpbb/template/twig/extension/icon.php
```diff
@@ -110,6 +133,20 @@ public function icon(environment $environment, $type, $icon, $title = '', $hidde
...
+		if ($not_found)
+		{
+			if (empty($base_path))
+			{
+				$board_url = defined('PHPBB_USE_BOARD_URL_PATH') && PHPBB_USE_BOARD_URL_PATH;
+				$base_path = $board_url ? generate_board_url() . '/' : $environment->get_web_root_path();
+			}
+
+			$source = "{$base_path}styles/chameleon/theme/icons/png/404.png";
+			$type = 'png';
```
hanakin: can we make this dynamic. check for 404.svg if none then use 404.png.

phpBB/phpbb/template/twig/extension/icon.php
```diff
@@ -110,6 +133,20 @@ public function icon(environment $environment, $type, $icon, $title = '', $hidde
...
+		// If no PNG or SVG icon was found, display a default 404 PNG icon.
+		if ($not_found)
+		{
+			if (empty($base_path))
+			{
+				$board_url = defined('PHPBB_USE_BOARD_URL_PATH') && PHPBB_USE_BOARD_URL_PATH;
+				$base_path = $board_url ? generate_board_url() . '/' : $environment->get_web_root_path();
+			}
+
+			$source = "{$base_path}styles/chameleon/theme/icons/png/404.png";
```
hanakin: this will be in the all/imgs folders not chameleon/theme.

phpBB/phpbb/template/twig/extension/icon.php
mrgoldy: I thought the idea was to provide an icon/image that is always there. So there shouldn't be a need to verify existance of the 'not found icon'. So not sure what you want here? Either display a `404.svg` if the original type was `svg` or what?

hanakin: while we may provide 1 for both this could be customized by someone so we need to check both folders. if there is an 404.svg display it. if not display the 404.png. It will still only display the same thing.

mrgoldy: well, this changes things slightly as that would mean we would have to verify the existance of the icons and otherwise still end up with no icon.
Isn't more logical to provide a 404.svg and a 404.png in the `all` folder, which should just always be there as backup. 
What's customized are the styles, and they should just provide their own icons and when designed correctly it should never even show these 404 icons.
I thought the 404 icon was intended as an ultimate backup.

hanakin: sorry guess i was getting confused a little by the theme folder. to simplify lets just use the png for everything as originally discussed that way it works on any browser/system.

phpBB/styles/all/templates/macros/icons/font.html
```diff
@@ -0,0 +1,4 @@
+{% spaceless %}
```
hanakin: think these are supposed to be in `template` not `templates`

hanakin: they should also be twig files not html files

mrgoldy: I mean .. https://github.com/phpbb/phpbb/pull/5545#issuecomment-552680502
> @mrgoldy just noticed that you have all the templates in the template folder. This is not right the icons folder needs moved to all/templates/macros

If you let me know where you want them I’ll adjust it. Along with the twig file extension.

hanakin: it belongs in the current folder that exsists in the software https://github.com/phpbb/phpbb/tree/master/phpBB/styles/all/template

hanakin: templates was a typo so just move the macros/iocns folders into the template folder

marc1706: We'll get there eventually :)

phpBB/phpbb/template/twig/extension/icon.php
```diff
@@ -0,0 +1,255 @@
...
+	/**
+	 * Prepare an SVG for usage in the template icon.
+	 *
+	 * This removes any <?xml ?> and <!DOCTYPE> elements,
+	 * aswell as any <svg> and <title> elements.
+	 *
+	 * @param \Twig\TemplateWrapper	$file	The SVG file loaded from the environment
+	 * @return string						The cleaned SVG
+	 */
+	protected function prepare_svg(\Twig\TemplateWrapper $file)
```
hanakin: This is messing up the viewBox parameter. Which needs to be pulled from the original source. Somehow you have it defaulting to "0 0 24 24" which is incorrect

phpBB/styles/all/templates/macros/icons/svg.html
```diff
@@ -0,0 +1,9 @@
+{% spaceless %}
+	{% set TITLE_ID = TITLE ? TITLE|lower|replace({' ': '_'}) ~ '-' ~ random() %}
+
+	<svg class="o-icon o-icon-svg svg-{{ ICON ~ (CLASSES ? ' ' ~ CLASSES) }}" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"{% if TITLE %}{% if S_HIDDEN %} aria-hidden="true"{% endif %} aria-labelledby="{{ TITLE_ID }}"{% endif %} role="img"{{ ATTRIBUTES }}>
```
hanakin: as stated above the viewBox needs to be passed from the original svg

phpBB/phpbb/template/twig/extension/icon.php
```diff
@@ -0,0 +1,321 @@
...
+	 * @param string		$type			Icon type (font|iconify|png|svg)
+	 * @param string		$icon			Icon name (eg. "bold")
+	 * @param string		$title			Icon title
+	 * @param bool			$hidden			Hide the icon title from view
+	 * @param string		$classes		Additional classes (eg. "fa-fw")
+	 * @param array			$attributes		Additional attributes for the icon, where the key is the attribute.
+	 *                      				{'data-ajax': 'mark_forums'} results in ' data-ajax="mark_forums"'
+	 * @return string
+	 */
+	public function icon(environment $environment, $type, $icon, $title = '', $hidden = false, $classes = '', array $attributes = [])
```
rxu: So, currently you need to explicitly define icon type to render it. Can we add some `auto` mode to automatically detect icon type and render it then? Or even split such code into a function and then use it here instead of `switch`.

hanakin: No that makes no sense and defeats the purpose.  How would you even automatically detect an icon type? There are too different.

mrgoldy: Automatically detecting will also not return one definitive result, as a lot of the icon sets use the same names. So, that would only be possible indeed if some 'default' was configured. And then you can just use that variable in the template instead: `{{ icon(DEFAULT_ICON, 'user') }}`.

PR #5555, opened 2019-03-27
[ticket/15257] Provide extension not enableable messages

senky:
PHPBB3-15257

Replaces #4863, in honor of javiexin.

I took different approach - `is_enableable` now allows to return array apart from boolean. `true` means the ext can be enabled, `false` that it cannot be enabled and should report generic reason, array means it cannot be enabled and reasons are attached.

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15257


tests/extension/ext/vendor5/foo/ext.php
```diff
@@ -0,0 +1,19 @@
...
+	{
+		return array('Reason 1', 'Reason 2');
+	}
+
+	public function enable_step($old_state)
+	{
+		self::$enabled = true;
+		return self::$enabled;
+	}
+}
```
mrgoldy: No new line at end of file?

phpBB/includes/acp/acp_extensions.php
```diff
@@ -1130,4 +1126,14 @@ public function output_metadata_to_template($metadata)
 			));
 		}
 	}
+
+	protected function check_is_enableable($enableable)
```
marc1706: Please add a docblock to this.

tests/extension/ext/vendor5/foo/composer.json
```diff
@@ -0,0 +1,23 @@
...
+	"require": {
+		"php": ">=5.3"
+	},
+	"extra": {
+		"display-name": "phpBB Bar Extension",
+		"soft-require": {
+			"phpbb/phpbb": "3.3.*@dev"
+		}
+	}
+}
```
marc1706: Add new line at the end of the file

tests/extension/ext/vendor5/foo/ext.php
marc1706: Should be added

phpBB/includes/acp/acp_extensions.php
VSEphpbb: Just a thought... Another approach (which I think is may be more cognitive/readable, could be like):
```php
$extension = $this->ext_manager->get_extension($ext_name);
if ($message = $extension->is_enableable() !== true)
{
	$this->show_error($message);
}

/**
* Trigger an error message for an extension that could not be enabled.
* Error message can be set by the extension.
*
* @param bool|array $message Array of reasons (set by the extension) 
* or false for default reason.
*/
public function show_error($message)
{
	if (is_array($message))
	{
		$message = implode('<br />', $message)
	}
	else if (empty($message)
	{
		$message = $this->user->lang('EXTENSION_NOT_ENABLEABLE');
	}

	trigger_error($message . adm_back_link($this->u_action), E_USER_WARNING);
}
```

marc1706:
@VSEphpbb better now?

PR #5556, opened 2019-03-31
[ticket/16004] Add check-in for Emojis in Username

3D-I:
PHPBB3-16004

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16004


phpBB/includes/functions_user.php
```diff
@@ -1709,6 +1709,14 @@ function validate_username($username, $allowed_username = false)
 		return false;
 	}
 
+	// The very first check is for
+	// out-of-bounds characters that are currently
+	// not supported by utf8_bin in MySQL
+	if (preg_match('/[\x{10000}-\x{10FFFF}]/u', $username))
+	{
+		return 'INVALID_EMOJIS_USERNAME';
```
VSEphpbb: It should just return `INVALID_EMOJIS` sine `_USERNAME` is appended to the result of this function

VSEphpbb:
Should test this in the validate_username_test test

phpBB/includes/functions_user.php
3D-I: Done, thx.

3D-I:
I will see to add such test therein.

VSEphpbb:
Also, just a thought... Is it true Emoji are not allowed in a name? Why not?

DavidIQ:
Well I think I'd rather support them than deny them. But perhaps that's something that we need to push off for a future version as it would require the username column data type to be changed, a likely expensive and time-consuming operation.

VSEphpbb:
But it would be kind of hard to know what to call a user named: 🐸

DavidIQ:
"Kermit"? "None of my business meme"? "Pepe"? "Trigger"? Take your pick 😆

3D-I:
If you allow emojis in usernames then many other tables will break, like Logs, Notifications etc.

3D-I:
The test is a courtesy of @mrgoldy :+1:

PR #5562, opened 2019-04-08
[ticket/15961] Use newer TLS in SMTP

senky:
PHPBB3-15961

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15961


mrgoldy:
You're changing the setting to 'not include' the protocol. While that's fine for new users that are going to configure that setting, but users with that setting already configured are most definitely going to overlook that.
Perhaps it's better to check if no server protocol was provided, and if so, either throw an error or strip the protocol yourself?

senky:
Why? There is a fallback which stops creating new TSL connection ([here](https://github.com/phpbb/phpbb/pull/5562/files#diff-b16f574b60c586a459ead8be7e5c296bR1588)) when one has already been created by including protocol. So user can include protocol but is discouraged to do so.

PR #5575, opened 2019-04-20
[ticket/14459] Check language input for group and fix template logic

rubencm:
Checklist:

- [x] Tests pass

Replaces #4198 

https://tracker.phpbb.com/browse/PHPBB3-14459

### DIFF
--- a/phpBB/adm/style/acp_attachments.html
+++ b/phpBB/adm/style/acp_attachments.html
@@ -421,26 +421,25 @@
 	</tr>
 	</thead>
 	<tbody>
-	<!-- BEGIN attachments -->
+	{% for attachments in attachments %}
 		<tr>
 			<td>
-				<!-- IF ! attachments.S_IN_MESSAGE -->{# File info of attachments in private messages is hidden #}
-					<a href="{attachments.U_FILE}"><strong>{attachments.REAL_FILENAME}</strong></a><br />
-					{L_EXTENSION_GROUP}{L_COLON} <strong><!-- IF attachments.EXT_GROUP_NAME -->{attachments.EXT_GROUP_NAME}<!-- ELSE -->{L_NO_EXT_GROUP}<!-- ENDIF --></strong><br />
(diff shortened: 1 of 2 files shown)


3D-I:
So the language object it is used (added) in this file just for this specific language stuff while in the rest of it is the user object.

rubencm:
That can be done in other pull request, make a lot of changes in the file can cause conflicts in other pull requests

3D-I:
So why not to use `$language = $this->phpbb_container->get('language');` only and beforewhere it is needed like in the rest of the file? Then `$language->lang...`

rubencm:
Because it will have to be injected with dependency injection, so is better to leave the variable in the constructor

rubencm:
@3D-I

3D-I:
I'll take a look at it tonight. Thanks.

3D-I:
I posted everything at discord ;)

3D-I:
immagine

3D-I:
Also, a language key should be removed from the language file now

`    'NO_EXT_GROUP'                => 'None',`

phpBB/adm/style/acp_attachments.html
```diff
@@ -421,17 +421,25 @@ <h2>{L_UPLOADING_FILES}</h2>
 	</tr>
 	</thead>
 	<tbody>
-	<!-- BEGIN attachments -->
+	{% for attachments in attachments %}
```
rubencm: @3D-I do this work? for attachments in attachments

phpBB/includes/acp/acp_attachments.php
```diff
@@ -27,6 +27,9 @@ class acp_attachments
 	/** @var \phpbb\config\config */
 	protected $config;
 
+	/** @var \phpbb\language\language */
+	protected $lang;
```
marc1706: Please use `$language` as that is what we usually use.

PR #5589, opened 2019-05-11
[captcha/qa] fix warning around using count() with NULL variable

nunoplopes:
The QA captcha plugin gives the following warnings when the user's language doesn't have any questions set. The code runs `count()` on a NULL variable:
```
[phpBB Debug] PHP Warning: in file [ROOT]/phpbb/captcha/plugins/qa.php on line 87: count(): Parameter must be an array or an object that implements Countable
[phpBB Debug] PHP Warning: in file [ROOT]/phpbb/captcha/plugins/qa.php on line 104: count(): Parameter must be an array or an object that implements Countable
```

### DIFF
--- a/phpBB/phpbb/captcha/plugins/qa.php
+++ b/phpBB/phpbb/captcha/plugins/qa.php
@@ -21,7 +21,7 @@ class qa
 {
 	var $confirm_id;
 	var $answer;
-	var $question_ids;
+	var $question_ids = [];
 	var $question_text;
 	var $question_lang;
 	var $question_strict;


rxu:
The file has another several occurrences of ```!count($this->question_ids)```. I guess it would be enough to assign it to array in the var definition, e.g.
```php
var $question_ids = [];
```
and leave the rest of code as is.

mrgoldy:
Please don't use `var` 😢 
I would prefer `public/protected`.

rxu:
It's not me but the current core code 😉

3D-I:
Any way.. the commit message is wrong.

nunoplopes:
> Any way.. the commit message is wrong.

Could you please expand? I'm not a regular contributor, so need some guidance here.

How do you prefer to move this forward? Remove all those `count`? Just initializer the field to an empty array?  I can prepare the patch once there's an agreement.
Thanks!

rubencm:
Hello @nunoplopes , here, the commit messages should follow this format https://area51.phpbb.com/docs/dev/master/development/git.html#commit-messages and reference a ticket in the tracker https://tracker.phpbb.com

3D-I:
here's the ticket ID
https://tracker.phpbb.com/browse/PHPBB3-15982

EA117:
> How do you prefer to move this forward? Remove all those `count`?
> Just initialize the field to an empty array?

The approach currently in the code for other similar fixes, such as https://github.com/phpbb/phpbb/pull/5111/files, was to declare as "var $question_ids = array();"

nunoplopes:
I've updated the patch, thank you!

marc1706:
merged via https://github.com/phpbb/phpbb/commit/d3c3dc24fe900773ecaf32ea107c29a0af562d4d

PR #5637, opened 2019-07-28
[ticket/16054] Restore ability to login from any template.

EA117:
PHPBB3-16054

Moving the login form's add_form_key() work into page_header(), so that the
template variables required for presenting a login form are again available
to any template that chooses to consume them.

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16054

### DIFF
--- a/phpBB/includes/functions.php
+++ b/phpBB/includes/functions.php
@@ -2526,9 +2526,6 @@ function login_box($redirect = '', $l_explain = '', $l_success = '', $admin = fa
 		));
 	}
 
-	// Add form token for login box
-	add_form_key($form_name, '_LOGIN');
-
 	$s_hidden_fields = build_hidden_fields($s_hidden_fields);
 
 	$login_box_template_data = array(
@@ -2662,9 +2659,6 @@ function login_forum_box($forum_data)
 	extract($phpbb_dispatcher->trigger_event('core.login_forum_box', compact($vars)));
 
 	page_header($user->lang['LOGIN']);
-
-	// Add form token for login box
-	add_form_key('login', '_LOGIN');
 
 	$template->assign_vars(array(
 		'FORUM_NAME'			=> isset($forum_data['forum_name']) ? $forum_data['forum_name'] : '',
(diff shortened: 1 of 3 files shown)


EA117:
@Derky Hoping maybe you could review the changes in https://github.com/phpbb/phpbb/pull/5637/files to confirm they maintain everything expected regarding the earlier SECURITY-228 fix.  Thanks.

The changes themselves aren't big, but the meaning and intention may be important, so a brief explanation:

Originally "[ticket/security/228] Add form token to login box" (https://github.com/phpbb/phpbb/commit/b8368980162392bf9f97496ecec18abe2bd34fad) introduced use of add_form_key() & check_form_key() to the login form(s), and created a S_FORM_TOKEN_LOGIN template variable for login forms to consume.

To create a temporary backwards-compatibility story, "[ticket/16042] Use S_LOGIN_REDIRECT to output login form token" (https://github.com/phpbb/phpbb/commit/3c822556c137a8517dd6a34963bced89b2a150ed) was added to "piggy-back" the new S_FORM_TOKEN_LOGIN fields into the existing S_LOGIN_REDIRECT variable.  So that for now even phpBB 3.2.5-compatible templates could still login successfully after the SECURITY-228 fix, rather than so many phpBB 3.2.6+ users getting "locked out" after updating.

It was then also determined that not even all of the proSilver-defined login forms supported S_FORM_TOKEN_LOGIN yet, and this was addressed in "[ticket/16036] Add S_FORM_TOKEN_LOGIN to all login forms" (https://github.com/phpbb/phpbb/commit/540a6b7d2e7bfb9f2c24636a37127b16bbf6bc46)  It's possible we're still missing S_FORM_TOKEN_LOGIN on one more proSilver login form, described in https://tracker.phpbb.com/browse/PHPBB3-16046.


The behavior intended to be addressed by "[ticket/16054] Restore ability to login from any template"  (this PR) and the changes shown in https://github.com/phpbb/phpbb/pull/5637/files is the fact that prior to the original SECURITY-228 fix, "a style could choose to put a login form on any page it wanted to."

Meaning, all of the template variables required to present a login form (S_LOGIN_REDIRECT) were part of page_header(), and available to every page template.

Now, after SECURITY-228, "only those pages where proSilver saw fit to present a login dialog" can support a login dialog, because those are the only pages the SECURITY-228 fix added add_form_key() calls to.  Styles which presented login forms on additional pages (compared to proSilver) now fail with "form invalid", and "it can't be solved by fixing the style templates."


In https://github.com/phpbb/phpbb/pull/5637/files the add_form_key() call is being moved to the central page_header() location, such that both S_LOGIN_REDIRECT and now S_FORM_TOKEN_LOGIN are both available to any template for any page which desires to consume them.  Meaning any template can present a login form if desired, same as it was in phpBB 3.2.5 and earlier.

The expectation is that this maintains every aspect of the previous security fix, and introduces at worst a small overhead by now generating the "_LOGIN" key hash to be available for every page template.

PR #5645, opened 2019-08-05
[ticket/16124] Fix users search by last visit time in memberlist.php

rxu:
Checklist:

- [x] Tests pass

<a href="https://tracker.phpbb.com/browse/PHPBB3-16124">PHPBB3-16124</a>.

### DIFF
--- a/phpBB/memberlist.php
+++ b/phpBB/memberlist.php
@@ -1073,11 +1073,11 @@ switch ($mode)
 
 				if ($active_time !== false)
 				{
-					if ((int) $active[0] == 0 && (int) $active[1] == 0 && (int) $active[2] == 0)
+					if ($active_select === 'lt' && (int) $active[0] == 0 && (int) $active[1] == 0 && (int) $active[2] == 0)
 					{
 						$sql_where .= ' AND u.user_lastvisit = 0';
 					}
-					else if ($active_select != 'lt')
+					else if ($active_select === 'gt')
 					{
 						$sql_where .= ' AND u.user_lastvisit ' . $find_key_match[$active_select] . ' ' . $active_time;
 					}
--- a/phpBB/memberlist.php
+++ b/phpBB/memberlist.php
@@ -1073,7 +1073,18 @@ switch ($mode)
 
 				if ($active_time !== false)
 				{
-					$sql_where .= " AND u.user_lastvisit " . $find_key_match[$active_select] . ' ' . $active_time;
+					if ((int) $active[0] == 0 && (int) $active[1] == 0 && (int) $active[2] == 0)
+					{
+						$sql_where .= ' AND u.user_lastvisit = 0';
+					}
+					else if ($active_select != 'lt')
+					{
+						$sql_where .= ' AND u.user_lastvisit ' . $find_key_match[$active_select] . ' ' . $active_time;
+					}
+					else
+					{
+						$sql_where .= ' AND (u.user_lastvisit > 0 AND u.user_lastvisit < ' . $active_time . ')';
+					}
 				}
 			}
 


EA117:
Something doesn't seem right about bringing the _' AND u.user_lastvisit = 0'_ case over from acp_prune.php, at least in my opinion.

Currently in memberlist.php I can search for either "users who have never logged on" (before 0000-00-00), or "only users who have logged on" (after 0000-00-00).  But with the proposed code change, now we're ignoring "before" or "after" simply because the date specified is 0000-00-00.

This is intentional in the acp_prune.php code, and is specifically cited in the ACP Prune Users UI as _"Enter 0000-00-00 to prune users who never logged in, Before and After conditions will be ignored."_  What's not clear is how this is helpful or resolves issues for membership.php.  At face value it seems to simply remove flexibility which is currently present.

The "positive change" seen in bringing the logic over from acp_purge.php is that the "less than a specific non-zero date" case will now always exclude "never logged on" users.  (i.e. _"less than N"_ where _"N > 0"_ will never match "0" now.)  Meaning when searching with a specific last visit date (other than 0000-00-00), the member must actually _have_ a last visit date, else that member will not match.

But its unclear how that benefit would explain or resolve the PHPBB3-16124 issue ("incorrect results, or incorrectly empty results"), so perhaps I still don't fully understand the original issue, the code, or both.  Some more details in PHPBB3-16124 regarding specifics of the original problem to be solved might help illuminate additional benefits or confirm concerns with the current proposal.

rxu:
For example, if you search users visited before 2019-12-12, you'll get users never visited (having 000-00-00) which shouldn't be the case.

EA117:
That makes sense, and is solved directly by the third case in the copied acp_prune.php logic.

It was mainly the "or no results when there shouldn't be" aspect of PHPBB3-16124 which I can't explain, since this proposed logic would only remove 0000-00-00 matches from the results (as you've indicated), and wouldn't "now makes a non-empty result set where there was previously an empty set."  If that's not really an issue, then we can ignore that.

So the behavior being lost by using the proposed code as-is would be "there isn't a easy way to specify wanting to see only users who have visited the site."  Before I could search "after 0000-00-00" to see this in memberlist.php, but with the proposed code, this query would now show me "only the never visited users" (regardless of "before" or "after" being specified), due to how the acp_prune.php logic intentionally works.

If we want to just add the same disclaimer to the memberlist.php search UI as exists in acp_prune.php UI ("specify 0000-00-00 and Before or After will be ignored"), that is an option.  And someone wanting to list "only users who have a last visit date" would need to instead search with some real date that pre-dates their board start date; instead of being able to specify 0000-00-00 to get that same result.

My suggestion, for what it's worth, would be to not change the UI and instead just update the logic to "not be literally what acp_prune.php needed", and allow for the "after 0000-00-00" case to continue to be honored in memberlist.php.

Meaning we would add a _"if ($active_select != 'gt')"_ condition to the first case, such that the replacement code is:

```
if ($active_select != 'gt' && (int) $active[0] == 0 && (int) $active[1] == 0 && (int) $active[2] == 0)
{
	$sql_where .= ' AND u.user_lastvisit = 0';
}
else if ($active_select != 'lt')
{
	$sql_where .= ' AND u.user_lastvisit ' . $find_key_match[$active_select] . ' ' . $active_time;
}
else
{
	$sql_where .= ' AND (u.user_lastvisit > 0 AND u.user_lastvisit < ' . $active_time . ')';
}
```

Such that if someone specifies "before 0000-00-00" (or the impossible to specify "equals" case), we fall into the "u.user_lastvisit = 0" query.  Someone specifying "after 0000-00-00" falls into the same case as anyone searching with a real date.  And only someone searching with a real date would fall into the last case, "before YYYY-MM-DD", where the 0000-00-00 users will now be automatically excluded.

rxu:
So what you suggest is falling into 3 cases:
- lt && 0000-00-00 -> search only for never logged in users;
- gt - search for users logged in ever (if the date specified is 0000-00-00) or later than the date specified (other than 0000-00-00), never-logged-in users are excluded;
- lt && date specified != 0000-00-00 - search for users logged in before the date specified excluding never-logged-in users.

Well, that doesn't differ much from what I meant to implement, so looks fine for me.

rxu:
AppVeyor tests config needs fixing.

EA117:
Agreed on the updated summary of expected outcomes in all three cases.  The updated commit code works successfully here too, on phpBB 3.2.7 and PHP 7.2.19.  Thanks for considering.

sfinexer:
Sorry, I don’t know English, use a translator ....

I discovered this inaccuracy. And my opinion is that the admin code is not correct, and it needs to be corrected.
I also think that it is necessary to make a FLAG that the user did not visit the forum when searching.
Most importantly, this does not violate the structure and database of the forum.

rxu:
ACP is unrelated to the current PR. Also you don't state what exactly is incorrect in ACP.

sfinexer:
> ACP is unrelated to the current PR. Also you don't state what exactly is incorrect in ACP.

Что такое ACP и что такое PR, я непонимаю............
Прошу прощения..........
I beg your pardon.

marc1706:
Restarted appveyor build

PR #5650, opened 2019-08-07
[ticket/16127] Add UI for email $max_chunk_size

3D-I:
PHPBB3-16127

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16127


EA117:
Notwithstanding that "Email maximum chunk size" is exactly the right name for the database setting we know needs to be controlled, it's not as clear that simply mirroring this name as the "human-readable" name of this ACP setting is the most intuitive choice.

The same argument could be made for "Email package size", but I don't think we want to propose changing the name of an existing established setting.  Still, "Maximum number of queued emails to send at the same time" might have been a more intuitive name for what that setting controls, too.

But since we are introducing the "Email maximum chunk size" setting for the first time in ACP, could we eliminate the non-technical and non-intuitive "chunk" reference, and instead make the setting name in ACP "Maximum number of recipients per email".

Or something similar, with simply the goal to provide a more intuitive name.  The description can then be used to say exactly what we mean, such as:

```
	'EMAIL_MAX_CHUNK_SIZE'			=> 'Maximum number of recipients per email',
	'EMAIL_MAX_CHUNK_SIZE_EXPLAIN'	=> 'Controls the number of recipients allowed on any single email.  If the total list of recipients exceeds this limit, multiple emails will be sent, each with a portion of the list that is within this limit.',
```

dark-1:
@3D-I & @EA117 ,

What about just 3 words : `Email Maximum Recipients` .

Best regards 👍

3D-I:
!set php
!set Language/Text

EA117:
> What about just 3 words : `Email Maximum Recipients` .

That is certainly more intuitive than "chunk" to me too, and in combination with appropriate "explain" text, the intended meaning should be clear.  Bottom line in my mind is just that someone can walk away with the understanding that this configuration controls how many recipients we'll _use_ on any single email before breaking the send up into multiple emails.  As opposed language which could be interpreted as "the maximum email recipients you'll be allowed to specify."

mrgoldy:
"Email maximum recipients" is not correct in my opinion. As it's not the maximum of recipients. It can be way over the chunk size. I suppose it would be better to have "Maximum recpients per e-mail".

phpBB/includes/acp/acp_board.php
```diff
@@ -450,6 +450,7 @@ function main($id, $mode)
 						'email_enable'			=> array('lang' => 'ENABLE_EMAIL',			'validate' => 'bool',	'type' => 'radio:enabled_disabled', 'explain' => true),
 						'board_email_form'		=> array('lang' => 'BOARD_EMAIL_FORM',		'validate' => 'bool',	'type' => 'radio:enabled_disabled', 'explain' => true),
 						'email_package_size'	=> array('lang' => 'EMAIL_PACKAGE_SIZE',	'validate' => 'int:0',	'type' => 'number:0:99999', 'explain' => true),
+						'email_max_chunk_size'	=> array('lang' => 'EMAIL_MAX_CHUNK_SIZE',	'validate' => 'int:0',	'type' => 'number:1:99999', 'explain' => true),
```
mrgoldy: You validate the number with `min = 0`, yet the `type` (which creates the HTML) has a `min = 1 && max = 99999`. Seems inconsistent.

3D-I:
> 
> 
> "Email maximum recipients" is not correct in my opinion. As it's not the maximum of recipients. It can be way over the chunk size. I suppose it would be better to have "Maximum recpients per e-mail".

It's all explained into the ... explaination string?

phpBB/includes/acp/acp_board.php
3D-I: Thx m8 :+1:

EA117:
> ...it's all explained into the ... explaination string?

Agree that the "shorter" setting name of "Email maximum recipients" or "Maximum allowed email recipients" _can be_ sufficient, if the explanation then clarifies what we really mean.  But I'm not sure the latest explanation text fully achieves this:

```
	'EMAIL_MAX_CHUNK_SIZE'			=> 'Maximum allowed email recipients',
	'EMAIL_MAX_CHUNK_SIZE_EXPLAIN'	=> 'If necessary, set this to not exceed the maximum number of recipients that your email server will allow in one email message.',
```
We've confirmed the "why" someone might want to set this, but not what the setting actually does.  The language leaves open for interpretation what MrGoldy appears to be reading into it as well; that this setting could mean "you're not allowed to specify more than this number of phpBB mail recipients".

When what we really mean is, we're not going to _use_ more than this number of recipients -- out of whatever larger recipient list may have been specified -- and will generate additional emails, each with recipients within this limit, until the entire specified list has been emailed.

Which is what the original proposed text had been intending to make clear, not that this is the only way we could make it clear:

```
	'EMAIL_MAX_CHUNK_SIZE'			=> 'Maximum number of recipients per email',
	'EMAIL_MAX_CHUNK_SIZE_EXPLAIN'	=> 'Controls the number of recipients allowed on any single email.  If the total list of recipients exceeds this limit, multiple emails will be sent, each with a portion of the list that is within this limit.',
```

Its fine by me if we want the actual setting name to be shorter and perhaps still ambiguous as to the real meaning; so long as the actual explain text makes it clear.  I'm not scared of making the actual setting name longer too, since there is no down-side or "users will have to type in this long setting name in order to use it" or similar.  But whatever we think would be best.

I do like your intention to include the "why" someone may do this, so here is an updated proposal which includes the "why" in addition to the "how":

```
	'EMAIL_MAX_CHUNK_SIZE'			=> 'Maximum recipients to include per email',
	'EMAIL_MAX_CHUNK_SIZE_EXPLAIN'	=> 'Controls the number of recipients included on any single email, if necessary to avoid a mail server's per-email recipient limit.  If the total list of recipients exceeds this limit, multiple emails will be generated, each with a portion of the list that is within this limit.',
```

3D-I:
> If the total list of recipients exceeds this limit, multiple emails will be generated, each with a portion of the list that is within this limit.
Curios to see the part of code related to your assertion here.
The issue instead it is a PHP error in most case.

EA117:
> Curios to see the part of code related to your assertion here.

https://github.com/phpbb/phpbb/blob/9bb0275781d674f9f2321cb3494ec7dc0cbade48/phpBB/includes/acp/acp_email.php#L167

When $i reaches the $max_chunk_size (or if we need to switch to generating an email with a different user language), we stop adding recipients to the current $email_list[] row and begin building a new one.

$email_list[] being the array we eventually walk through one row at a time to call our $messenger instance at https://github.com/phpbb/phpbb/blob/9bb0275781d674f9f2321cb3494ec7dc0cbade48/phpBB/includes/acp/acp_email.php#L260

3D-I:
> The issue instead it is a PHP error in most case.

EA117:
> When using mass emails, BCC recipients.
> Which is not what this PR is for, I am just adding an UI to a native old configuration.

Well I certainly cannot rule out that I am the one who is confused.  "UI for Mass email" is literally in the name of the ticket though, and I felt like I understood this clearly.

To my assertion the code links provided above are referencing _the one and only place_ within phpBB where this currently database-only "email_max_chunk_size" configuration is being used.

https://github.com/phpbb/phpbb/blob/9bb0275781d674f9f2321cb3494ec7dc0cbade48/phpBB/includes/acp/acp_email.php#L156

Meaning this PR cannot be about anything else; because the database configuration you're making an ACP setting for is not _used_ for anything else except the acp_email.php page.

_"When using mass email, the recipients are BCC"_ is certainly true.  But those are still recipients.  They will be held to the limit configured here, and/or are potentially subject to a recipient limit imposed by the mail server.

Which is what you're trying to provide an ACP configuration for, correct?  Since that's what your current commit's explain text says, too.

3D-I:
The text is ok by my side.

PR #5652, opened 2019-08-10
[ticket/11327] Implement forgot password functionality via form

marc1706:
Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-11327


phpBB/phpbb/ucp/controller/reset_password.php
```diff
@@ -0,0 +1,436 @@
...
+		$username	= $this->request->variable('username', '', true);
+		$email		= strtolower($this->request->variable('email', ''));
+
+		add_form_key('ucp_reset_password');
+
+		if ($submit)
+		{
+			if (!check_form_key('ucp_reset_password'))
+			{
+				return $this->helper->message('FORM_INVALID');
```
mrgoldy: You're using `$helper->message()` to show an error message.
While I understand that you are in the UCP and can not throw exceptions, I still think `trigger_error()` is currently better. Will be easier to find and replace later on with proper exceptions

phpBB/phpbb/ucp/controller/reset_password.php
```diff
@@ -0,0 +1,436 @@
...
+				'email',
+				'username',
+				'sql_array',
+			];
+			extract($this->dispatcher->trigger_event('core.ucp_remind_modify_select_sql', compact($vars)));
+
+			$sql = $this->db->sql_build_query('SELECT', $sql_array);
+			$result = $this->db->sql_query_limit($sql, 2); // don't waste resources on more rows than we need
+			$rowset = $this->db->sql_fetchrowset($result);
+
```
mrgoldy: I would just free the result here, you already have the entire `$rowset` anyway.

phpBB/phpbb/ucp/controller/reset_password.php
```diff
@@ -0,0 +1,436 @@
...
+				]);
+			}
+			else
+			{
+				$message = $this->language->lang('PASSWORD_RESET_LINK_SENT') . '<br /><br />' . $this->language->lang('RETURN_INDEX', '<a href="' . append_sid("{$this->root_path}index.{$this->php_ext}") . '">', '</a>');
+
+				$user_row = empty($rowset) ? [] : $rowset[0];
+				$this->db->sql_freeresult($result);
+
+				if (!$user_row)
```
mrgoldy: I would do all this in a slightly different order.
```php
if ($rowset === false)
{
    // No account found
}

$user_row = $rowset[0];
```
Either rowset is an array which means `0` is set or it is `false`.

phpBB/phpbb/ucp/controller/reset_password.php
```diff
@@ -0,0 +1,436 @@
...
+				$reset_token = strtolower(gen_rand_string(32));
+
+				$sql_ary = [
+					'reset_token'				=> $reset_token,
+					'reset_token_expiration'	=> strtotime('+1 day'),
+				];
+
+				$sql = 'UPDATE ' . $this->tables['users'] . '
+					SET ' . $this->db->sql_build_array('UPDATE', $sql_ary) . '
+					WHERE user_id = ' . $user_row['user_id'];
```
mrgoldy: Eventhough directly retrieved from the db, might be better to typecast `(int)` as for all id's in all SQL queries

phpBB/phpbb/ucp/controller/reset_password.php
```diff
@@ -0,0 +1,436 @@
...
+				$messenger->set_addresses($user_row);
+
+				$messenger->anti_abuse_headers($this->config, $this->user);
+
+				$messenger->assign_vars([
+						'USERNAME'			=> htmlspecialchars_decode($user_row['username']),
+						'U_RESET_PASSWORD'	=> generate_board_url(true) . $this->helper->route('phpbb_ucp_reset_password_controller', [
+							'u'		=> $user_row['user_id'],
+							'token'	=> $reset_token,
+						], false)
```
mrgoldy: Why not use `$reference_type = UrlGeneratorInterface::ABSOLUTE_URL` instead of the `generate_board_url()`? 
```
    /**
     * Generates an absolute URL, e.g. "http://example.com/dir/file".
     */
    const ABSOLUTE_URL = 0;
```

phpBB/phpbb/ucp/controller/reset_password.php
```diff
@@ -0,0 +1,436 @@
...
+				$messenger->send($user_row['user_notify_type']);
+
+				return $this->helper->message($message);
+			}
+		}
+
+		$this->template->assign_vars([
+			'USERNAME'			=> $username,
+			'EMAIL'				=> $email,
+			'S_PROFILE_ACTION'	=> $this->helper->route('phpbb_ucp_forgot_password_controller'),
```
mrgoldy: Shouldn't `<form>` action attributes not just be considered as url's and thus prefixed with `U_`?

phpBB/phpbb/ucp/controller/reset_password.php
```diff
@@ -0,0 +1,436 @@
...
+			return $this->helper->message('NO_USER');
+		}
+
+		add_form_key('ucp_reset_password');
+
+		$sql_array = [
+			'SELECT'	=> 'user_id, username, user_permissions, user_email, user_jabber, user_notify_type, user_type,'
+				. ' user_lang, user_inactive_reason, reset_token, reset_token_expiration',
+			'FROM'		=> [$this->tables['users'] => 'u'],
+			'WHERE'		=> 'user_id = ' . $user_id,
```
mrgoldy: Eventhough it's already typecasted by `$request`, perhaps here again to keep SQL queries consistent?

phpBB/phpbb/ucp/controller/reset_password.php
```diff
@@ -0,0 +1,436 @@
...
+		}
+
+		if ($user_row['reset_token_expiration'] < time())
+		{
+			$this->remove_reset_token($user_id);
+
+			return $this->helper->message($message);
+		}
+
+		$error = [];
```
mrgoldy: Seeing it is an array, I would consider using `$errors` instead.

phpBB/phpbb/ucp/controller/reset_password.php
```diff
@@ -0,0 +1,436 @@
...
+				'new_password'		=> [
+					['string', false, $this->config['min_pass_chars'], $this->config['max_pass_chars']],
+					['password'],
+				],
+				'password_confirm'	=> ['string', true, $this->config['min_pass_chars'], $this->config['max_pass_chars']],
+			];
+			$error = array_merge($error, validate_data($data, $check_data));
+			if (strcmp($data['new_password'], $data['password_confirm']) !== 0)
+			{
+				$error[] = ($data['password_confirm']) ? 'NEW_PASSWORD_ERROR' : 'NEW_PASSWORD_CONFIRM_EMPTY';
```
mrgoldy: Redundant `()` ?

phpBB/phpbb/ucp/controller/reset_password.php
```diff
@@ -0,0 +1,436 @@
...
+					$user_row['username']
+				]);
+				meta_refresh(3, append_sid("{$this->root_path}index.{$this->php_ext}"));
+				return $this->helper->message($this->language->lang('PASSWORD_RESET'));
+			}
+		}
+
+		$this->template->assign_vars([
+			'S_IS_PASSWORD_RESET'	=> true,
+			'ERROR'					=> !empty($error) ? implode('<br />', array_map([$this->language, 'lang'], $error)) : '',
```
mrgoldy: I think you'll make Hanakin happy if we stop doing this in PHP.
We can just assign the array to the template and do it in twig.
```html
{% if ERROR %}
    {% for error in ERROR %}{{ error }}<br>{% endfor %}
{% endif %}
```
So eventually he can do all the styling and stuff in HTML / CSS.

phpBB/includes/functions.php
```diff
@@ -2519,11 +2519,14 @@ function login_box($redirect = '', $l_explain = '', $l_success = '', $admin = fa
...
 
+	/** @var \phpbb\controller\helper $controller_helper */
+	$controller_helper = $phpbb_container->get('controller.helper');
+
 	$login_box_template_data = array(
 		'LOGIN_ERROR'		=> $err,
 		'LOGIN_EXPLAIN'		=> $l_explain,
 
-		'U_SEND_PASSWORD' 		=> ($config['email_enable']) ? append_sid("{$phpbb_root_path}ucp.$phpEx", 'mode=sendpassword') : '',
+		'U_SEND_PASSWORD' 		=> ($config['email_enable']) ? $controller_helper->route('phpbb_ucp_forgot_password_controller') : '',
```
mrgoldy: redundant `()`?

phpBB/phpbb/ucp/controller/reset_password.php
marc1706: Symfony ends up creating this kind of URL: `//example.com/dir/file`

While perfectly valid when used as src in `<a>`, it'll most definitely cause issues when inside a mail.

phpBB/phpbb/ucp/controller/reset_password.php
```diff
@@ -0,0 +1,436 @@
...
+	/**
+	 * Init controller
+	 */
+	protected function init_controller()
+	{
+		$this->language->add_lang('ucp');
+
+		if (!$this->config['allow_password_reset'])
+		{
+			trigger_error($this->language->lang('UCP_PASSWORD_RESET_DISABLED', '<a href="mailto:' . htmlspecialchars($this->config['board_contact']) . '">', '</a>'));
```
rubencm: In controller you should throw `phpbb/exception/http_exception` if I don't remember wrong

marc1706: Hm, using the http_exception doesn't really work in this case due to the mailto link.

phpBB/phpbb/ucp/controller/reset_password.php
marc1706: I can actually already use exceptions for this as the pages are handled via app.php

phpBB/phpbb/ucp/controller/reset_password.php
marc1706: This is taken directly from the previous code but yeah, it's very inconsistent with the intended use of the S_ prefix.

phpBB/phpbb/ucp/controller/reset_password.php
marc1706: I think it's fine like this.

phpBB/includes/functions.php
marc1706: Commonly done like this and IMHO fine.

phpBB/phpbb/ucp/controller/reset_password.php
```diff
@@ -0,0 +1,442 @@
...
+	 * @param int $user_id User ID
+	 */
+	protected function remove_reset_token(int $user_id)
+	{
+		$sql_ary = [
+			'reset_token'				=> '',
+			'reset_token_expiration'	=> 0,
+		];
+
+		$sql = 'UPDATE ' . $this->tables['users'] . '
```
CHItA: It is more of a preference thing, but you only user `users` so maybe only inject that one?

phpBB/phpbb/ucp/controller/reset_password.php
```diff
@@ -0,0 +1,442 @@
...
+
+			if (empty($email))
+			{
+				return $this->helper->message('NO_EMAIL_USER');
+			}
+
+			$sql_array = [
+				'SELECT'	=> 'user_id, username, user_permissions, user_email, user_jabber, user_notify_type, user_type,'
+								. ' user_lang, user_inactive_reason, reset_token, reset_token_expiration',
+				'FROM'		=> [$this->tables['users'] => 'u'],
```
CHItA: Same as above.

phpBB/phpbb/ucp/controller/reset_password.php
```diff
@@ -0,0 +1,442 @@
...
+
+				// Generate reset token
+				$reset_token = strtolower(gen_rand_string(32));
+
+				$sql_ary = [
+					'reset_token'				=> $reset_token,
+					'reset_token_expiration'	=> strtotime('+1 day'),
+				];
+
+				$sql = 'UPDATE ' . $this->tables['users'] . '
```
CHItA: Same as above.

phpBB/phpbb/ucp/controller/reset_password.php
```diff
@@ -0,0 +1,442 @@
...
+		{
+			return $this->helper->message('NO_USER');
+		}
+
+		add_form_key('ucp_reset_password');
+
+		$sql_array = [
+			'SELECT'	=> 'user_id, username, user_permissions, user_email, user_jabber, user_notify_type, user_type,'
+				. ' user_lang, user_inactive_reason, reset_token, reset_token_expiration',
+			'FROM'		=> [$this->tables['users'] => 'u'],
```
CHItA: Same as above.

phpBB/language/en/ucp.php
```diff
@@ -415,7 +416,8 @@
 	'PASS_TYPE_SYMBOL_EXPLAIN'	=> 'Password must be between %1$s and %2$s long, must contain letters in mixed case, must contain numbers and must contain symbols.',
 	'PASSWORD'					=> 'Password',
 	'PASSWORD_ACTIVATED'		=> 'Your new password has been activated.',
-	'PASSWORD_UPDATED_IF_EXISTED'	=> 'If your account exists, a new password was sent to your registered email address. If you do not receive an email, it may be because you are banned, your account is not activated, or you are not allowed to change your password. Contact admin if any of those reasons apply. Also, check your spam filter.',
+	'PASSWORD_RESET'			=> 'Your password has been successfully reset.',
+	'PASSWORD_RESET_LINK_SENT'	=> 'If your account exists, a password reset link was sent to your registered email address. If you do not receive an email, it may be because you are banned, your account is not activated, you have requested multiple password resets within a short time frame, or you are not allowed to change your password. Contact an admin if any of those reasons apply. Also, please check your spam filter.',
```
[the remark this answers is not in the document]
marc1706: We switch around but I guess I can change it to administrator

phpBB/phpbb/ucp/controller/reset_password.php
```diff
@@ -0,0 +1,442 @@
...
+				}
+
+				// Do not create multiple valid reset tokens
+				if (!empty($user_row['reset_token']) && (int) $user_row['reset_token_expiration'] >= time())
+				{
+					return $this->helper->message($message);
+				}
+
+				// Check users permissions
+				$auth2 = new auth();
```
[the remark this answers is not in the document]
marc1706: This stems from when we used to have a global $auth also defined. Removed the appended 2 as it's not needed inside a class method.

phpBB/phpbb/ucp/controller/reset_password.php
```diff
@@ -0,0 +1,443 @@
...
+	 * @param template $template
+	 * @param user $user
+	 * @param string $users_table
+	 * @param string $root_path
+	 * @param string $php_ext
+	 */
+	public function __construct(config $config, driver_interface $db, dispatcher $dispatcher, helper $helper,
+								language $language, log_interface $log, manager $passwords_manager,
+								request_interface $request, template $template, user $user, string $users_table,
+								string $root_path, string $php_ext)
```
mrgoldy: will the `string` type declaration work for all php versions supported?

marc1706: That's supported since PHP 7.0 so yes.

PR #5653, opened 2019-08-11
[ticket/16129] Enhance attachment's ALT tag

3D-I:
PHPBB3-16129

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16129

### DIFF
--- a/phpBB/styles/prosilver/template/attachment.html
+++ b/phpBB/styles/prosilver/template/attachment.html
@@ -8,14 +8,14 @@
 
 		<!-- IF _file.S_THUMBNAIL -->
 		<dl class="thumbnail">
-			<dt><a href="{_file.U_DOWNLOAD_LINK}"><img src="{_file.THUMB_IMAGE}" class="postimage" alt="{_file.DOWNLOAD_NAME}" title="{_file.DOWNLOAD_NAME} ({_file.FILESIZE} {_file.SIZE_LANG}) {_file.L_DOWNLOAD_COUNT}" /></a></dt>
+			<dt><a href="{_file.U_DOWNLOAD_LINK}"><img src="{_file.THUMB_IMAGE}" class="postimage" alt="{% if _file.COMMENT %}{{ _file.COMMENT }}{% else %}{{ _file.DOWNLOAD_NAME }}{% endif %}" title="{_file.DOWNLOAD_NAME} ({_file.FILESIZE} {_file.SIZE_LANG}) {_file.L_DOWNLOAD_COUNT}" /></a></dt>
 			<!-- IF _file.COMMENT --><dd> {_file.COMMENT}</dd><!-- ENDIF -->
 		</dl>
 		<!-- ENDIF -->
 
 		<!-- IF _file.S_IMAGE -->
 		<dl class="file">
-			<dt class="attach-image"><img src="{_file.U_INLINE_LINK}" class="postimage" alt="{_file.DOWNLOAD_NAME}" onclick="viewableArea(this);" /></dt>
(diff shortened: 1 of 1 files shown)


3D-I:
!set Styles

phpBB/styles/prosilver/template/attachment.html
```diff
@@ -8,14 +8,14 @@
 
 		<!-- IF _file.S_THUMBNAIL -->
 		<dl class="thumbnail">
-			<dt><a href="{_file.U_DOWNLOAD_LINK}"><img src="{_file.THUMB_IMAGE}" class="postimage" alt="{_file.DOWNLOAD_NAME}" title="{_file.DOWNLOAD_NAME} ({_file.FILESIZE} {_file.SIZE_LANG}) {_file.L_DOWNLOAD_COUNT}" /></a></dt>
+			<dt><a href="{_file.U_DOWNLOAD_LINK}"><img src="{_file.THUMB_IMAGE}" class="postimage" alt="{% if _file.COMMENT %}{{ _file.COMMENT }}{% else %}{{ _file.DOWNLOAD_NAME }}{% endif %}" title="{_file.DOWNLOAD_NAME} ({_file.FILESIZE} {_file.SIZE_LANG}) {_file.L_DOWNLOAD_COUNT}" /></a></dt>
```
VSEphpbb: Alternatively you can use ternary syntax:
`{{ _file.COMMENT ?: _file.DOWNLOAD_NAME }}`

mrgoldy: I believe Hanakin is not too fond of that though.

DavidIQ: What happens if there is a quote in the name or comment? Isn't there something else needed here for escaping?

3D-I: @DavidIQ I will test it and amend the code if necessary.

3D-I: @VSEphpbb Yes but the main reason for this is code's readability, in any case let's see what @hanakin says about that.

3D-I: @DavidIQ it is ok.

immagine

DavidIQ: That just looks like it's rendering out the quote inside of the alt property making the markup invalid.

3D-I: Hmm, I commented with 

immagine

VSEphpbb: Needs to use an escape, like `_file.COMMENT|e('html')`

3D-I: Just on the comment side?

VSEphpbb: yes. File download name's shouldn't have html chars in them

3D-I: How the screen reader will "read" this?

immagine

VSEphpbb: That's proper HTML markup now.

3D-I: Ok

3D-I:
Thx guys. :+1:

phpBB/styles/prosilver/template/attachment.html
hanakin: Please no ternary in template files

marc1706:
While the escaped characters might not "sound" great when using a screenreader, I think they're fine. In essence one can only hope that users will set a proper comment for the files they upload.

PR #5656, opened 2019-08-13
[ticket/16019] Deny prosilver's uninstallation

3D-I:
PHPBB3-16019

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16019


3D-I:
!set php

3D-I:
!set Language/Text

phpBB/includes/acp/acp_styles.php
```diff
@@ -259,6 +259,19 @@ protected function action_uninstall()
...
+		$sql = 'SELECT style_id
+			FROM ' . STYLES_TABLE . "
+			WHERE style_name = '" . $this->db->sql_escape('prosilver') . "'";
+		$result = $this->db->sql_query($sql);
+		$prosilver_id = (int) $this->db->sql_fetchrow($result);
+		$this->db->sql_freeresult($result);
+
+		if (in_array($prosilver_id, $ids))
+		{
+			trigger_error($this->user->lang['UNINSTALL_PROSILVER'] . adm_back_link($this->u_action), E_USER_WARNING);
```
rxu: it would be probably better to use  ```language->lang()``` method instead of ```user->lang``` array for the sake of 3.3/4.0 compatibility.

3D-I: Yes, but the language object is not in use there, still.

I believe this bit should be adressed on another PR if not already exists, like "moving *CPs to controller".

phpBB/includes/acp/acp_styles.php
```diff
@@ -259,6 +259,19 @@ protected function action_uninstall()
 		// Get list of styles to uninstall
 		$ids = $this->request_vars('id', 0, true);
 
+		// Don't remove prosilver, you can still deactivate it.
+		$sql = 'SELECT style_id
+			FROM ' . STYLES_TABLE . "
+			WHERE style_name = '" . $this->db->sql_escape('prosilver') . "'";
+		$result = $this->db->sql_query($sql);
+		$prosilver_id = (int) $this->db->sql_fetchrow($result);
```
rxu: Does it actually return the  ```style_id```? I think ```sql_fetchrow``` returns an array. But ```sql_fetchfield``` would return a value directly 😉

3D-I: Indeed, same here.. that's old code like

```
		// Check if style has child styles
		$sql = 'SELECT style_id
			FROM ' . STYLES_TABLE . '
			WHERE style_parent_id = ' . (int) $id . " OR style_parent_tree = '" . $this->db->sql_escape($path) . "'";

		$conflict = $this->db->sql_fetchrow($result);
```

I will use your suggestion though, just is a bit late here.

phpBB/includes/acp/acp_styles.php
rxu: Okay, than probably ```user->lang()``` method instead of ```user->lang[]``` array?

3D-I: It's the same as used everywhere there.

3D-I: see the function below

rxu: @3D-I I see, but the common sense is using new code in new PRs,. But whatever you like.

3D-I: Right, let me work on this one a bit more, thx m8.

3D-I:
@rxu what do you think about now?

rxu:
Looks good.
Okay, now it works like: in UI user has an option to select prosilver for uninstallation (by uninstall direct link or a checkbox), but then is getting error message that it can't be done.
Thus, the suggestion is to get rid of uninstall direct link and corresponding checkbox in prosilver row additionally.

3D-I:
> Thus, the suggestion is to get rid of uninstall direct link and corresponding checkbox in prosilver row additionally.

Yeah, I thought about that. Let me see if I can do it now or wait for tomorrow :smiley_cat:

3D-I:
immagine

Now to find the checkbox :smiley_cat:

EA117:
For consideration:  Everything in the current commit is great, and should stay.  There are multiple paths for invoking the uninstall, and keeping a failure in the uninstall function (same as when they try to uninstall the default style) is appropriate.

But acp_styles.php's list_style() function is already building an $actions[ ] array, and is already applying conditional logic for what actions should be offered for a given style in the styles list.

Wouldn't it be "easy" to apply our new "detect prosilver" logic to the building of the $actions[ ] array, and skip even _offering_ the uninstall option in this circumstance where we don't intend to allow uninstall?  i.e. An extra visual cue of "you can't uninstall prosilver."

The additional logic wouldn't even need to be SQL-based like in the current commit.  Detecting prosilver when building the $actions[ ] array would simply be a matter of PHP string compare against $style['style_name'].

I would say arguably the same could be done for the current default style, too.  But to be fair, which style is default isn't obvious when viewing the styles list.  So that doesn't need to be part of this ticket or PR.  Which style is prosilver, on the other hand, is obvious.  So if we didn't want people to uninstall it, "why did we offer the uninstall option"?

3D-I:
Please see the above progress, I am working already on the  list_style() function.
Perhaps in the meantime you wrote a so long comment ;)

3D-I:
immagine

@rxu does it all looks good to you?

EA117:
Thanks @3D-I 😝  Whether it was the long message itself, or the time I spent inspecting code before making sure I was correct in hitting Comment, I did miss the further activity.  Great that it was already noted and taken up for further enhancement.

3D-I:
No probs m8 :+1:

phpBB/includes/acp/acp_styles.php
```diff
@@ -259,6 +259,22 @@ protected function action_uninstall()
...
 
+		// Don't remove prosilver, you can still deactivate it.
+		$sql = 'SELECT style_id
+			FROM ' . STYLES_TABLE . "
+			WHERE style_name = '" . $this->db->sql_escape('prosilver') . "'";
+		$result = $this->db->sql_query($sql);
+		$prosilver_id = (int) $this->db->sql_fetchfield('style_id');
+		$this->db->sql_freeresult($result);
+
+		if ($prosilver_id)
```
rxu: To avoid nested conditions, maybe just
```php
if ($prosilver_id && in_array($prosilver_id, $ids))
```

3D-I:
This PR completes this other one BTW: https://github.com/phpbb/phpbb/pull/5654

phpBB/includes/acp/acp_styles.php
3D-I: Oh yes, I like it ;)

rxu:
And just for common discussion: does it make any sense to deny uninstalling prosilver only in case it has inheriting styles?
Is it technically possible to install a style inheriting from prosilver while prosilver is uninstalled and deleted from the board?

3D-I:
> Is it technically possible to install a style inheriting from prosilver while prosilver is uninstalled and deleted from the board?

Not if you deleted the files. AFAIK.

EA117:
If there are child styles currently installed, phpBB will not let you uninstall prosilver.  It's gives a green "Information" message of "Style "prosilver" cannot be uninstalled because it has one or more child styles."

If you successfully uninstall prosilver from the styles list (but do not delete the files), phpBB will not let you install any style which cites "prosilver" as it's parent style.  The "Install styles" page simply doesn't offer the installation option, as shown here:

styles

Agree that once you have deleted from the file system, prosilver isn't even offered in the "Install styles" section any more, so there would be no way to install a child style at that point, either.

If prosilver is simply "disabled"  rather than uninstalled, you can still install styles which cite "prosilver" as their parent style.  Only "uninstall" prevents installing prosilver-dependent styles.

rxu:
So it seems that all possible cases got handled in the current code, why are we going to deny uninstalling prosilver at all then?

EA117:
rxu, asking the real questions. 😁

From a power user or expert perspective, indeed, there is probably no reason to disallow removal of prosilver.  Someone who knows what they're doing, and knows they can remove it, can certainly do so without "doing anything wrong."

From a "supportability" standpoint of phpBB at large, I understand the desire to "prevent users from thinking that uninstalling prosilver is supposed to be a good idea."  We ask people to "test and compare against prosilver" frequently, which wouldn't be possible if they "cleaned up their styles list" to no longer include prosilver, "just because they could."

There is a "delete from file system" checkbox easily offered to them when they uninstall prosilver, which then makes it harder for the novice user to undo this action which was otherwise easy to invoke.  And specifically for prosilver, "you're going to get this style back during the next update anyway", so it's been largely "all short-term risk and no long-term reward" by offering its removal.

But, @rxu, your question does perhaps bring a different alternative to mind.  What if we detected "prosilver" and _only_ removed the "delete from file system" option on the uninstall screen?

So you "can" uninstall prosilver, but phpBB will prevent you from doing that unless (a) another style is already default, and (b) none of the installed styles cite prosilver as the parent.

But if and when you _do_ uninstall prosilver, the change we make here "guarantees" that prosilver remains listed on the "Install styles" page, such that the action can be "undone" when necessary.  i.e. No more option to "delete from filesystem" for prosilver.

Leaving the only way prosilver could disappear completely is because someone deleted it via FTP, and not because of an option that phpBB ACP itself offered.

@3D-I If the "delete from filesystem" option was suppressed for prosilver, what is your opinion of the existing checks that phpBB already performs?  Are they enough that we could rest assured "someone who uninstalled prosilver must have been in a configuration where this was an option", else phpBB would have prevented it?  Or do you still see a reason why it would be important to prevent "literally everyone" from being able to invoke uninstall on prosilver, even if "they know what they're doing?"

3D-I:
I'd like to know the opposite, and that's why not?

However, to drastically reduce support requests for "prosilver", as should be done for the "en" language somehow, not investigated yet. Support questions of any kind I mean.

The PHP part seems to me to be worthily covered, the change made for the checkbox in HTML is still covered by what was done in PHP. It means that even if an obsolete style made you see the checkbox anyway you will find yourself in front of the error programmed in PHP.

If you use a direct link won't work because the style's ID is hashed. AFAIK.

marc1706:
If someone really knows what they are doing, they can still remove prosilver. I think it's good to prevent admins from doing that as it would prevent a lot of potential issues we have faced where a test with prosilver could have easily helped. prosilver is essentially the fallback solution if everything else goes wrong so it should not be removed.
The point is also that while one would think that only people knowing what they're doing would remove prosilver, it's usually the people that are not aware of why keeping prosilver might make sense opt for removing it just to clean up.

rxu:
Then it probably makes sense to extend the ```ACP_STYLES_EXPLAIN``` entry to state the point that prosilver can't be uninstalled for the sake of the board operation stability or something like that.

EA117:
Fair enough; it's fine to proceed on the original course from my perspective, too.

The "only remove the delete from file system" approach (and then simply let existing phpBB checks handle all other situations) was just realized as "a minimum change which still keeps us out of the supportability problem we're trying to avoid."  A more extensive change intended to meet this same goal -- like the approach 3-DI was already on -- is fine too.

@3D-I Agree with your assessment of the checkbox scenario.  It will get complicated to try and make a "logical" behavior for the checkbox options, and that is probably not worth our time.  As you said, if someone uses the checkbox, simply let the failure message happen for that case.  Rather than trying to "prevent them" from pressing the Uninstall button when prosilver is included in the checkbox-marked styles.

Having successfully removed the "Uninstall" link that appears on the style row itself I think communicates enough visually about the intentions here.

3D-I:
Ok, I think I am done here. Thank you all.

3D-I:
> 
> 
> Then it probably makes sense to additionally extend the `ACP_STYLES_EXPLAIN` entry to state the point that prosilver can't be uninstalled for the sake of the board operation stability or something like that.

Surely it makes sense, the problem is that I am not a native speaker and to open a long and tedious debate on the words to use seems excessive for this PR. :smile:

rxu:
We can poke then some native speakers to dig in and help 😉

3D-I:
`ACP_STYLES_EXPLAIN` should be slightly rewritten IMO, I would prefer it to be done in a separate PR, unless I am quickly given the final text then I would push it into this, an approved text in short.

This is the present text
```
Here you can manage the available styles on your board. You may alter existing styles, delete, deactivate, reactivate, install new ones. You can also see what a style will look like using the preview function. Also listed is the total user count for each style, note that overriding user styles will not be reflected here.
```

rxu:
> Seems a remain of Olympus

Indeed.
I'd go with something like that

```
Here you can manage the available styles on your board. You may uninstall, deactivate, reactivate existing styles, install new ones. You can also see what a style will look like using the preview function. Also listed is the total user count for each style, note that overriding user styles will not be reflected here.
Note that in order to provide the proper board operation prosilver style can not be uninstalled.
```

3D-I:
`Here you can manage the styles available on your board. You can delete, deactivate, activate or reactivate them. Or install new ones. Please note that for obvious reasons you cannot delete the "prosilver" style completely. You can also see what a style will look like by using the preview function. Also listed is the total user count for each style, note that overwritten user styles will not be reflected here.
`

That's what I came up with.

rxu:
@VSEphpbb @DavidIQ @marc1706 
Would be nice of you to review ```ACP_STYLES_EXPLAIN``` suggested changes for the language correctness if possible 😃

3D-I:
That's what it looks like

immagine

EA117:
I don't think there are any English issues with the text.  I do think we could do better simply in terms of "user documentation" though.  A statement like "for obvious reasons" just stands out as "that's not going to be true for the people reading it."  So I would say either tell the reasons (which would take too much room), or don't mention them and simply declare what the behavior is.

This is how I would re-word the entire existing text, with our new statement added:

> Here you can install, activate, deactivate, uninstall or delete styles on your board.
> You can also preview a style, see the number of users who are using a specific style,
> or view and manage other details about each style.  Note that the "prosilver" style
> included with phpBB cannot be uninstalled or deleted, but can be deactivated
> if necessary.

For your and/or david / vse / marc's consideration.

3D-I:
It says `for obvious reasons like provide proper board operation`, so at least the main reason is given.

The fact that can be de-activated is pretty clear since its link is there.

3D-I:
!set Styles

VSEphpbb:
The English was a little jacked and stating the obvious. You don't need a paragraph explaining what is clearly in front of us. This is all it needs to say:

> Here you can manage the styles available on your board.  Please note you cannot uninstall the “prosilver” style as it is phpBB's default and primary parent style.

rxu:
Nice. Short and clear.

EA117:
If we're saying this paragraph isn't necessary, then I wouldn't see why we need more than VSE's "Here you can manage the styles available on your board."  Full stop.

Because uninstall of prosilver is just one more example of explaining something like the previous paragraph did, which "should already be clear" about the page.  Since there will no longer be an "uninstall" link next to prosilver to click on.

I'm not sure that "we" are the audience of this paragraph, or of the similar explanations on ACP General, ACP Forums, etc.?  They feel like introductions that existed on the page for users who are seeing them for the first time.

Fine if we're saying "but we're going to stop doing that", and just go with a terse confirmation of "you're on the Styles page."

3D-I:
@EA117 

I would say that the discussions on how to do this or that in the future find their logical place for a debate at area51. In fact, that's what I wanted to avoid happening here, as I said before.

There is also a guide already exhaustive enough that could be modified if desired.
https://www.phpbb.com/support/docs/en/3.2/ug/adminguide/customise_styles/

Thank you.

3D-I:
immagine

EA117:
@3D-I Absolutely agree that this is exactly what you had already stated you wanted to avoid.  I can understand not wanting to do it, but there is still a change being introduced by this PR.  Not that I'm in any way "an approver" or a longer-term stakeholder than anyone else; but I don't feel out of place discussing concerns or issues with the changes being initiated here, either.

If you're saying "discussion is closed" and keeping the short explanation shown in the current commit, then of course we'll have to live with that, assuming approval.

But I think we still have one remaining issue, even if we set aside ALL of the previous points.  This was going to be a moot point if we had gone with only the short or long re-writes; but now that it looks like we're keeping the current commit, I'm raising it:

In addition to making the short statement _"Here you can manage the styles available on your board"_, we have also changed & included the following statement, _"Overwritten user styles will not be reflected here."_

Recommend we return this to it's original English statement, _"Note that overriding user styles will not be reflected here."_  Or, given the language we have removed from the overall paragraph, perhaps this now needs to be _"Note that overriding user styles will not be reflected in the user counts shown here."_  Since the prefix that mentioned the context of "user counts" has been removed.

This sentence is in reference to the _"Override user style: [ ] Yes [ ] No"_ configuration that exists in ACP Board Settings.  "Overwritten" is not the correct word to use in referencing this setting, and creates ambiguity as to what kind of "overwriting" we might actually be referring to.

Thanks for considering.

3D-I:
immagine

PR #5658, opened 2019-08-14
[ticket/13175] Check if account is already linked when using OAuth

rubencm:
PHPBB3-13175

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-13175


3D-I:
Wouldn't it be better to allow multiusers with some option?
We did it in an Oauth2 extension already.

rubencm:
Well, I was seeing how oauth works a little in depth, and I just realized this error.

About the extension, which extension is, to see how it is implemented? If is linkedaccounts, that looks to me like a completely new (but related) feature.
Create a dialog asking with which account login is not a bad idea

marc1706:
Yes, I think the initial implementation did not expect users to link more than one account. It's probably best to create new ticket for the possibility to have more than one account and implement that later.

3D-I:
@rubencm Is an ext of us, not public. And yes I do agree with @marc1706 here.

PR #5663, opened 2019-08-22
[ticket/16138] Eliminate redundant parameters from board URLs

v12mike:
Most internal board urls had redundant f= or t= paramters.
This change eliminates redundant paramters from urls generated
by phpBB, and any such parameters encountered in incoming URLs
will be ignored.

Some minor refactoring of top-level scripts (especially posting.php)
was done to ensure that only necessary URL parameters are used.

Test cases have been updated to remove the redundent url paramters.

PHPBB3-16138

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16138


Crizz0:
Are the old 3.0/3.1/3.2 URLs still working, but not yet used any more?

v12mike:
Yes, the old links still work, but the redundant parameters are ignored.  
I have been testing the main part of this (viewforum.php, viewtopic.php, search.php etc) on a live forum for several days, and have not had any reports of dead links.

dark-1:
@v12mike ,

you need to change :
`https://tracker.phpbb.com/browse/PHPBB3-12345`
to :
`https://tracker.phpbb.com/browse/PHPBB3-16138`
...

dark-1:
@v12mike , 
you need to fix your email ID in your commits with the one on Github.

Best regards 👍

v12mike:
> @v12mike ,
> you need to fix your email ID in your commits with the one on Github.
> 
> Best regards 👍

I don't understand.  Can you explain that a bit more?  
Which system is being upset by this? 
Do I need to redo my commits?

phpBB/includes/functions_mcp.php
```diff
@@ -51,8 +51,7 @@ function phpbb_module_warn_url($mode, $module_row)
 	{
 		global $forum_id, $post_id;
 
-		$url_extra = ($forum_id) ? "&amp;f=$forum_id" : '';
-		$url_extra .= ($post_id) ? "&amp;p=$post_id" : '';
+		$url_extra .= ($post_id) ? "&amp;p=$post_id" : (($forum_id) ? "&amp;f=$forum_id" : '');
```
rubencm: Better use conditional blocks instead of nested ternary operators to make it more readable.

Also, php [behaves weird](http://www.phpsadness.com/sad/30) when there are nested ternary operators, I wouldn't use them besides the parenthesis

v12mike: OK, will do.

3D-I:
`$url_extra .= $post_id ? "&amp;p=$post_id" : ($forum_id ? "&amp;f=$forum_id" : '');`

phpBB/includes/functions_mcp.php
3D-I: `$url_extra .= $post_id ? "&amp;p=$post_id" : ($forum_id ? "&amp;f=$forum_id" : '');` ?

mrgoldy:
Aren't those parameters also to save on SQL queries?

v12mike:
> Aren't those parameters also to save on SQL queries?

It may be that it was the intention to save sql queries, but it does not work that way because url parameters are not trustworthy, and so need to be checked against any corresponding values in the database anyway.

In some scripts, e.g the previous version of posting.php line 265, the code had already been changed to reduce the reliance on url parameters, but this had not been done systematically.

dark-1:
> Aren't those parameters also to save on SQL queries?



> Aren't those parameters also to save on SQL queries?

This is also my point of view,
coz I have seen in many places in 3.2.x that it uses `f` & `t` url queries to store & use respective ID's ,
if those queries are not found then only make extra 1 or 2 SQL queries to get them from post ID or topic ID.
This saves few SQL queries , everytime a page loads...

Best regards 👍

v12mike:
> > Aren't those parameters also to save on SQL queries?
> 
> > Aren't those parameters also to save on SQL queries?
> 
> This is also my point of view,
> coz I have seen in many places in 3.2.x that it uses `f` & `t` url queries to store & use respective ID's ,
> if those queries are not found then only make extra 1 or 2 SQL queries to get them from post ID or topic ID.
> This saves few SQL queries , everytime a page loads...
> 
> Best regards 👍

Can you give an example for a common use-case (like viewing a topic or posting) where a query for finding the forum id from the database was not already done regardless of the supplied parameters?  I don't think that there were many.

dark-1:
@v12mike ,

Like Here : 
https://github.com/phpbb/phpbb/blob/3.2.x/phpBB/viewtopic.php#L72
the SQL query in that `if bock` will not be executed if `...&f=...` is given in URL..

Also I still don't get your security concerns regarding this...
In `.../viewtopic.php` , if `...&f=...` is give any random value , then it will either display a Topic or give an Error , depending on flow of code... But will not do any harm to DB or anything else...

Best regards 👍

v12mike:
Removing the redundant url parameters is for me a matter of good programming practice.  I was taught that user input should never be trusted and should always be validated.  Ignoring unnecessary parameters is safer and more efficient than validating them.

Aside form that, one of my motivations for making this change is to make the internal urls generated by phpBB to be (as far as practicable)  consistent with the canonical urls of the relevant pages.  That essentially means removing the unnecessary parameters.

I just checked the timings for the path that was indicated above (http://forum.com/viewtopic.php?t=6&view=unread), and on my production board, the sql query time for the extra query is 0.00016s.  Note that this (specific) extra query is only run in the case where the first unread post in the topic is selected.

phpBB/includes/functions_posting.php
```diff
@@ -2490,7 +2490,7 @@ function submit_post($mode, $subject, $username, $topic_type, &$poll_ary, &$data
 *				- 'topic_last_poster_name'
 *				- 'topic_last_poster_colour'
 * @param int $bump_time The time at which topic was bumped, usually it is a current time as obtained via time().
-* @return string An URL to the bumped topic, example: ./viewtopic.php?forum_id=1&amptopic_id=2&ampp=3#p3
+* @return string An URL to the bumped topic, example: ./viewtopic.php?amptopic_id=2&ampp=3#p3
```
Louis7777: ./viewtopic.php?topic_id=2&amp;p=3#p3

v12mike: Actually, to be consistent with the new code, it should be:
./viewtopic.php?p=3#p3

phpBB/viewtopic.php
```diff
@@ -142,9 +140,10 @@
 		}
 		else
 		{
+			$forum_id = $row['forum_id'];
 			$sql = 'SELECT topic_id, forum_id
 				FROM ' . TOPICS_TABLE . '
-				WHERE forum_id = ' . $row['forum_id'] . "
+				WHERE forum_id = ' . $forum_id . "
```
3D-I: type casting:

`WHERE forum_id = ' . (int) $forum_id . '`

same applies to any occurance here.

v12mike: Just for clarity, are you asking for all cases of (int) php variables in sql expressions to be cast?  Including those not otherwise touched by this PR?

3D-I: That's up to you, at least yours.

v12mike:
Closing this PR as I have found some issues which require further work.  I will re-open when I have sorted them.

v12mike:
Added new commit to ensure that all integer url parameters are cast to (int) before use.

phpBB/includes/acp/acp_permission_roles.php
```diff
@@ -48,7 +48,7 @@ function main($id, $mode)
 		$this->tpl_name = 'acp_permission_roles';
 
 		$submit = (isset($_POST['submit'])) ? true : false;
-		$role_id = $request->variable('role_id', 0);
+		$role_id = (int) $request->variable('role_id', 0);
```
3D-I: No need to type cast requests since them all pass via the `type_cast_helper` class.

In this specific case `, 0` defines INT, `, '0'` defines STRING instead.

You could see it using a simple test file to run from the root of your devs.

```
<?php
define('IN_PHPBB', true);
$phpbb_root_path = (defined('PHPBB_ROOT_PATH')) ? PHPBB_ROOT_PATH : './';
$phpEx = substr(strrchr(__FILE__, '.'), 1);
include($phpbb_root_path . 'common.' . $phpEx);

global $request;

$test = $request->variable('blah', 0);

var_dump($test);

```

v12mike:
Will revert

v12mike:
I have reverted the unnecessary casting that I had added.

3D-I:
> 
> 
> I have reverted the unnecessary casting that I had added.

I am not quite sure Mike, I am seeing the same stuffs.

v12mike:
> > I have reverted the unnecessary casting that I had added.
> 
> I am not quite sure Mike, I am seeing the same stuffs.

Sorry, I had pushed my changes while the PR was closed, I just made a forced push of the same stuff, and it seems to have updated the PR correctly.

mrgoldy:
I would like to raise my concern about the amended MCP url's.
With the current set up, some modules take their permission (and display) depending on the forum identifier parameter. So that might change by removing them, labeling them as 'redundant'.

While I do agree that user input should never be trusted, but if that leads to not working pages and/or modules, then that's on them. If if it's by following link created by the core code, than that specific code should be adjusted.

v12mike:
> I would like to raise my concern about the amended MCP url's.
> With the current set up, some modules take their permission (and display) depending on the forum identifier parameter. So that might change by removing them, labeling them as 'redundant'.

With the updated code, the permissions are still taken from the forum ID, but the forum ID is taken from the database, which can be assumed to be 100% correct.  There is (as far as I know) no case where normal forum behaviour is different after this change.
> 
> While I do agree that user input should never be trusted, but if that leads to not working pages and/or modules, then that's on them. If if it's by following link created by the core code, than that specific code should be adjusted.

The problem is not that the forum has been producing wrong urls, but that the urls with redundant parameters are confusing for search engines.  The forums have not required the presence of the "redundant" t= and f= parameters for some time.  The few cases where the "redundant" parameters could be used to influence the behaviour were (I am sure) accidental due to the evolution of the software over time.

I note that in the months that my forums have been running with this change, Google search index has been slowly forgetting the previous "canonical" urls with the redundant parameters, and the number of index stats more closely match the actual number of pages in the forum (as supplied by my sitemap).

v12mike:
I started this PR (I think) before the 3.3.x branch was made.  
Is it appropriate for me to close this PR, rebase to the current 3.3.x and open a new PR against it?

v12mike:
This chnge is now on PR https://github.com/phpbb/phpbb/pull/5760

PR #5664, opened 2019-08-25
[ticket/16076] addFileFilter to check max file size per mime type

mrgoldy:
PHPBB3-16076

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16076

### DIFF
--- a/phpBB/assets/javascript/plupload.js
+++ b/phpBB/assets/javascript/plupload.js
@@ -449,13 +449,13 @@ phpbb.plupload.initialize();
 /**
  * Add a file filter to check for max file sizes per mime type.
  */
-plupload.addFileFilter('mime_types_max_file_size', function(types, file, cb) {
+plupload.addFileFilter('mime_types_max_file_size', function(types, file, callback) {
 	if (file.size !== 'undefined') {
 		$(types).each(function(i, type) {
 			let extensions = [],
-				exts_array = type.extensions.split(',');
+				extsArray = type.extensions.split(',');
 
-			$(exts_array).each(function(i, extension) {
+			$(extsArray).each(function(i, extension) {
 				/^\s*\*\s*$/.test(extension) ? extensions.push("\\.*") : extensions.push("\\." + extension.replace(new RegExp("[" + "/^$.*+?|()[]{}\\".replace(/./g, "\\$&") + "]", "g"), "\\$&"));
 			});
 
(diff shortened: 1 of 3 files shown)


EA117:
What will be the intended relationship between the per-Attachment Extension Group "Maximum file size" and the main ACP Attachment Settings "Maximum File Size"?

If the per-group value is set to zero, does this mean "I want to be limited by whatever the main ACP Attachment Settings maximum file size is set to"?  Or is setting a per-group value to zero an assertion that "even though other downloads may be limited, this particular Attachment Extension Group needs to be unlimited"?

The proposed change hasn't stopped passing _"max_file_size: '{FILESIZE}b'"_ in the plupload config, so I'm just trying to picture what the actual effect, interaction and intention will be.

mrgoldy:
All files are limited to the general "maximum file size".
Per group you can lower the maximum file size.

The general one, is the 'overall' rule for attachment filesize, each and every attachment has to adhere to. Otherwise you will start working 'exceptions' and then you lose track rather quickly.

This also simply means, setting a group's maximum filesize to 0, that a group does not specific maximum filesize, other than the overall one.

The provides a linear way of thinking / managing: 
"overall" is the absolute limit and you can lower them per group.

phpBB/assets/javascript/plupload.js
```diff
@@ -446,6 +446,44 @@ phpbb.plupload.fileError = function(file, error) {
...
 phpbb.plupload.initialize();
 
+/**
+ * Add a file filter to check for max file sizes per mime type.
+ */
+plupload.addFileFilter('mime_types_max_file_size', function(types, file, cb) {
+	if (file.size !== 'undefined') {
+		$(types).each(function(i, type) {
+			let extensions = [],
+				exts_array = type.extensions.split(',');
```
marc1706: We're using camelCase in JS
```suggestion
				extsArray = type.extensions.split(',');
```

phpBB/assets/javascript/plupload.js
```diff
@@ -446,6 +446,44 @@ phpbb.plupload.fileError = function(file, error) {
 phpbb.plupload.uploader = new plupload.Uploader(phpbb.plupload.config);
 phpbb.plupload.initialize();
 
+/**
+ * Add a file filter to check for max file sizes per mime type.
+ */
+plupload.addFileFilter('mime_types_max_file_size', function(types, file, cb) {
```
marc1706:
```suggestion
plupload.addFileFilter('mime_types_max_file_size', function(types, file, callback) {
```
Please rename all occurences to `callback`. There is just no point in calling it `cb` unless one wants to reduce readability.

PR #5667, opened 2019-08-27
[ticket/16141] plupload chunk_size incorrect when 'unlimited' is involved.

EA117:
PHPBB3-16141

Change get_chunk_size() calculation to correctly calculate limits without
letting a zero "unlimited" value always win.  Also ensure get_chunk_size()
can only return zero if all of the limits were in fact set to unlimited.

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16141

### DIFF
--- a/phpBB/phpbb/plupload/plupload.php
+++ b/phpBB/phpbb/plupload/plupload.php
@@ -314,7 +314,7 @@ class plupload
 		{
 			$max = min($limit_post, ($max ? $max : $limit_post));
 		}
-		
+
 		// $config['max_filesize'] is not a limiter to chunk size.
 
 		return floor($max / 2);
--- a/phpBB/phpbb/plupload/plupload.php
+++ b/phpBB/phpbb/plupload/plupload.php
@@ -276,46 +276,36 @@ class plupload
 	}
 
 	/**
-	* Checks various php.ini values to determine the maximum chunk
-	* size a file should be split into for upload.
-	*
-	* The intention is to calculate a value which reflects whatever
-	* the most restrictive limit is set to.  And to then set the chunk
-	* size to half that value, to ensure any required transfer overhead
-	* and POST data remains well within the limit.  Or, if all of the
-	* limits are set to unlimited, the chunk size will also be unlimited.
-	*
-	* @return int
-	*/
+	 * Checks various php.ini values to determine the maximum chunk
+	 * size a file should be split into for upload.
+	 *
+	 * The intention is to calculate a value which reflects whatever
+	 * the most restrictive limit is set to.  And to then set the chunk
+	 * size to half that value, to ensure any required transfer overhead
+	 * and POST data remains well within the limit.  Or, if all of the
+	 * limits are set to unlimited, the chunk size will also be unlimited.
+	 *
+	 * @return int
+	 *
+	 * @access public
+	 */
 	public function get_chunk_size()
 	{
 		$max = 0;
 
-		// unlimited is -1 for memory_limit. 0 should be an invalid configuration.
-		$limit_memory = $this->php_ini->getBytes('memory_limit');
+		$limits = [
+			$this->php_ini->getBytes('memory_limit'),
+			$this->php_ini->getBytes('upload_max_filesize'),
+			$this->php_ini->getBytes('post_max_size'),
+		];
 
-		if ($limit_memory > 0)
+		foreach ($limits as $limit_type)
 		{
-			$max = $limit_memory;
+			if ($limit_type > 0)
+			{
+				$max = ($max !== 0) ? min($limit_type, $max) : $limit_type;
+			}
 		}
-
-		// For all remaining limits, 0 means "unlimited".
-
-		$limit_upload = $this->php_ini->getBytes('upload_max_filesize');
-
-		if ($limit_upload > 0)
-		{
-			$max = min($limit_upload, ($max ? $max : $limit_upload));
-		}
-
-		$limit_post = $this->php_ini->getBytes('post_max_size');
-
-		if ($limit_post > 0)
-		{
-			$max = min($limit_post, ($max ? $max : $limit_post));
-		}
-
-		// $config['max_filesize'] is not a limiter to chunk size.
 
 		return floor($max / 2);
 	}
--- a/phpBB/phpbb/plupload/plupload.php
+++ b/phpBB/phpbb/plupload/plupload.php
@@ -283,15 +283,51 @@ class plupload
 	*/
 	public function get_chunk_size()
 	{
-		$max = min(
-			$this->php_ini->getBytes('upload_max_filesize'),
-			$this->php_ini->getBytes('post_max_size'),
-			max(1, $this->php_ini->getBytes('memory_limit')),
-			$this->config['max_filesize']
-		);
+		$max = 0;
 
-		// Use half of the maximum possible to leave plenty of room for other
-		// POST data.
+		$limit = $this->php_ini->getBytes('memory_limit');
+
+		// unlimited is -1 for memory_limit. 0 would be an invalid configuration.
+
+		if ($limit > 0)
+		{
+			$max = $limit;
+		}
+
+		// For all remaining limits, 0 means "unlimited".
+
+		// For each limit, if there is a non-unlimited value to
+		// apply, apply the limit if it's less than whatever non-
+		// unlimited max value is currently set.  Also, apply the
+		// limit if the current max value is otherwise unlimited.
+
+		$limit = $this->php_ini->getBytes('upload_max_filesize');
+
+		if ($limit > 0)
+		{
+			$max = min($limit, max($max, $limit));
+		}
+
+		$limit = $this->php_ini->getBytes('post_max_size');
+
+		if ($limit > 0)
+		{
+			$max = min($limit, max($max, $limit));
+		}
+
+		$limit = $this->config['max_filesize'];
+
+		if ($limit > 0)
+		{
+			$max = min($limit, max($max, $limit));
+		}
+
+		// Only if every limit was 0/unlimited will we still
+		// have a zero value in $max at this point.
+
+		// Use half of the maximum possible to leave plenty of
+		// room for other POST data and be well under limits.
+
 		return floor($max / 2);
 	}
 
--- a/phpBB/phpbb/plupload/plupload.php
+++ b/phpBB/phpbb/plupload/plupload.php
@@ -276,8 +276,14 @@ class plupload
 	}
 
 	/**
-	* Checks various php.ini values and the maximum file size to determine
-	* the maximum size chunks a file can be split up into for upload
+	* Checks various php.ini values to determine the maximum chunk
+	* size a file should be split into for upload.
+	*
+	* The intention is to calculate a value which reflects whatever
+	* the most restrictive limit is set to.  And to then set the chunk
+	* size to half that value, to ensure any required transfer overhead
+	* and POST data remains well within the limit.  Or, if all of the
+	* limits are set to unlimited, the chunk size will also be unlimited.
 	*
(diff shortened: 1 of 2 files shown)


phpBB/phpbb/plupload/plupload.php
```diff
@@ -283,15 +283,51 @@ public function generate_resize_string()
...
+		{
+			$max = $limit;
+		}
+
+		// For all remaining limits, 0 means "unlimited".
+
+		// For each limit, if there is a non-unlimited value to
+		// apply, apply the limit if it's less than whatever non-
+		// unlimited max value is currently set.  Also, apply the
+		// limit if the current max value is otherwise unlimited.
```
3D-I: `// `should be used for one/two-liners.

phpBB/phpbb/plupload/plupload.php
```diff
@@ -283,15 +283,51 @@ public function generate_resize_string()
...
+		if ($limit > 0)
+		{
+			$max = min($limit, max($max, $limit));
+		}
+
+		// Only if every limit was 0/unlimited will we still
+		// have a zero value in $max at this point.
+
+		// Use half of the maximum possible to leave plenty of
+		// room for other POST data and be well under limits.
```
3D-I: better group the comments in a docblok like style. As per above.

phpBB/phpbb/plupload/plupload.php
```diff
@@ -283,15 +283,51 @@ public function generate_resize_string()
...
+		{
+			$max = min($limit, max($max, $limit));
+		}
+
+		$limit = $this->config['max_filesize'];
+
+		if ($limit > 0)
+		{
+			$max = min($limit, max($max, $limit));
+		}
```
3D-I: No way to beautify this code?

phpBB/phpbb/plupload/plupload.php
3D-I: Even better a shorter descro.

phpBB/phpbb/plupload/plupload.php
dark-1: This code looks good for me, as simple as possible...
It's also noob friendly... & Readable...

3D-I:
```

$limit_memory_limit			= $this->php_ini->getBytes('memory_limit');
$limit_upload_max_filesize	= $this->php_ini->getBytes('upload_max_filesize');
$limit_max_filesize			= (int) $this->config['max_filesize'];
$limit_post_max_size		= $this->php_ini->getBytes('post_max_size');

// If true is more than 0
if ($limit_memory_limit)
{
	$max = $limit_memory_limit;
}
if ($limit_upload_max_filesize)
{
	$max = min($limit_upload_max_filesize, max($max, $limit_upload_max_filesize));
}
if ($limit_post_max_size)
{
	$max = min($limit_post_max_size, max($max, $limit_post_max_size));
}
if ($limit_max_filesize)
{
	$max = min($limit_max_filesize, max($max, $limit_max_filesize));
}

echo floor($max / 2);

```

3D-I:
On a secondary note, the calculation results from your logic and the native one is IMHO wrong.
I made 2 stand-alone scripts to test what it.

And, the native code also returns ´0´ if all the values are null, also if only the `memory_limit` is `-1`.

3D-I:
> Without looking at the specific example values, the answer is always "the smallest one."

That's what the native logic does but yours :)

If you need my scripts to test it what I am saying I could gist them.

3D-I:
> Except that the native logic also thinks "whichever one is 0" is "the smallest one." When that's not what "0" means. "0" is effectively "the largest one."

The native logic returns 0 if there is a 0 in the values.

EA117:
> Therefore your logic seems buggy to me.

Agreed.  I see what you're saying now.  Yes, the max() I was using with the intention of "make sure we account for the fact $max might still be zero" is having the unintended effect you identified.  Closing to re-work.

phpBB/phpbb/plupload/plupload.php
```diff
@@ -276,22 +276,47 @@ public function generate_resize_string()
...
-			$this->php_ini->getBytes('post_max_size'),
-			max(1, $this->php_ini->getBytes('memory_limit')),
-			$this->config['max_filesize']
-		);
+		$max = 0;
+
+		// unlimited is -1 for memory_limit. 0 should be an invalid configuration.
+		$limit_memory = $this->php_ini->getBytes('memory_limit');
+
+		if ($limit_memory > 0)
```
3D-I: If not null is returned by `php_ini->getBytes()` it is safe to assume that true is more than 0.

`if ($limit_memory)`

Applies everywhere here.

phpBB/phpbb/plupload/plupload.php
```diff
@@ -276,22 +276,47 @@ public function generate_resize_string()
...
+		}
+
+		$limit_post = $this->php_ini->getBytes('post_max_size');
+
+		if ($limit_post > 0)
+		{
+			$max = min($limit_post, $max ? $max : $limit_post);
+		}
+		
+		// $config['max_filesize'] is not a limiter to chunk size.
```
3D-I: References?

phpBB/phpbb/plupload/plupload.php
```diff
@@ -276,22 +276,47 @@ public function generate_resize_string()
...
+			$max = $limit_memory;
+		}
+
+		// For all remaining limits, 0 means "unlimited".
+
+		$limit_upload = $this->php_ini->getBytes('upload_max_filesize');
+
+		if ($limit_upload > 0)
+		{
+			$max = min($limit_upload, $max ? $max : $limit_upload);
```
3D-I: Ternary is better to be surrounded to avoid any issues.

`$max = min($limit_upload, ($max ? $max : $limit_upload));`

EA117:
Re-worked to acknowledge that the _"max( $max, $limit )"_ actually needed to be _"$max ? $max : $limit"_.  Meaning we want to min() ourselves against the current $max value, unless the current $max value is zero.  Attached is the test script that was used to verify this change in the math.

[get_chunk_size_test.php.txt](https://github.com/phpbb/phpbb/files/3553481/get_chunk_size_test.php.txt)

**$this->config['max_filesize'] has been removed from the calculation.**  When re-reviewing the logic, it was realized that phpBB's "Maximum file size:" configuration has no bearing on helping ensure the plupload _chunk_size_ selection is successful.

_max_filesize_ is simply the phpBB admin's assertion of what size file they wish to accept, and does not determine any part of "what size transfer will be safe or technically possible."_  Only the PHP's _memory_limit_, PHP's _upload_max_filesize_, and PHP's _post_max_size_ impose the type of limits that _chunk_size_ would fail if not honored.

Note plupload is already being provided with _max_filesize_ as part of _"max_file_size: '{FILESIZE}b'"_, which prevents a file larger than _max_filesize_ from being selected in the first place.  And the phpBB server itself will reject any attachment larger than _max_filesize_, once the attachment is successfully uploaded to the server.  But the _chunk_size_ used when transferring the file to the phpBB server is unrelated to either of those decisions.

3D-I:
> I will give this a shot again, tomorrow. :)

phpBB/phpbb/plupload/plupload.php
3D-I: Yeah,
`if (true === $limit_memory)`, nitpicking though.

Agreed that negative values are always `true` in PHP.

phpBB/phpbb/plupload/plupload.php
```diff
@@ -276,22 +276,47 @@ public function generate_resize_string()
...
+		if ($limit_post > 0)
+		{
+			$max = min($limit_post, $max ? $max : $limit_post);
+		}
+		
+		// $config['max_filesize'] is not a limiter to chunk size.
 
-		// Use half of the maximum possible to leave plenty of room for other
-		// POST data.
 		return floor($max / 2);
```
3D-I: Old comments should be kepf in place, amended if necessarily due to changes.

3D-I:
Code sniffer says: `[if] Error in IfTask`

EA117:
I suspect the IfTask error (not that I have any specific information from the log to go on) was simply because I did two back-to-back "close and re-open" actions on the PR, and thereby aborted an already in-progress test pass. I saw it happen, but just wasn't in a hurry to re-start the the tests again until after feedback occurred.

But the ternary is encapsulated at this point too; so if that was actually it, this is solved too.

The comment is still there; it's with the block that was expanded at the beginning of the function.  That comment regarding the "divide by two" goal, plus now the other goals, were put together into a single summary.

3D-I:
There is another error too

```
FILE: /home/travis/build/phpbb/phpbb/phpBB/phpbb/plupload/plupload.php

----------------------------------------------------------------------

FOUND 1 ERROR AFFECTING 1 LINE

----------------------------------------------------------------------

 317 | ERROR | [x] Whitespace found at end of line
```

pretty clear.

3D-I:
That's it.

3D-I:
The logic does now what's written on the tin.

3D-I:
The Studio's review proposal:

```
	/**
	 * Checks various php.ini values to determine the maximum chunk
	 * size a file should be split into for upload.
	 *
	 * The intention is to calculate a value which reflects whatever
	 * the most restrictive limit is set to.  And to then set the chunk
	 * size to half that value, to ensure any required transfer overhead
	 * and POST data remains well within the limit.  Or, if all of the
	 * limits are set to unlimited, the chunk size will also be unlimited.
	 *
	 * @return int
	 * @access public
	*/
	public function get_chunk_size()
	{
		$max = 0;

		$limit_memory	= $this->php_ini->getBytes('memory_limit');
		$limit_upload	= $this->php_ini->getBytes('upload_max_filesize');
		$limit_post	= $this->php_ini->getBytes('post_max_size');

		$limits = [$limit_memory, $limit_upload, $limit_post];

		foreach ($limits as $limit_type)
		{
			if ($limit_type > 0)
			{
				$max = ($max !== 0) ? min($limit_type, $max) : $limit_type;
			}
		}

		return floor($max / 2);
	}
```

EA117:
That looks fine to me, too.  Attached is an updated script used to validate that zero and non-zero values at setting will generate the expected outcomes.  Thanks for the additional review time.

[get_chunk_size_test.php.txt](https://github.com/phpbb/phpbb/files/3559683/get_chunk_size_test.php.txt)

3D-I:
>  Attached is an updated script

Use https://gist.github.com/ next time, is way better.

Instead that's my simple test file. ;)

```
<?php

$max = 0;

$limits = [100, 200, 300];

foreach ($limits as $limit)
{
	if ($limit > 0)
	{
		$max = ($max !== 0) ? min($limit, $max) : $limit;
	}
}

echo floor($max / 2);

```

EA117:
Thanks, appreciate the pointer.  This is my first github-hosted project, as well as my first PHP-based project, so I'm definitely ignorant on many details and still learning.

marc1706:
I've added a test that should cover your .txt file. Please confirm that those values are also what you were expecting.

EA117:
Thanks @marc1706 .  I agree the test data constructed here matches the input values and the expected output values of the six tests in the .txt file attached.  What I realize now when looking back at those tests though is that there really should have been seven tests to cover all the basic cases.

We have the "every configuration value was zero/unlimited" test, followed by three tests with "zero/unlimited" present at each of the possible configuration value, then a test where "upload_max divided by two" is the expected answer, and finally a test where "post_max divided by two" is the expected answer.

We're missing a test where "the memory limit divided by two" was the expected output.   So if you didn't mind adding one more set of test data to the array, I would suggest:

```
	[[
		'memory_limit' => 1000,
		'upload_max_filesize' => 2000,
		'post_max_size' => 3000,
	], 500],
```

Thanks.

PR #5682, opened 2019-09-08
[ticket/16151] Enable Emojis and rich text in forum name

3D-I:
PHPBB3-16151

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16151

### DIFF
--- a/phpBB/includes/acp/acp_forums.php
+++ b/phpBB/includes/acp/acp_forums.php
@@ -986,10 +986,30 @@ class acp_forums
 			$errors[] = $user->lang['FORUM_NAME_EMPTY'];
 		}
 
-		// No Emojis
+		/**
+		 * Replace Emojis and other 4bit UTF-8 chars not allowed by MySql to NCR.
+		 * Using their Numeric Character Reference's Hexadecimal notation.
+		 * Doesn't interfere with Japanese or Cyrillic etc.
+		 *
+		 * @see https://www.w3.org/TR/xml11/
+		 * @see https://www.opentag.com/xfaq_charrep.htm
+		 */
+		if (preg_match_all('/[\x{10000}-\x{10FFFF}]/u', $forum_data_ary['forum_name'], $matches))
+		{
+			foreach ($matches as $key => $emoji)
+			{
+				$forum_data_ary['forum_name'] = str_replace($emoji, utf8_encode_ncr($emoji), $forum_data_ary['forum_name']);
+			}
+		}
+
+		/**
+		 * This should never happen again.
+		 * Leaving the fallback hre just in case there will be the need of it.
+		 */
 		if (preg_match_all('/[\x{10000}-\x{10FFFF}]/u', $forum_data_ary['forum_name'], $matches))
 		{
 			$character_list = implode('<br>', $matches[0]);
+
 			$errors[] = $user->lang('FORUM_NAME_EMOJI', $character_list);
 		}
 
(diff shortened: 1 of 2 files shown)


3D-I:
!set php

3D-I:
immagine

immagine

immagine

immagine

3D-I:
https://github.com/phpbb/phpbb/pull/5622

phpBB/includes/acp/acp_forums.php
```diff
@@ -986,10 +986,30 @@ function update_forum_data(&$forum_data_ary)
...
-		// No Emojis
+		/**
+		 * Replace Emojis and other 4bit UTF-8 chars not allowed by MySql to NCR.
+		 * Using their Numeric Character Reference's Hexadecimal notation.
+		 * Doesn't interfere with Japanese or Cyrillic etc.
+		 *
+		 * @see https://www.w3.org/TR/xml11/
+		 * @see https://www.opentag.com/xfaq_charrep.htm
+		 */
+		if (preg_match_all('/[\x{10000}-\x{10FFFF}]/u', $forum_data_ary['forum_name'], $matches))
```
JoshyPHP: Instead of running as many `str_replace` as there are replacements, you should be able to run a single replacement with the `utf8_encode_ncr_callback`.

```php
function utf8_encode_sp($text)
{
	return preg_replace_callback('/[\\xF0-\\xF4].../', 'utf8_encode_ncr_callback', $text);
}
```

Then the whole block becomes a single assignment:
```php
$forum_data_ary['forum_name'] = utf8_encode_sp($forum_data_ary['forum_name']);
```

3D-I: I don't see the need to add a new function IMHO.

Let's not forget that we are in ACP, the admin is choosing a name for a forum that most likely will never change, if he ever did I doubt that it is a constant operation over time.

Why do I have to pass under the press a string if there is no need to do it? :thinking: 
Your regex seems to me to evaluate the same range of reserved characters (SMP) as mine, the native function receives and evaluates only the characters extracted from my regex. 

I admit I thought about a solution like yours during the developmment of this idea, I chose to go this way though.

If, and could be possible indeed, I am missing something pls let me know, thanks @JoshyPHP .

JoshyPHP: It's a good idea to have a function for it so you can use it in [your other PR](https://github.com/phpbb/phpbb/pull/5685/files#diff-2f48d04a5ad1975b5751ca778161d39b) instead of having the same code twice or more. It's got fewer lines and a lower cyclomatic complexity.

The regexp is functionally the same, it's just cheaper to operate on bytes than on codepoints.

3D-I: I am bound to open 2 more PRs about Emojis, I will see what a do after some test. Thx Joshy.

marc1706: I also think adding a function for this makes more sense. If the counter argument is only "oh no, we have to add a function", then I'd definitely go for it. :tada:

3D-I: Okay. I will add a function, no probs.

3D-I:
@marc1706 I've implemented the function in this PR, once this single PR is merged I will take care to amend the other one (topic title). Some PR more will follow which uses this new function.

marc1706:
done :)

PR #5683, opened 2019-09-08
[ticket/16149] Missing profile field data after registration

3D-I:
PHPBB3-16149

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16149

### DIFF
--- a/phpBB/includes/ucp/ucp_register.php
+++ b/phpBB/includes/ucp/ucp_register.php
@@ -102,7 +102,9 @@ class ucp_register
 		/* @var $cp \phpbb\profilefields\manager */
 		$cp = $phpbb_container->get('profilefields.manager');
 
-		$error = $cp_data = $cp_error = array();
+		$cp_data = ['user_id' => ''];
+
+		$error = $cp_error = array();
 		$s_hidden_fields = array();
 
 		// Handle login_link data added to $_hidden_fields
--- a/phpBB/phpbb/profilefields/manager.php
+++ b/phpBB/phpbb/profilefields/manager.php
@@ -364,7 +364,7 @@ class manager
 
 			foreach ($user_ids as $user_id)
 			{
-				if (!isset($user_fields[$user_id][$used_ident]) && $this->profile_cache[$used_ident]['field_show_novalue'])
+				if (!isset($user_fields[$user_id][$used_ident]) && !$this->profile_cache[$used_ident]['field_show_novalue'])
 				{
 					$user_fields[$user_id][$used_ident]['value'] = '';
 					$user_fields[$user_id][$used_ident]['data'] = $this->profile_cache[$used_ident];


3D-I:
!set php

3D-I:
!set WIP

3D-I:
@kasimi Thoughts?

* the first commit solves the registration issue

* the second one is "hackish" IMO, returns an array with empty values for each field if the user_id is not in the table. But still is not there, such user_id.

```
array (size=1)
  33 => 
    array (size=7)
      'phpbb_location' => 
        array (size=2)
          'value' => string '' (length=0)
          'data' => 
            array (size=28)
              ...
      'phpbb_website' => 
        array (size=2)
          'value' => string '' (length=0)
          'data' => 
            array (size=28)
              ...
      'phpbb_facebook' => 
        array (size=2)
          'value' => string '' (length=0)
          'data' => 
            array (size=28)
              ...
      'phpbb_twitter' => 
        array (size=2)
          'value' => string '' (length=0)
          'data' => 
            array (size=28)
              ...
      'phpbb_skype' => 
        array (size=2)
          'value' => string '' (length=0)
          'data' => 
            array (size=28)
              ...
      'phpbb_youtube' => 
        array (size=2)
          'value' => string '' (length=0)
          'data' => 
            array (size=28)
              ...
      'phpbb_googleplus' => 
        array (size=2)
          'value' => string '' (length=0)
          'data' => 
            array (size=28)
              ...
```

Else a cron-job or a migration?

kasimi:
Thanks for looking into this, I will test it soon.

As an alternative idea, I wonder what happens if the second condition here is removed: https://github.com/phpbb/phpbb/blob/4025a49a59f64116acf9f0999a39cca475fe6203/phpBB/phpbb/profilefields/manager.php#L367

If `!$this->profile_cache[$used_ident]['field_show_novalue']` is removed, the method returns all CPFs with an empty value for users who have never updated their profile, as would be expected, but it would mean that all callers of the `grab_profile_fields_data()` method that previously relied on getting an empty array would need to check the `field_show_novalue`. As far as I can see, the core already does this anyway.

3D-I:
I will be back on this as soon as I can, let me know pls if in the meantime you tested it.
There are a couple of things I'd like to discuss, later on.

kasimi:
I tested your code changes in `phpbb/profilefields/manager.php` and it's returning unexpected results as well. For example, if I check "Show field if no value was selected" (the `field_show_novalue` option) for a PF, this PF's information is missing in the newly registered user's PF data that is returned by `grab_profile_fields_data()` even though it should be there. I don't understand what `field_show_novalue` has to do with fetching default CP data.

As for your changes in `includes/ucp/ucp_register.php` I'm not sure why they are needed. The `user_id` is already set in the `user_add()` function: https://github.com/phpbb/phpbb/blob/4db585a4cb2e5359074a82ef088574609155294b/phpBB/includes/functions_user.php#L302 I don't think registration needs fixing. No empty row should be added to the `phpbb_profile_fields_data` table for newly registered users, which is how it currently works.

3D-I:
Closing ATM.

PR #5684, opened 2019-09-09
[ticket/12539] Live member search improvements

mrgoldy:
PHPBB3-12539

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-12539

---

This will add the ability to navigate the search results in the _Live Member Search_ in the _Memberlist > Find a Member_ functionallity.

Able to navigate with the `arrow keys` (up and down).
Set the value of the selected row with `enter`
Close the dropdown with `escape`.
Show the loading indicator

--- 

Branch is set to 3.3.x as it is rather pointless for 4.0 as it is stated that entire JS will be redone for that version.

### DIFF
--- a/phpBB/assets/javascript/core.js
+++ b/phpBB/assets/javascript/core.js
@@ -563,7 +563,6 @@ phpbb.search.setValue = function($input, value, multiline) {
 phpbb.search.setValueOnClick = function($input, value, $row, $container) {
 	$row.click(function() {
 		phpbb.search.setValue($input, value.result, $input.attr('data-multiline'));
-
 		phpbb.search.closeResults($input, $container);
 	});
 };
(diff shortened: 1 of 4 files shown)


marc1706:
Can we also prevent the caret from moving while navigating through the list in the dropdown? Right now the caret jumps to the beginning of the text one has entered when pressing the up key. This will result in users then accidentally typing at the beginning of the entered string instead of appending to the end.

mrgoldy:
The cursor (or caret) is no longer repositioning upon navigating the search results, for both the up (start) and down (end) arrow. It now just sticks to its current position.

marc1706:
Last question: do you think applying the selected username with Enter makes more sense than using the Tab key? I somehow would want to use the tab key but not sure how others feel about it.

mrgoldy:
We can also do both, if you want.

 But if you look at it in general, ‘enter’ is selecting and ‘tab’ is going to the next input. Not setting a value. 

As for that matter, if you want to implement ‘tab’ somewhere, it sounds more logical to use it to navigate through the search results. Acting similar to arrow down.

VSEphpbb:
I Don't know about tab. My brain says up and down arrows to navigate, enter to select, and ESC to close

PR #5691, opened 2019-09-17
[ticket/15985] List ban duration in logs

stevendegroote:
Log the length of a ban, so moderators can check afterwards how long previous
bans were.

PHPBB3-15985

Checklist:

- [ ] Tests pass


marc1706:
Hm, I don't really like the approach you're taking. You're adding handling for a very specific type of log to the log class whereas that class should only take care about adding and displaying logs, not about handling specifics inside those logs. Also, I think you're current approach will cause at least PHP warnings with displaying bans that were created before this change as `$log_data_ary[2]` and `$log_data_ary[3]` will not be set.

stevendegroote:
Thanks @marc1706  for looking into it.

First of all, there are no warnings in case the arguments are not set. On lines 675-700 in log.php, there is code that completes missing arguments for the printf method, so that's covered (I was frankly surprised we have such code, but anyway)

I think this change is very much needed. I'm doing workarounds for years on this matter, and there's no true possibility to do this via an extension. To me, it's a vital functionality that's missing for moderators.

I'd by happy to re-implement though, but what do you propose then?

PR #5692, opened 2019-09-17
[ticket/16159] Wrap post times in html time tag

stevendegroote:
PHPBB3-16159

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16159


3D-I:
IMHO that's wrong.

`$user->format_date()` already formats the date depending on users' choice in UCP.

Why would you want to force a date format for everybody?
Not to mention that this will break BC.

Or am I missing something?

phpBB/includes/functions_display.php
```diff
@@ -547,11 +547,12 @@ function display_forums($root_data = '', $display_moderators = true, $return_mod
 				$last_post_subject = $last_post_subject_truncated = '';
 			}
 			$last_post_time = $user->format_date($row['forum_last_post_time']);
+			$last_post_time_rfc3339 = gmdate("Y-m-d H:i:sP", $row['forum_last_post_time']);
```
3D-I: Why?

stevendegroote:
The output format for the user is still configurable, that doesn't change at all.
What changes is that the outputted dates are put in a html <time> tag.

Please see https://tracker.phpbb.com/browse/PHPBB3-16159 for an explanation.

3D-I:
How can it be congigurable if you are forcing a format?

`gmdate("Y-m-d H:i:sP"`

3D-I:
@stevendegroote 

I had the opportunity to discuss the subject, sorry for my healthy doubts due to an extemporaneous objective vision. Thank you :+1:

phpBB/includes/functions_display.php
hanakin: What’s with the rfc3339 stuff? Are you fixing all date times I. The theme to work with the time and datetime setup in this pr?

stevendegroote: No no. Just check the ticket description in the issue tracker, and check in the template files where these RFC3339 dates are used. They don't change anything visually, and maintain the ability for users to configure their own date format for output.

3D-I: @hanakin 
I had the same impression I have to admit but that's just a wrapper for the sake of search engines, at the end of all.

Why not to use DATE_ATOM instead? That's the same as of DATE_RFC3339.

https://www.php.net/manual/en/class.datetimeinterface.php#datetime.constants.rfc3339

stevendegroote: @3D-I You have a point about the format. Maybe we can use some PHP constant. According to MDN documentation for <time>, there are several valid formats. 

I'll recheck the exact format. It looks like the format I'm using is actually not fully RFC3339 (as it's missing the T between date and time now).

3D-I: **RFC3339_EXTENDED** is for PHP >= 7 if I correctly recall?

stevendegroote: I replaced the date formats with DATE_RFC3339 (available on required PHP versions, and the actual name of the standard). ATOM is a feed standard, which happens to use RFC3339 as its format for dates.

marc1706:
Thanks for your PR. Good job.

PR #5696, opened 2019-09-22
[ticket/16165] Refactor viewforum.php into services

CHItA:
PHPBB3-16165

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16165

### DIFF
--- a/phpBB/phpbb/forum/enumeration/forum_options.php
+++ b/phpBB/phpbb/forum/enumeration/forum_options.php
@@ -1,0 +1,30 @@
+<?php
+/**
+ *
+ * This file is part of the phpBB Forum Software package.
+ *
+ * @copyright (c) phpBB Limited <https://www.phpbb.com>
+ * @license GNU General Public License, version 2 (GPL-2.0)
+ *
+ * For full copyright and license information, please see
+ * the docs/CREDITS.txt file.
+ *
+ */
+
+namespace phpbb\forum\enumeration;
+
+/**
+ * Enumeration type for forum options.
+ */
+class forum_options
+{
+	/**
+	 * @var int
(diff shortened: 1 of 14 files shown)


phpBB/phpbb/forum/enumeration/forum_options.php
```diff
@@ -0,0 +1,31 @@
...
+
+define('FORUM_OPTION_FEED_NEWS', 1);
+define('FORUM_OPTION_FEED_EXCLUDE', 2);
+
+class forum_options
+{
+	/**
+	 * @var int
+	 */
+	const FORUM_OPTION_FEED_NEWS = 1;
```
JoshyPHP: I'm about 90% sure that you can use global constants in class constant declarations on every version of PHP supported by phpBB. It may be a good idea to start specifying the constant's visibility since it's in PHP 7.1.

```php
public const FORUM_OPTION_FEED_NEWS = FORUM_OPTION_FEED_NEWS;
```

Although come to think of it, I think it's a better idea to do it the other way around and define the global constant based on the class constant. It seems more correct and it may make a difference in Opcache's optimizer.

```php
define('FORUM_OPTION_FEED_NEWS', forum_options::FORUM_OPTION_FEED_NEWS);
class forum_options
{
	public const FORUM_OPTION_FEED_NEWS = 1;
}
```

CHItA: the defines were left in by accident.

apple-fan-pl:
Someone should fix this ASAP as its quite important change

rubencm:
Will be outdated after #5696 is merged

PR #5704, opened 2019-10-04
[ticket/16172] Add group rank label to group view

senky:
PHPBB3-16172

Before:
Snímka obrazovky 2019-10-04 o 14 19 30

After:
Snímka obrazovky 2019-10-04 o 14 19 39

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16172

### DIFF
--- a/phpBB/styles/prosilver/template/memberlist_body.html
+++ b/phpBB/styles/prosilver/template/memberlist_body.html
@@ -29,7 +29,7 @@
 			<!-- IF AVATAR_IMG -->{AVATAR_IMG}<!-- ENDIF -->
 			{% EVENT memberlist_body_group_rank_before %}
 			<!-- IF RANK_IMG -->{RANK_IMG}<!-- ENDIF -->
-			<!-- IF GROUP_RANK -->{GROUP_RANK}<!-- ENDIF -->
+			<!-- IF GROUP_RANK -->{{ lang('GROUP_RANK') ~ lang('COLON') }} {GROUP_RANK}<!-- ENDIF -->
 			{% EVENT memberlist_body_group_rank_after %}
 		</p>
 	<!-- ELSE -->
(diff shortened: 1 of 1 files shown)


mrgoldy:
Will this not look weird with a rank image?

dark-1:
I think this should be like :
```php
<!-- IF RANK_IMG or GROUP_RANK -->{{ lang('GROUP_RANK') ~ lang('COLON') }}<!-- ENDIF -->
<!-- IF RANK_IMG -->{RANK_IMG}<!-- ENDIF -->
<!-- IF GROUP_RANK -->{GROUP_RANK}<!-- ENDIF -->
```

senky:
I don't think so. But you are right that it is weird with rank image. I decided to display the label only when the image is not present (because it is the only situation when it is unclear what the text is about).

phpBB/styles/prosilver/template/memberlist_body.html
```diff
@@ -29,7 +29,12 @@ <h2 class="group-title"<!-- IF GROUP_COLOR --> style="color:#{GROUP_COLOR};"<!--
 			<!-- IF AVATAR_IMG -->{AVATAR_IMG}<!-- ENDIF -->
 			{% EVENT memberlist_body_group_rank_before %}
 			<!-- IF RANK_IMG -->{RANK_IMG}<!-- ENDIF -->
-			<!-- IF GROUP_RANK -->{{ lang('GROUP_RANK') ~ lang('COLON') }} {GROUP_RANK}<!-- ENDIF -->
+			<!-- IF GROUP_RANK -->
+				<!-- IF not RANK_IMG -->
+					{{ lang('GROUP_RANK') ~ lang('COLON') }}
+				<!-- ENDIF -->
+				{GROUP_RANK}
+			<!-- ENDIF -->
```
3D-I: Twig should be used on every bit of code added or amended.

PR #5715, opened 2019-10-19
[ticket/16078] Try using retries for timeout issue

marc1706:
The timeout only seems to occur on the first test so let's see what a
retry can do to help with this issue.

PHPBB3-16078

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16078

### DIFF
--- a/tests/test_framework/phpbb_ui_test_case.php
+++ b/tests/test_framework/phpbb_ui_test_case.php
@@ -159,20 +159,23 @@ class phpbb_ui_test_case extends phpbb_test_case
 	public function visit($path)
 	{
 		// Retry three times on curl issues, e.g. timeout
-		try
+		$attempts = 0;
+		$retries = 3;
+
+		while (true)
 		{
-			$this->getDriver()->get(self::$root_url . $path);
-		}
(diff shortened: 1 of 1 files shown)


tests/test_framework/phpbb_ui_test_case.php
```diff
@@ -158,7 +158,23 @@ public function getDriver()
 
 	public function visit($path)
 	{
-		$this->getDriver()->get(self::$root_url . $path);
+		// Retry three times on curl issues, e.g. timeout
```
rubencm: I would do something like this to avoid repeating the same 3 times

```php
$attemps = 0;
$retries = 3;

while(true) {
	$attemps++;
	try {
		$this->getDriver()->get(self::$root_url . $path);
		break;
	} catch(Facebook\WebDriver\Exception\WebDriverCurlException $exception) {
		if($attemps == $retries) {
			throw $exception;
		}
	}
}
```

marc1706: We can do this in case it actually works. Was aiming for better debuggability in case it errors. ;)

PR #5718, opened 2019-10-24
[ticket/16189] Deprecate inet_ntop and inet_pton wrappers

rubencm:
PHPBB3-16189

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16189


phpBB/includes/functions.php
```diff
@@ -2200,7 +2200,7 @@ function confirm_box($check, $title = '', $hidden = '', $html_body = 'confirm_bo
 		return false;
 	}
 
-	// re-add sid / transform & to &amp; for user->page (user->page is always using &)
+	// re-add sid / transform & to &amp; for user->page (uhttps://github.com/rubencm/phpbb/pull/new/ticket/15958ser->page is always using &)
```
3D-I: errors in the url, should be:
`// re-add sid / transform & to &amp; for user->page (https://github.com/rubencm/phpbb/pull/new/ticket/15958 user->page is always using &)`

marc1706:
Do the now deprecated functions behave differently than their replacement? If not, shouldn't the functions in functions_compatibility.php just be wrappers to the now used functions?

phpBB/includes/functions.php
```diff
@@ -2851,10 +2851,13 @@ function get_preg_expression($mode)
 		// Whoa these look impressive!
 		// The code to generate the following two regular expressions which match valid IPv4/IPv6 addresses
 		// can be found in the develop directory
+
+		// @deprecated
```
marc1706: I think there isn't that much of a point to deprecate these. The preg expressions are still valid and might also be used by extensions.

rubencm: It have no sense to keep that when php already do that natively with filter_var

marc1706:
Can you rebase onto latest 3.3.x? Thanks.

rubencm:
Done

PR #5728, opened 2019-11-03
[ticket/15294] Fix session_gc() selecting expired sessions for unique users

rxu:
Checklist:

- [x] Tests pass

<a href="https://tracker.phpbb.com/browse/PHPBB3-15294">PHPBB3-15294</a>.

### DIFF
--- a/phpBB/phpbb/session.php
+++ b/phpBB/phpbb/session.php
@@ -954,8 +954,6 @@ class session
 	{
 		global $db, $config, $phpbb_container, $phpbb_dispatcher;
 
-		$batch_size = 10;
-
 		if (!$this->time_now)
 		{
 			$this->time_now = time();
@@ -968,14 +966,21 @@ class session
 		$db->sql_query($sql);
 
 		// Get expired sessions, only most recent for each user
-		$sql = 'SELECT session_user_id, session_page, MAX(session_time) AS recent_time
-			FROM ' . SESSIONS_TABLE . '
-			WHERE session_time < ' . ($this->time_now - $config['session_length']) . '
-			GROUP BY session_user_id, session_page';
-		$result = $db->sql_query_limit($sql, $batch_size);
+		// Inner SELECT gets most recent expired sessions for unique session_user_id
+		// Outer SELECT gets session_page for them
+		$sql = 'SELECT s1.session_page, s1.session_user_id, s1.session_time AS recent_time
+			FROM ' . SESSIONS_TABLE . ' AS s1
+			INNER JOIN (
+				SELECT session_user_id, MAX(session_time) AS recent_time
+				FROM ' . SESSIONS_TABLE . '
+				WHERE session_time < ' . ($this->time_now - (int) $config['session_length']) . '
+				GROUP BY session_user_id
+			) AS s2
+			ON s1.session_user_id = s2.session_user_id
+				AND s1.session_time = s2.recent_time';
+		$result = $db->sql_query($sql);
 
 		$del_user_id = array();
-		$del_sessions = 0;
 
 		while ($row = $db->sql_fetchrow($result))
 		{
@@ -985,7 +990,6 @@ class session
 			$db->sql_query($sql);
 
 			$del_user_id[] = (int) $row['session_user_id'];
-			$del_sessions++;
 		}
 		$db->sql_freeresult($result);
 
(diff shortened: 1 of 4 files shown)


rxu:
I'm not sure if we need to select and update info (last session page, user last visit) in the users table for anonymous user here.

phpBB/phpbb/session.php
```diff
@@ -954,72 +954,71 @@ function session_gc()
...
-
-		while ($row = $db->sql_fetchrow($result))
-		{
-			$sql = 'UPDATE ' . USERS_TABLE . '
-				SET user_lastvisit = ' . (int) $row['recent_time'] . ", user_lastpage = '" . $db->sql_escape($row['session_page']) . "'
-				WHERE user_id = " . (int) $row['session_user_id'];
-			$db->sql_query($sql);
+		// Delete all expired sessions
+		$sql = 'DELETE FROM ' . SESSIONS_TABLE . '
+			WHERE session_time < ' . ($this->time_now - $config['session_length']);
```
mrgoldy: Above you cast this variable to `(int)`, might want to do it here aswell.

phpBB/phpbb/session.php
```diff
@@ -954,72 +954,71 @@ function session_gc()
...
+			case 'sqlite3':
+			case 'mysqli':
+				$sql = 'UPDATE ' . USERS_TABLE . " AS u,
+					$sql_select
+					SET u.user_lastvisit = s3.recent_time, u.user_lastpage = s3.session_page
+					WHERE u.user_id = s3.session_user_id";
+			break;
+
+			default:
+			$sql = 'UPDATE ' . USERS_TABLE . "
```
mrgoldy: Needs additional indentation

phpBB/phpbb/session.php
```diff
@@ -954,72 +954,71 @@ function session_gc()
...
 			$this->time_now = time();
 		}
 
-		// Firstly, delete guest sessions
-		$sql = 'DELETE FROM ' . SESSIONS_TABLE . '
-			WHERE session_user_id = ' . ANONYMOUS . '
-				AND session_time < ' . (int) ($this->time_now - $config['session_length']);
+		// Get expired sessions, only most recent for each registered user
+		// Inner SELECT gets most recent expired sessions data for unique session_user_id
+		// Outer SELECT gets also session_page for them
```
mrgoldy: Perhaps it's better to make this a block comment, `/** ... */`.

mrgoldy:
I believe the travis error is cause of the derived table alias, which could/should not have an `AS` infront.

rxu:
> I believe the travis error is cause of the derived table alias, which could/should not have an `AS` infront.

I've tested locally on SQLite 3.20 and it's erroring both ways (with and withous `AS`). It seems that SQLite doesn't support `UPDATE` from `SELECT`, at least in such way, regrettably.

3D-I:
> Are there any SQLite experts around?

Not really, just found a related topic here: https://stackoverflow.com/questions/10027468/how-can-i-update-table-by-using-select-statement-in-sqlite#10027800

rxu:
That would be possible in SQLite using CTE (common table expressions) but it's supported since SQLite 3.8.3 while phpBB 3.3 requires 3.6.15.

3D-I:
> 
> 
> That would be possible in SQLite using CTE (common table expressions) but it's supported since SQLite 3.8.3 while phpBB 3.3 requires 3.6.15.

Other way around could be rewrite that SQL to be DBAL compliant at once,
I mean diving the query and what not?

3D-I:
> I guess I'll better reset to the previous splitted solution.
That's what I meant.

rxu:
Well, current approach looks better from the point it gets rid of executing `UPDATE` queries in a `while` loop. But there's no other choice for the moment I guess.

3D-I:
> 
> 
> Well, current approach looks better from the point it gets rid of executing `UPDATE` queries in a `while` loop. But there's no other choice for the moment I guess.

Or `foreach` maybe... wondering those approachs though will need a `sql->transaction` sort of loop.

rxu:
> executing `UPDATE` queries in a `while` loop.

So, this is probably why the `$batch_size` was needed.

rxu:
Still thinking on better implementation.
Please don't merge.

rxu:
Tested on a live production board (v.3.2.8) for a week, and it seems to be fine.
`phpbb_sessions` table size redused like 2,5 MB -> 200 kB, `user_lastvisit` is getting correct.
No issues with login etc.

marc1706:
I've also looked at the query in MySQL and so far it seems alright with regards to using indexes etc.

PR #5769, opened 2019-12-07
[ticket/16248] swap logo for new svg version

hanakin:
Checklist:

- [x] Tests pass

PHPBB3-16248

https://tracker.phpbb.com/browse/PHPBB3-16248

Screenshot 2019-12-07 07 59 52


marc1706:
Will this change anything with regards to people wanting to use a custom logo instead of the phpBB one? E.g. one that might be a PNG.

hanakin:
it should not. That is why i added it via css for now. in 4.0 we can switch it out for the twig function

mrgoldy:
Well, it does affect it, doesn't it? Prosilver now looks for `site_logo.svg` instead of `site_logo.gif`, so users can not just upload a new 'gif logo'.

hanakin:
Trivial. Most users do not just do that as they are usually different sizes and current ext will still work for it as it’s using the same class names.

marc1706:
While I generally approve of moving to SVG, I think there is currently no big reason to do this only for the logo. I think for now we can live without the SVG logo and rather update the currently used logo with the new one in PNG format. For 4.0 we should and will definitely move over to SVG.

vinny:
@marc1706 @mrgoldy it will not affect users to use custom logo of different formats (png, gif, jpg)

hanakin:
Do not understand the hesitation? But I can make a png version but it will require more code to make it retina which is more complicated?

3D-I:
I've been thinking a little more about this topic and I don't see how it can prevent any customization by users. Personally, therefore, I feel like approving these changes and not postponing them, just my opinion anyway. The KB clearly explains already how to change it despite of the file type.

marc1706:
Ok, as all seem to agree that this change does not have a big effect I'll merge it. @vinny I think it would still make sense to maybe inform users about this in the styles forums? Some might still just replace site_logo.png and will probably be a bit surprised by this change.

hanakin:
@marc1706 can we just put that info in the changes topic in the styles forum?

phpBB/styles/prosilver/template/overall_header.html
```diff
@@ -84,7 +84,9 @@
 			<div class="inner">
 
 			<div id="site-description" class="site-description">
-				<a id="logo" class="logo" href="<!-- IF U_SITE_HOME -->{U_SITE_HOME}<!-- ELSE -->{U_INDEX}<!-- ENDIF -->" title="<!-- IF U_SITE_HOME -->{L_SITE_HOME}<!-- ELSE -->{L_INDEX}<!-- ENDIF -->"><span class="site_logo"></span></a>
+				<a id="logo" class="logo" href="<!-- IF U_SITE_HOME -->{U_SITE_HOME}<!-- ELSE -->{U_INDEX}<!-- ENDIF -->" title="<!-- IF U_SITE_HOME -->{L_SITE_HOME}<!-- ELSE -->{L_INDEX}<!-- ENDIF -->">
```
3D-I: Moving to Twig syntax is the preferred thing to do here.

vinny:
> @marc1706 can we just put that info in the changes topic in the styles forum?

a fixed topic would make more sense.

phpBB/styles/prosilver/template/overall_header.html
```diff
@@ -84,7 +84,9 @@
 			<div class="inner">
 
 			<div id="site-description" class="site-description">
-				<a id="logo" class="logo" href="<!-- IF U_SITE_HOME -->{U_SITE_HOME}<!-- ELSE -->{U_INDEX}<!-- ENDIF -->" title="<!-- IF U_SITE_HOME -->{L_SITE_HOME}<!-- ELSE -->{L_INDEX}<!-- ENDIF -->"><span class="site_logo"></span></a>
+				<a id="logo" class="logo" href="{% if U_SITE_HOME %}{{ U_SITE_HOME }}{% else %}{{ U_INDEX }}{% endif %}" title="{% if U_SITE_HOME %}{{ lang(L_SITE_HOME) }}{% else %}{{ lang(L_INDEX) }}{% endif %}">
```
3D-I: `{{ lang(L_SITE_HOME) }}` and `{{ lang(L_INDEX) }}`
should be
`{{ lang('SITE_HOME') }}` and `{{ lang('INDEX') }}`

3D-I:
@hanakin please see: https://github.com/phpbb/phpbb/pull/5769#discussion_r359071608

phpBB/styles/prosilver/template/overall_header.html
```diff
@@ -84,7 +84,7 @@
 			<div class="inner">
 
 			<div id="site-description" class="site-description">
-				<a id="logo" class="logo" href="{% if U_SITE_HOME %}{{ U_SITE_HOME }}{% else %}{{ U_INDEX }}{% endif %}" title="{% if U_SITE_HOME %}{{ lang(L_SITE_HOME) }}{% else %}{{ lang(L_INDEX) }}{% endif %}">
+		<a id="logo" class="logo" href="{% if U_SITE_HOME %}{{ U_SITE_HOME }}{% else %}{{ U_INDEX }}{% endif %}" title="{% if U_SITE_HOME %}{{ lang('SITE_HOME') }}{% else %}{{ lang('INDEX') }}{% endif %}">
```
3D-I: That should be

`<a id="logo" class="logo" href="{% if U_SITE_HOME %}{{ U_SITE_HOME }}{% else %}{{ U_INDEX }}{% endif %}" title="{% if U_SITE_HOME %}{{ L_SITE_HOME }}{% else %}{{ L_INDEX }}{% endif %}">`

because those 2 vars are automatically generated by the PHP's logic, there are not equivalents in any language file.

PR #5770, opened 2019-12-07
[ticket/16250] Add a service to check BBCodes safeness in ACP

JoshyPHP:
This PR is against 3.3.x, not sure whether this is considered a new feature.

This PR adds a new service used in the ACP. It doesn't change the ACP's behaviour but it changes what kinds of BBCodes produce an error.

PHPBB3-16250

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16250
https://area51.phpbb.com/phpBB/viewtopic.php?f=126&t=72401

### DIFF
--- a/phpBB/includes/acp/acp_bbcodes.php
+++ b/phpBB/includes/acp/acp_bbcodes.php
@@ -174,15 +174,15 @@ class acp_bbcodes
 
 				$acp_utils   = $phpbb_container->get('text_formatter.acp_utils');
 				$bbcode_info = $acp_utils->analyse_bbcode($bbcode_match, $bbcode_tpl);
-				$warn_unsafe = ($bbcode_info['status'] === 'unsafe');
+				$warn_unsafe = ($bbcode_info['status'] === $acp_utils::BBCODE_STATUS_UNSAFE);
 
-				if ($bbcode_info['status'] === 'invalid_definition')
-				{
-					trigger_error($user->lang['BBCODE_INVALID'] . adm_back_link($this->u_action), E_USER_WARNING);
-				}
-				if ($bbcode_info['status'] === 'invalid_template')
+				if ($bbcode_info['status'] === $acp_utils::BBCODE_STATUS_INVALID_TEMPLATE)
 				{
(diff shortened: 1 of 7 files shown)


phpBB/language/en/acp/posting.php
```diff
@@ -56,6 +56,7 @@
 
 	'BBCODE_INVALID_TAG_NAME'	=> 'The BBCode tag name that you selected already exists.',
 	'BBCODE_INVALID'			=> 'Your BBCode is constructed in an invalid form.',
+	'BBCODE_INVALID_TEMPLATE'	=> 'Your BBCode\'s template is invalid.',
```
marc1706: Use the utf8 version `’`

JoshyPHP: Changed.

phpBB/includes/acp/acp_bbcodes.php
```diff
@@ -172,14 +172,25 @@ function main($id, $mode)
 				);
 				extract($phpbb_dispatcher->trigger_event('core.acp_bbcodes_modify_create', compact($vars)));
 
-				$warn_text = preg_match('%<[^>]*\{text[\d]*\}[^>]*>%i', $bbcode_tpl);
+				$acp_utils   = $phpbb_container->get('text_formatter.acp_utils');
+				$bbcode_info = $acp_utils->analyse_bbcode($bbcode_match, $bbcode_tpl);
+				$warn_unsafe = ($bbcode_info['status'] === 'unsafe');
```
marc1706: I'm not the biggest fan of using strings for the status unless explicitly need. What do you think about adding constants to your interface instead and using these?

JoshyPHP: I don't think it really offers any benefits but if you think it's worth it please feel free to add it to the interface directly.

JoshyPHP: Actually, check out the latest commit and modify as desired. Without an enum type I don't think it really matters what the constants are but if the goal is to program defensively then I think it's a good idea to rework the conditionals as in d4d93160a7e40cd91b0816dcf52ed8365e1d8864. Instead of emitting an error if the status shows it's invalid/unsafe, it throws an error if the status is not valid.

PR #5778, opened 2019-12-17
[ticket/15233] Standardize avatar output variables

gijsmartens:
Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15233

### DIFF
--- a/phpBB/adm/style/acp_groups.html
+++ b/phpBB/adm/style/acp_groups.html
@@ -113,7 +113,7 @@
 		<legend>{L_GROUP_AVATAR}</legend>
 	<dl>
 		<dt><label>{L_CURRENT_IMAGE}{L_COLON}</label><br /><span>{L_AVATAR_EXPLAIN}</span></dt>
-		<dd>{AVATAR}</dd>
+		<dd>{% if AVATAR_HTML %}{{ AVATAR_HTML }}{% else %}<img src="{{ ADMIN_ROOT_PATH ~ 'images/no_avatar.gif' }}" alt="">{% endif %}</dd>
 		<dd><label for="avatar_delete"><input type="checkbox" name="avatar_delete" id="avatar_delete" /> {L_DELETE_AVATAR}</label></dd>
 	</dl>
 	<dl>
--- a/phpBB/includes/acp/acp_groups.php
+++ b/phpBB/includes/acp/acp_groups.php
@@ -721,8 +721,6 @@ class acp_groups
 					}
 				}
 
-				$group_avatar = $group_helper->get_avatar($group_row, 'GROUP_AVATAR', true);
-
 				if (isset($phpbb_avatar_manager) && !$update)
 				{
 					// Merge any avatar errors into the primary error array
@@ -741,6 +739,12 @@ class acp_groups
 						$u_back = $this->u_action;
 					break;
 				}
+
+				/** @var \phpbb\avatar\helper $avatar_helper */
+				$avatar_helper = $phpbb_container->get('avatar.helper');
+
+				$group_avatar = $avatar_helper->get_group_avatar($group_row, 'GROUP_AVATAR', true);
+				$template->assign_vars($avatar_helper->get_template_vars($group_avatar));
 
 				$template->assign_vars(array(
 					'S_EDIT'			=> true,
@@ -771,13 +775,6 @@ class acp_groups
 
 					'S_RANK_OPTIONS'		=> $rank_options,
 					'S_GROUP_OPTIONS'		=> group_select_options(false, false, (($user->data['user_type'] == USER_FOUNDER) ? false : 0)),
-					'AVATAR'				=> empty($group_avatar['html']) ? '<img src="' . $phpbb_admin_path . 'images/no_avatar.gif" alt="" />' : $group_avatar['html'],
-					'AVATAR_LAZY'			=> $group_avatar['lazy'],
-					'AVATAR_SRC'			=> $group_avatar['src'],
-					'AVATAR_TITLE'			=> $group_avatar['title'],
-					'AVATAR_TYPE'			=> $group_avatar['type'],
-					'AVATAR_WIDTH'			=> $group_avatar['width'],
-					'AVATAR_HEIGHT'			=> $group_avatar['height'],
 					'AVATAR_MAX_FILESIZE'	=> $config['avatar_filesize'],
 
 					'GROUP_TYPE_FREE'		=> GROUP_FREE,
--- a/phpBB/includes/acp/acp_groups.php
+++ b/phpBB/includes/acp/acp_groups.php
@@ -721,7 +721,7 @@ class acp_groups
 					}
 				}
 
-				$avatar = phpbb_get_group_avatar($group_row, 'GROUP_AVATAR', true);
+				$group_avatar = $group_helper->get_avatar($group_row, 'GROUP_AVATAR', true);
 
 				if (isset($phpbb_avatar_manager) && !$update)
 				{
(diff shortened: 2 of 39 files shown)


phpBB/includes/acp/acp_users.php
```diff
@@ -1966,12 +1966,22 @@ function main($id, $mode)
...
+				$avatar_helper = $phpbb_container->get('avatar.helper');
+
+				$avatar = $avatar_helper->get_user_avatar($user_row, 'USER_AVATAR', true);
 
 				$template->assign_vars(array(
 					'S_AVATAR'	=> true,
 					'ERROR'			=> (!empty($error)) ? implode('<br />', $error) : '',
-					'AVATAR'		=> (empty($avatar) ? '<img src="' . $phpbb_admin_path . 'images/no_avatar.gif" alt="" />' : $avatar),
+
+					'AVATAR'			=> empty($avatar['html']) ? '<img src="' . $phpbb_admin_path . 'images/no_avatar.gif" alt="" />' : $avatar['html'],
```
hanakin: there is not to be any html in the backend this needs factored out. You should be instead returning a path to the generic avatar since that is how all avatars will work. In the template we will have img class=o-avatar t-avatar-{{avatar-type}} src={{avatar-src}} alt={{avatar-alt}}

phpBB/includes/functions.php
```diff
@@ -4138,7 +4149,13 @@ function page_header($page_title = '', $display_online_list = false, $item_id =
 		'RECORD_USERS'					=> $l_online_record,
 
 		'PRIVATE_MESSAGE_COUNT'			=> (!empty($user->data['user_unread_privmsg'])) ? $user->data['user_unread_privmsg'] : 0,
-		'CURRENT_USER_AVATAR'			=> phpbb_get_user_avatar($user->data),
+		'CURRENT_USER_AVATAR'			=> $avatar['html'],
```
hanakin: Should not need this given comment above

phpBB/includes/functions.php
```diff
@@ -4138,7 +4149,13 @@ function page_header($page_title = '', $display_online_list = false, $item_id =
...
+		'CURRENT_USER_AVATAR_LAZY'		=> $avatar['lazy'],
+		'CURRENT_USER_AVATAR_SOURCE'	=> $avatar['src'],
+		'CURRENT_USER_AVATAR_TITLE'		=> $avatar['title'],
+		'CURRENT_USER_AVATAR_TYPE'		=> $avatar['type'],
+		'CURRENT_USER_AVATAR_WIDTH'		=> $avatar['width'],
```
hanakin: I do not think we really need width and height? Here width and height is more for upload restrictions not for display. The dimensions will be barcodes by the classes for all use cases

phpBB/includes/acp/acp_groups.php
```diff
@@ -771,10 +771,14 @@ function main($id, $mode)
 
 					'S_RANK_OPTIONS'		=> $rank_options,
 					'S_GROUP_OPTIONS'		=> group_select_options(false, false, (($user->data['user_type'] == USER_FOUNDER) ? false : 0)),
-					'AVATAR'				=> empty($avatar) ? '<img src="' . $phpbb_admin_path . 'images/no_avatar.gif" alt="" />' : $avatar,
+					'AVATAR'				=> empty($group_avatar['html']) ? '<img src="' . $phpbb_admin_path . 'images/no_avatar.gif" alt="" />' : $group_avatar['html'],
+					'AVATAR_LAZY'			=> $group_avatar['lazy'],
```
hanakin: What’s lazy?

phpBB/phpbb/avatar/helper.php
```diff
@@ -0,0 +1,221 @@
...
+	 *
+	 * Created for Backwards Compatibility (BC).
+	 * Styles should generate their own HTML element instead.
+	 *
+	 * @deprecated 4.1.0				After admin style is reworked aswell
+	 *
+	 * @param array		$data			The avatar data array
+	 * @return string					The avatar's HTML <img> element
+	 */
+	protected function get_avatar_html(array $data)
```
hanakin: Remove all of this again no html allowed in php

mrgoldy:
##### html
That is still needed as the avatars should also be displayed in the ACP, and the `adm/style` *(the ACP style)* is as far as I am aware, not being done for 4.0. So we still need html being generated in php until then. I have already added a deprecated notice to it though.

#### lazy
Lazy is the indicator for *lazy loading*. Currently utilised by avatars inside notifications. Meaning that the element is created as `` and then jQuery swaps the sources around. Meaning that those images do not take up any *(read: minimal)* time during page loads.

#### Dimensions
If dimensions are no longer necessary in 4.0, then I would suggest doing that in a separate PR after this one. As that will take a lot more chances, including database migrations to drop the columns from the user table, avatar driver adjustements and possibly more adjustments. 

#### Variables
Currently I am just assigning the variables to template that are available from the back-end, so the front-end can do with them what it wants and has all options available to it. You mentioned that you wanted individual variables, that's why I did it like this. It's also simply possible to assign an array to the template in a single variable.

#### Function
I was already under the assumption that you would be creating a macro or a block for things like these, where you can just input the variables and you get the desired outcome.
Speaking of functions, a while back I've added an `{{ avatar() }}` twig extension that is a wrapper for `phpbb_get_user_avatar()`, not knowing / having a full understanding all of this (the full overhaul). That mostlikely has to go then and probably be replaced with either a macro or something else.

mrgoldy:
I am not sure what you mean by the last paragraph. What is it you want me to change?

hanakin:
Sorry nothing pending having one of the other team members review it it looks good to me. I wa just saying I do not think a twig function is nessecary

phpBB/adm/style/acp_groups.html
```diff
@@ -113,7 +113,7 @@ <h3>{L_WARNING}</h3>
 		<legend>{L_GROUP_AVATAR}</legend>
 	<dl>
 		<dt><label>{L_CURRENT_IMAGE}{L_COLON}</label><br /><span>{L_AVATAR_EXPLAIN}</span></dt>
-		<dd>{AVATAR}</dd>
+		<dd>{% if AVATAR_HTML %}{{ AVATAR_HTML }}{% else %}<img src="{{ ADMIN_ROOT_PATH ~ 'images/no_avatar.gif' }}" alt="">{% endif %}</dd>
```
rubencm, line 116: If I'm not wrong, this constant don't work with routes, you have to use web_root_path

phpBB/includes/functions.php
```diff
@@ -3579,6 +3579,8 @@ function phpbb_quoteattr($data, $entities = null)
 /**
 * Get user avatar
 *
+* @deprecated 4.0.0 Use \phpbb\avatar\helper::get_user_avatar() instead
```
rubencm, line 3582: Move this functions to functions_compatibility.php if you can

phpBB/adm/style/acp_groups.html
mrgoldy: It's set in the [adm_page_header](https://github.com/phpbb/phpbb/blob/master/phpBB/includes/functions_acp.php#L80), and seeing currently there are no routes and controllers possible within in control panels, this should be fine. And once CP's are moved, these template variables will have to be updated anyway, so then it will still work.

phpBB/includes/functions.php
mrgoldy: Okay, so simply move them to there? Will do.

hanakin:
just re-looking at this is there any reason we did not convert to TWIG syntax?

marc1706:
Rebased on top of latest master.

PR #5790, opened 2019-12-26
[ticket/16266] Fix argon2 driver issue for Sodium implementation

rxu:
Checklist:

- [x] Tests pass

<a href="https://tracker.phpbb.com/browse/PHPBB3-16266">PHPBB3-16266</a>.

### DIFF
--- a/phpBB/phpbb/passwords/driver/argon2i.php
+++ b/phpBB/phpbb/passwords/driver/argon2i.php
@@ -38,13 +38,14 @@ class argon2i extends base_native
 		parent::__construct($config, $helper);
 
 		/**
-		 * For Sodium implementation of argon2 algorithm, set special cost factor values (since PHP 7.4)
+		 * For Sodium implementation of argon2 algorithm (since PHP 7.4), set special value of 1 for "threads" cost factor
 		 * See https://wiki.php.net/rfc/sodium.argon.hash and PHPBB3-16266
 		 * Don't allow cost factors to be below default settings where possible
 		 */
-		$this->memory_cost = $this->is_sodium() ? max($memory_cost, 256 * 1024)	: max($memory_cost, 1024);
-		$this->threads     = $this->is_sodium() ? 1								: max($threads, 2);
-		$this->time_cost   = $this->is_sodium() ? max($time_cost, 3)			: max($time_cost, 2);
+		$this->memory_cost = max($memory_cost, PASSWORD_ARGON2_DEFAULT_MEMORY_COST);
+		$this->time_cost   = max($time_cost, PASSWORD_ARGON2_DEFAULT_TIME_COST);
+		$this->threads     = (defined('PASSWORD_ARGON2_PROVIDER') && PASSWORD_ARGON2_PROVIDER == 'sodium') ?
+									PASSWORD_ARGON2_DEFAULT_THREADS : max($threads, PASSWORD_ARGON2_DEFAULT_THREADS);
 	}
 
 	/**
--- a/phpBB/phpbb/passwords/driver/argon2i.php
+++ b/phpBB/phpbb/passwords/driver/argon2i.php
@@ -37,9 +37,11 @@ class argon2i extends base_native
 	{
 		parent::__construct($config, $helper);
 
-		// For Sodium implementation, set special cost factor values (since PHP 7.4)
-		// See https://wiki.php.net/rfc/sodium.argon.hash and PHPBB3-16266
-		// Otherwise don't allow cost factors to be below default settings
+		/**
+		 * For Sodium implementation of argon2 algorithm, set special cost factor values (since PHP 7.4)
+		 * See https://wiki.php.net/rfc/sodium.argon.hash and PHPBB3-16266
+		 * Don't allow cost factors to be below default settings where possible
+		 */
 		$this->memory_cost = ($this->is_sodium()) ? max($memory_cost, 256*1024)	: max($memory_cost, 1024);
 		$this->threads     = ($this->is_sodium()) ? 1							: max($threads, 2);
 		$this->time_cost   = ($this->is_sodium()) ? max($time_cost, 3)			: max($time_cost, 2);
--- a/phpBB/phpbb/passwords/driver/argon2i.php
+++ b/phpBB/phpbb/passwords/driver/argon2i.php
@@ -37,10 +37,21 @@ class argon2i extends base_native
 	{
 		parent::__construct($config, $helper);
 
-		// Don't allow cost factors to be below default settings
-		$this->memory_cost = max($memory_cost, 1024);
-		$this->threads     = max($threads,     2);
-		$this->time_cost   = max($time_cost,   2);
+		if ($this->is_sodium())
+		{
+			// For Sodium implementation, set special cost factor values (since PHP 7.4)
+			// See https://wiki.php.net/rfc/sodium.argon.hash and PHPBB3-16266
+			$this->memory_cost = max($memory_cost, 256*1024);
+			$this->threads     = 1;
+			$this->time_cost   = max($time_cost,   3);
+		}
+		else
+		{
+			// Otherwise don't allow cost factors to be below default settings
+			$this->memory_cost = max($memory_cost, 1024);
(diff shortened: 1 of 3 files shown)


rxu:
The patch can be implemented in several ways (using ternaries; no `is_sodium()` function; etc), so I'm open for suggestions :)

3D-I:
Ternaries imo, smth like 
```
$is_sodium		= (bool) (defined('PASSWORD_ARGON2_PROVIDER') && PASSWORD_ARGON2_PROVIDER == 'sodium');
$memory_cost	= $is_sodium ? max($memory_cost, 256 * 1024) : max($memory_cost, 1024);
$threads		= $is_sodium ? 1 : max($threads, 2);
$time_cost		= $is_sodium ? max($time_cost, 3) : max($time_cost, 2);

$this->memory_cost = $memory_cost;
$this->threads = $threads;
$this->time_cost = $time_cost;
```

rxu:
@marc1706 ^

marc1706:
I think the question comes down to whether we'd need this for other drivers. If the answer to that is no then I'd go with the proposed ternary.

rxu:
Well, currently it's only argon2 but who knows what they will implement in the future.
I can go with alternative variant leaving `is_sodium` but using ternary instead of if-else.

phpBB/phpbb/passwords/driver/base_native.php
```diff
@@ -57,6 +57,25 @@ public function hash($password)
...
+	*
+	* @link https://wiki.php.net/rfc/sodium.argon.hash
+	*
+	* @return bool
+	*/
+	public function is_sodium()
+	{
+		static $is_sodium;
+
+		if (empty($is_sodium))
```
rubencm: I would use isset better instead of empty for undefined variables

rxu: Makes perfect sense for bool var. Good catch.

phpBB/phpbb/passwords/driver/argon2i.php
```diff
@@ -37,21 +37,12 @@ public function __construct(\phpbb\config\config $config, helper $helper, $memor
...
-		else
-		{
-			// Otherwise don't allow cost factors to be below default settings
-			$this->memory_cost = max($memory_cost, 1024);
-			$this->threads     = max($threads,     2);
-			$this->time_cost   = max($time_cost,   2);
-		}
+		// For Sodium implementation, set special cost factor values (since PHP 7.4)
+		// See https://wiki.php.net/rfc/sodium.argon.hash and PHPBB3-16266
+		// Otherwise don't allow cost factors to be below default settings
```
3D-I: Multiline comment best included in a docblock

```
		/**
		 * Set special cost factor values (since PHP 7.4) for Sodium implementation.
		 * @see https://wiki.php.net/rfc/sodium.argon.hash		Documentation.
		 * @see https://tracker.phpbb.com/browse/PHPBB3-16266	Tracker ticket.
		 *
		 * Don't allow cost factors to be below default settings otherwise.
		 */
```

rxu: It's not a function.

3D-I: `//` should be used for one/two-liners, `/* */` comment blocks otherwise.

phpBB/phpbb/passwords/driver/argon2i.php
```diff
@@ -37,21 +37,12 @@ public function __construct(\phpbb\config\config $config, helper $helper, $memor
...
+		$this->memory_cost = ($this->is_sodium()) ? max($memory_cost, 256*1024)	: max($memory_cost, 1024);
```
mrgoldy: Do we really need the redundant `()` around `$this->is_sodium()` ?
And I think there should be spaces between `256*1024`.

marc1706: Shouldn't it be possible to use the default constants that were introduced in PHP, e.g. `PASSWORD_ARGON2_DEFAULT_MEMORY_COST`? IIRC that shouldn't be an issue as this portion of the code shouldn't be exected if we're using an unsupported PHP version due to lazy loading? Could you check that maybe or should I do that?

rxu:
It turns out sodium defines standard argon2 constants with sodium required values https://github.com/php/php-src/blob/master/ext/sodium/sodium_pwhash.c#L193-L195
So the only problem to solve with sodium is to fix `threads` factor at its default value.

3D-I:
Looks good. :+1:  I can test it with Sodium and PHP 7.4 later on today.

rxu:
@3D-I Not tested myself yet, but travis tests did and now fail 😃

3D-I:
> 
> 
> @3D-I Not tested myself yet, but travis tests did and now fail smiley

Nah, travis and appveyor are failing since hours, see my 2 other PRs :smiley_cat: 
I will test it then.

rxu:
Yeah but tests fail exactly on password drivers as well.

3D-I:
I see now

```
1) phpbb_passwords_helper_test::test_helper_encode64 with data set #0 ('foobar', 6, 'axqPW3aQ')

Use of undefined constant PASSWORD_ARGON2_DEFAULT_MEMORY_COST - assumed 'PASSWORD_ARGON2_DEFAULT_MEMORY_COST' (this will throw an Error in a future version of PHP)

/home/travis/build/phpbb/phpbb/phpBB/phpbb/passwords/driver/argon2i.php:45

/home/travis/build/phpbb/phpbb/tests/passwords/drivers_test.php:26
```

Should the test be amended?

rxu:
Probably yes for unsupported PHP versions, but it also fails on PHP 7.3&7.4, where tests should be adjusted with argon2 predefined constants.

3D-I:
Argon2 has been introduced in PHP 7.2?

rxu:
As default hashing algorithm - yes. But I guess it was optionally available (via PECL) before.

3D-I:
Strange enough. I installed 3.3.0-rc1 on PHP 7.4.1 and Sodium without this PR - on Windows.

immagine


immagine

rxu:
Native implementation has a priority over Sodium, so on Windows it's unlikely Sodium can be tested as a standalone argon2 provider.

3D-I:
I can enable sodium on our live server, I will do later on tonight though.

rxu:
Native `libargon` should be unloaded for Sodium to load its own argon2 implementation.

3D-I:
Well, I just enabled the extension in php.ini. Honestly I don't know hot to unload libargon.

Perhaps is better to ask the reporter to do that test?

rxu:
@marc1706 It seems that in some circumstances argon2 default values `m/t/p` arent't `1024/2/2`, that's why tests fail on `needs_rehash()`.
F.e. on my localhost (PHP 7.2.26) they're `65536/4/2`. I'm not sure if it's possible but it looks like PHP sets it dynamically depending on the system resources.

rxu:
I'm out of ideas for now. TravisCI 7.2 test fails for whatever reason.

3D-I:
Roll back to https://github.com/phpbb/phpbb/pull/5790/commits/d000717d341a2c12099b0fba3ab677bbb0f2340c ?

marc1706:
Check if the constants are defined for memory cost and time cost as well. No clue why this happens but I guess that should fix it.

rxu:
All failures fail on `$this->login();` of the functional test case. I've tried to find out the reason but failed 😈

marc1706:
Well, that's one way to solve it :man_shrugging: :+1:

CrazyC0d3r:
As implemented, a phpBB 3.3 board using PHP 7.3 has a very real chance of invalidating every Argon2 password if PHP is updated to 7.4. Lanes (which you're incorrectly call threads) MUST be hardcoded to 1 and removed from the services_password.yml file. You are also setting excessively low memory usage values which will render the hashes less secure than had bcrypt been used.

The following archive contains all the files necessary to add proper Argon2id support to both phpBB 3.2 and 3.3: https://forum.dion-designs.com/docs/argon2id.zip

PR #5794, opened 2019-12-27
[ticket/16273] Check whether the index exists in memberlist - PHP 7.4

3D-I:
PHPBB3-16273

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16273

### DIFF
--- a/phpBB/memberlist.php
+++ b/phpBB/memberlist.php
@@ -656,8 +656,8 @@ switch ($mode)
 
 		$result = $db->sql_query($sql);
 		$row = $db->sql_fetchrow($result);
-		$foe = $row['foe'] ?? false;
-		$friend = $row['friend'] ?? false;
+		$foe = (bool) $row['foe'] ?? false;
+		$friend = (bool) $row['friend'] ?? false;
 		$db->sql_freeresult($result);
 
 		if ($config['load_onlinetrack'])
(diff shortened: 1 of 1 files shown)


rubencm:
phpBB 3.3 should be compatible with php7.3 too

3D-I:
> 
> 
> phpBB 3.3 should be compatible with php7.3 too

Tested also with PHP 7.3.5, am I missing something?

rubencm:
I'm not sure, but don't you have to cast it to a boolean now?, since database return an integer

rubencm:
Somehow I think this will fail sooner or later. @3D-I 

PHP 7.3:
```php
false['']; // return null, no error
```

PHP 7.4:
```php
false['']; // php notice: "Trying to access array offset on value of type bool"
false['']??false; // return false, no error
```

I propose change it to this:
```php
$var = $row ? (bool) $row['foo'] : false;
```

3D-I:
What about?

```
if ($row)
{
	$foe = (bool) $row['foe'];
	$friend = (bool) $row['friend'];
}
```

rubencm:
Yes, is the same

3D-I:
Dunno what to do here, should I open a new ticket? :thinking:

rubencm:
Probably the same is ok

3D-I:
Nope, your approach is ok.

PR #5798, opened 2019-12-29
[ticket/16277] Move from each() function

rxu:
Checklist:

- [x] Tests pass

<a href="https://tracker.phpbb.com/browse/PHPBB3-16277">PHPBB3-16277</a>.

### DIFF
--- a/phpBB/includes/diff/engine.php
+++ b/phpBB/includes/diff/engine.php
@@ -285,7 +285,7 @@ class diff_engine
 				$matches = $ymatches[$line];
 
 				reset($matches);
-				while (list(, $y) = each($matches))
+				while ($y = current($matches))
 				{
 					if (empty($this->in_seq[$y]))
 					{
@@ -296,7 +296,7 @@ class diff_engine
 				}
 
 				// no reset() here
-				while (list(, $y) = each($matches))
+				while ($y = current($matches))
 				{
 					if ($y > $this->seq[$k - 1])
 					{
--- a/phpBB/includes/diff/engine.php
+++ b/phpBB/includes/diff/engine.php
@@ -287,6 +287,7 @@ class diff_engine
 				reset($matches);
 				while ($y = current($matches))
 				{
+					next($matches);
 					if (empty($this->in_seq[$y]))
 					{
 						$k = $this->_lcs_pos($y);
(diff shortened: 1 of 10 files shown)


3D-I:
`ROOT\phpbb\request\type_cast_helper.php` at line 109, is it ok?

rxu:
@3D-I ?

3D-I:
I mean this line is ok or should it be fixed?

https://github.com/phpbb/phpbb/blob/036871cd9535503f1f7456096e6368544bdd8fc5/phpBB/phpbb/request/type_cast_helper.php#L109

rxu:
Ah. It is alredy in this PR? https://github.com/phpbb/phpbb/pull/5798/files#diff-e45a34a00aaa203b1577640fbed16597R86-R124

3D-I:
> **Edit:** sorry @rxu I missed you already did it.

Yeah sorry, I saw it 10 minutes ago.

phpBB/includes/diff/engine.php
```diff
@@ -285,7 +285,7 @@ function _diag($xoff, $xlim, $yoff, $ylim, $nchunks)
 				$matches = $ymatches[$line];
 
 				reset($matches);
-				while (list(, $y) = each($matches))
+				while ($y = current($matches))
```
marc1706: That's not the equivalent. `current()` will not move the pointer and therefore will always return the first element in this case.

phpBB/includes/diff/engine.php
```diff
@@ -296,7 +296,7 @@ function _diag($xoff, $xlim, $yoff, $ylim, $nchunks)
 				}
 
 				// no reset() here
-				while (list(, $y) = each($matches))
+				while ($y = current($matches))
```
marc1706: Same as above.

phpBB/includes/diff/engine.php
marc1706: I think a simple `foreach ($matches as $y)` would be the correct replacement.

mrgoldy:
You can also have a look at my CP PR's, as I've fixed all of these in there already.

rxu:
Thanks. I just was under impression that `current()` moves array pointer forward. I guess I misread the manual.

phpBB/includes/diff/engine.php
rxu: The issue with `foreach` is that this 1st `while` loop is to be broken on `empty($this->in_seq[$y])` and the 2nd `while` loop (right below) starts with no pointer reset from the current array position.
I wonder what's gonna happen if we use `foreach` for both.

rxu: The logic should be restored now.

PR #5815, opened 2020-01-03
[ticket/16234] Fix & enhance Sphinx search backend functionality

KYPREO:
Addresses multiple reported issues with search functionality when using Sphinx search backend and adds new functionality by taking advantage of Sphinx's extended query syntax and advanced configuration options.

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16234
https://tracker.phpbb.com/browse/PHPBB3-16233
https://tracker.phpbb.com/browse/PHPBB3-15367
https://tracker.phpbb.com/browse/PHPBB3-13958


KYPREO:
I'm not sure what happened here. I rebased my forked repository from the main 3.2.x branch but it still did not pick up my change to the same files from ticket 16269 (adding the line html_strip = 1), so the changes could not be automatically merged.

rxu:
You need to rebase your local `ticket/16234` branch on the top of your **updated** local `3.2.x` branch and then force push to your remote.
Alternatively, you can recreate `ticket/16234` branch again off the **updated** `3.2.x`.

KYPREO:
I saw the failed test due to trailing whitespaces. Thanks very much. I'll fix it all up.

phpBB/phpbb/search/fulltext_sphinx.php
```diff
@@ -331,11 +331,16 @@ protected function config_generate()
 				array('docinfo',					'extern'),
 				array('morphology',					'none'),
 				array('stopwords',					''),
+				array('wordforms',					'  # optional, specify path to wordforms file. See ./docs/sphinx_wordforms.txt for example'),
```
mrgoldy: Should these comments be inside the strings? 
Wouldn't it be more logical to have them behind them?
```
array('wordforms', ''), // optional, specify path to wordforms file. See ./docs/sphinx_wordforms.txt for example
```

phpBB/phpbb/search/fulltext_sphinx.php
```diff
@@ -430,22 +435,27 @@ protected function config_generate()
 	*/
 	public function split_keywords(&$keywords, $terms)
 	{
+		// Keep quotes and new lines
+		$keywords = str_replace(array('&quot;', "\n"), array('"', ' '), trim($keywords));
```
mrgoldy: If possible, please use the new array syntax `[]` instead of the old `array()`.
For here and other edited / added code below.

phpBB/phpbb/search/fulltext_sphinx.php
```diff
@@ -430,22 +435,27 @@ protected function config_generate()
...
+			// Replaces verbal operators OR and NOT with special characters | and -, unless appearing within quotation marks
+			$match		= array('#\sor\s(?=([^"]*"[^"]*")*[^"]*$)#i', '#\snot\s(?=([^"]*"[^"]*")*[^"]*$)#i');
+			$replace	= array(' | ', ' -');
 
 			$keywords = preg_replace($match, $replace, $keywords);
 			$this->sphinx->SetMatchMode(SPH_MATCH_EXTENDED);
 		}
 		else
 		{
+			$match = array ( '\\', '(',')','|','!','@','~', '/', '^', '$', '=','&amp;', '&lt;', '&gt;');
```
mrgoldy: This should be without spacing, `array('\\',` instead of `array ( '\\',`.
And ideally with new array syntax `[]`.

phpBB/phpbb/search/fulltext_sphinx.php
```diff
@@ -430,22 +435,27 @@ protected function config_generate()
...
+			$replace = array ( ' ', ' ', ' ', ' ',' ',' ',' ', ' ', ' ', ' ', ' ', ' ', ' ', ' ');
+
+			$keywords = str_replace ($match, $replace, $keywords);
```
mrgoldy: Remove the space in `str_replace (`.

phpBB/phpbb/search/fulltext_sphinx.php
KYPREO: They are inside strings because the sample configuration file is generated within a window in the ACP (for the user to copy and paste into a configuration file) and so the comments need to appear there (rather than within fulltext_sphinx.php).

phpBB/phpbb/search/fulltext_sphinx.php
```diff
@@ -455,6 +465,49 @@ public function split_keywords(&$keywords, $terms)
...
+	* (b) proximity search: "hello world"~10 -- finds hello and world within 10 words of each other. Number can be between 1 and 99.
+	* (c) strict word order: aaa << bbb << ccc -- finds "aaa" only where it appears before "bbb" and only where "bbb" appears before "ccc".
+	* (d) exact match operator: if lemmatizer or stemming enabled, search will find exact match only and ignore other grammatical forms of the same word stem..
+	* eg raining =cats and =dogs -- will not return "raining cat and dog" 
+	* eg ="search this exact phrase" -- will not return "searched this exact phrase", "searching these exact phrases".
+	* 5. Special characters /, ~, << and = not complying with the correct syntax and other reserved operators are escaped and searched literally.
+	* Special characters not explicitly listed in charset_table or blend_chars in sphinx.conf will not be indexed and keywords containing them will be ignored by Sphinx.
+	* By default, only $, %, & and @ characters are indexed and searchable.
+	* String transformation is in backend only and not visible to the end user nor reflected in the results page URL or keyword highlighting.
+	*/
```
mrgoldy: Props on the well detailed docblock, makes a lot clear.
Some sentences are however a bit long and it's missing the `@param` and `@return` statement.

Perhaps something like this is a possibility:
```
/**
 * Clean search string.
 *
 * Cleans search query passed into Sphinx search engine, as follows:
 * 1. Hyphenated words are replaced with keyword search for either the exact phrase with spaces
 *    or as a single word without spaces eg search for "know-it-all" becomes ("know it all"|"knowitall*")
 * 2. Words with apostrophes are contracted eg "it's" becomes "its"
 * 3. <, >, " and & are decoded from HTML entities.
 * 4. Following special characters used as search operators in Sphinx are preserved when used with correct syntax:
 *    (a) quorum matching: "the world is a wonderful place"/3
 *        Finds 3 of the words within the phrase. Number must be between 1 and 9.
 *    (b) proximity search: "hello world"~10
 *        Finds hello and world within 10 words of each other. Number can be between 1 and 99.
 *    (c) strict word order: aaa << bbb << ccc
 *        Finds "aaa" only where it appears before "bbb" and only where "bbb" appears before "ccc".
 *    (d) exact match operator: if lemmatizer or stemming enabled,
 *        search will find exact match only and ignore other grammatical forms of the same word stem..
 *        eg. raining =cats and =dogs
 *            will not return "raining cat and dog"
 *        eg. ="search this exact phrase"
 *            will not return "searched this exact phrase", "searching these exact phrases".
 * 5. Special characters /, ~, << and = not complying with the correct syntax
 *    and other reserved operators are escaped and searched literally.
 *    Special characters not explicitly listed in charset_table or blend_chars in sphinx.conf
 *    will not be indexed and keywords containing them will be ignored by Sphinx.
 *    By default, only $, %, & and @ characters are indexed and searchable.
 *    String transformation is in backend only and not visible to the end user
 *    nor reflected in the results page URL or keyword highlighting.
 *
 * @param string	$search_string
 * @return string
 */
```

phpBB/phpbb/search/fulltext_sphinx.php
mrgoldy: Oh okay, that makes perfect sense! Cheers.

phpBB/phpbb/search/fulltext_sphinx.php
KYPREO: Thank you. I thought it might be a bit wordy but I wanted to be transparent about what I was trying to achieve. I will take up all of your suggestions and make some further changes.

rxu:
I guess merge commit shouldn't present in PR's, so rebasing PR or squashing its commits would be a good idea.

KYPREO:
> I guess merge commit shouldn't present in PR's, so rebasing PR or squashing its commits would be a good idea.

I am happy to do whatever is required to ensure the pull request is in order (assuming there is in fact a problem), but there is the potential to make matters worse by attempting to squash the commits (and I'd lose the history above).  Apparently, the repository manager can squash all the commits on the pull request through the "Squash & merge" option, which might be the neater, more foolproof option here?

KYPREO:
Just checking in to see whether I am required to do anything further on this, or whether it's just in the queue to be reviewed by the phpBB dev team in due course.

mrgoldy:
I believe you'll have to remove your own copyright, as it can not be merged like that.
But indeed, you'll have to wait for a further review / status update.

KYPREO:
> I believe you'll have to remove your own copyright, as it can not be merged like that.
> But indeed, you'll have to wait for a further review / status update.

Thanks mrgoldy.  My logic on the copyright notice on the supplied text files was that the douments do not actually form part of phpBB code. They are effectively third party add-ons I have made from scratch for use by another third party program, Sphinx Search. It is therefore like vendor files and indeed the supplied Sphinx API (sphinxapi.php).

But as you suggest, I'll wait for a further view. No doubt there may be other matters I need to address as well.

phpBB/phpbb/search/fulltext_sphinx.php
```diff
@@ -430,22 +435,27 @@ protected function config_generate()
...
+			// Replaces verbal operators OR and NOT with special characters | and -, unless appearing within quotation marks
+			$match		= ['#\sor\s(?=([^"]*"[^"]*")*[^"]*$)#i', '#\snot\s(?=([^"]*"[^"]*")*[^"]*$)#i'];
+			$replace	= [' | ', ' -'];
 
 			$keywords = preg_replace($match, $replace, $keywords);
 			$this->sphinx->SetMatchMode(SPH_MATCH_EXTENDED);
 		}
 		else
 		{
+			$match = ['\\', '(',')','|','!','@','~', '/', '^', '$', '=','&amp;', '&lt;', '&gt;'];
```
marc1706:
```suggestion
			$match = ['\\', '(',')', '|', '!', '@', '~', '/', '^', '$', '=', '&amp;', '&lt;', '&gt;'];
```

phpBB/phpbb/search/fulltext_sphinx.php
```diff
@@ -430,22 +435,27 @@ protected function config_generate()
...
+			$replace = [' ', ' ', ' ', ' ',' ',' ',' ', ' ', ' ', ' ', ' ', ' ', ' ', ' '];
+
+			$keywords = str_replace($match, $replace, $keywords);
```
marc1706: As you're replacing all of thise with an empty string, I think you should rather use this:
```suggestion
			$keywords = str_replace($match, ' ', $keywords);
```
And get rid of the `$replace` array.

marc1706:
With regards to the copyright information I also think that the current copyright statement is not in line with our own licensing.

First of, you need to consider that other third party files are actually kept in a separate folder and each carry their own copyright & license information in their respective folders. Their source code is also not kept in the phpBB repository.

Yes, one might consider sample files for a separate program - in this case Sphinx search - as independent work. However, if we distribute these not as separate work but as part of phpBB, then they IMHO need to be covered under our license.
Again, the aggregation of third party code like the vendor system we use is a distinctively different than adding the sample files to phpBB.

Even if this does not apply, files would at least need to be compatible with phpBB's license, e.g. as listed here http://gplv3.fsf.org/wiki/index.php/Compatible_licenses#GPLv2-compatible_licenses

However, putting your copyright info into the files and not specifying any further license effectively marks them as proprietary and hence causes them to be incompatible.

KYPREO:
Thanks for the input Marc. I suppose I could include a GPLv2 licence but I think it's just easier and less controversial to remove the copyright notice altogether. I pushed this change through.

One final thought.  Those users who already have Sphinx configured may need to update their configuration files to fully take advantage of these change. This file resides outside of phpBB and so there is no way of changing the file through a phpBB update.  For example, the phpBB community board has not updated its sphinx.conf file as the search index still includes HTML markup, even though this was fixed in 3.2.9. 

Perhaps a notice might be included in the official release notes/announcement for the next version to inform Sphinx users of the need to compare their existing configuration file to the new one and adopt changes they consider necessary?

marc1706:
Thanks. I'll try to keep this in mind and mention the need to update the config file in the release topic.

PR #5827, opened 2020-01-11
[ticket/16314] Fix database connections on controllers

3D-I:
PHPBB3-16314

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-16314

### DIFF
--- a/phpBB/includes/functions.php
+++ b/phpBB/includes/functions.php
@@ -4493,6 +4493,20 @@ function page_footer($run_cron = true, $display_template = true, $exit_handler =
 		$template->display('body');
 	}
 
+	/**
+	* Do not lose database connection on controllers.
+	* @todo verify if garbage_collection() still is needed.
+	*/
+	global $symfony_request;
+
+	$script_name = $symfony_request->getScriptName();
+	$page_name = substr($script_name, -1, 1) == '/' ? '' : utf8_basename($script_name);
+
+	if ($page_name !== 'app.php')
+	{
+		garbage_collection();
+	}
+
 	if ($exit_handler)
 	{
 		exit_handler();
--- a/phpBB/includes/functions.php
+++ b/phpBB/includes/functions.php
@@ -4493,8 +4493,6 @@ function page_footer($run_cron = true, $display_template = true, $exit_handler =
 		$template->display('body');
 	}
 
-	garbage_collection();
-
 	if ($exit_handler)
 	{
 		exit_handler();


3D-I:
Not sure if I have to move it to

```
		if ($task)
		{
			$url = $task->get_url();
			$template->assign_var('RUN_CRON_TASK', '<img src="' . $url . '" width="1" height="1" alt="cron" />');

			garbage_collection();
		}
```

I cqan do it in case.

3D-I:
!set WIP

3D-I:
!unset WIP

marc1706:
IMHO this is just a hack for a rather bigger problem: Why are controllers even calling `page_footer()`?
The controller helper `render()` method calls it:
```
page_footer(true, false, false);
```
Which means cron is run, but the template is not displayed and the exit_handler() is not called.
As a result of that, the following tasks are executed in `page_footer()`:
- cron image link is generated
- sql report is output
- call to `garbage_collection()` (hence closing SQL connection)

On controller pages, we currently have two events that are being triggered on kernel terminate, hence after the page is output:
- cron_runner_listener: only executed on cron pages. This might try to access the sql connection which will already be closed as `page_footer()` was called in the controller helper
- kernel_terminate_subscriber: executed on all pages. Calls only `exit_handler()`, hence what was skipped in the controller helper `render()` method.

Based on this, I think the correct solution should be:
- Stop calling `page_footer()` from the controller helper `render()` method
- Add method for outputting cron image link and sql report in controller helper
- Add SQL close call to kernel_terminate_subscriber

rxu:
@marc1706 👍 
1st has an obvious implementation.
2nd can be implemented either as a controller helper method or via refactoring relevant code into method from the `page_footer()`.
3rd can be done simply by adding `garbage_collection()` call before the `exit_handler()` call here https://github.com/phpbb/phpbb/blob/3.3.x/phpBB/phpbb/event/kernel_terminate_subscriber.php#L32

rxu:
Applied the solution path locally and it seems to be solving the issue.

rxu:
Or even better, page_footer() already has an injection of controller_helper service. So the whole call cron script can be transferred to the method in controller helper and then called in `render()` as well as in `page_footer()`.
Also I wonder if it's worth to change the [cron related div](https://github.com/phpbb/phpbb/blob/3.3.x/phpBB/styles/prosilver/template/overall_footer.html#L61-L64) in `overall_footer.html` to something like
```html
<!-- IF RUN_CRON_TASK and not S_IS_BOT-->
<div>
	<a id="bottom" class="anchor" accesskey="z"></a>
	{RUN_CRON_TASK}
</div>
<!-- ENDIF -->
```
because now the div is rendered even if no cron is getting called.

3D-I:
Thanks guys, 
Am I correct understanding that the PR https://github.com/phpbb/phpbb/pull/5831 should replace this one?

rxu:
> Am I correctly understanding that the PR #5831 should replace this one?

Yeah, I guess so.

3D-I:
Closing then :+1:

PR #5831, opened 2020-01-19
[ticket/16314] Improve handling of page_footer output for controllers

marc1706:
Moved the common template output to controller helper and added
garbage_collection handling to kernel_terminate_subscriber.

PHPBB3-16314

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16314


phpBB/includes/functions.php
```diff
@@ -4403,8 +4403,7 @@ function phpbb_generate_debug_output(\phpbb\db\driver\driver_interface $db, \php
 */
 function page_footer($run_cron = true, $display_template = true, $exit_handler = true)
 {
-	global $db, $config, $template, $user, $auth, $cache, $phpEx;
-	global $request, $phpbb_dispatcher, $phpbb_admin_path;
+	global $phpbb_dispatcher, $phpbb_container, $template;
```
mrgoldy: Why is the entire function not moved to the controller helper, or preferably even a different file within the controller namespace such as `phpb\controller\functions` and called from the `$helper->render()` function. Other 'old functions' that still call `page_footer()` can just call that object & function instead.
As we are going to move more and more to controllers, so will be getting more and more functions for the controller aswell. So moving them to a separate file will help with maintainability.

phpBB/phpbb/controller/helper.php
```diff
@@ -22,53 +34,101 @@
...
+	/** @var dispatcher */
+	protected $dispatcher;
+
+	/** @var language */
+	protected $language;
+
 	/**
 	* Template object
-	* @var \phpbb\template\template
+	* @var template
```
mrgoldy: Any change we could update all these docblocks from here downwards aswell to use `/** class */`, lines rather than blocks with double `**`.

phpBB/phpbb/controller/helper.php
```diff
@@ -22,53 +34,101 @@
...
+		$this->db = $db;
+		$this->dispatcher = $dispatcher;
+		$this->language = $language;
 		$this->template = $template;
 		$this->user = $user;
 		$this->config = $config;
 		$this->symfony_request = $symfony_request;
 		$this->request = $request;
 		$this->routing_helper = $routing_helper;
+		$this->admin_path = $admin_path;
```
mrgoldy: Since you're injecting so many new services anyway and have to amend tests already, any chance we can reorder the existing services aswell. As in, aphabetically.

tests/pagination/pagination_test.php
```diff
@@ -58,8 +57,27 @@ public function setUp(): void
...
+			$db,
+			new phpbb_mock_event_dispatcher(),
+			new \phpbb\language\language(new \phpbb\language\language_file_loader($phpbb_root_path, $phpEx)),
+			$this->template,
+			$this->user,
+			$this->config,
+			$symfony_request,
+			$request,
+			$this->routing_helper,
+			'',
```
mrgoldy: I understand it's not needed here but why not send the correct admin path anyway?

phpBB/includes/functions.php
marc1706: I haven't moved the entire function as this is for 3.3 and would otherwise result in a BC break. IMHO this function should be removed / deprecated in master.

rxu:
The patch worked fine for me.

phpBB/phpbb/controller/helper.php
```diff
@@ -192,4 +243,86 @@ public function get_current_url()
...
+	}
+
+	/**
+	 * Display SQL report
+	 *
+	 * @return void
+	 */
+	protected function display_sql_report()
+	{
+		if ($this->sql_explain && $this->request->variable('explain', false) && $this->auth->acl_get('a_'))
```
Derky: This code is now in two places, also still in **functions.php** under the function name `phpbb_check_and_display_sql_report` https://github.com/phpbb/phpbb/blob/8bdad7d73e6c0876110bee033bfa382d16856c1d/phpBB/includes/functions.php#L4319

Maybe let that function call this one or add a deprecation message?

3D-I: Agreed :+1:

PR #5832, opened 2020-01-20
[ticket/16329] Add configuration options for Plupload

KYPREO:
Adds ACP options to configure the compression level used by Plupload (default 85%) and whether Exif metadata is stripped from image attachments (default yes).

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16329
https://tracker.phpbb.com/browse/PHPBB3-16330

### DIFF
--- a/phpBB/includes/acp/acp_attachments.php
+++ b/phpBB/includes/acp/acp_attachments.php
@@ -170,7 +170,7 @@ class acp_attachments
 						'img_min_thumb_filesize'	=> array('lang' => 'MIN_THUMB_FILESIZE',	'validate' => 'int:0:999999999999999',	'type' => 'number:0:999999999999999', 'explain' => true, 'append' => ' ' . $user->lang['BYTES']),
 						'img_max'					=> array('lang' => 'MAX_IMAGE_SIZE',		'validate' => 'int:0:9999',	'type' => 'dimension:0:9999', 'explain' => true, 'append' => ' ' . $user->lang['PIXEL']),
 						'img_strip_metadata'		=> array('lang' => 'IMAGE_STRIP_METADATA',	'validate' => 'bool',	'type' => 'radio:yes_no', 'explain' => true),
-						'img_quality'				=> array('lang' => 'IMAGE_QUALITY',			'validate' => 'int:1:100',	'type' => 'number:1:100', 'explain' => true, 'append' => '&nbsp;&percnt;'),
+						'img_quality'				=> array('lang' => 'IMAGE_QUALITY',			'validate' => 'int:50:90',	'type' => 'number:50:90', 'explain' => true, 'append' => ' &percnt;'),
 						'img_link'					=> array('lang' => 'IMAGE_LINK_SIZE',		'validate' => 'int:0:9999',	'type' => 'dimension:0:9999', 'explain' => true, 'append' => ' ' . $user->lang['PIXEL']),
 					)
 				);
--- a/phpBB/includes/acp/acp_attachments.php
+++ b/phpBB/includes/acp/acp_attachments.php
@@ -169,6 +169,8 @@ class acp_attachments
 						'img_max_thumb_width'		=> array('lang' => 'MAX_THUMB_WIDTH',		'validate' => 'int:0:999999999999999',	'type' => 'number:0:999999999999999', 'explain' => true, 'append' => ' ' . $user->lang['PIXEL']),
 						'img_min_thumb_filesize'	=> array('lang' => 'MIN_THUMB_FILESIZE',	'validate' => 'int:0:999999999999999',	'type' => 'number:0:999999999999999', 'explain' => true, 'append' => ' ' . $user->lang['BYTES']),
 						'img_max'					=> array('lang' => 'MAX_IMAGE_SIZE',		'validate' => 'int:0:9999',	'type' => 'dimension:0:9999', 'explain' => true, 'append' => ' ' . $user->lang['PIXEL']),
+						'img_strip_metadata'		=> array('lang' => 'IMAGE_STRIP_METADATA',	'validate' => 'bool',	'type' => 'radio:yes_no', 'explain' => true),
+						'img_quality'				=> array('lang' => 'IMAGE_QUALITY',			'validate' => 'int:1:100',	'type' => 'number:1:100', 'explain' => true, 'append' => '&nbsp;&percnt;'),
 						'img_link'					=> array('lang' => 'IMAGE_LINK_SIZE',		'validate' => 'int:0:9999',	'type' => 'dimension:0:9999', 'explain' => true, 'append' => ' ' . $user->lang['PIXEL']),
 					)
 				);
--- a/phpBB/install/schemas/schema_data.sql
+++ b/phpBB/install/schemas/schema_data.sql
@@ -155,7 +155,7 @@ INSERT INTO phpbb_config (config_name, config_value) VALUES ('img_max_height', '
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('img_max_thumb_width', '400');
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('img_max_width', '0');
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('img_min_thumb_filesize', '12000');
-INSERT INTO phpbb_config (config_name, config_value) VALUES ('img_quality', '90');
+INSERT INTO phpbb_config (config_name, config_value) VALUES ('img_quality', '85');
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('img_strip_metadata', '1');
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('ip_check', '3');
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('ip_login_limit_max', '50');
--- a/phpBB/install/schemas/schema_data.sql
+++ b/phpBB/install/schemas/schema_data.sql
@@ -156,7 +156,7 @@ INSERT INTO phpbb_config (config_name, config_value) VALUES ('img_max_thumb_widt
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('img_max_width', '0');
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('img_min_thumb_filesize', '12000');
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('img_quality', '85');
-INSERT INTO phpbb_config (config_name, config_value) VALUES ('img_strip_metadata', '1');
+INSERT INTO phpbb_config (config_name, config_value) VALUES ('img_strip_metadata', '0');
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('ip_check', '3');
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('ip_login_limit_max', '50');
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('ip_login_limit_time', '21600');
--- a/phpBB/install/schemas/schema_data.sql
+++ b/phpBB/install/schemas/schema_data.sql
@@ -155,6 +155,8 @@ INSERT INTO phpbb_config (config_name, config_value) VALUES ('img_max_height', '
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('img_max_thumb_width', '400');
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('img_max_width', '0');
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('img_min_thumb_filesize', '12000');
+INSERT INTO phpbb_config (config_name, config_value) VALUES ('img_quality', '90');
+INSERT INTO phpbb_config (config_name, config_value) VALUES ('img_strip_metadata', '1');
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('ip_check', '3');
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('ip_login_limit_max', '50');
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('ip_login_limit_time', '21600');
--- a/phpBB/language/en/acp/attachments.php
+++ b/phpBB/language/en/acp/attachments.php
@@ -111,6 +111,10 @@ $lang = array_merge($lang, array(
 
 	'IMAGE_LINK_SIZE'			=> 'Image link dimensions',
 	'IMAGE_LINK_SIZE_EXPLAIN'	=> 'Display image attachment as an inline text link if image is larger than this. To disable this behaviour, set the values to 0px by 0px.',
+	'IMAGE_QUALITY'				=> 'Quality for image compression',
+	'IMAGE_QUALITY_EXPLAIN'		=> 'Higher value means lower compression/larger file size. Default 90%. Setting is only effective if max image dimensions set to value other than 0px v 0px, but will then apply to all uploaded JPEG images',
+	'IMAGE_STRIP_METADATA'		=> 'Strip image metadata',
+	'IMAGE_STRIP_METADATA_EXPLAIN'	=> 'Strip Exif metadata eg author name, GPS coordinates & camera details (JPEG only). Setting is only effective if max image dimensions set to value other than 0px v 0px.',
 
 	'MAX_ATTACHMENTS'				=> 'Maximum number of attachments per post',
 	'MAX_ATTACHMENTS_PM'			=> 'Maximum number of attachments per private message',
--- a/phpBB/language/en/acp/attachments.php
+++ b/phpBB/language/en/acp/attachments.php
@@ -111,10 +111,10 @@ $lang = array_merge($lang, array(
 
 	'IMAGE_LINK_SIZE'			=> 'Image link dimensions',
 	'IMAGE_LINK_SIZE_EXPLAIN'	=> 'Display image attachment as an inline text link if image is larger than this. To disable this behaviour, set the values to 0px by 0px.',
-	'IMAGE_QUALITY'				=> 'Quality for image compression',
-	'IMAGE_QUALITY_EXPLAIN'		=> 'Higher value means lower compression/larger file size. Default 90%. Setting is only effective if max image dimensions set to value other than 0px v 0px, but will then apply to all uploaded JPEG images',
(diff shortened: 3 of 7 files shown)


KYPREO:
I submitted this PR against the 3.2.x branch. If accepted, corresponding changes should also go into the 3.3.x branch. Should I create an additional commit and PR into the 3.3.x branch with the migration file for 3.3.0 to 3.3.1?

phpBB/phpbb/plupload/plupload.php
```diff
@@ -263,11 +263,24 @@ public function generate_resize_string()
...
+			}
+			else
+			{
+				$resize = sprintf(
+					'resize: {width: %d, height: %d, quality: %d},',
+					(int) $this->config['img_max_width'],
+					(int) $this->config['img_max_height'],
+					(int) $this->config['img_quality']
+				);
+			}
```
3D-I: Doesn't work this way?

```
$resize = sprintf(
	'resize: {width: %d, height: %d, quality: %d, preserve_headers: %s},',
	(int) $this->config['img_max_width'],
	(int) $this->config['img_max_height'],
	(int) $this->config['img_quality'],
	(bool) $this->config['img_strip_metadata']
);
```

phpBB/phpbb/db/migration/data/v32x/enable_plupload_config.php
```diff
@@ -0,0 +1,32 @@
...
+		return array(
+			'\phpbb\db\migration\data\v32x\v329',
+		);
+	}
+
+	public function update_data()
+	{
+		return array(
+			array('config.add', array('img_quality', '90')),
+			array('config.add', array('img_strip_metadata', '1')),
```
3D-I: Why true as per default? Current behaviour is does not strip them.

phpBB/phpbb/db/migration/data/v32x/enable_plupload_config.php
```diff
@@ -0,0 +1,32 @@
...
+*
+*/
+
+namespace phpbb\db\migration\data\v32x;
+
+class enable_plupload_config extends \phpbb\db\migration\migration
+{
+	static public function depends_on()
+	{
+		return array(
```
3D-I: Suggested using PHP short array syntax.

phpBB/install/schemas/schema_data.sql
```diff
@@ -155,6 +155,8 @@ INSERT INTO phpbb_config (config_name, config_value) VALUES ('img_max_height', '
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('img_max_thumb_width', '400');
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('img_max_width', '0');
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('img_min_thumb_filesize', '12000');
+INSERT INTO phpbb_config (config_name, config_value) VALUES ('img_quality', '90');
+INSERT INTO phpbb_config (config_name, config_value) VALUES ('img_strip_metadata', '1');
```
3D-I: Should be IMHO false per default.

3D-I:
`phpbb/tests/plupload/plupload_test.php` at line 55 needs attention.

KYPREO:
> `phpbb/tests/plupload/plupload_test.php` at line 55 needs attention.

Failing the plupload test was expected, as the test configuration does not include the value img_quality in phpbb_config which is required by the amended version of plupload.php I have submitted.

Do I need to amend the test itself so that the test passes and then push that updated test to my branch?

3D-I:
You should amend the test in your local fork and commit it into this same branch.

phpBB/phpbb/plupload/plupload.php
KYPREO: I looked at whether I could simplify this, so thanks for the suggestion.

I'd also need to invert the value, or invert the config setting (eg from strip metadata to preserve metadata). But in any event, it doesn't work. It is ok when preserve_headers: true, but when false the attachment function fails. After uploading the file, there is an extra empty file upload line and the image cannot be placed inline (see below).

failed upload

phpBB/phpbb/db/migration/data/v32x/enable_plupload_config.php
KYPREO: Metadata should be stripped by default. The current behaviour is wrong, as users and board administrators have not consented to sensitive metadata being uploaded, nor are they notified it is taking place.  This should be an opt-in feature, as preserving metadata is likely to be required for exceptional cases only.  Weighing up the value of perserving metadata vs the risk of inadvertent disclosure and misuse by bad actors, stripping metadata is the better option.  This is why most platforms that allow image uploading strip metadata by default. Boards that require it can enable the feature.

phpBB/phpbb/plupload/plupload.php
3D-I: Try something like, no need to compare the config since it is true if 1 false otherwise.

```
$my_var = $this->config['img_strip_metadata'] ? 'false' : 'true';

$resize = sprintf(
	'resize: {width: %d, height: %d, quality: %d, preserve_headers: %s},',
	(int) $this->config['img_max_width'],
	(int) $this->config['img_max_height'],
	(int) $this->config['img_quality'],
	$my_var
);
```

phpBB/phpbb/db/migration/data/v32x/enable_plupload_config.php
3D-I: Usually the old behaviour is kept though.

phpBB/phpbb/plupload/plupload.php
KYPREO: Yep, this way did it. Thanks very much for the tip.  I have amended and will push the change when I fix the plupload test.

KYPREO:
Plupload test updated to include new config options and tests now pass.

phpBB/phpbb/db/migration/data/v32x/enable_plupload_config.php
KYPREO: Yes normally it would make sense to preserve old default behaviour, however here the motivator for the PR was to change the old behaviour, as discussed in https://www.phpbb.com/community/viewtopic.php?f=436&t=2528276

phpBB/includes/acp/acp_attachments.php
```diff
@@ -169,6 +169,8 @@ function main($id, $mode)
 						'img_max_thumb_width'		=> array('lang' => 'MAX_THUMB_WIDTH',		'validate' => 'int:0:999999999999999',	'type' => 'number:0:999999999999999', 'explain' => true, 'append' => ' ' . $user->lang['PIXEL']),
 						'img_min_thumb_filesize'	=> array('lang' => 'MIN_THUMB_FILESIZE',	'validate' => 'int:0:999999999999999',	'type' => 'number:0:999999999999999', 'explain' => true, 'append' => ' ' . $user->lang['BYTES']),
 						'img_max'					=> array('lang' => 'MAX_IMAGE_SIZE',		'validate' => 'int:0:9999',	'type' => 'dimension:0:9999', 'explain' => true, 'append' => ' ' . $user->lang['PIXEL']),
+						'img_strip_metadata'		=> array('lang' => 'IMAGE_STRIP_METADATA',	'validate' => 'bool',	'type' => 'radio:yes_no', 'explain' => true),
+						'img_quality'				=> array('lang' => 'IMAGE_QUALITY',			'validate' => 'int:1:100',	'type' => 'number:1:100', 'explain' => true, 'append' => '&nbsp;&percnt;'),
```
3D-I: I'd remove that unbreakable space.

`' &percnt;'),` see for example how spaces are handled here `' ' . $user->lang['PIXEL']),`

phpBB/language/en/acp/attachments.php
```diff
@@ -111,6 +111,10 @@
 
 	'IMAGE_LINK_SIZE'			=> 'Image link dimensions',
 	'IMAGE_LINK_SIZE_EXPLAIN'	=> 'Display image attachment as an inline text link if image is larger than this. To disable this behaviour, set the values to 0px by 0px.',
+	'IMAGE_QUALITY'				=> 'Quality for image compression',
+	'IMAGE_QUALITY_EXPLAIN'		=> 'Higher value means lower compression/larger file size. Default 90%. Setting is only effective if max image dimensions set to value other than 0px v 0px, but will then apply to all uploaded JPEG images',
```
Crizz0: I don't like this 0px V 0 px wording. Maybe something more like "0px by 0py"?

Crizz0:
The initial idea to hard limit this to 85 was, that 100% increase the file size and threw the "file is to large" error during upload.

This is not prevented by this patch, so when I set the quality to 100 or  95 the problem will be present again?

phpBB/language/en/acp/attachments.php
KYPREO: Thanks for the pickup, yes that's a typo. If you mean it should read "0px by 0px" (like the line above) then I agree.

KYPREO:
As a sanity check, I just randomly sampled around 20 stock JPEG images with file size ranging from 10kB to 5MB but under the maximum allowed resoultion.  Using 90% image quality and preserving EXIF, all files were comfortably under the original size and I was therefore able to successfully upload with a test account (with ordinary user privileges). 

92-93% was also fine, but 94%+ produced larger files than original if the resolution was unchanged.

I am therefore confident 90% quality is a safe upper value.

For me, this has highlighted a catch with the old fix.  What if a person had a low resolution limit (eg 640x480) but large maximum attachment file size? They would be able to safely use a 100% quality figure, as most images would reduced in file size simply due to the resolution drop, and the 100% setting might help preserve image quality in this instance.  In reality, the error only appears where someone has a large maximum image dimension setting but relatively low maximum file attachment file size, which is a silly combination IMHO.

Nonetheless, I can see why we want to avoid displayed unexpected errors to users and until Plupload updates its image compression algorithm, I don't think there is a perfect solution.

Crizz0:
The problem for the UX is:
- limit is 450 kiB
- user uploads a 440 kiB image
- upload with 95-100% quality generates a 500 kiB and denies the upload.
- the user can simply not explain, what went wrong and will look for help

So imho a growing image file size during the upload should be avoided.

KYPREO:
No problem. This issue can be avoided. I have made a change so that it is impossible for an admin to set quality higher than 90%. I will commit  the change if you think this is acceptable.

Crizz0:
One last thought:
Will the user be able to choose 0 to 90 or 0 to 100, which is calculated to 0 to 90 in the backend?

Because so input 100% equals 90 quality, 50% input equals 45 quality?

mrgoldy:
> One last thought: Will the user be able to choose 0 to 90 or 0 to 100, which is calculated to 0 to 90 in the backend? Because so input 100% equals 90 quality, 50% input equals 45 quality?

That would make things only more confusing.
Just put it as 0-90%. "Image quality". _"90% is the limit as otherwise it will increase image filesize"_

KYPREO:
Yes setting 100% and having 90% in the backend should possible and would offer more intuitive UX for the admin. Let me work on it and I will see if I can submit a solution along those lines.

3D-I:
I'd go starting from 50 to 90, IMHO makes no sense to allow from 0 to 50

KYPREO:
In response to the above feedback, I have:

- made the default quality 85% to match the current hard-coded setting.
- limited the allowable quality setting to 50-90%
- updated the explanatory text for the ACP to refer to the 90% limit and clear up that quality and metadata settings only affected uploaded JPEG (as per Plupload's documentation). PNG does not have metadata and cannot be compressed.

As 3Di says, it's hopefully now a matter of the powers at be to decide whether or not EXIF should be stripped by default. If Ideas board is anything to go by, majority were in favour. Some of the downvotes may have voted yes if it was admin-configurable, which this is.

It should also be noted that:

- the existing maximum image dimensions setting only truly affects JPEG and PNG, as Plupload only provides client-side resizing for those image types.  GIF and BMP are not resized which may confuse some users. Perhaps this might be addressed in the ACP by adding (JPEG & PNG only) to the setting description

- Plupload v 2.3.6 used by phpBB reportedly features bilinear resampling for much higher image quality compression.  Plupload's documentation has not been updated to reflect this fact. Actually, the feature was first introduced into Plupload 2.2.1: https://github.com/moxiecode/plupload/releases/tag/v2.2.1  phpBB therefore first got this feature when Plupload v  2.3.3 was shipped with phpBB 3.2.2.  This means that Plupload may be using a different compression algorithm to what it was when ticket PHPBB3-14770 was addressed.  It might also not be enabled as of now.

In Plupload 2.3.6 there is a flag for resize to change the resampling option from 'default' to 'bilinear' to 'nearest', but after testing these options in phpBB I cannot see any different in the images at all.🤷‍♂️ I will continue looking at this and if there is an improvement to be made, I'll report back.

phpBB/includes/ucp/ucp_pm_compose.php
```diff
@@ -665,12 +665,6 @@ function compose_pm($id, $mode, $action, $user_folders = array())
 		$subject = (!$subject && $action != 'post') ? $user->lang['NEW_MESSAGE'] : $subject;
 		$message = $request->variable('message', '', true);
```
3D-I: Why are you removing this?

KYPREO: I stuffed up.  My ticket ended up on a detached head, not a branch, and it's ended up with unwanted merge commits. I was trying to fix the merge commit and accidentally reverted it.  I'm still new to Git and finding it so confusing. My apologies.

3D-I: You can try 

`git reset --hard HEAD~1`
then
`git push -f origin ticket/16329`

to reset the last commit.

KYPREO: Thanks. I think this has fixed it.

phpBB/install/schemas/schema_data.sql
marc1706: I agree with @3D-I as this would be in line with current behavior and might otherwise come as a surprise to users of phpBB.

phpBB/phpbb/db/migration/data/v32x/enable_plupload_config.php
```diff
@@ -0,0 +1,30 @@
...
+*
+*/
+
+namespace phpbb\db\migration\data\v32x;
+
+class enable_plupload_config extends \phpbb\db\migration\migration
+{
+	static public function depends_on()
+	{
+		return ['\phpbb\db\migration\data\v32x\v329',];
```
marc1706:
```suggestion
		return ['\phpbb\db\migration\data\v32x\v329'];
```

phpBB/language/en/acp/attachments.php
```diff
@@ -111,6 +111,10 @@
 
 	'IMAGE_LINK_SIZE'			=> 'Image link dimensions',
 	'IMAGE_LINK_SIZE_EXPLAIN'	=> 'Display image attachment as an inline text link if image is larger than this. To disable this behaviour, set the values to 0px by 0px.',
+	'IMAGE_QUALITY'				=> 'Quality of uploaded image attachments (JPEG only)',
+	'IMAGE_QUALITY_EXPLAIN'		=> 'Specify value between 50% (smaller file size) and 90% (higher quality). Quality higher than 90% increases filesize and is disabled. Setting only applies if max image dimensions set to value other than 0px by 0px.',
+	'IMAGE_STRIP_METADATA'		=> 'Strip image metadata (JPEG only)',
+	'IMAGE_STRIP_METADATA_EXPLAIN'	=> 'Strip Exif metadata eg author name, GPS coordinates & camera details. Setting only applies if max image dimensions set to value other than 0px by 0px.',
```
marc1706:
```suggestion
	'IMAGE_STRIP_METADATA_EXPLAIN'	=> 'Strip Exif metadata, e.g. author name, GPS coordinates and camera details. Setting only applies if maximum image dimensions are set to a value other than 0px by 0px.',
```

phpBB/language/en/acp/attachments.php
```diff
@@ -111,6 +111,10 @@
 
 	'IMAGE_LINK_SIZE'			=> 'Image link dimensions',
 	'IMAGE_LINK_SIZE_EXPLAIN'	=> 'Display image attachment as an inline text link if image is larger than this. To disable this behaviour, set the values to 0px by 0px.',
+	'IMAGE_QUALITY'				=> 'Quality of uploaded image attachments (JPEG only)',
+	'IMAGE_QUALITY_EXPLAIN'		=> 'Specify value between 50% (smaller file size) and 90% (higher quality). Quality higher than 90% increases filesize and is disabled. Setting only applies if max image dimensions set to value other than 0px by 0px.',
```
marc1706:
```suggestion
	'IMAGE_QUALITY_EXPLAIN'		=> 'Specify value between 50% (smaller file size) and 90% (higher quality). Quality higher than 90% increases filesize and is disabled. Setting only applies if maximum image dimensions are set to a value other than 0px by 0px.',
```

phpBB/phpbb/db/migration/data/v32x/enable_plupload_config.php
```diff
@@ -0,0 +1,30 @@
...
+* @license GNU General Public License, version 2 (GPL-2.0)
+*
+* For full copyright and license information, please see
+* the docs/CREDITS.txt file.
+*
+*/
+
+namespace phpbb\db\migration\data\v32x;
+
+class enable_plupload_config extends \phpbb\db\migration\migration
```
marc1706: Please rename this file to `add_plupload_config.php` as this IMHO better describes the purpose of this migration.

phpBB/install/schemas/schema_data.sql
KYPREO: I personally think that most users of phpBB would be surprised that the default behaviour has been to preserve personally-identifiable metadata as this has never been disclosed to administrators or end-users. However, I understand your reasoning and will amend.

phpBB/phpbb/db/migration/data/v32x/enable_plupload_config.php
KYPREO: Thank you @marc1706 for your review and feedback.  I have made all these changes. Would you like me to create another migration file for 3.3.0 > 3.3.1 (or whichever version this makes it into)? Or will the file be replicated and amended for correct version number when merged into 3.3.x branch?

marc1706: The changes will be merged upstream into 3.3 and master so there is no need for that.

PR #5835, opened 2020-01-21
[ticket/15766] Check for role on changing permission settings

mrgoldy:
Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15766

Perhaps I should explain a bit better what this actually does. 
When you change a permission, it checks if the new group of permissions matches a permission role and indicates it as such in the dropdown. Rather than always defaulting to ‘No role assigned...’

### DIFF
--- a/phpBB/adm/style/permissions.js
+++ b/phpBB/adm/style/permissions.js
@@ -322,21 +322,21 @@ function match_role_settings(id)
 {
 	var fs = document.getElementById(id),
 		cbs = fs.getElementsByTagName('input'),
-		xyz = {};
+		set = {};
 
 	for (var i = 0; i < cbs.length; i++) {
 		var matches = cbs[i].id.match(/setting\[\d+]\[\d+]\[([a-z_]+)]/);
 
 		if (matches !== null && cbs[i].checked && cbs[i].value !== '-1') {
-			xyz[matches[1]] = parseInt(cbs[i].value);
+			set[matches[1]] = parseInt(cbs[i].value);
 		}
 	}
 
-	xyz = sort_and_stringify(xyz);
+	set = sort_and_stringify(set);
 
 	for (var r in role_options)
 	{
-		if (sort_and_stringify(role_options[r]) === xyz) {
+		if (sort_and_stringify(role_options[r]) === set) {
 			return r;
 		}
 	}
--- a/phpBB/adm/style/permissions.js
+++ b/phpBB/adm/style/permissions.js
@@ -278,11 +278,26 @@ function reset_role(id) {
 		return;
 	}
 
-	t.options[0].selected = true;
+	var parent = t.parentNode,
+		roleId = match_role_settings(id.replace('role', 'perm')),
+		text = no_role_assigned,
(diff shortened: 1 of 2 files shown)


phpBB/adm/style/permissions.js
```diff
@@ -278,11 +278,26 @@ function reset_role(id) {
...
 
-	t.options[0].selected = true;
+	var parent = t.parentNode,
+		roleId = match_role_settings(id.replace('role', 'perm')),
+		text = no_role_assigned,
+		index = 0;
+
+	if (roleId) {
+		for (var i = 0; i < t.options.length; i++) {
+			if (t.options[i].value == roleId) {
```
marc1706:
```suggestion
			if (t.options[i].value === roleId) {
```

phpBB/adm/style/permissions.js
```diff
@@ -302,3 +317,36 @@ function set_role_settings(role_id, target_id) {
 		mark_one_option(target_id, r, (settings[r] === 1) ? 'y' : 'n');
 	}
 }
+
+function match_role_settings(id)
```
marc1706: Please add docblocks for newly added functions.

phpBB/adm/style/permissions.js
```diff
@@ -302,3 +317,36 @@ function set_role_settings(role_id, target_id) {
...
+{
+	var fs = document.getElementById(id),
```
marc1706: I understand that these might be in line with some of the older JS but you should use variable names that actually mean something. These are not helpful at all and remind me a bit of really old C code. ;)

phpBB/adm/style/permissions.js
```diff
@@ -302,3 +317,36 @@ function set_role_settings(role_id, target_id) {
...
+		cbs = fs.getElementsByTagName('input'),
+		set = {};
+
+	for (var i = 0; i < cbs.length; i++) {
+		var matches = cbs[i].id.match(/setting\[\d+]\[\d+]\[([a-z_]+)]/);
+
+		if (matches !== null && cbs[i].checked && cbs[i].value !== '-1') {
+			set[matches[1]] = parseInt(cbs[i].value);
```
marc1706: Should add the radix parameter:
```suggestion
			set[matches[1]] = parseInt(cbs[i].value, 10);
```

phpBB/adm/style/permissions.js
```diff
@@ -302,3 +317,36 @@ function set_role_settings(role_id, target_id) {
...
+	{
+		if (sort_and_stringify(role_options[r]) === set) {
+			return r;
+		}
+	}
+
+	return 0;
+}
+
+function sort_and_stringify(obj) {
```
marc1706: Same as above.

3D-I:
Wait.. I haven't tried this one but a question arose: 

* Is the ‘No role assigned...’ thinghy not available any more here?

marc1706:
It is still available, it'll however display the role you picked when selecting the same settings as a current role.

PR #5838, opened 2020-01-25
[ticket/16339] Fix Poll's voting options refresh

3D-I:
PHPBB3-16339

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16339

### DIFF
--- a/phpBB/styles/prosilver/template/viewtopic_body.html
+++ b/phpBB/styles/prosilver/template/viewtopic_body.html
@@ -77,7 +77,7 @@
 <!-- EVENT viewtopic_body_poll_before -->
 
 <!-- IF S_HAS_POLL -->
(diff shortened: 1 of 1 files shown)


mrgoldy:
This looks like it can not be the solution. The `data-refresh="true"` attribute should not refresh the page, unless a `REFRESH_DATA => ['url' => '', 'time' => 0]` is present in the JSON response. Which is not the case for casting a vote in the poll. Not to mention, that refreshing the page will make the ajax callback completely redundant.

I believe this should probably fixed in some other manner.

3D-I:
As per the tracker ticket, it does refresh the voting `form` only in case of error, there is no need to add data to the json array as I have seen and tested.

PR #5841, opened 2020-01-26
[ticket/14463] Add Doctrine DBAL as a driver

CHItA:
This probably breaks all db tools etc, so it is mostly opened to gather feedback.

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-14463


mrgoldy:
I have played with this a few months back aswell. 
In case you’re interested: https://github.com/mrgoldy/phpbb/tree/ticket/14463

phpBB/phpbb/db/result_iterator.php
```diff
@@ -0,0 +1,254 @@
...
+
+		$this->position = $this->target_position;
+	}
+
+	/**
+	 * Fetch an element before the current iterator position.
+	 *
+	 * @return void
+	 */
+	private function iterate_backward()
```
CHItA: As of now, this method is just horse shit mixed with snake oil, as `fetch_all()` doesn't fetch all but all remaining rows from the current cursor position. The problem here is saving all rows (which is one option to actually support going backwards) would add some overhead on all queries. I'm not sure if breaking BC here would make more sense, as I guess this might be some functionality that is not used at all by anyone (certainly not by the core).

@marc1706 @Derky @Nicofuma thoughts?

marc1706: IMHO if there is no good way to achieve this then don't. Especially if it's not even used in the core. We should however make sure to properly document this BC break, e.g. in the documentation. I can prepare a page for this (similar to how symfony is documenting those changes: https://github.com/symfony/symfony/blob/master/UPGRADE-5.0.md).

CHItA: The only solution I see is to store all fetched rows in the iterator. I doubt that there would be any other solution to this.

CHItA: Although we could reexecute the query as well when backward iteration is called for, thus only having any overhead when it is absolutely required.

mrgoldy: Is it not possible to save queries that are made per page in a protected property? This will also allow extensions to access SQL queries that are already made by the core, without having to rerun them if they're not cached.
For example saving the SQL queries in `$queries` and all the SQL results in `$results`.
```php
protected $queries = [
    0 => 'SELECT * FROM `phpbb_forums`,
];

protected $results = [
    0 => [
        0 => ['forum_id => 1', '...' => '...'],
    ],
];
```
So you can iterate over them and check before querying to see if it already exists.

CHItA: I don't think we would want to add any new features to this, as the goal here is to keep BC as much as possible. Also, if you re-execute a query, chances are it comes from the cache which should be much faster. Thirdly, extensions should not really execute the same queries as that data probably is accessible to them from somewhere.

mrgoldy: Well, that's not completely true as a lot of queries in the core are not cached and not accessible for extensions, if not in tedious ways. But it was just an idea.

mrgoldy:
What I not fully understand, is that the idea was to 'nuke dbal' and to eventually get rid of the phpBB part all together. Wouldn't it then make more sense to nuke it completely. Namely, replacing phpBB DBAL with Doctrine DBAL and implementing a phpBB wrapper class (for BC) that can eventually be deprecated and removed, rather than keeping the phpBB DBAL and adding a Doctrine driver?

phpBB/phpbb/db/result_iterator.php
CHItA: Extension authors can add events if they want to.

CHItA:
> What I not fully understand, is that the idea was to 'nuke dbal' and to eventually get rid of the phpBB part all together. Wouldn't it then make more sense to nuke it completely. Namely, replacing phpBB DBAL with Doctrine DBAL and implementing a phpBB wrapper class (for BC) that can eventually be deprecated and removed, rather than keeping the phpBB DBAL and adding a Doctrine driver?

That is exactly what this PR does. See the `@depreceted` annotation on `doctrine`. I'm not really sure if this was your question.

mrgoldy:
Ah, I kind of misinterpreted this PR. As you are indeed doing it.
However, in a different fashion that I had imagined. I would of thought of adding a wrapperClass rather than using a factory. Both work I suppose and with this you keep the class `\phpbb\db\driver\driver_interface` alive. Okay I get it I think.

phpBB/phpbb/di/container_builder.php
```diff
@@ -548,12 +548,12 @@ protected function inject_dbal_driver()
...
 
 		$config_data = $this->config_php_file->get_all();
-		if (!empty($config_data))
+		/*if (!empty($config_data))
 		{
 			if ($this->dbal_connection === null)
 			{
 				$dbal_driver_class = $this->config_php_file->convert_30_dbms_to_31($this->config_php_file->get('dbms'));
-				/** @var \phpbb\db\driver\driver_interface $dbal_connection */
+				/** @var \phpbb\db\driver\driver_interface $dbal_connection * /
```
3D-I: This commented line within an already commented block could lead to a 
`Parse error: syntax error, unexpected '*' ` IMHO

CHItA: It certainly could, hence the WIP tag on this ticket.

rubencm:
https://github.com/doctrine/dbal/search?p=1&q=getTruncateTableSQL&type=code

In case you want to implement #5329 here

PR #5844, opened 2020-01-30
[ticket/16343] Fix Quick reply subject

3D-I:
PHPBB3-16343

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16343

### DIFF
--- a/phpBB/viewtopic.php
+++ b/phpBB/viewtopic.php
@@ -2364,7 +2364,7 @@ if ($s_can_vote || $s_quick_reply)
 			'S_QUICK_REPLY'			=> true,
 			'U_QR_ACTION'			=> append_sid("{$phpbb_root_path}posting.$phpEx", "mode=reply&amp;f=$forum_id&amp;t=$topic_id"),
 			'QR_HIDDEN_FIELDS'		=> build_hidden_fields($qr_hidden_fields),
-			'SUBJECT'				=> 'Re: ' . censor_text($topic_data['topic_title']),
+			'SUBJECT'				=> 'Re: ' . censor_text($topic_data['topic_last_post_subject']),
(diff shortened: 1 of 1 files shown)


3D-I:
immagine

rxu:
Personally, I don't like this change as last post subject is editable by the last post's author so can be even empty. phpBB is flat so users reply to a topic and not to post.
Also, if you, say, merge posts/topic into another topic, you'll get merged post's/topic's last post subject here instead of the "parent" topic's title and have to edit last post subject after merging as such.

3D-I:
If the topic has one post only the `post subject` is the topic title, if it has 2 or more posts then it is the `last post subject`.

When you reply with the button to a topic you get the `topic's last post subject` or the `post subject` into the form depending on the above conditions, that's what the quick reply has to do: **obey to the standard.** IMHO :smiley_cat: 

Moreover, when you parse via CLI the _posts/topics/forums/PMs_ this are the indexes related, which are parsed:

```
		'forums'	=> 'forum_last_post_subject',
		'topics'	=> 'topic_last_post_subject',
		'posts'		=> 'post_subject',
		'pms'		=> 'message_subject',
```

mrgoldy:
Not sure if I follow, but when replying to a topic (regardless of how many replies that topic has) I always get the `Re: <topic title>` and never the `Re: <last post subject>`

3D-I:
> 
> 
> Not sure if I follow, but when replying to a topic (regardless of how many replies that topic has) I always get the `Re: <topic title>` and never the `Re: <last post subject>`

Yes you are correct, same goes to @rxu - I missed something here, my bad.

Closing as such.

PR #5845, opened 2020-02-02
[ticket/16363] Allow replying to locked topics if user has m_lock permissions

bwoebi:
An user may not be able to edit - but lock and unlock.
It is terribly inconvenient for users being only able to un/lock threads to have to first unlock a topic, then reply and then re-lock it.

Thus in reply mode we check for lock permissions instead of edit.

This also includes editing if the user could normally edit his post if it
weren't for the lock.

Checklist:

- [x] Tests pass

Tracker ticket (set the ticket ID to **your ticket ID**): https://tracker.phpbb.com/browse/PHPBB3-16363

### DIFF
--- a/phpBB/posting.php
+++ b/phpBB/posting.php
@@ -434,7 +434,7 @@ if ($post_data['forum_type'] != FORUM_POST && in_array($mode, array('post', 'bum
 }
 
 // Forum/Topic locked?
-if (($post_data['forum_status'] == ITEM_LOCKED || (isset($post_data['topic_status']) && $post_data['topic_status'] == ITEM_LOCKED)) && !$auth->acl_get('m_edit', $forum_id))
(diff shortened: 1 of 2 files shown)


mrgoldy:
Yes, you **need** a tracker ticket.
Then you'll have to create a branch in your fork with the ticket name, so for example the branch name will be `ticket/12345` (where 12345 is replaced with the actual ticket id).
Make sure the new branch is against the correct 'base branch' (3.2 / 3.3 for fixes, 4.0 for new features).
Then you'll have to commit your changes with a specific commit message:
```
[ticket/12345] A short description of the commit

PHPBB3-12345

```
Where *12345* is replaced again with the actual ticket ID.

Most of this is explained in the PR Posting template you overrid ;) .

marc1706:
I've fixed your commit message and rebased your changes on top of 3.3.x

bwoebi:
@marc1706 Thanks! Also amended the commit to include the display of the edit button in locked view (if he normally would be allowed to).

marc1706:
And merged. Thanks for the PR.

PR #5846, opened 2020-02-02
[ticket/16346] Small improvements

rubencm:
PHPBB3-16346

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16346

### DIFF
--- a/phpBB/includes/bbcode.php
+++ b/phpBB/includes/bbcode.php
@@ -704,5 +704,4 @@ class bbcode
 
 		return $return;
 	}
-
 }
--- a/phpBB/includes/bbcode.php
+++ b/phpBB/includes/bbcode.php
@@ -33,6 +33,8 @@ class bbcode
 
 	var $template_bitfield;
 
+	protected $template_filename;
+
 	/**
 	* Constructor
 	*/
@@ -149,7 +151,7 @@ class bbcode
 	*/
 	function bbcode_cache_init()
 	{
-		global $user, $phpbb_dispatcher, $phpbb_extension_manager, $phpbb_container, $phpbb_filesystem;
+		global $user, $phpbb_dispatcher, $phpbb_extension_manager, $phpbb_container;
 
 		if (empty($this->template_filename))
 		{
@@ -702,4 +704,5 @@ class bbcode
 
 		return $return;
 	}
+
 }
--- a/phpBB/includes/constants.php
+++ b/phpBB/includes/constants.php
@@ -61,7 +61,6 @@ define('ACL_YES', 1);
 define('ACL_NO', -1);
 
 // Login error codes
-//define('LOGIN_CONTINUE', 1);
 define('LOGIN_BREAK', 2);
 define('LOGIN_SUCCESS', 3);
 define('LOGIN_SUCCESS_CREATE_PROFILE', 20);
@@ -120,10 +119,6 @@ define('POST_NORMAL', 0);
 define('POST_STICKY', 1);
 define('POST_ANNOUNCE', 2);
 define('POST_GLOBAL', 3);
-
-// Lastread types
-//define('TRACK_NORMAL', 0);
-//define('TRACK_POSTED', 1);
 
 // Notify methods
 define('NOTIFY_EMAIL', 0);
--- a/phpBB/includes/constants.php
+++ b/phpBB/includes/constants.php
@@ -61,7 +61,7 @@ define('ACL_YES', 1);
 define('ACL_NO', -1);
 
 // Login error codes
-define('LOGIN_CONTINUE', 1);
+//define('LOGIN_CONTINUE', 1);
 define('LOGIN_BREAK', 2);
 define('LOGIN_SUCCESS', 3);
 define('LOGIN_SUCCESS_CREATE_PROFILE', 20);
@@ -122,8 +122,8 @@ define('POST_ANNOUNCE', 2);
 define('POST_GLOBAL', 3);
 
 // Lastread types
-define('TRACK_NORMAL', 0);
-define('TRACK_POSTED', 1);
+//define('TRACK_NORMAL', 0);
+//define('TRACK_POSTED', 1);
 
 // Notify methods
 define('NOTIFY_EMAIL', 0);
--- a/phpBB/includes/functions.php
+++ b/phpBB/includes/functions.php
@@ -915,7 +915,7 @@ function markread($mode, $forum_id = false, $topic_id = false, $post_time = 0, $
 /**
 * Get topic tracking info by using already fetched info
 */
-function get_topic_tracking($forum_id, $topic_ids, &$rowset, $forum_mark_time, $global_announce_list = false)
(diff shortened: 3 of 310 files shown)


phpBB/includes/bbcode.php
```diff
@@ -702,4 +704,5 @@ function bbcode_second_pass_by_extension()
 
 		return $return;
 	}
+
```
marc1706, line 707: Don't think this line is needed

phpBB/includes/constants.php
```diff
@@ -61,7 +61,7 @@
 define('ACL_NO', -1);
 
 // Login error codes
-define('LOGIN_CONTINUE', 1);
+//define('LOGIN_CONTINUE', 1);
```
marc1706, line 64: No longer used? If yes, we should think about removing them.

rxu:
Probably move the whole codebase to the short array syntax as a small inprovement? :)

phpBB/includes/functions.php
```diff
@@ -4317,19 +4296,6 @@ function exit_handler()
...
 
-/**
-* Casts a numeric string $input to an appropriate numeric type (i.e. integer or float)
-*
-* @param string $input		A numeric string.
-*
-* @return int|float			Integer $input if $input fits integer,
-*							float $input otherwise.
-*/
-function phpbb_to_numeric($input)
```
marc1706, line 4328: Still being used here: https://github.com/phpbb/phpbb/blob/master/phpBB/phpbb/php/ini.php#L138

rubencm:
@rxu I used a tool to find possible problems and fixed them manually, so there wasn't a way to do it in parts, but I can make other Pr's to make this smaller

phpBB/includes/functions.php
rubencm: I move that function to functions_compatibility because that class is deprecated

phpBB/install/startup.php
```diff
@@ -227,6 +227,7 @@ function installer_shutdown_function($display_errors)
 
 // Include files
 phpbb_require_updated('includes/compatibility_globals.' . $phpEx, $phpbb_root_path);
+phpbb_require_updated('includes/constants.' . $phpEx, $phpbb_root_path);
```
rxu, line 230: I guess it should be already done here https://github.com/phpbb/phpbb/blob/master/phpBB/phpbb/install/helper/container_factory.php#L187

rubencm: Yes, I have some errors doing that, that is why i have set wip. But withtout that there is an error during install when you have error_reporting(-1)

`
    phpbb: [phpBB Debug] "Use of undefined constant ANONYMOUS - assumed 'ANONYMOUS' (this will throw an Error in a future version of PHP)" in file /home/vagrant/phpbb/phpBB/phpbb/install/console/command/install/install.php on line 69
    phpbb: [phpBB Debug] "Use of undefined constant ANONYMOUS - assumed 'ANONYMOUS' (this will throw an Error in a future version of PHP)" in file /home/vagrant/phpbb/phpBB/phpbb/install/console/command/install/install.php on line 69
    phpbb: 
    phpbb:       0/13  ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░    0%
`

rubencm: Ignore the last commit for now

rxu: @rubencm probably you have the source code outdated a bit. The bug was fixed with this commit https://github.com/phpbb/phpbb/commit/09e308a06d7fcaa624afe7e4a86e178bab84a362#diff-8bfd43635044b97335b9c309cd081a1a

rubencm: No, i have that updated

rubencm: In fact, you can see that the error is the same, but not in the same file.

rxu: Indeed. That is because of `\phpbb\user` initialization in the command constructor without `ANONYMOUS` constant being defined.
Okay.

rubencm: If you want to look at it, it happens during installation with cli. I will remove this commit.

rxu: the real issue for both cases is this line https://github.com/phpbb/phpbb/blob/1e40c814f39c986de6c0e67723026beb4f078e34/phpBB/phpbb/session.php#L30

I couldn't find any use cases of this property excepting a couple of assignments in session.php https://github.com/phpbb/phpbb/blob/1e40c814f39c986de6c0e67723026beb4f078e34/phpBB/phpbb/session.php#L368 and https://github.com/phpbb/phpbb/blob/1e40c814f39c986de6c0e67723026beb4f078e34/phpBB/phpbb/session.php#L655
but it looks like `->id` is not used anywhere. Probably just drop it in 4.0?

rxu: Hmm. It was added by @mrgoldy not a long ago here https://github.com/phpbb/phpbb/commit/e5c80e46b09b90ec2f5e181032c66480af598e51#diff-ef9a28b1448b3c0a926f41845507b78fR28

phpBB/includes/functions_compatibility.php
```diff
@@ -847,3 +847,37 @@ function phpbb_delete_user_pms($user_id)
...
+	return ($input > PHP_INT_MAX) ? (float) $input : (int) $input;
+}
+
+/**
+* Check and display the SQL report if requested.
+*
+* @param \phpbb\request\request_interface		$request	Request object
+* @param \phpbb\auth\auth						$auth		Auth object
+* @param \phpbb\db\driver\driver_interface		$db			Database connection
+ *
```
marc1706, line 872: Indentation needs to be fixed

phpBB/phpbb/filesystem/helper.php
```diff
@@ -233,7 +233,7 @@ public static function resolve_path($path, $prefix = '', $absolute = false, $ret
 			$path = str_replace(DIRECTORY_SEPARATOR, '/', $path);
 		}
 
-		trim($path, '/');
+		//$path = trim($path, '/'); // TODO: Check this
```
marc1706, line 236: I don't see a reason to remove this. It could prevent potentially worry-some absolute paths.

phpBB/includes/functions_display.php
```diff
@@ -767,7 +767,7 @@ function generate_forum_nav(&$forum_data_ary)
 		return;
 	}
 
-	$navlinks_parents = $forum_template_data = array();
```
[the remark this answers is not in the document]
rubencm: The variable is overwritten later, so the initialization there do nothing

phpBB/phpbb/filesystem/helper.php
rubencm:
```php
trim($path, '/');```

Was doing nothing, because the trimmed value was being returned to nowhere. I just modified it and commented it, because it fails in windows if I don't remember wrong, it has to be investigated, but when doing that I haven't modified the current behavior.

marc1706 (approved):

PR #5865, opened 2020-02-27
[ticket/16377] Check navlinks existance

mrgoldy:
PHPBB3-16377

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16377


phpBB/styles/prosilver/template/navbar_header.html
```diff
@@ -192,7 +192,7 @@
 			{% EVENT overall_header_breadcrumb_prepend %}
 				<span class="crumb" {{ MICRODATA }}><a itemprop="item" href="{{ U_INDEX }}" accesskey="h" data-navbar-reference="index">{% if not U_SITE_HOME %}<i class="icon fa-home fa-fw"></i>{% endif %}<span itemprop="name">{{ L_INDEX }}</span></a><meta itemprop="position" content="{{ navlink_position }}{% set navlink_position = navlink_position + 1 %}" /></span>
 
-			{% for navlinks in navlinks %}
+			{% for navlinks in loops.navlinks %}
```
3D-I: Since no `loops.` are to be used in the codebase, will

```
{% if navlinks|length %}
	{% for navlinks in navlinks %}
		{% set NAVLINK_NAME .....
	{% endfor %}
{% endif %}
```
do the trick here?

mrgoldy: Did not know `loops.` was to be avoided.
I'll update it then.

Reason I went for this in the first place is cause I am not a fan of using the same variable as the loop name, so this would solve both at once. This would be the equivalant in PHP of `foreach ($navlinks as $navlinks)` which is wrong, so I would that rather change it to `{% for navlink in navlinks %}` with the `{% if %}` statement surrounding it.

3D-I: Yeah, in this case was possible to change the variable name, in some other case it wasn't.

mrgoldy:
When running my fork on my `development` environment, I get this error aswell on 3.2
Should I rebase against 3.2 or create an additional PR for 3.2?

dark-1:
> When running my fork on my `development` environment, I get this error aswell on 3.2
> Should I rebase against 3.2 or create an additional PR for 3.2?

On branch `3.2.x`, would better...

3D-I:
Please change the base branch from 3.2.x to 3.3.x back again.

Changíng branch here caused issues at the tracker.
immagine

mrgoldy:
Apparently I triggered a whole lot of emails here from Jira.
My apoligies.

3D-I:
Yeah no probs :smile_cat: 

immagine

PR #5870, opened 2020-02-27
[ticket/16381] Add core event to Modify Topics SQL

dark-1:
PHPBB3-16381

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16381

### DIFF
--- a/phpBB/viewforum.php
+++ b/phpBB/viewforum.php
@@ -716,7 +716,7 @@ if (count($topic_list))
 	/**
 	* Event to modify the SQL query before obtaining topics/stickies
 	*
-	* @event core.viewforum_modify_topics_sql
+	* @event core.viewforum_modify_topic_list_sql
 	* @var	int		forum_id			The forum ID
 	* @var	array	forum_data			Data about the forum
 	* @var	array	topic_list			Topic ids array
@@ -724,13 +724,13 @@ if (count($topic_list))
 	*
 	* @since 3.2.10-RC1
 	*/
-	$vars = array(
+	$vars = [
 		'forum_id',
 		'forum_data',
 		'topic_list',
 		'sql_array',
-	);
-	extract($phpbb_dispatcher->trigger_event('core.viewforum_modify_topics_sql', compact($vars)));
+	];
+	extract($phpbb_dispatcher->trigger_event('core.viewforum_modify_topic_list_sql', compact($vars)));
 
 	// If store_reverse, then first obtain topics, then stickies, else the other way around...
 	// Funnily enough you typically save one query if going from the last page to the middle (store_reverse) because
--- a/phpBB/viewforum.php
+++ b/phpBB/viewforum.php
@@ -710,9 +710,27 @@ if (count($topic_list))
 		'SELECT'		=> $sql_array['SELECT'],
 		'FROM'			=> $sql_array['FROM'],
 		'LEFT_JOIN'		=> $sql_array['LEFT_JOIN'],
-
 		'WHERE'			=> $db->sql_in_set('t.topic_id', $topic_list),
 	);
+
+	/**
+	* Event to modify the SQL query before obtaining topics/stickies
+	*
+	* @event core.viewforum_modify_topics_sql
+	* @var	int		forum_id			The forum ID
+	* @var	array	forum_data			Data about the forum
+	* @var	array	topic_list			Topic ids array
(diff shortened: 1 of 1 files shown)


dark-1:
!set event

phpBB/viewforum.php
```diff
@@ -710,10 +710,28 @@
...
 	);
 
+	/**
+	* Event to modify the SQL query before obtaining topics/stickies
+	*
+	* @event core.viewforum_modify_topics_sql
+	* @var	int		forum_id			The forum ID
+	* @var	array	forum_data			Data about the forum
+	* @var	array	topic_list			Topic ids array
+	* @var	array	sql_array			SQL query array for obtaining topics/stickies
```
mrgoldy: I would prefer `@var array` with a single space, rather than a tab.

phpBB/viewforum.php
```diff
@@ -710,10 +710,28 @@
 		'SELECT'		=> $sql_array['SELECT'],
 		'FROM'			=> $sql_array['FROM'],
 		'LEFT_JOIN'		=> $sql_array['LEFT_JOIN'],
-
 		'WHERE'			=> $db->sql_in_set('t.topic_id', $topic_list),
 	);
 
+	/**
+	* Event to modify the SQL query before obtaining topics/stickies
```
mrgoldy: I would prefer multiline comments to have the asterix indented with a space:
```
/*
 * A comment
 */
```
rather than
```
/*
* A comment
*/
```

phpBB/viewforum.php
```diff
@@ -710,10 +710,28 @@
...
+	*
+	* @event core.viewforum_modify_topics_sql
+	* @var	int		forum_id			The forum ID
+	* @var	array	forum_data			Data about the forum
+	* @var	array	topic_list			Topic ids array
+	* @var	array	sql_array			SQL query array for obtaining topics/stickies
+	*
+	* @since 3.2.10-RC1
+	*/
+	$vars = array(
```
mrgoldy: Please use the new array syntax (`[]`) instead.

dark-1: This I'll do.

phpBB/viewforum.php
dark-1: Ok. 
But what about this event : https://github.com/dark-1/phpbb/blob/ticket/16381/phpBB/phpbb/session.php#L1090
there it's used tabs & it's `@since 3.2.9-RC1` ???

phpBB/viewforum.php
dark-1: Same here ,
in this event : https://github.com/dark-1/phpbb/blob/ticket/16381/phpBB/phpbb/session.php#L1090
there it's used & it's `@since 3.2.9-RC1` ???

mrgoldy: Yes and there are many more places.
Unfortunately there is no consistency in those yet, hence I remarked it. 
We should move towards a more uniform way. 
Not sure how others feel about this, but I feel this is the way to go.

These comments apply to all your event PR's.

dark-1: For this particular case,
If I go by this : https://area51.phpbb.com/docs/32x/coding-guidelines.html
& this : https://area51.phpbb.com/docs/33x/coding-guidelines.html
I have to use :
```php
/**
* A comment
*/
```
So I think it's better to keep as-per the `coding-guidelines` ...

mrgoldy: And that's what I have been trying to discuss for quite some time, see http://area51.phpbb.com/phpBB/viewtopic.php?f=81&t=70561 aswell
Not indenting the asterix with a space is very old code style.
All proper editors automatically indent these block comments.
Have a look through the vendor folder, and see how many (if not all) use it.

Those coding guidelines need to be updated.
And the code has to be cleaned up throughout the core aswell, some is just several years old.

dark-1: So should I wait till coding guidelines is updated ?

dark-1: I think I'll keep it as per `coding-guidelines` , once your discussion topic comes on a final decision, and the `coding-guidelines` are updated , I'll follow that henceforth.

phpBB/viewforum.php
dark-1: I think I'll keep it as per `coding-guidelines` , once your discussion topic comes on a final decision, and the `coding-guidelines` are updated , I'll follow that henceforth.

PR #5873, opened 2020-02-28
[ticket/16325] Ensure lock file is opened for updating

marc1706:
PHPBB3-16325

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16325

### DIFF
--- a/phpBB/phpbb/lock/flock.php
+++ b/phpBB/phpbb/lock/flock.php
@@ -73,7 +73,7 @@ class flock
 		// one file for writing simultaneously
 		if (file_exists($this->path . '.lock'))
 		{
-			$mode = 'rb';
+			$mode = 'rb+';
 		}
 		else
 		{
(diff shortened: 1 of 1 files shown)


rubencm:
I think open it in binary is not necessary

Crizz0:
I have one report from a user using a Strato webspace, that this fix solved the problem and phpBB 3.3.x is installable.

marc1706:
> I think open it in binary is not necessary

I don't think binary open adds any overhead and in theory I'd expect the opposite to be the case.

rubencm:
Maybe we can rely on the symfony implementation for this, so we don't have to worry about this
https://github.com/symfony/lock/blob/master/Store/FlockStore.php

marc1706:
Probably a good idea to try and use the symfony implementation. However, I do think this is out of the scope of this issue and I'd rather have a look at this in another pull request.

marc1706:
I have created a ticket for that: https://tracker.phpbb.com/browse/PHPBB3-16390

PR #5891, opened 2020-03-10
[ticket/16391] Add extensions info_acp_ lang file to MCP logs

mrgoldy:
PHPBB3-16391

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16391

---

The ticket suggested including the `info_mcp_` into the ACP.
However, the core includes the `acp/common` into the MCP.
So I opted to stick with that approach, where `$p_master` is also already available.

### DIFF
--- a/phpBB/includes/mcp/mcp_logs.php
+++ b/phpBB/includes/mcp/mcp_logs.php
@@ -39,6 +39,7 @@ class mcp_logs
 		global $config, $phpbb_container, $phpbb_log;
 
 		$user->add_lang('acp/common');
+		$this->p_master->add_mod_info('acp');
 
 		$action = $request->variable('action', array('' => ''));
 


kasimi:
Tested and works! 👍 

Is it documented that language keys for log entries should always be added in an `info_acp_` language file, regardless of the log type? If not, should we also fix the code to cover the case where an extension adds its language strings for the MCP log entries in an `info_mcp_` language file? In this scenario, currently, the raw language key is displayed in _ACP => Maintenance => Moderator log_.

mrgoldy:
I think it is best to tell extensions to add all log language strings in the ACP file. 
Makes the most sense code-wide, but even viewing as my own perspective as extension writer.
Might just have to be 'announced' so everybody is fully aware, or something like that.

3D-I:
I encountered the problem just today, I had to create an `info_mcp` file just for a single key. :disappointed: 

 :+1:

PR #5892, opened 2020-03-10
[ticket/16397] Expose user identifier

mrgoldy:
PHPBB3-16397

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16397

### DIFF
--- a/phpBB/phpbb/session.php
+++ b/phpBB/phpbb/session.php
@@ -25,6 +25,7 @@ class session
 	var $forwarded_for = '';
 	var $host = '';
 	var $session_id = '';
+	public $id = ANONYMOUS;
 	var $ip = '';
 	var $load = 0;
 	var $time_now = 0;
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/session.php
```diff
@@ -25,6 +25,7 @@ class session
 	var $forwarded_for = '';
 	var $host = '';
 	var $session_id = '';
+	public $id = ANONYMOUS;
```
marc1706: Design-wise I think it makes more sense to have this default to an invalid value, i.e. 0.

mrgoldy: Usually I would completely agree, you set base variables to 'nothing' (`null`, `''`, `0`, etc).
However, considering the amount of checks that are done with `$user_id <> ANONYMOUS` I would stick with this. Cause if this is set to `0` and somehow not changed, all of a sudden in various places the user is not considered a guest. Should never really happen but still.

rxu: Now we have issues with this default assignment in installer (fixed [here](https://github.com/phpbb/phpbb/commit/09e308a06d7fcaa624afe7e4a86e178bab84a362#diff-8bfd43635044b97335b9c309cd081a1a)) and in console [here](https://github.com/phpbb/phpbb/pull/5846#discussion_r436789271).

marc1706: Yep, ran into this while testing for 3.3.1-RC1.

PR #5894, opened 2020-03-10
[ticket/16398] Disable AJAX for extension actions

mrgoldy:
PHPBB3-16398

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16398

---

This adds a 'disable AJAX actions' in the upper right corner when viewing the extensions list.
This will add a `&ajax=off` GET parameter to the module.
If it's set to `off`, the `data-ajax=""` attributes are not added on that page load, and the actions are 'regular' requests. 
This will affect one action, AJAX is automatically turned back on the next page load when viewing extensions. It is not persistent.

### DIFF
--- a/phpBB/adm/style/acp_ext_actions.html
+++ b/phpBB/adm/style/acp_ext_actions.html
@@ -1,7 +1,7 @@
 {% for action in enabled.actions %}
-	<a href="{{ action.U_ACTION }}"{% if action.L_ACTION_EXPLAIN %} title="{{ action.L_ACTION_EXPLAIN }}"{% endif %}{% if action.COLOR %} style="color: {{ action.COLOR }};"{% endif %} data-ajax="{{ action.ACTION_AJAX }}" data-refresh="true">{{ action.L_ACTION }}</a>{% if not action.S_LAST_ROW %}&nbsp;|&nbsp;{% endif %}
+	<a href="{{ action.U_ACTION }}"{% if action.L_ACTION_EXPLAIN %} title="{{ action.L_ACTION_EXPLAIN }}"{% endif %}{% if action.COLOR %} style="color: {{ action.COLOR }};"{% endif %}{% if not S_AJAX_DISABLED %} data-ajax="{{ action.ACTION_AJAX }}" data-refresh="true"{% endif %}>{{ action.L_ACTION }}</a>{% if not action.S_LAST_ROW %}&nbsp;|&nbsp;{% endif %}
 {% endfor %}
 
 {% for action in disabled.actions %}
-	<a href="{{ action.U_ACTION }}"{% if action.L_ACTION_EXPLAIN %} title="{{ action.L_ACTION_EXPLAIN }}"{% endif %}{% if action.COLOR %} style="color: {{ action.COLOR }};"{% endif %} data-ajax="{{ action.ACTION_AJAX }}" data-refresh="true">{{ action.L_ACTION }}</a>{% if not action.S_LAST_ROW %}&nbsp;|&nbsp;{% endif %}
+	<a href="{{ action.U_ACTION }}"{% if action.L_ACTION_EXPLAIN %} title="{{ action.L_ACTION_EXPLAIN }}"{% endif %}{% if action.COLOR %} style="color: {{ action.COLOR }};"{% endif %}{% if not S_AJAX_DISABLED %} data-ajax="{{ action.ACTION_AJAX }}" data-refresh="true"{% endif %}>{{ action.L_ACTION }}</a>{% if not action.S_LAST_ROW %}&nbsp;|&nbsp;{% endif %}
 {% endfor %}
--- a/phpBB/adm/style/acp_ext_list.html
+++ b/phpBB/adm/style/acp_ext_list.html
@@ -7,7 +7,17 @@
 	<p>{L_EXTENSIONS_EXPLAIN}</p>
 
 	<fieldset class="quick">
-		<span class="small"><a href="https://www.phpbb.com/go/customise/extensions/3.3" target="_blank">{L_BROWSE_EXTENSIONS_DATABASE}</a> &bull; <a href="{U_VERSIONCHECK_FORCE}">{L_VERSIONCHECK_FORCE_UPDATE_ALL}</a> &bull; <a href="javascript:phpbb.toggleDisplay('version_check_settings');">{L_SETTINGS}</a></span>
+		<span class="small">
+			{% if not S_AJAX_DISABLED %}
+				<a href="{{ U_AJAX_DISABLE }}" class="uses-js">{{ lang('DISABLE_AJAX_ACTIONS') }}</a>
+				&bull;
+			{% endif %}
+			<a href="https://www.phpbb.com/go/customise/extensions/3.3" target="_blank">{L_BROWSE_EXTENSIONS_DATABASE}</a>
+			&bull;
+			<a href="{U_VERSIONCHECK_FORCE}">{L_VERSIONCHECK_FORCE_UPDATE_ALL}</a>
+			&bull;
+			<a href="javascript:phpbb.toggleDisplay('version_check_settings');">{L_SETTINGS}</a>
+		</span>
 	</fieldset>
 
 	<form id="version_check_settings" method="post" action="{U_ACTION}" style="display:none">
--- a/phpBB/adm/style/admin.css
+++ b/phpBB/adm/style/admin.css
@@ -129,6 +129,10 @@ hr {
 	display: none;
 }
 
+.nojs .uses-js {
+	display: none;
+}
+
 @media only screen and (max-width: 800px), only screen and (max-device-width: 800px)
 {
 	body {
--- a/phpBB/includes/acp/acp_extensions.php
+++ b/phpBB/includes/acp/acp_extensions.php
@@ -152,6 +152,8 @@ class acp_extensions
 				$this->list_available_exts();
 
 				$this->template->assign_vars(array(
+					'S_AJAX_DISABLED'		=> $this->request->variable('ajax', '', true) === 'off',
+					'U_AJAX_DISABLE'		=> $this->u_action . '&amp;action=list&amp;ajax=off',
 					'U_VERSIONCHECK_FORCE' 	=> $this->u_action . '&amp;action=list&amp;versioncheck_force=1',
 					'FORCE_UNSTABLE'		=> $this->config['extension_force_unstable'],
 					'U_ACTION' 				=> $this->u_action,
--- a/phpBB/language/en/acp/extensions.php
+++ b/phpBB/language/en/acp/extensions.php
@@ -116,6 +116,7 @@ $lang = array_merge($lang, array(
 	'AUTHOR_HOMEPAGE'		=> 'Homepage',
 	'AUTHOR_ROLE'			=> 'Role',
 
+	'DISABLE_AJAX_ACTIONS'	=> 'Disable AJAX actions',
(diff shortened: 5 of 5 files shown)


DavidIQ:
I don't really get the point of this. AJAX should be able to handle whatever you're trying to fix by disabling it. Even upgrading phpBB uses AJAX and that is a much more intensive operation. Maybe fix the problem with AJAX instead?

mrgoldy:
The problem isn't AJAX. The problem is users installing extensions that have failing migrations. And not all exceptions are caught and rendered properly. These uncaught exceptions are shown properly and fully when the requests are not AJAX. However, if the request is AJAX, a generic message is shown with "Something went wrong processing your request". 

So, what ends up happening is that users start posting in the Support forums, without any proper error to show. So the first two pages ends up explaining how to get the actual error. Whether it is in the support forum or in the extension's support or development forum.

DavidIQ:
We should just show those errors, not have them try it, fail, click on the "Disable AJAX" link, then make them try it again.

3D-I:
Agreed with @DavidIQ - hence remove Ajax from there seems to be the solution?

DavidIQ:
@3D-I this isn't removing anything. It's adding a way to disable it after the fact.

3D-I:
> 
> 
> @3D-I this isn't removing anything. It's adding a way to disable it after the fact.

I know, I am suggesting to remove it instead, at least I tried.

https://tracker.phpbb.com/browse/PHPBB3-16270

DavidIQ:
You did fine, sorry. I don't think this or removing AJAX altogether is the solution though. Should probably be fixing AJAX to show the error messages. This feels very hacky.

3D-I:
> 
> 
> You did fine, sorry. I don't think this or removing AJAX altogether is the solution though. Should probably be fixing AJAX to show the error messages. This feels very hacky.

Also agreed, unfortunately I do not know how yet. That would be TEH solution ideed.

3D-I:
As a last thought, how many times an user needs to disable/enable extensions? Is it really Ajax needed there?

DavidIQ:
Doesn't happen a whole lot, but we want to be more "modern" I guess.

If we're rendering the error nicely without AJAX then it's already being handled. I think it probably just needs to be spewed out with the AJAX response.

3D-I:
> 
> 
> Doesn't happen a whole lot, but we want to be more "modern" I guess.
> 
> If we're rendering the error nicely without AJAX then it's already being handled. I think it probably just needs to be spewed out with the AJAX response.

Also @mrgoldy stated:

> This will affect one action, AJAX is automatically turned back on the next page load when viewing extensions. It is not persistent.

mrgoldy:
Yes, it is possible to catch all exceptions, for that matter we can even add a `catch (\Exception $e)` block. However, that's missing the point. If a non-caught exception is thrown, it means there is something wrong with the migrations of the extension and this can be anything. 

To generate a meaningful AJAX message as you are suggesting, would mean that we would have to create a debug back trace of our own, displaying all the required information in a tiny AJAX alert. Otherwise we will end up with just another incomplete message as we currently have.

Rather than trying to create our own, why not use the already available stack trace that the uncaught exception will show, on a full page. So it is not accidentally closed by pressing ESC, ENTER or clicking somewhere on the page for that matter.

The idea is so that everything behaves as normal, extensions actions are ajaxified. However, if there are errors during an action, which are not (or can not) be handled by the extensions module/manager, the user has the option to disable AJAX for a single action, the next action. 

This was added to aid support team members and extension writers, to better assist users who are having errors. Rather than having to explain how to currently get the full error message, which is quite hard for users to follow and - to be frank - not a place they should have to visit: Developer tools > Network > Request > Response OR Developer tools > Settings > Disable javascript. They can now just click a single button and try again.

DavidIQ:
As I said, if we are already displaying migration errors or any error for that matter in a nice way when there is no AJAX then they are already being handled, but not being returned to the client when AJAX is enabled. If fixing it properly proves too difficult then I would have to agree with @3D-I and just remove the AJAX. This proposed solution just isn't good UX.

3D-I:
I vote for the removal of Ajax if we agree, I do not see how it is of such extreme importance to have it in that location despite the benefits of not having it at all as it was, starting with the requests for support that I predicted came true.

dark-1:
> which is quite hard for users to follow and - to be frank - not a place they should have to visit: Developer tools > Network > Request > Response OR Developer tools > Settings > Disable javascript. They can now just click a single button and try again.

You could just tell them to <kbd>Right Click</kbd> and <kbd>Open link in new Tab/Window</kbd>,
Instead of `Disable javascript`.
It gives the proper error and all , coz it follows previous behavior.



> As I said, if we are already displaying migration errors or any error for that matter in a nice way when there is no AJAX then they are already being handled, but not being returned to the client when AJAX is enabled. If fixing it properly proves too difficult then I would have to agree with @3D-I and just remove the AJAX. This proposed solution just isn't good UX.

When `SQL General Error` is thrown while enabling the extension,
AJAX does not show any error at all... Instead goes on to do required DOM changes on HTML & then refreshes the page.

Side Note :
The AJAX doing DOM changes on HTML & then refresh the page for any action like Enable, Disable, Delete data.
Instead just refresh the page directly, which will also save some time...

So back to Topic :
If AJAX can not handle the Errors , then I think AJAX should be disabled on that page,
But if possible then I suggest to keep it , because it's a modern approach & of-coarse it looks COOL... 😜 

Best regards 👍

3D-I:
@dark-1 that is not relevant to the discussion IMHO.

dark-1:
> @dark-1 that is not relevant to the discussion IMHO.

@3D-I ,
Which Part ???

3D-I:
The whole except the last as in: `So back to Topic :`

dark-1:
> The whole except the last as in: `So back to Topic :`

@3D-I ,
Even this part ???

> > which is quite hard for users to follow and - to be frank - not a place they should have to visit: Developer tools > Network > Request > Response OR Developer tools > Settings > Disable javascript. They can now just click a single button and try again.
> 
> You could just tell them to Right Click and Open link in new Tab/Window,
> Instead of `Disable javascript`.
> It gives the proper error and all , coz it follows previous behavior.


Coz I thought this is better solution then manually "Disabling javascript" in "Developer tools" or making this PR...

Best regards 👍

3D-I:
We are trying to tell users what to do, it should be easy as it was.

It throws an error and the user goes to ask the extension's developer what's going on there instead ofasking at the support forums.

We all developers we know how to handle such situations, but the end users.

dark-1:
> We are trying to tell users what to do, it should be easy as it was.
> 
> It throws an error and the user goes to ask the extension's developer what's going on there instead of asking at the support forums.
> 
> We all developers we know how to handle such situations, but the end users.

For the same reason & easier solution for "end users" , I am saying...
For `Enable/Disable/Delete data` of Extension link :
On Desktop : <kbd>Right Click</kbd> and <kbd>Open link in new Tab</kbd>
On Touch : <kbd>Long Press</kbd> and <kbd>Open link in new Tab</kbd>

I don't think so the above instruction are tough for them...

Best regards 👍

3D-I:
We do not have to instruct them, enough said. :smile_cat:

dark-1:
> We do not have to instruct them, enough said. 😸

Ho!! you mean , not telling them anything???

Then disable the Damn AJAX on that page...
Which is way better than this PR...

Best regards 👍

3D-I:
Sometimes I think I was not as clear as the discussion itself, someone help me. Thank you. :disappointed:

dark-1:
@3D-I , 
doesn't matter... as long as it's constructive convo...

So my conclusion to get this solved is :
1. Make AJAX handle all the errors , but this is out of scope of this PR.
**OR**
2. Disable AJAX permanently on this page & not one time thing, this can be done in this PR.

Best regards 👍

3D-I:
Wow, that's what we are after! Do you reckon mate? :smile:

dark-1:
That what I am telling...

I also gave  a temporally solution... with "Right click"...

What's happening here mate ???

3D-I:
We are discussing if drop Ajax or implement a button to "disable javascript for just one action".
We do not need to instruct end users as in: "we developers we know how to handle such situations".

It seems to me that the topic got focus already.

dark-1:
Yes!! you are right !!! 👍

mrgoldy:
I still feel like AJAX is a nice addition for extension actions and for a lot of administrators it will properly work flawlessly, especially if they only decide to install validated or recognized extensions. So completely removing AJAX and having them go through the various pages seems redundant.

However, that's optimistic thinking and we know exceptions occur during the installation of extensions. So that's why I proposed this solution, where we can have the best of both worlds. This will not affect administrator that 'live by the book' and even most of the others aswell, it just allows for a far easier debug method.

DavidIQ:
It's still bad workflow and am not in favor of this. It screams "I don't know how to fix this properly so here's a workaround _built-in_ to our product".

dark-1:
> It's still bad workflow and am not in favor of this. It screams "I don't know how to fix this properly so here's a workaround _built-in_ to our product".

Exactly ...
that's the reason I am saying ,
**EITHER** remove AJAX completely
**OR** keep AJAX and solve the problem
**OR** keep AJAX as-is and give user the workaround for debug to get an Err0r :
For `Enable/Disable/Delete data` of Extension link :
On Desktop : <kbd>Right Click</kbd> and <kbd>Open link in new Tab</kbd>
On Touch : <kbd>Long Press</kbd> and <kbd>Open link in new Tab</kbd>

As simple as that...
This PR is not the professional way to tackle.

Best regards 👍

neuropass:
agree with @3D-I @DavidIQ . Can you guys just simply revert Senky's previous PR? @mrgoldy this one: https://github.com/phpbb/phpbb/pull/5223

kasimi:
Here's another vote to entirely remove ajaxified extension actions. I don't feel they speed up the process, but they do make it harder to understand errors.

The confirmation/success pages probably are not adding much value in most cases, but instead of randomly ajaxifying things, they should be dealt with on a broader scale.

VSEphpbb:
I agree with @DavidIQ in that if the problem is AJAX related failures don't produce the same user feedback as non-AJAX failures, the correct solution here is not to add a band-aid disable for AJAX (or to remove it) but instead to make sure whatever errors are consistent for both AJAX and non-AJAX interactions. 

And what could be nicer about keeping AJAX, besides the better experience, is when PHP failes completely, instead of a blank screen as happens w/o AJAX, at least with AJAX we can instead display a meaningful message that the operation failed and the user should examine their server PHP error logs, etc.

3D-I:
I agree with that too, but I have to say that Ajax with these 3 functions is not necessary at all since you are deceived during the refresh of the page while you think that everything is over and you try to do another action but you can't. So in the end no, I remain of my opinion since reproducing all the possible errors in Ajax doesn't seem to be possible I opt for the switch, or the total removal of Ajax.

The latter the better IMHO.

neuropass:
yes revert this patch. I just don't get why you guys have so long for a ticket and always get 1000 replies? debating things to a pulp. Nothing gets done like that. its clear the majority here wants ajax removed. So be it and revert it unless someone can provide a pull-request addressing all the ajax issues. Which doesn't look like it. so please revert that original PR.

3D-I:
Because removing a feature it is not the same as adding it.

mrgoldy:
@neuropass Because the majority here (which has a whopping 7 participants) can not 'quickly' make a decision for thousands of people. You're more than welcome to participate in discussions and development to speed things up. We do our best, in our free time, to please everyone using phpBB..

PR #5902, opened 2020-03-14
[ticket/15937] Google reCAPTCHA v3 Plugin

mrgoldy:
### Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15937

### Other tickets:
These tickets are however made for reCAPTCHA v2, but implemented in this PR for v3. 
Not sure if they should be implemented for v2 plugin aswell, cause then I'll have to refactor.
- [Use "global" reCAPTCHA domain to circumvent blocking in some countries](https://tracker.phpbb.com/browse/PHPBB3-16105)
- [Google reCaptcha URL wrong when behind an HTTPS proxy](https://tracker.phpbb.com/browse/PHPBB3-15426)
- [Allow selecting ReCaptcha's "RequestMethod" rather than using the default POST method](https://tracker.phpbb.com/browse/PHPBB3-14988)

### 3.3 Branch
I made this ticket against the 3.3 branch as I feel like we should not wait until 4.0 to provide this captcha plugin.

### Migration
Actually my first time having to create a migration for phpBB core (not sure how I've managed so many PR's without one), so please verify if everything is correct.

### Tests
I've tested this locally and it's working. Both v2 and v3.
I'll see if I can set it up on one of our live test boards aswell.
But if people can test this aswell (locally and/or live), that would be appreciated.
Please test both v3 and v2, as I've slightly adjusted the JS for v2 aswell (to prevent repitition).

### DIFF
--- a/phpBB/phpbb/captcha/plugins/recaptcha_v3.php
+++ b/phpBB/phpbb/captcha/plugins/recaptcha_v3.php
@@ -262,7 +262,7 @@ class recaptcha_v3 extends captcha_abstract
 
 			'RECAPTCHA_ACTION'		=> self::$actions[$this->type] ?? reset(self::$actions),
 			'RECAPTCHA_KEY'			=> $config['recaptcha_v3_key'] ?? '',
-			'U_RECAPTCHA_SCRIPT'	=> sprintf('//%s/recaptcha/api.js?render=%s', $domain, $render),
+			'U_RECAPTCHA_SCRIPT'	=> sprintf('//%1$s/recaptcha/api.js?render=%2$s', $domain, $render),
 
 			'S_CONFIRM_CODE'		=> true,
 			'S_RECAPTCHA_AVAILABLE'	=> $this->is_available(),
--- a/phpBB/phpbb/captcha/plugins/recaptcha_v3.php
+++ b/phpBB/phpbb/captcha/plugins/recaptcha_v3.php
@@ -14,18 +14,27 @@
 namespace phpbb\captcha\plugins;
 
 /**
- * Google reCaptcha v3 plugin.
+ * Google reCAPTCHA v3 plugin.
  */
 class recaptcha_v3 extends captcha_abstract
 {
-	const CURL		= 'curl';
-	const POST		= 'post';
-	const SOCKET	= 'socket';
+	/**
+	 * Possible request methods to verify the token.
+	 */
+	const CURL   = 'curl';
+	const POST   = 'post';
+	const SOCKET = 'socket';
 
-	const GOOGLE	= 'google.com';
-	const RECAPTCHA	= 'recaptcha.net';
+	/**
+	 * Possible domain names to load the script and verify the token.
+	 */
+	const GOOGLE    = 'google.com';
+	const RECAPTCHA = 'recaptcha.net';
 
+	/** @var string Default action when no other applies */
 	static protected $action = 'default';
+
+	/** @var array CAPTCHA types mapped to their action with threshold */
 	static protected $actions = [
 		CONFIRM_REG		=> 'register',
 		CONFIRM_LOGIN	=> 'login',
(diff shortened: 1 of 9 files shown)


mrgoldy:
I've set it up on one of our live test sites, feel free to test it out.
It is activated during registration, and the 'second' login attempt.
Meaning if you enter your email address falsely once, it will prompt the reCAPTCHA for login aswell.

https://oberon.phpbbstudio.com/

3D-I:
This is what it looks like in the ACP.

immagine

DrDeath:
Teste this PR on https://oberon.phpbbstudio.com/
Seems to work for me

phpBB/phpbb/db/migration/data/v330/google_recaptcha_v3.php
```diff
@@ -0,0 +1,63 @@
...
+		$data = [
+			['config.add', ['recaptcha_v3_key', '']],
+			['config.add', ['recaptcha_v3_secret', '']],
+			['config.add', ['recaptcha_v3_domain', \phpbb\captcha\plugins\recaptcha_v3::GOOGLE]],
+			['config.add', ['recaptcha_v3_method', \phpbb\captcha\plugins\recaptcha_v3::POST]],
+		];
+
+		foreach (\phpbb\captcha\plugins\recaptcha_v3::get_actions() as $action)
+		{
+			$data[] = ['config.add', "recaptcha_v3_threshold_{$action}", 0.5];
```
EA117: I believe this is supposed to be:

`$data[] = ['config.add', ["recaptcha_v3_threshold_{$action}", 0.5]];`

i.e. Second $data[] dimension is also supposed to be an array.

phpBB/phpbb/db/migration/data/v330/google_recaptcha_v3.php
```diff
@@ -0,0 +1,63 @@
...
+		$data = [
+			['config.remove', ['recaptcha_v3_key']],
+			['config.remove', ['recaptcha_v3_secret']],
+			['config.remove', ['recaptcha_v3_domain']],
+			['config.remove', ['recaptcha_v3_method']],
+		];
+
+		foreach (\phpbb\captcha\plugins\recaptcha_v3::get_actions() as $action)
+		{
+			$data[] = ['config.remove', "recaptcha_v3_threshold_{$action}"];
```
EA117: I believe this is supposed to be:

`$data[] = ['config.remove', ["recaptcha_v3_threshold_{$action}", '']];`

i.e. Second $data[] dimension is also supposed to be an array.

phpBB/phpbb/db/migration/data/v330/google_recaptcha_v3.php
3D-I: It is correct, $data is the first array. You can use this file and test it for yourself.

```
<?php
define('IN_PHPBB', true);
$phpbb_root_path = (defined('PHPBB_ROOT_PATH')) ? PHPBB_ROOT_PATH : './';

$action = 'test';
$data[] = ['config.add', ["recaptcha_v3_threshold_{$action}", 0.5]];
var_dump($data);

//https://area51.phpbb.com/docs/dev/3.2.x/migrations/tools/config.html#example
$a = array(array('config.add', array('foo', 'bar')));
var_dump($a);

```

phpBB/phpbb/db/migration/data/v330/google_recaptcha_v3.php
3D-I: As per above, plus you do not have to add the type while removing configurations.

https://area51.phpbb.com/docs/dev/3.2.x/migrations/tools/config.html#delete-config-setting

phpBB/phpbb/db/migration/data/v330/google_recaptcha_v3.php
3D-I: Ops, I used your example which it seems correct.

phpBB/phpbb/db/migration/data/v330/google_recaptcha_v3.php
```diff
@@ -0,0 +1,63 @@
...
+		}
+
+		return $data;
+	}
```
3D-I: The function `revert_data()` can be removed IMHO since configurations are automatically removed like permissions.

mrgoldy: I was thinking that this way any possible new actions would be removed aswell, but that's also already handled by the `update_data` inverse. So removed.

phpBB/phpbb/captcha/plugins/recaptcha_v3.php
```diff
@@ -0,0 +1,350 @@
...
+
+		$domain = $config['recaptcha_v3_domain'] ?? self::GOOGLE;
+		$render = $config['recaptcha_v3_key'] ?? '';
+
+		$template->assign_vars([
+			'CONFIRM_EXPLAIN'		=> $language->lang($explain, '<a href="' . $contact . '">', '</a>'),
+
+			'RECAPTCHA_ACTION'		=> self::$actions[$this->type] ?? reset(self::$actions),
+			'RECAPTCHA_KEY'			=> $config['recaptcha_v3_key'] ?? '',
+			'U_RECAPTCHA_SCRIPT'	=> sprintf('//%s/recaptcha/api.js?render=%s', $domain, $render),
```
marc1706, line 265: Please use the same method as in language files, e.g.
```suggestion
			'U_RECAPTCHA_SCRIPT'	=> sprintf('//%1$s/recaptcha/api.js?render=%2$s', $domain, $render),
```

marc1706:
Apart from that tested to be working fine for both recaptcha v2 and v3.

PR #5914, opened 2020-03-17
[ticket/16319] Responsive post profile

mrgoldy:
PHPBB3-16319

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16319

### DIFF
--- a/phpBB/styles/prosilver/theme/responsive.css
+++ b/phpBB/styles/prosilver/theme/responsive.css
@@ -552,22 +552,18 @@
 	}
 }
 
-@media (min-width: 700px) {
-	.postbody { width: 70%; }
-}
-
-@media (min-width: 701px) and (max-width: 850px) {
+@media (min-width: 701px) {
 	.postbody {
-		width: 76%;
+		/** deduct postprofile width (including border width, margin and padding) */
+		width: calc(100% - 200px - 1px);
 	}
 
 	.postprofile {
-		width: 22%;
+		width: 200px;
 	}
 }
 
 @media (min-width: 701px) and (max-width: 950px) {
-
 	ul.topiclist dt {
     	margin-right: -410px;
 	}
--- a/phpBB/styles/prosilver/theme/responsive.css
+++ b/phpBB/styles/prosilver/theme/responsive.css
@@ -150,7 +150,7 @@
 		margin-right: -5px;
 	}
 
-	.cp-main .forabg, .cp-main .forumdb, .cp-main .post, .cp-main .panel {
+	.cp-main .forabg, .cp-main .forumbg, .cp-main .post, .cp-main .panel {
 		border-radius: 7px;
 	}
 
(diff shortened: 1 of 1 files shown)


phpBB/styles/prosilver/theme/responsive.css
```diff
@@ -150,7 +150,7 @@
 		margin-right: -5px;
 	}
 
-	.cp-main .forabg, .cp-main .forumdb, .cp-main .post, .cp-main .panel {
+	.cp-main .forabg, .cp-main .forumbg, .cp-main .post, .cp-main .panel {
```
hanakin: wow 7 year old error there made it through 3 patches

phpBB/styles/prosilver/theme/responsive.css
```diff
@@ -557,14 +556,14 @@
...
-
+@media (min-width: 701px) and (max-width: 850px) {
+	.postbody {
+		width: 76%;
+	}
 
+	.postprofile {
+		width: 22%;
+	}
 }
```
hanakin: this is a failed implementation. The postprofile should not be shrunk in-fact as I stated in the post on .com it should be hardcoded 200px with a 10px left margin and the postbody should be scaled down when necessary. This is why I said it would require a complete refactor for the responsiveness of these two elelements.

hanakin: it will probably require 3 or adjustments under 1100px or so

mrgoldy: Okay, so what is it you want me to do here?
You obviously have something specific in mind?

And isn't hardcoding the width to 200px and then setting the `.postbody` to `calc(100% - 210px)` an option?

hanakin: Yes you can do that in 3.3 or 4.0, but still have to factor out all the other stuff. But 3.2 still supports ie 10/11 and technically 3.3 still supports 11 partially which has problems with calc

hanakin:
As stated in the topic on .com rebase this on 3.3 forget about 3.2 its EOL soon enough. 3.3 does not require IE10. Test it using calc as discussed above on all the necessary browsers including IE11 and see if it works.

3.3 Browser support https://browserl.ist/?q=%3E+1%25 (We support 1% browsers but we only test on the following)
```json
	"ie 11",
	"edge >= 18",
	"ff >= 72",
	"chrome >= 79",
	"safari >= 13",
	"ios >= 12.4"
```
All the other browsers in the link above are considered outliers

mrgoldy:
Okay, rebased on 3.3.x
Made the `.postprofile` a fixed `200px`.
Commented the `calc()` for the `.postbody` width.

Tested this on Chrome, Firefox, Edge and IE11 and works for me.
If someone on a Mac could test this for me aswell, that would be greatly appreciated.

hanakin:
I’ll test it when I get a chance

PR #5916, opened 2020-03-18
[ticket/13426] Improve user timezone initialization

rxu:
Checklist:

- [x] Tests pass

<a href="https://tracker.phpbb.com/browse/PHPBB3-13426">PHPBB3-13426</a>.

### DIFF
--- a/phpBB/phpbb/user.php
+++ b/phpBB/phpbb/user.php
@@ -227,15 +227,7 @@ class user extends \phpbb\session
 
 		$this->language->set_user_language($user_lang_name);
 
-		try
-		{
-			$this->timezone = new \DateTimeZone($user_timezone);
-		}
-		catch (\Exception $e)
-		{
-			// If the timezone the user has selected is invalid, we fall back to UTC.
-			$this->timezone = new \DateTimeZone('UTC');
-		}
+		$this->create_timezone($user_timezone);
 
 		$this->add_lang($lang_set);
 		unset($lang_set);
@@ -635,7 +627,7 @@ class user extends \phpbb\session
 		if (!$format_date_override)
 		{
 			$time = new $this->datetime($this, '@' . (int) $gmepoch, $utc);
-			$time->setTimezone($this->timezone);
+			$time->setTimezone($this->create_timezone());
 
 			return $time->format($format, $forcedate);
 		}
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/user.php
```diff
@@ -645,6 +637,36 @@ function format_date($gmepoch, $format = false, $forcedate = false)
...
 
+	/**
+	 * Create a DateTimeZone object in the context of the current user
+	 *
+	 * @param string $user_timezone Time zone of the current user.
+	 * @return DateTimeZone DateTimeZone object linked to the current users locale
+	 */
+	public function create_timezone($user_timezone = null)
+	{
+		if (!isset($this->timezone))
```
mrgoldy: I know this works, but I am not a fan of checking if a variable is `null` with `isset()`.
`isset()` should be used to check if a variable is defined, not if it is `null`. Atleast in my eyes.
This contradicts with the actual function name, cause it is set, it just does not have the desired value.
Personally I would use `if ($this->timezone === null)`.

phpBB/phpbb/user.php
```diff
@@ -645,6 +637,36 @@ function format_date($gmepoch, $format = false, $forcedate = false)
...
+		{
+			global $config;
```
mrgoldy: Strictly speaking this `global` is only needed inside the `if` block below.

phpBB/phpbb/user.php
```diff
@@ -645,6 +637,36 @@ function format_date($gmepoch, $format = false, $forcedate = false)
...
+			{
+				$this->timezone = new \DateTimeZone($user_timezone);
+			}
+			catch (\Exception $e)
+			{
+				// If the timezone the user has selected is invalid, we fall back to UTC.
+				$this->timezone = new \DateTimeZone('UTC');
+			}
+		}
+		return $this->timezone;
```
mrgoldy: Could we have an empty line before the `return` statement.

phpBB/phpbb/user.php
rxu: > This contradicts with the actual function name, cause it is set

Where?

dark-1: This is actually ok.

"Determine if a variable is considered set, this means if a variable is declared and is different than NULL."
- https://www.php.net/manual/en/function.isset.php

mrgoldy: https://github.com/phpbb/phpbb/blob/master/phpBB/phpbb/user.php#L35
As I said, I understand this will work as it currently is, but I am not a fan of using `isset()` to check if a variable is `null`, rather than if the variable actually exists.
```php
if ($something)
{
    $variable = 'i am set';
}

$var = isset($variable) ?: 'is set' : 'is not set';
```
In such a manner `isset()` makes sense to me, as the variable might be completely not defined, not initialised.
But that is not the case for `$timezone`, it is initialised, just with a value of `null`.

But as I mentioned, this might just be my view on how `isset()` should be used.

rxu: Simply `!$this->timezone` can be used instead.

mrgoldy: That's exactly what I am saying @dark-1 .. I know it is correct, but that doesn't mean we should use it that way. That's the entire discussion I am bringing up.
I am indeed more in favor of `$this->timezone === null` or `!$this->timezone` for that matter.

dark-1: > Simply `!$this->timezone` can be used instead.

Nice... 👍 

---

> That's exactly what I am saying @dark-1 .. I know it is correct, but that doesn't mean we should use it that way. ...

No worries 👍

PR #5919, opened 2020-03-21
[ticket/16338] Add DuckDuckGo bot

AlfredoRamos:
Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-16338

### DIFF
--- a/phpBB/phpbb/db/migration/data/v33x/bot_update.php
+++ b/phpBB/phpbb/db/migration/data/v33x/bot_update.php
@@ -1,0 +1,91 @@
+<?php
+/**
+ *
+ * This file is part of the phpBB Forum Software package.
+ *
+ * @copyright (c) phpBB Limited <https://www.phpbb.com>
+ * @license GNU General Public License, version 2 (GPL-2.0)
+ *
+ * For full copyright and license information, please see
+ * the docs/CREDITS.txt file.
+ *
+ */
+
+namespace phpbb\db\migration\data\v33x;
+
+class bot_update extends \phpbb\db\migration\migration
+{
+	static public function depends_on()
+	{
+		return ['\phpbb\db\migration\data\v330\v330'];
+	}
+
+	public function update_data()
+	{
+		return [
+			['custom', [[$this, 'add_duckduckgo_bot']]],
+		];
+	}
+
+	public function add_duckduckgo_bot()
+	{
+		$bot_name = 'DuckDuckGo [Bot]';
+		$bot_name_clean = utf8_clean_string($bot_name);
+
+		$sql = 'SELECT user_id
+			FROM ' . $this->table_prefix . 'users
+			WHERE ' . $this->db->sql_build_array('SELECT', ['username_clean' => $bot_name_clean]);
+		$result = $this->db->sql_query($sql);
+		$bot_exists = (bool) $this->db->sql_fetchfield('user_id');
+		$this->db->sql_freeresult($result);
+
+		if (!$bot_exists)
+		{
+			$bot_agent = 'DuckDuckBot/';
+			$bot_ip = '';
+			$sql = 'SELECT group_id, group_colour
+				FROM ' . $this->table_prefix . 'groups
+				WHERE ' . $this->db->sql_build_array('SELECT', ['group_name' => 'BOTS']);
+			$result = $this->db->sql_query($sql);
+			$group_row = $this->db->sql_fetchrow($result);
+			$this->db->sql_freeresult($result);
+
+			// Default fallback, should never get here
+			if (!$group_row)
+			{
+				$group_row['group_id'] = 6;
+				$group_row['group_colour'] = '9E8DA7';
+			}
+
+			if (!function_exists('user_add'))
+			{
+				include($this->phpbb_root_path . 'includes/functions_user.' . $this->php_ext);
+			}
+
+			$user_row = [
+				'user_type'				=> USER_IGNORE,
+				'group_id'				=> $group_row['group_id'],
+				'username'				=> $bot_name,
+				'user_regdate'			=> time(),
+				'user_password'			=> '',
+				'user_colour'			=> $group_row['group_colour'],
+				'user_email'			=> '',
+				'user_lang'				=> $this->config['default_lang'],
+				'user_style'			=> $this->config['default_style'],
+				'user_timezone'			=> 0,
+				'user_dateformat'		=> $this->config['default_dateformat'],
+				'user_allow_massemail'	=> 0,
+			];
+
+			$user_id = user_add($user_row);
+			$sql = 'INSERT INTO ' . $this->table_prefix . 'bots ' . $this->db->sql_build_array('INSERT', [
+				'bot_active'	=> 1,
+				'bot_name'		=> (string) $bot_name,
+				'user_id'		=> (int) $user_id,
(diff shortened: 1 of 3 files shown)


phpBB/phpbb/db/migration/data/v33x/bot_update.php
```diff
@@ -0,0 +1,91 @@
...
+			$sql = 'INSERT INTO ' . BOTS_TABLE . ' ' . $this->db->sql_build_array('INSERT', [
+				'bot_active'	=> 1,
+				'bot_name'		=> (string) $bot_name,
+				'user_id'		=> (int) $user_id,
+				'bot_agent'		=> (string) $bot_agent,
+				'bot_ip'		=> (string) $bot_ip,
+			]);
+			$this->db->sql_query($sql);
+		}
+	}
```
3D-I: Wondering if we need to `$cache->destroy('_bots');` here?

phpBB/phpbb/db/migration/data/v33x/bot_update.php
```diff
@@ -0,0 +1,91 @@
...
+				'user_email'			=> '',
+				'user_lang'				=> $this->config['default_lang'],
+				'user_style'			=> $this->config['default_style'],
+				'user_timezone'			=> 0,
+				'user_dateformat'		=> $this->config['default_dateformat'],
+				'user_allow_massemail'	=> 0,
+			];
+
+			$user_id = user_add($user_row);
+			$sql = 'INSERT INTO ' . BOTS_TABLE . ' ' . $this->db->sql_build_array('INSERT', [
```
3D-I: What if the bots table does not exists? Edge case.
Not sure about the use of constants here, on a side note.

phpBB/phpbb/db/migration/data/v33x/bot_update.php
AlfredoRamos: I don't think that's really needed.

At least migrations that change configuration options, do not purge the cache.

3D-I: Yeah we do not.

phpBB/phpbb/db/migration/data/v33x/bot_update.php
AlfredoRamos: It will fail, like any other migration where a column or table is missing. I would treat a missing bots table like a broken install.

I'm not sure about the constant either, but I was following the code style from recent migration files, like this one.

https://github.com/phpbb/phpbb/blob/master/phpBB/phpbb/db/migration/data/v330/remove_attachment_flash.php

I don't even like the final comma in the array elements, but that's in the coding guidelines.

phpBB/phpbb/db/migration/data/v33x/bot_update.php
```diff
@@ -0,0 +1,91 @@
...
+{
+	static public function depends_on()
+	{
+		return ['\phpbb\db\migration\data\v330\v330'];
+	}
+
+	public function update_data()
+	{
+		return [
+			['custom', [[&$this, 'add_duckduckgo_bot']]],
```
mrgoldy, line 26: Any particular reason why this object is passed by reference?
Looks like a copy and paste and not providing any actual functionality.

phpBB/phpbb/db/migration/data/v33x/bot_update.php
mrgoldy: Preferably use the `$this->table_prefix` parameter which is already available. I know it isn't used in the other migration either, but we have to start somewhere. Also for the table constants above.

phpBB/phpbb/db/migration/data/v33x/bot_update.php
AlfredoRamos: > Looks like a copy and paste and not providing any actual functionality.

Indeed.

phpBB/phpbb/db/migration/data/v33x/bot_update.php
dark-1: > Preferably use the `$this->table_prefix` parameter which is already available. I know it isn't used in the other migration either, but we have to start somewhere. Also for the table constants above.

May I know where this discussion was made with regards to this?

Best regards 👍

mrgoldy: The idea is to move away from defined constants. So if there is an option to use the table prefix, ideally that should be used.

dark-1: So no actual discussion topic was made ?
Or it was made & decided that `$this->table_prefix` is to be used ?

Thanks for the responce.

Best regards 👍

mrgoldy: There are several discussions throughout, but there have been quite a few, not sure where they all are located. Table prefix is still not the ideal solution, that would be to use the tables parameter (the array). But that isn't added to 3.3 base migration yet, it is already in 4.0.

dark-1: Ahh! ok.

Thanks for your kind info. 😄 

Best regards 👍

phpBB/phpbb/db/migration/data/v33x/bot_update.php
```diff
@@ -0,0 +1,91 @@
...
+			['custom', [[$this, 'add_duckduckgo_bot']]],
+		];
+	}
+
+	public function add_duckduckgo_bot()
+	{
+		$bot_name = 'DuckDuckGo [Bot]';
+		$bot_name_clean = utf8_clean_string($bot_name);
+
+		$sql = 'SELECT user_id
```
Crizz0, line 35: Question: Does it make sense ot check for $bot_agent in the "bot_exists" SQL-query. Because when some user added it himself, the name can be more or less anything, but not he user agent for the same bot.

AlfredoRamos: @Crizz0 Totally, however `bot_agent` is not guaranteed to be unique in the table, you can have two bots with the same `bot_agent` but different username.

You cannot, however, have two users with the same name, hence the check in the users table.

I could check for bots with that `bot_agent` and update its bot and user data if it already exists, in case a user tried to impersonate a bot, however what should I do if there's more than one? delete other to leave only one and edit that one?

3D-I: I have Duck there as `DuckDuckBot/1.0; (+http://duckduckgo.com/duckduckbot.html)`

immagine

AlfredoRamos: @3D-I Yeah, that's another problem, bots with different `bot_agent`, however that would be a little easier to "fix" with a `preg_match()`, because they both start with `DuckDuckBot/`.

In my opinion, the `bot_agent` should be a partial match, because the version (for example) can change overtime. However that partial match should be specific enough to allow to identify that and only that bot.

In fact there's a `DuckDuckBot/1.1` user agent ([reference])

[reference]: https://github.com/duckduckgo/duckduckgo-publisher/blob/master/share/site/duckduckgo/duckduckbot.tx

3D-I: Yeah, your proposal is right as it is IMO. :+1:

PR #5921, opened 2020-03-24
[ticket/16413] Add an option to define CPF FontAwesome contact icon

rxu:
Checklist:

- [x] Tests pass

<a href="https://tracker.phpbb.com/browse/PHPBB3-16413">PHPBB3-16413</a>.

image

image

https://github.com/user-attachments/assets/ac4e55f7-51c6-4899-a6c3-d567ef3911c4

image


hanakin:
We are transitioning how we work with icons in 4.0. So this may need to be implemented different. I’m not exactly sure what this is yet. Give me sometime and I’ll get back to you after I have time to fully look things over.

hanakin:
ok so yeah not going to make it into 3.3, this will have to go into 4.0. It does require some changes to utilize the new icon system though. Not sure how yet. essentially we would have to expand this to default fields as well. With the ability to edit the icons for them in the ACP. so that we can simply just run a loop through them in the template to call the icon function. As for the acp changes. the js you are using will need to be redone as we are no longer using font-awsome alone. We are now supporting these format svg/png/iconify/font for icons. This means you would have to serve back the icon in one of these formats based on another field. It would still work very similar in that the name of the icon would be the name provided in the icon field. For reference the loading of an icon looks like this respectively. 

SVG: load based on input.svg
PNG: load based on input.png
ICONIFY: `<i class="iconify o-icon-src-iconify o-icon" data-icon="{{ input }}" data-inline="true"></i>`
FONTAWSOME: `<i class="o-icon o-icon-font fa-{{ input }} fw` etc

you can reference the icon function here: https://github.com/phpbb/phpbb/pull/5545/files

rxu:
So am I getting it right that this PR is ok for 3.3 but for 4.0 there should be another one with the use of new icon system? @hanakin

hanakin:
No there. It will not work for 3.3

rxu:
I see, so 2 different PRs for 3.3 and 4.0? 😃

hanakin:
No it will no5 make it into 3.3 just work on one for 4.0

rxu:
Oh. Regret, so postponed for several years ahead :)
But whatever, ok then.

mrgoldy:
If you don't mind me asking, but why is it not possible for 3.3?
Cause it sounds like it has to be implemented in some fashion for 4.0 anyway, then atleast with this there is a basis for it. Which might have to be altered and adjusted, but the groundwork has been done.

hanakin:
Ok let me put it another way. This branch is for 3.3.x released which are security and bug fixes only. All new features need to go against master branch. Regardless of what version we are calling it be it 3.4 or 4.0. Which means that he can take advantage of the icon function. Which means no need to do a version without.

rxu:
Since there's no way to automatically detect icon type in 4.0, there's no way to implement this feature.

hanakin:
What do You mean by detect icon type? Why do you need to do that?

rxu:
Because only icon name input is not enough to render the icon now.

hanakin:
So? Add a field for type

rxu:
2 new DB fields for a single icon? :) Nice.
I'll think about it.

hanakin:
Well you do not need two, you can repurpose the name field as the icon field

mrgoldy:
Perhaps it's possible to store the icon name the same way as Iconify determines it:
`fa:user`
`mdi:user`
`svg:user`

hanakin:
Well you have to do that anyway if your using iconify. But you mean for svg and png? Not sure how that would work for fonts. font-fa

rxu:
> Perhaps it's possible to store the icon name the same way as Iconify determines it

I don't think phpBB is going to use it like that, is it?

hanakin:
Not sure who you are responding to. We will do whatever makes the most sense to work. The default icons will be served via iconify where possible. We will have some custom icons. If we can not get them into iconify then they would be served via svg

rxu:
I don't think I understand what are you talking about. Looking at #5545 I see almost nothing from described at https://iconify.design.
So I have no idea what are you offering me to implement in this PR, sorry. I guess I'll hold it up to some 4.0-beta or so, thanks.

mrgoldy:
There are various types that the `{{ Icon() }}` function takes: `font`, `iconify`, `png`, `svg`.
You want to provide an interface in the ACP for selecting an icon for a custom profile field.
If you want to use FontAwesome icons for the contact icons, you can add the FontAwesome picker as you're doing here. Then in the styles you'll be able to call `{{ Icon('font', '<the icon>') }}`.
If you want to allow more than just FA icons, you can add a 'select icon type' dropdown before the input. So users can select from `font | iconify | png | svg`. And then still insert their desired icon. You can store this in a single column in the db, concatenating them with colon `:`, eg: `font:user`. Then when displaying the icon for the profile field, you'll be able to grab the part before the first colon and use that as your icon type: `{{ Icon('font', 'user') }}`. Only thing then would be that iconify icons will have a 2nd colon that will need to be handled, eg: `iconify:mdi:user`.

hanakin:
That created the icon function. Iconify works by adding a js file to the footer of the new theme. It it looks for an html element with the class iconify and a data attribute with the icon library:name. We have not added any js yet if that is what you are asking. The or for switching over the prosIlver is one will do that, but it’s not merged yet.

hanakin:
so to clarify fonts is only supported for font-awesome at the moment for BC
with prosilver. I would just not support it here as we will eventually not
support it anymore down the road. You can use iconify to access most if not
all font-awsome anyway. You will also not be able to have a picker and that
is not something I think that we need here. It would be way to resource
costly to implement and a nightmare to maintain.

phpBB/adm/style/ajax.js
```diff
@@ -429,12 +429,49 @@ $(function() {
 		$field_icon.after('<i style="font-size: 16px; margin:0 6px; vertical-align: middle;"></i>');
 	}
 
-	$('#field_icon').on('keyup blur', function() {
+	$field_icon.on('keyup blur', function() {
 		var input = $(this).val();
 		var $icon = $(this).next('i');
-		$icon.attr('class', 'o-icon o-icon-font fa-fw fa-' + input + ' fas');
+		$icon.attr('class', 'o-icon o-icon-font fa-fw fa-' + input + ' fas acp-icon');
```
iMattPro, line 435:
```
if (fieldIcon) {
	fieldIcon.addEventListener('keyup', updateIconClass);
	fieldIcon.addEventListener('blur', updateIconClass);
}

function updateIconClass(event) {
	const input = event.target.value;
	const icon = event.target.nextElementSibling;

	if (icon && icon.tagName.toLowerCase() === 'i') {
		icon.className = `o-icon o-icon-font fa-fw fa-${input} fas acp-icon`;
	}
}
```

phpBB/adm/style/ajax.js
```diff
@@ -429,12 +429,49 @@ $(function() {
...
+	};
+
+	const handleInput = ({ target }) => {
+		if (target === colorPicker) {
+			colorText.value = target.value.substring(1);
+		} else {
+			syncColors(colorText, colorPicker);
+		}
+		var icon = $field_icon.next('i');
+		icon.css('color', colorPicker.value);
```
iMattPro, line 464:
```
const icon = fieldIcon?.nextElementSibling;
if (icon && icon.tagName.toLowerCase() === 'i') {
	icon.style.color = colorPicker.value;
}
```

phpBB/adm/style/ajax.js
rxu: That works, but in case of empty icon name field leaves `fa-` in classes.

phpBB/language/en/acp/profile.php
```diff
@@ -93,6 +93,9 @@
 	'FIELD_DESCRIPTION'			=> 'Field description',
 	'FIELD_DESCRIPTION_EXPLAIN'	=> 'The explanation for this field presented to the user.',
 	'FIELD_DROPDOWN'			=> 'Dropdown box',
+	'FIELD_ICON'				=> 'Field icon',
+	'FIELD_ICON_COLOR'			=> 'Icon colour',
+	'FIELD_ICON_EXPLAIN'		=> 'Enter the name of a Font Awesome icon to use if displaying as a contact field option above is set. Optionally, set the icon colour in 6-value hexadecimal format or choose it using the colour picker next to the icon colour field (only effective when Font Awesome icon name is set). Leave Font Awesome icon field blank to use phpBB default contact field icon if any (this will also empty the icon colour field).',
```
iMattPro, line 98:
```
Enter a Font Awesome icon name to display with this contact field. Optionally, set its colour using a 6-digit hex code or the colour picker. Leave blank to use phpBB’s default icon and clear the colour.
```

PR #5932, opened 2020-04-05
[ticket/15572] Fix W3C validation error for quote BBCode

AlfredoRamos:
Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-15572

### DIFF
--- a/phpBB/styles/prosilver/template/bbcode.html
+++ b/phpBB/styles/prosilver/template/bbcode.html
@@ -13,6 +13,9 @@
 <!-- BEGIN quote_close --></div></blockquote><!-- END quote_close -->
 <!-- BEGIN quote_extended -->
 <blockquote>
+	<xsl:if test="@post_url">
+		<xsl:attribute name="cite"><xsl:value-of select="@post_url"/></xsl:attribute>
+	</xsl:if>
 	<xsl:if test="not(@author)">
 		<xsl:attribute name="class">uncited</xsl:attribute>
 	</xsl:if>
@@ -42,7 +45,7 @@
 					<a href="{@msg_url}" data-msg-id="{@msg_id}">&#8593;</a>
 				</xsl:if>
 				<xsl:if test="@date">
-					<div class="responsive-hide"><xsl:value-of select="@date"/></div>
+					<span class="responsive-hide"><xsl:value-of select="@date"/></span>
 				</xsl:if>
 			</cite>
 		</xsl:if>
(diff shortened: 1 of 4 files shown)


AlfredoRamos:
!set HTML
!set CSS
!set Styles

AlfredoRamos:
!set WIP

3D-I:
Does it also take into consideration the quotes in PMs? `@msg_url`

AlfredoRamos:
@3D-I I thought about it but I'm not sure if I should also check for the `@msg_url` (`cite` attribute) since that URL links to a private content, which is not publicly accessible.

AlfredoRamos:
!unset WIP

dark-1:
Well !!! in PM , there is no way for a W3C validator to do any validation.
So this PR is perfect for `@post_url`.

Also nice that you changed `div` to `span` 😄 

Best regards 👍

rxu:
After this change, what will happen with BBCode in old posts (already saved in database)? Will they require reparsing?

AlfredoRamos:
@rxu no, it only requires the user to clear the cache so TextFormatter can read that file to setup those BBCodes again.

rxu:
What I meant is whether quote BBCode in 'old' posts will be broken. So if they won't, that's good.

AlfredoRamos:
@rxu in my tests I didn't had problems with previously posted messages.

Should I keep the `blockquote cite > div` CSS selector just to be sure? Because the worst thing that can happen is that date won't be located at the right of the quoted message.

rxu:
Tested locally: posted a message and then applied the changes from the PR.
Before:
image

After:
image

dark-1:
> After:
> image

I think you forgot the CSS part.
Here : https://github.com/phpbb/phpbb/pull/5932/files#diff-640d4c568fbd9176eefbdfbd5782e31aR487

Best regards 👍

rxu:
I did apply HTML/CSS part while recompiling outdated templates setting is on, but purging cache made it right, my bad :)
image

dark-1:
> I did apply both of HTML/CSS parts while recompiling outdated templates setting is on, but purging cache made it right, my bad :)

Ahh! Nice 😍 .

Best regards 👍

mrgoldy:
> @3D-I I thought about it but I'm not sure if I should also check for the `@msg_url` (`cite` attribute) since that URL links to a private content, which is not publicly accessible.
Could you expand a bit please?
Cause just because it's not accessible for bots, shouldn't mean it can have invalid markup.

AlfredoRamos:
@mrgoldy The missing `cite` attribute (`<blockquote cite="...">`) doesn't make the markup invalid.

I said that I didn't include that attribute on private messages because I didn't see the point for it.

marc1706:
Agreed, the cite attribute will definitely help the search engines there. :)

PR #5938, opened 2020-04-08
[ticket/16430] Sort permissions

mrgoldy:
PHPBB3-16430

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16430

And the _(old)_ RFC:
http://area51.phpbb.com/phpBB/viewtopic.php?f=108&t=43295

### DIFF
--- a/phpBB/includes/acp/auth.php
+++ b/phpBB/includes/acp/auth.php
@@ -1134,6 +1134,8 @@ class auth_admin extends \phpbb\auth\auth
 		/* @var $phpbb_permissions \phpbb\permissions */
 		$phpbb_permissions = $phpbb_container->get('acl.permissions');
 
+		$order = array_flip(array_keys($phpbb_permissions->get_permissions()));
+
 		foreach ($category_array as $cat => $cat_array)
 		{
 			if (!$phpbb_permissions->category_defined($cat))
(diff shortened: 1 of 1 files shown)


mrgoldy:
To quickly expand on how this works.

Permissions are now ordered by the way they are defined in [phpbb\permissions](https://github.com/phpbb/phpbb/blob/master/phpBB/phpbb/permissions.php#L222).
This is a more natural way of ordering permissions than by array keys (`f_read`, `f_sticky`, etc) and aswell as by their translated language string. This way we can sort them by order of importance / logical follow ups.

Extension permissions are added through an event listener listening to the `core.permissions` event. Which would append the extension's permissions to the end of the array (when done the 'common way' with `$event['permissions'] = array_merge($event['permissions', [...])`. So extensions' permissions are added to the end of a permission list by default, which I would argue is a good default so admins know where to look for new/extra permissions. 
However, this does allow extensions to also position their permissions more precisely, if they insert their permissions in a specific position in the permissions array. And extensions between themselves can use the event listener's priority functionality.

To visualize; **BEFORE**
Screenshot_3
Screenshot_2

**AFTER**
Screenshot_5
Screenshot_4

kasimi:
I tested these changes, works great in the "global permissions", "forum based permissions" and "permission masks" sections, but no sorting is happening when editing any of the "permission roles" (ACP Permission tab, 3rd category in the left sidebar). So when editing these roles, the permissions are still sorted the old way. I tested on 3.2.9 and 3.3.0, same issue on both. Is there any reason why this couldn't be targeting 3.2.x?

3D-I:
That's more a fix than a new feature and should be addressed to 3.2.x IMHO.

marc1706:
@mrgoldy maybe reconsider this for 3.3.2?

mrgoldy:
Yeah, haven't had the time yet. Will rebase on 3.3 and implement it in permission roles.

PR #5978, opened 2020-05-12
[ticket/15851] Instant update

gijsmartens:
PHPBB3-15851

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15851

---

Slight alternative of #5419, @senky mentioned he didn't have time to continue it.

This will require that the download link and sha256 checksum are added to [versions.json](https://github.com/phpbb/versioncheck/blob/master/phpbb/versions.json):
```json
"3.3": {
      "current": "3.3.0",
      "announcement": "https://www.phpbb.com/community/viewtopic.php?f=14&t=2534541",
      "download": "https://download.phpbb.com/pub/release/3.3/3.3.0/phpBB-3.3.0.zip",
      "checksum": "5d01d12961ed9b67d87ce2408cdefbbe8e6f00f24cbc75ce10d410f473179b86",
      "eol": null,
      "security": false
}
```

### DIFF
--- a/phpBB/includes/acp/acp_update.php
+++ b/phpBB/includes/acp/acp_update.php
@@ -167,7 +167,7 @@ class acp_update
 		{
 			$this->instant_update_error(
 				$this->language->lang('DIRECTORY_NOT_WRITABLE') . '<br>' .
-				$this->language->lang('MAKE_FOLDER_WRITABLE', $this->root_path . 'store')
+				$this->language->lang('MAKE_FOLDER_WRITABLE', '/store')
 			);
 		}
 
--- a/phpBB/includes/acp/acp_update.php
+++ b/phpBB/includes/acp/acp_update.php
@@ -1,45 +1,85 @@
 <?php
 /**
-*
-* This file is part of the phpBB Forum Software package.
-*
-* @copyright (c) phpBB Limited <https://www.phpbb.com>
-* @license GNU General Public License, version 2 (GPL-2.0)
-*
-* For full copyright and license information, please see
-* the docs/CREDITS.txt file.
-*
-*/
+ *
+ * This file is part of the phpBB Forum Software package.
+ *
+ * @copyright (c) phpBB Limited <https://www.phpbb.com>
+ * @license GNU General Public License, version 2 (GPL-2.0)
+ *
+ * For full copyright and license information, please see
+ * the docs/CREDITS.txt file.
+ *
+ */
 
 /**
-* @ignore
-*/
+ * @ignore
+ */
 if (!defined('IN_PHPBB'))
 {
 	exit;
 }
 
+/**
+ * ACP Module: phpBB update.
+ */
 class acp_update
 {
-	var $u_action;
+	/** @var string Page title */
+	public $page_title;
 
+	/** @var string Template file name */
+	public $tpl_name;
+
+	/** @var string Custom form action */
+	public $u_action;
+
+	/** @var \phpbb\config\config */
+	protected $config;
+
+	/** @var \Symfony\Component\DependencyInjection\ContainerInterface */
+	protected $container;
+
+	/** @var \phpbb\language\language */
+	protected $language;
+
+	/** @var string phpBB root path */
+	protected $root_path;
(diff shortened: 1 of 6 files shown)


rxu:
What this implementation currently does on update is:

- creates temporary directory
- downloads phpBB archive to that dir
- extract the archive to the tempdir/phpBB3
- deletes `config.php`, `styles`, `images`, `store` from extracted package
- copies extracted files to the board root
- redirects to `/install` to perform manual update.

What it doesn't do is:
- doesn't make a backup before overwriting files in the board root
- doesn't delete `.htaccess` while those often are custom (and no backup made)
- doesn't perform actual `instant` update

So, currently it's rather just a download helper yet.

mrgoldy:
It's an instant update, as in that it downloads the package and deletes the necessary files. Which was the outcome of the discussion(s) that were linked in the previous PR and ticket. 

The database migration requires a different container and checks, which can not be replicated. So the user is redirected to the 'Update database' page, where he has to click two additional buttons.
It is possible to automatically create a database update and backup of the files, I've managed to do that locally, but that doesn't change the fact that then they are still in the `/store` folder, where the Admin will still have to download them. I can implement nonetheless, so the Admin only have to download them through the FTP. It will not provide a direct download, as that was intentionally removed.

The `.htaccess` file can be deleted from the downloaded package, that's no problem.

This is not a 'full/complete/automated/one-click' update.
This merely does the part that apparently most administrators find tedious.
Downloading the package, deleting the required files, setting the correct board defaults, etc.

rxu:
> The database migration requires a different container and checks, which can not be replicated.

Yeah. That would be done just like installer currently is which has its own container. In that world updater could be able to perform complete update with minimal user participation, but that's probably requires much more job to be done in code.

koraldon:
Great to see this, but I think the name instant update is confusing, it seems similar to Wordpress automatic update, while this is actually a sort of 1 click update.

phpBB/language/en/acp/common.php
```diff
@@ -458,6 +458,7 @@
 	'VERSIONCHECK_INVALID_ENTRY'	=> 'Latest version information contains an unsupported entry.',
 	'VERSIONCHECK_INVALID_URL'		=> 'Latest version information contains invalid URL.',
 	'VERSIONCHECK_INVALID_VERSION'	=> 'Latest version information contains an invalid version.',
+	'VERISONCHECK_INVALID_SHA256'	=> 'Latest version information contains an invalid sha256 checksum.',
```
3D-I, line 461: That should be `VERSIONCHECK_INVALID_SHA256`

phpBB/includes/acp/acp_update.php
```diff
@@ -1,45 +1,85 @@
...
+	 * @return void
+	 */
 	function main($id, $mode)
 	{
-		global $config, $user, $template, $request;
-		global $phpbb_root_path, $phpEx, $phpbb_container;
+		global $config, $language, $request, $template;
+		global $phpbb_container, $phpbb_root_path, $phpEx;
 
-		$user->add_lang('install');
```
rubencm, line 31: Properties initialization should go in the constructor

phpBB/includes/acp/acp_update.php
```diff
@@ -49,40 +89,235 @@ function main($id, $mode)
...
+
+		// Verify the sha256 checksum of the download
+		if ($update['checksum'] !== hash_file('sha256', $filesystem->realpath($temp_zip)))
+		{
+			$filesystem->remove($temp_zip);
+
+			$this->instant_update_error($this->language->lang('INSTANT_UPDATE_ERROR_CHECKSUM_MISMATCH'));
+		}
+
+		$zip = new \ZipArchive;
```
rubencm, line 200: Add ext-zip to composer.json or better, use the phpbb zip implementation in functions_compress.php

rubencm:
I would do this in multiple steps, because if it takes more than the http server time_limit, the update will remain unfinished.
Also, it would be useful to abstract the functionality to be able to use it from cli too.

PR #5979, opened 2020-05-12
[ticket/15769] Crop avatar on upload

gijsmartens:
Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15769


dark-1:
> Tracker ticket (set the ticket ID to **your ticket ID**):
> 
> https://tracker.phpbb.com/browse/PHPBB3-12345

The ticket link should be : `https://tracker.phpbb.com/browse/PHPBB3-15769`

Best regards 👍

neuropass:
I don't know why there is a need to add more bloating to this. No need to crop anything no need to add more bloating. Please don't merge.

mrgoldy:
Not sure why you would comment that.
This is literally the most voted idea on phpBB.com: https://www.phpbb.com/community/viewtopic.php?f=436&t=2437796
This will save users having to resize images on their computer, to meet the requirements for a specific board.

neuropass:
because I dont understand the need of constantly bloating the core code base.

The extension system was created for this purpose. Extend the code base at "will".
This has been said over and over by people like @DavidIQ @VSEphpbb etc. dont take my words for it. Why would this be different now?

Just because half of the planet is smoking doesn't make it good decision to start smoking...

Can you make into an extension instead? request the core events if they are not there?

hanakin:
@neuropass 
This is absolutely required and not bloat. Its is a UI requirement to have an avatar going forward with any application today. In order to facilitate that you have to require stricter adherence to avatar standards this is why this is required. This is why facebook, github and any major application has this exact feature or something like it. The force use of an avatar adds a level of control over responsive design and UI that we currently are seriously lacking and in order to fix that moving forward. If you do not understand that or agree with that, then that is your opinion. However its a flawed one that does not change the fact that in order to provide the best front-end experience we require this in the core. The avatar is no longer just a decorative, user personalization. Its an alternative representation of the user and their name. Its purpose is now a 1:1 representation in the standard for application design. We will also be forcing the requirement of avatars in 4.0 so that will also be a change going forward. Hopefully you understand we are making these changes for the future of the app.

neuropass:
yes it is required and I'm not arguing that. 

What a "flawed one" is the fact you and even the team cannot get a general agreement at the least over what is being done with all these PR's. Then later on back and forth to undue what was merged because it causes issues. 

A "flawed one" is not sticking to practice what you preach. You go on and on about building tools, standards and the need to have a better expendable core moving forward and yet you fail to stop a PR that will just make your life harder in the future for refactoring. 

A flawed one, is telling people to work on extensions because that is what are for and then allowed this type of PR's just because its convenient at the moment.

A flawed one is thinking you know better when you let this go through and say "you are wrong or you don't understand" what gives you the presumption to say you don't understand? (perhaps is language barrier there?)

Since when phpBB was at the mercy at whatever comes in just because someone submits something?  Where did the development path and/or goals go? 
 
Use the extension system instead. This is what it was meant for since day 1. make this an official extension if this is what what you want to do seeing are it is the most voted feature? 
Or this changed somewhere along the way?

With your own words, what do you not understand about not bloating the core? The implementation in the core of this is what "flawed" means. There are better way to do this.  Starting from creating a standalone extension admin can choose to use if they wish to do so.

phpBB/phpbb/image/image_cropper.php
```diff
@@ -0,0 +1,259 @@
+<?php
```
CHItA, line 1: Cannot this be done all in the front-end? If you can do all these (and probably more) from JS, I think it should be done there and just save the file.

hanakin:
what you do not understand, is that it cannot be an ext if it is required by another aspect of the core.

If the templating system requires it then it can not be an ext as everyone is forced to install that ext for the app to work.

this was not submitted at a whim its a planned feature by the development team of which mrgoldy is a member https://area51.phpbb.com/phpBB/viewtopic.php?f=81&t=61766&p=318491#p318491

There are a-lot of other changes required to go with this to how we are using avatars but in short there will be less options for the admin for them to manage as more of it is required and dictated by the front-end requirements.

Now your comments here are appreciated but distracting from the point for a pr, which is to comment on/review the code. If you wish to discuss then can you please either open up a conversation on area51 or comment in the ides topic linked above.

thanks

phpBB/phpbb/image/image_cropper.php
mrgoldy: Well, it is possible to do it client side, most definitely.
However, that would enforce two things, one of which is fine and the other is debatable.
This would mean we would have to use `.toBlob()` on the image, and blobs can only be send with ajax request. So then we would have to ajaxify the form submission, which is fine.
However, the following:
> Uses the Browser's native [canvas.toBlob](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toBlob) API to do the compression work, which means it is **lossy compression**. For better image quality, you can upload the original image and the cropped data to a server and do the crop work on the server.

CHItA: Wait, you just send it as a hidden field. Why cannot we just set the front-end generated image as a value? Is it the problem that this library doesn't support that?

hanakin:
for this to work properly we need to establish the requirements for the avatars. First we want all avatars to be physically 2x the size we need to use them. We also do not want to have to load images that are very large in the style when they do not need to be. This means we should have different sized avatars saved upon upload. similar to how gravatar and GitHub do. Currently we have a large full sized one 200px which means 400px and then the smaller versions 40px which means 80px planned for chameleon

phpBB/phpbb/image/image_cropper.php
mrgoldy: No, the problem is that that will be at the cost of quality.
There are two possibilities:
1. Do the cropping client side and send a Blob, which is ready to go, but possibly losing quality.
2. Send the original image, with the cropping data, and do the cropping server side, to keep the quality.
I am currently using the 2nd option, so just sending the original image with the cropping data in a hidden field.

CHItA: You don't have to loose quality at all. Canvas can return whatever image compression quality you [want](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toDataURL).

phpBB/assets/css/cropper.css
```diff
@@ -0,0 +1,304 @@
+/*!
+ * Cropper.js v1.5.6
```
marc1706, line 2: Please update this to 1.5.7

phpBB/assets/css/cropper.min.css
```diff
@@ -0,0 +1,9 @@
+/*!
+ * Cropper.js v1.5.6
```
marc1706, line 2: Update to 1.5.7

phpBB/assets/javascript/cropper.js
```diff
@@ -0,0 +1,3616 @@
+/*!
+ * Cropper.js v1.5.6
```
marc1706, line 2: I think you catch my drift here :)

phpBB/phpbb/avatar/driver/upload.php
```diff
@@ -89,11 +90,16 @@ public function prepare_form($request, $template, $user, $row, &$error)
 			return false;
 		}
 
-		$template->assign_vars(array(
-			'S_UPLOAD_AVATAR_URL' => ($this->config['allow_avatar_remote_upload']) ? true : false,
-			'AVATAR_UPLOAD_SIZE' => $this->config['avatar_filesize'],
+		$use_board = defined('PHPBB_USE_BOARD_URL_PATH') && PHPBB_USE_BOARD_URL_PATH;
```
marc1706, line 93: Slightly more readable name:
```suggestion
		$use_board_url = defined('PHPBB_USE_BOARD_URL_PATH') && PHPBB_USE_BOARD_URL_PATH;
```

phpBB/phpbb/image/image_cropper.php
marc1706: I'd also definitely opt for client side cropping

phpBB/assets/javascript/phpbb-avatars.js
```diff
@@ -0,0 +1,174 @@
...
+
+	/**
+	 * Bind a function to the avatar file upload <input> element.
+	 *
+	 * If a file was chosen and it is a valid image file, the cropper is initialised.
+	 * Otherwise the cropper is destroyed.
+	 */
+	bindInput: function() {
+		this.input.on('change', function() {
+			var fileReader = new FileReader;
```
marc1706, line 80: Let's use new ES style syntax, i.e. `let` or `const`

In this case I think you can actually use const

phpBB/assets/javascript/phpbb-avatars.js
```diff
@@ -0,0 +1,174 @@
...
+	 *
+	 * Only buttons with a data-cropper-action attribute are recognized.
+	 * The value for this data attribute should be a function available in the cropper.
+	 * It also takes two optional parameters, imploded by a comma.
+	 * For example: data-cropper-action="move,0,10" which results in $().cropper('move', 0, 10)
+	 */
+	initButtons: function() {
+
+		this.buttons.show().find('[data-cropper-action]').off('click.phpbb.avatars').on('click.phpbb.avatars', function() {
+			var option = $(this).data('cropper-action').split(',');
```
marc1706, line 107:
```suggestion
			const option = $(this).data('cropper-action').split(',');
```

phpBB/assets/javascript/phpbb-avatars.js
```diff
@@ -0,0 +1,174 @@
...
+			var action = option[0];
```
marc1706, line 108:
```suggestion
			const action = option[0];
```

phpBB/assets/javascript/phpbb-avatars.js
```diff
@@ -0,0 +1,174 @@
...
+	 *
+	 * It also JSON encodes the data array and places it into an <input> element,
+	 * which will be requested server side, and crop the image accordingly.
+	 * Image cropping is done server side, to ensure the best image quality
+	 * and image blobs (from .toBlob()) can only be send through AJAX requests.
+	 *
+	 * @param {object} event
+	 */
+	onCrop: function(event) {
+		var data = phpbb.avatars.data.data();
```
marc1706, line 153: Same here, don't use `var` for new code

phpBB/assets/javascript/phpbb-avatars.js
```diff
@@ -11,11 +11,22 @@ phpbb.avatars = {
...
+	$data: $('#avatar-cropper-data'),
+
+	/** @type {jQuery} */
+	$input: $('#avatar_upload_file'),
+
+	/** @type {jQuery} */
+	$driver: $('#avatar_driver'),
+
+	/** @type {string} */
 	driverUpload: 'avatar_driver_upload',
```
hanakin, line 30: Is jshint driving this weird comment thing you are doing inline? Instead of in the jsdoc block? Also jshint needs removed as it's ancient and not well managed which is why I was working to replace it with xojs which is basically eslint in a better wrapper.

marc1706: We have a jsdoc block? This is in line with jsdoc btw, not based on jshint. If you think it makes more sense to sum this up into one block I'm open for suggestions, I did however like that this helps with static analysis in phpstorm.

hanakin: yes there is a jsdoc it works the same as all the others https://github.com/phpbb/chameleon/blob/master/all/js/temp.js

phpBB/styles/prosilver/template/ucp_avatar_options.html
```diff
@@ -8,7 +8,7 @@
 	<!-- IF ERROR --><p class="error">{ERROR}</p><!-- ENDIF -->
 		<dl>
 			<dt><label>{L_CURRENT_IMAGE}{L_COLON}</label><br /><span>{L_AVATAR_EXPLAIN}</span></dt>
-			<dd><!-- IF AVATAR -->{AVATAR}<!-- ELSE --><img src="{T_THEME_PATH}/images/no_avatar.gif" alt="" /><!-- ENDIF --></dd>
+			<dd id="avatar-box"><!-- IF AVATAR -->{AVATAR}<!-- ELSE --><img src="{T_THEME_PATH}/images/no_avatar.gif" alt="" /><!-- ENDIF --></dd>
```
hanakin, line 11: is this id required somewhere? we should use a class for styling and a data object for js...

hanakin (commented):
can you post some screenshots of the rendered elements so I can work on new front-ends for them for chameleon and part backport it here?

phpBB/assets/javascript/phpbb-avatars.js
marc1706: Unfortunately, other elements don't seem to work properly for type hinting in phpstorm. Can you maybe propose something that would allow for proper type hinting?

hanakin: I think it's because you are building a single obj. al your params are essential just objs and your methods are just objs. I do not think you can jsdoc internal to an obj.

marc1706:
Btw, this currently does not properly upload images. Yes, seems like I broke it. :)

CHItA (approved):

marc1706:
> can you post some screenshots of the rendered elements so I can work on new front-ends for them for chameleon and part backport it here?

image

phpBB/styles/prosilver/template/ucp_avatar_options.html
marc1706: It's being used but I guess we can replay the ID.

marc1706:
@hanakin should be ready for review again

package.json
```diff
@@ -40,6 +42,10 @@
 			],
 			"multiline-comment-style": "off",
 			"computed-property-spacing": "off",
+			"space-before-function-paren": [
```
hanakin, line 45: hmm this must have gotten removed somewhere?

phpBB/assets/css/cropper.css
```diff
@@ -0,0 +1,304 @@
...
+  position: relative;
+  -ms-touch-action: none;
+  touch-action: none;
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
+  user-select: none;
+}
+
+.cropper-container img {
```
hanakin, line 24: need to factor the img tag out here somehow...

phpBB/assets/css/cropper.css
```diff
@@ -0,0 +1,304 @@
...
+  display: block;
+  height: 100%;
+  image-orientation: 0deg;
+  max-height: none !important;
```
hanakin, line 28: also, why are all of these important required?

phpBB/assets/css/cropper.css
```diff
@@ -0,0 +1,304 @@
...
+}
+
+.cropper-hide {
+  display: block;
+  height: 0;
+  position: absolute;
+  width: 0;
+}
+
+.cropper-hidden {
```
hanakin, line 287: do not like that we have to use any of their CSS or classes...will need to completely refactor all to use proper CSS classes that we control going forward

hanakin (commented):
I looked over everything but can't do a full review as I am currently on vacation and do not have access to a full test install/review.

have to take a look in a couple of weeks

package.json
marc1706: Seems like it, this settings seems to be what we're aiming though.

phpBB/assets/css/cropper.css
marc1706: Hm ok, will have a look at how to best do that.

3D-I: > Hm ok, will have a look at how to best do that.

Let's forget chameleon for a while. What remains to do here?
I think it's clear that if there are no subtle issues such as the best css strategy or similar things the rest sounds good to me, so why not adopt it and possibly refine it in a future version of phpBB like 4.1 or similar?

hanakin:
@marc1706 apart from the conflicts being fixed. I think all else should be good based on our conversation about splitting things off. I will look into the conversion on my end.

CHItA (approved):

PR #5986, opened 2020-05-16
[ticket/16487] Fix URL for fontawesome 3.2.0 migration

3D-I:
PHPBB3-16487

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16487

### DIFF
--- a/phpBB/phpbb/db/migration/data/v320/font_awesome_update.php
+++ b/phpBB/phpbb/db/migration/data/v320/font_awesome_update.php
@@ -15,22 +15,22 @@ namespace phpbb\db\migration\data\v320;
 
 class font_awesome_update extends \phpbb\db\migration\migration
 {
-	static public function depends_on()
-	{
-		return array(
-			'\phpbb\db\migration\data\v320\dev',
-		);
-	}
-
 	public function effectively_installed()
 	{
 		return isset($this->config['load_font_awesome_url']);
 	}
 
+	static public function depends_on()
+	{
+		return [
+			'\phpbb\db\migration\data\v320\dev',
+		];
+	}
+
 	public function update_data()
 	{
-		return array(
-			array('config.add', array('load_font_awesome_url', 'https://maxcdn.bootstrapcdn.com/font-awesome/4.4.0/css/font-awesome.min.css')),
-		);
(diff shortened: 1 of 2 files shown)


DavidIQ:
Shouldn't this be a new migration?

3D-I:
Hmm, I think you are right @DavidIQ :thinking:

rxu:
Please don't. 3.3 already has a migration for FA 5, so if you add a new one for FA 4 it will be a mess completely.

3D-I:
Link to that FA-5 migration please?

rxu:
https://github.com/phpbb/phpbb/blob/3.3.x/phpBB/phpbb/db/migration/data/v33x/font_awesome_5_update.php

3D-I:
Ok, let me see. Thanks.

3D-I:
> 
> 
> Link to that FA-5 migration please?

That migration depends on 3.3, this one not. Moreover, I created a new migration which depends on 3.2, therefore the migration you linked above can not mess the things up IMHO.

Let me push and we will discuss it, no probs.

3D-I:
My thought is the FA 5 migration will overwrite this one which fixes the ancient one.

rxu:
Fixing FA4 URL for 3.3 simply makes no sense. If it's possible to merge this migration into 3.2.x only but not into 3.3.x/master, then it would be okay.

3D-I:
> And the new FA 5 migration should also depend on this one I guess.

rxu:
> Fixing FA4 URL for 3.3 simply makes no sense.

:)

3D-I:
Yeah, I see what you mean mate. 

It is just a question of consistence, but maybe that's just me. :)

marc1706:
How about adding an effectively installed check that marks the FA 4 migration as effectively installed if the config version is greater or equal to 3.3.0? And of course have the FA 5 migration depend on the new FA 4 migration.

3D-I:
That's indeed a good idea.

3D-I:
> And of course have the FA 5 migration depend on the new FA 4 migration.

This means a new PR for that, am I wrong?

PR #5990, opened 2020-05-25
[ticket/16496] Fix html special chars in custom BBcode help line

3D-I:
PHPBB3-16496

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16496

### DIFF
--- a/phpBB/adm/style/acp_posting_buttons.html
+++ b/phpBB/adm/style/acp_posting_buttons.html
@@ -44,7 +44,7 @@
 	</select>
 	<!-- EVENT acp_posting_buttons_custom_tags_before -->
 	<!-- BEGIN custom_tags -->
-	<input type="button" class="button2" name="addbbcode{custom_tags.BBCODE_ID}" value="{custom_tags.BBCODE_TAG}" onclick="bbstyle({custom_tags.BBCODE_ID})" title="{{ custom_tags.BBCODE_HELPLINE|e('html_attr') }}" />
+	<input type="button" class="button2" name="addbbcode{custom_tags.BBCODE_ID}" value="{custom_tags.BBCODE_TAG}" onclick="bbstyle({custom_tags.BBCODE_ID})" title="{{ custom_tags.BBCODE_HELPLINE }}" />
 	<!-- END custom_tags -->
 </div>
 <!-- EVENT acp_posting_buttons_after -->
--- a/phpBB/adm/style/acp_posting_buttons.html
+++ b/phpBB/adm/style/acp_posting_buttons.html
@@ -44,7 +44,7 @@
 	</select>
 	<!-- EVENT acp_posting_buttons_custom_tags_before -->
 	<!-- BEGIN custom_tags -->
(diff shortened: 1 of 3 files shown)


3D-I:
immagine


immagine

marc1706:
Somehow your example does not work as displayed above:
image

3D-I:
What commit are you using?

marc1706:
Latest in this PR. :)

3D-I:
I guess I was right then, wait pls.

3D-I:
Data's stored as in

immagine


In the database, thus we need to decode it.

kasimi:
1. BBCode help text entered in the ACP: `Same effect as > or < "HTML tags"`
2. The value stored in the database (`htmlspecialchars()` applied by the `type_cast_helper`): `Same effect as &gt; or &lt; &quot;HTML tags&quot;`
3. `htmlspecialchars_decode()` applied, as currently proposed in this PR: `Same effect as > or < "HTML tags"`. This is the value of the `BBCODE_HELPLINE` template variable.
4. Twig filter `e('html_attr')` applied in both `posting_buttons.html` and `acp_posting_buttons.html`: `Same&#x20;effect&#x20;as&#x20;&gt;&#x20;or&#x20;&lt;&#x20;&quot;HTML&#x20;tags&quot;`

While this works, I see two problems:
* There is unescaped data in the template, while usually all data assigned to the template (usernames, post text etc) has their HTML entities encoded. This means that the `BBCODE_HELPLINE` variable should not be used as is, but needs to be encoded when rendering, not just by the core but also by extension that use it. Again, this is not the default behavior.
* Why decode the entities in step 3 only to encode them again in step 4? It's perfectly fine to use the value in step 2 in an HTML attribute value without further escaping.

Unless I'm missing something, I'd suggest to remove the code from steps 3 and 4.

@marc1706 can you please test again?

3D-I:
> Unless I'm missing something, I'd suggest to remove the code from steps **3** and **4**.

Unless I am missing something I recall I have already tried it all, I am on phpBB 3.3.x testing here, if that matters. :thinking:

kasimi:
It's the final output.

image

3D-I:
Never seen here.

mrgoldy:
The usual way this is dealt with is running `htmlspecialchars_decode($var, ENT_COMPAT);` on the requested value when saving it. This ensures the 'correct' value is stored in the database, and not something encoded. Then it's just a matter of displaying and escaping it where it's needed.

kasimi:
Example in the core of a user-controller title attribute:

* Value is assigned to the template without decoding: https://github.com/phpbb/phpbb/blob/release-3.2.9/phpBB/includes/functions_display.php#L624
* Value is rendered without escaping: https://github.com/phpbb/phpbb/blob/release-3.2.9/phpBB/styles/prosilver/template/forumlist_body.html#L96

3D-I:
@kasimi Do you confirm this commit is right then?

https://github.com/phpbb/phpbb/pull/5990/commits/13914b91df40df99d826036a2a05f4f8653122cd

@marc1706 said it hasn't worked.

By my side both are working so whatever you decide is ok.

marc1706:
Looks good now.

PR #5995, opened 2020-06-01
[ticket/16505] Fix use of count() on non-Countable

rxu:
Checklist:

- [x] Tests pass

<a href="https://tracker.phpbb.com/browse/PHPBB3-16505">PHPBB3-16505</a>.

### DIFF
--- a/phpBB/phpbb/language/language.php
+++ b/phpBB/phpbb/language/language.php
@@ -292,10 +292,9 @@ class language
 	 *
 	 * @return string
 	 */
-	public function lang_array($key, $args = [])
+	public function lang_array($key, array $args = [])
 	{
 		$lang = $this->lang_raw($key);
-		$args = (array) $args;
 
 		if ($lang === $key)
 		{
--- a/phpBB/phpbb/language/language.php
+++ b/phpBB/phpbb/language/language.php
@@ -292,9 +292,10 @@ class language
 	 *
 	 * @return string
 	 */
-	public function lang_array($key, $args = array())
+	public function lang_array($key, $args = [])
 	{
 		$lang = $this->lang_raw($key);
+		$args = (array) $args;
 
 		if ($lang === $key)
 		{


phpBB/phpbb/language/language.php
```diff
@@ -292,9 +292,10 @@ public function lang_raw($key)
 	 *
 	 * @return string
 	 */
-	public function lang_array($key, $args = array())
+	public function lang_array($key, $args = [])
 	{
 		$lang = $this->lang_raw($key);
+		$args = (array) $args;
```
3D-I, line 298: What if we remove these lines instead?

https://github.com/phpbb/phpbb/blob/32932eadc4586356288c37ce4d37481fc417ad0d/phpBB/phpbb/language/language.php#L308-L311

rxu: It won't return a correct language entry then I guess.

3D-I: I am just assuming that the first condition rules it all?

rxu: 1st condition is on the different var ;)

3D-I: Then `if (empty($args))` due to 3.2.x, otherwise there are more options.

rxu: It's not needed if we cast `$args` to array for any of cases.

3D-I: Well I agree but seems a bit redundant to me. Anyway.

rxu:
!set WIP

phpBB/phpbb/language/language.php
```diff
@@ -292,10 +292,9 @@ public function lang_raw($key)
 	 *
 	 * @return string
 	 */
-	public function lang_array($key, $args = [])
+	public function lang_array($key, array $args = [])
```
3D-I, line 295: `public function lang_array($key, array $args = [])`
what if it is
`public function lang_array($key, array $args)`
and this PR moved to the 3.3.x branch?

rxu: Elaborate?

3D-I: Not so much to elaborate I think,  If we're not in a hurry, since this seems like an extreme case, I could work it out better though. I just need the time to reproduce this mistake and work on it, which I do not think I can at the moment... sorry. The seamless solution IMHO is using `empty()` ... as mentioned above and at first glance (coding reading) - it seems to me to be the optimal solution at the moment, we're not going to bandage our heads before breaking it hopefully.

rxu: Well, considering the ticket comments, that would be pointless. As long as we cast `$args` to `array` there're no cases we can't `count()` as it is now. So no need to replace it with `empty()` as it won't affect the result.

3D-I: Whatever you want no probs. I just expressed my opinion and I'm happy to learn. :smiley:

rxu:
!unset WIP

phpBB/phpbb/language/language.php
3D-I: IMHO that's an issue with 3.2.x and extension(s) and PHP greater than 7.2.x. 

 I do agree though changing that code... but just for 3.3.x or greater.
In this case I'd go the way I mentioned above.

> `public function lang_array($key, array $args)`
> and this PR rebased to 3.3.x

rxu: Let's just leave this "fix" as is. it's too paltry to complicate things in such a way :)

PR #5998, opened 2020-06-02
[ticket/16508] Fix WhoIs lookup

3D-I:
PHPBB3-16508

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16508


rxu:
The problem with `z` here is that while it is a query flag for `whois.arin.net`, it's not a flag for `whois.apnic.net` ReferralServer which the query is being readdressed to further along the function. So the query just fails.

3D-I:
immagine

CrazyC0d3r:
Whois remains broken for me when implementing these code changes on my 3.2.9 board.

I always get:

> This is the RIPE Database query service.
> % The objects are in RPSL format.
> %
> % The RIPE Database is subject to Terms and Conditions.
> % See http://www.ripe.net/db/support/db-terms-conditions.pdf
> 
> % Note: this output has been filtered.
> % To receive output for a database update, use the "-B" flag.
> 
> %ERROR:101: no entries found
> %
> % No entries found in source RIPE.
> 
> % This query was served by the RIPE Database Query Service version 1.97.1 (ANGUS)

I need to look into this further. :/

It works again when I remove:
```
	if (!filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6))
	{
		// Limit the query to all possible flags (whois.arin.net)
		$ip = 'z ' . $ip;
	}
```

rxu:
Well, if `z` is appended it will be forwarded to apnic in any case. F.e. currently I'm having issues with ipv4
```
% This is the RIPE Database query service.
% The objects are in RPSL format.
%
% The RIPE Database is subject to Terms and Conditions.
% See http://www.ripe.net/db/support/db-terms-conditions.pdf

% Note: this output has been filtered.
%       To receive output for a database update, use the "-B" flag.

%ERROR:101: no entries found
%
% No entries found in source RIPE.

% This query was served by the RIPE Database Query Service version 1.97.1 (HEREFORD)
```

3D-I:
immagine

rxu:
Glad you have no issues but it dosn't mean nobody has 😉

3D-I:
I know I am trying to get a clue mate.

3D-I:
I will do some test on my live test boards. Later on.

rxu:
@3D-I I'd suggest this way instead:
https://github.com/phpbb/phpbb/pull/5998/files#diff-e454d5a6345f3b7b1ff38548e003b4ffR1482
Try to change it to
```
fputs($fsk, "z $ip\r\n");
```

3D-I:
> 
> 
> @3D-I I'd suggest this way instead:
> https://github.com/phpbb/phpbb/blob/master/phpBB/includes/functions_user.php#L1473
> Try to change it to
> 
> ```
> fputs($fsk, "z $ip\r\n");
> ```

Seems ok by my side, what about you?

rxu:
Works for me on both ipv4/ipv6.

3D-I:
Fine thanks :+1:

rxu:
Add a simple test probably?

3D-I:
> 
> 
> Add a simple test probably?

Is there some example to work with?
Not a big prophecy expert here, still.

3D-I:
There is `phpbb_mcp_post_ip_test`

rxu:
Well, kinda quick code (untested):
Filename: `\tests\functions_user\whois_test.php`
```php
<?php
/**
*
* @package testing
* @copyright (c) 2020 phpBB Group
* @license http://opensource.org/licenses/gpl-2.0.php GNU General Public License v2
*
*/

require_once dirname(__FILE__) . '/../../phpBB/includes/functions_user.php';

class phpbb_functions_user_whois_test extends phpbb_database_test_case
{
	 public function ips_data()
	{
		return [
			['2001:4860:4860::8888'], // Google public DNS
			['64.233.161.139'], // google.com
		];
	}

	/**
	* @dataProvider ips_data
	*/
	public function test_ip_whois($ip)
	{
		$ip_whois = user_ipwhois($ip);
		$this->assertNotContains('Query terms are ambiguous', $ip_whois);
		$this->assertNotContains('no entries found', $ip_whois);
		$this->assertNotContains('ERROR', $ip_whois);
	}
}
```

rxu:
tests fail probably because of some another functions involved like `generate_board_url()` uses session/user object. That requires some more setup additions to the test file.

3D-I:
Failed on `phpbb_session_check_ban_test` as far as I can see

tests/functions_user/whois_test.php
```diff
@@ -0,0 +1,32 @@
...
+*
+* @package testing
+* @copyright (c) 2020 phpBB Group
+* @license http://opensource.org/licenses/gpl-2.0.php GNU General Public License v2
+*
+*/
+
+require_once dirname(__FILE__) . '/../../phpBB/includes/functions_user.php';
+
+class phpbb_functions_user_whois_test extends phpbb_functional_test_case
```
rxu, line 12: Oh, probably my fault, it should extend `phpbb_test_case` instead.

3D-I: I getting blind reading so much of code ehhe.

rxu:
Ip address is not to blame, that's the test file issue :)
Updated the test code above.

3D-I:
@rxu thanks. You can understand it is not that nice to work on things you do not know.

rxu:
@3D-I Updated again (`global` line).

rxu:
Well, those seeming to be unrelated profilefields tests fail anyway. Can't get why currently.
You might want to cancel adding test for now then to get the PR ready.

3D-I:
When I get some spare time I will do test here in local. Thanks for your support.

marc1706:
@3D-I rebased onto prep-release-3.2.10 ;)

rxu:
Figured out the test above was failing because of `make_clickable()` bug, should be fixed with #6002.

3D-I:
> 
> 
> Figured out the test above was failing because of `make_clickable()` bug, should be fixed with #6002.

Which is for 3.3.2, you might want to add that test there?

rxu:
Makes sense, added.

PR #5999, opened 2020-06-02
[ticket/16510] [3.2.x] Add config values to schema_data.sql

kasimi:
PHPBB3-16510

I tested a fresh 3.2.10-RC1 install with these changes and all config rows that were previously missing are now present in the `phpbb_config` table.

I also sorted the `INSERT` statements alphabetically, which I now realize makes it harder to understand the changes. Let me know and I'll undo the sorting.

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16510


kasimi:
1. All missing `INSERT` statements have been added to `schema_data.sql`. This makes sure all config entries are added for new installs. 
2. In the `v32x\add_missing_config.php` migration, there's a list of all core migrations that contain at least one call to `config.add` of a missing config entry. Their `update_data()` methods are executed only their `config.add` commands are executed again. Config entries that already exist in the DB are skipped by the core ([see here](https://github.com/phpbb/phpbb/blob/release-3.2.10-RC1/phpBB/phpbb/db/migration/tool/config.php#L54-L57)). This makes sure all config entries are added for board update.
3. When updating from 3.2.x to 3.2.10-RC1, the `v32x\font_awesome_update_cdn` migration threw an error because of the missing config value it wants to update. Therefore, its `depends_on()` was updated to make sure all config entries are added in step 2 before `v32x\font_awesome_update_cdn` runs. A new migration also updates the `migration_depends_on` field in the `phpbb_migrations` table.

Snippet of the dependency graph:

dep

3D-I:
A good job, it seems to me.

phpBB/phpbb/db/migration/data/v32x/add_missing_config.php
```diff
@@ -0,0 +1,64 @@
...
+
+use phpbb\db\migration\migration_interface;
+use phpbb\db\migrator;
+
+class add_missing_config extends \phpbb\db\migration\container_aware_migration
+{
+	static public function depends_on()
+	{
+		return [
+			'\phpbb\db\migration\data\v32x\v329',
```
marc1706, line 24: As this will be merged after 3.2.10-RC1, I think this should depend on v3210rc1

phpBB/phpbb/db/migration/data/v32x/v3210rc2.php
```diff
@@ -0,0 +1,38 @@
...
+* @license GNU General Public License, version 2 (GPL-2.0)
+*
+* For full copyright and license information, please see
+* the docs/CREDITS.txt file.
+*
+*/
+
+namespace phpbb\db\migration\data\v32x;
+
+class v3210rc2 extends \phpbb\db\migration\migration
```
marc1706, line 16: Shouldn't add that file, will be added upon release of 3.2.10-RC2. ;)

phpBB/phpbb/db/migration/data/v32x/add_missing_config.php
kasimi: This would result in a cyclic dependency: `add_missing_config` depends on `v3210rc1` depends on `font_awesome_update_cdn` depends on `add_missing_config`

marc1706: Ah ok, understood.

PR #6009, opened 2020-06-14
[ticket/16527] Add a role_exists method to permissions migrator tool

iMattPro:
PHPBB3-16527

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16527

### DIFF
--- a/phpBB/phpbb/db/migration/tool/permission.php
+++ b/phpBB/phpbb/db/migration/tool/permission.php
@@ -241,6 +241,25 @@ class permission implements \phpbb\db\migration\tool\tool_interface
 	}
 
 	/**
+	 * Check if a permission role exists
+	 *
+	 * @param string $role_name The role name
+	 *
+	 * @return int The id of the role if it exists, 0 otherwise
+	 */
+	public function role_exists($role_name)
+	{
+		$sql = 'SELECT role_id
+			FROM ' . ACL_ROLES_TABLE . "
+			WHERE role_name = '" . $this->db->sql_escape($role_name) . "'";
+		$result = $this->db->sql_query($sql);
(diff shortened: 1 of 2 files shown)


phpBB/phpbb/db/migration/tool/permission.php
```diff
@@ -240,6 +240,25 @@ public function remove($auth_option, $global = true)
...
+	 * @param string $role_name The role name
+	 *
+	 * @return int The id of the role if it exists, 0 otherwise
+	 */
+	public function role_exists($role_name)
+	{
+		$sql = 'SELECT role_id
+			FROM ' . ACL_ROLES_TABLE . "
+			WHERE role_name = '" . $this->db->sql_escape($role_name) . "'";
+		$result = $this->db->sql_query($sql);
```
3D-I, line 255: `$this->db->sql_query_limit($sql, 1);` maybe?

On a side note I am used to let the function return the bool, 

`return true if the role exists, false otherwise.`

iMattPro: If you look at the changes we actually need the role ID in some cases. No need to return a Boolean.

3D-I: Ah ok, I use a similar _(quite the same)_ query for the extensions, that's why I asked.
Wouldn't it be better, codewise, to limit that query though?

iMattPro: No. There will only be 1 result ever. And the roles table is VERY small. By default only 24 items. There would be no measurable performance gain. See (both same query time):
Screen Shot 2020-06-14 at 11 48 26 PM

PR #6017, opened 2020-06-30
[ticket/16543] Update query-string in stylesheet.css to hashes

3D-I:
PHPBB3-16543

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16543

### DIFF
--- a/build/update_stylesheet_querystrings.php
+++ b/build/update_stylesheet_querystrings.php
@@ -18,10 +18,10 @@ if (php_sapi_name() !== 'cli')
 
 if (version_compare(PHP_VERSION, '7.1.3', '<'))
 {
-	die('stylesheet_cache_busters.php requires at least PHP 7.1.3');
+	die('update_stylesheet_querystrings.php requires at least PHP 7.1.3');
 }
 
-// Usage: "$ php build/stylesheet_cache_busters.php"
+// Usage: "$ php build/update_stylesheet_querystrings.php"
 $targets = [dirname(dirname(__FILE__)) . '/phpBB/styles/prosilver/theme/stylesheet.css'];
 
 array_map('patch_glob', $targets);
--- a/build/update_stylesheet_querystrings.php
+++ b/build/update_stylesheet_querystrings.php
@@ -37,9 +37,9 @@ function patch_file(string $filepath): void
 	$old	= $file;
 	$new	= preg_replace_callback(
 		'(^@import\\s+url\\([\'"](?<basename>\\w++\\.css)\\?\\K(?:hash|v)=[^\'"]++)m',
-		function ($m) use ($filepath)
+		function ($match) use ($filepath)
 		{
-			$path = dirname($filepath) . DIRECTORY_SEPARATOR . $m['basename'];
+			$path = dirname($filepath) . DIRECTORY_SEPARATOR . $match['basename'];
 			$hash = sprintf('%08x', crc32(file_get_contents($path)));
 
 			return 'hash=' . $hash;
--- a/build/patch_stylesheet_cache_busters.php
+++ b/build/patch_stylesheet_cache_busters.php
@@ -1,0 +1,54 @@
+<?php declare(strict_types=1);
+/**
+*
+* This file is part of the phpBB Forum Software package.
+*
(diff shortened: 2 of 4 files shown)


3D-I:
Running the script will replace querystrings with hashes


immagine

DavidIQ:
I don't think the `patch_` prefix goes with the file.

3D-I:
> 
> 
> I don't think the `patch_` prefix goes with the file name.

Good catch, thx. :+1:

hanakin:
As we will not be using imports in 4.0 this should only be required in 3.x. It should have no conflicts either as the name of the loaded file will be theme.css just for clarification. So with that said pending test that it’s working looks/sounds good to me.

3D-I:
There are no tests for the `build/` utilities/tools afaik.

rxu:
What if we use `INCLUDECSS` instead?

3D-I:
I'd need to read an example first.

DavidIQ:
The way `T_STYLESHEET_LINK` gets generated is already kind of what `INCLUDECSS` does so it won't really help.

rxu:
> Where should it be the difference since the issue is in the browser of the board members?
It will add `assets_version` query.

> The way T_STYLESHEET_LINK gets generated is already kind of what INCLUDECSS does so it won't really help.
It's only applied to the `stylesheet.css`. What I'm talking about is adding every single css file with includecss (f.e. into `stylesheet.html` and then include this into `overall_header.html`).

3D-I:
Well, I see now what you mean. 
Whatever you guys prefer we can do as long as the mission is accomplished.

DavidIQ:
> It's only applied to the `stylesheet.css`. What I'm talking about is adding every single css file with includecss (f.e. into `stylesheet.html` and then include this into `overall_header.html`).

I figured that's what you were suggesting, but wasn't sure because we wouldn't do such a change in a current release. I think it's too invasive.

rxu:
No problem. The point is just to have an alternative way.

hanakin:
It’s also horrible for performance

3D-I:
I have it already manually tested, see screenshots.

JoshyPHP:
> I don't think the `patch_` prefix goes with the file name.

It was part of it, but you can name it anything you like, I haven't looked up whether phpBB has a convention for those files or where they're supposed to reside. I generally recommend starting with a verb that describes what the file does. `patch_` or `update_` are good candidates.

marc1706:
So if I understand this correctly, the script will always create a query string with `?hash=abcdefg` while you currently updated this to have `?v=3.3.1`? I would actually go with updating the query string for every release, i.e. run this as part of the release process. So for that we could have this even easier and just have the script replace for example `?v=3.3` with the version passed to the script which would be 3.3.1 here.

hanakin:
Or do we even need the version stuff? Just append random hash to end of every file on build that’s based on the release #

JoshyPHP:
If you use a hash, it means the script can be run any number of times with no side effects. It can be run during every build including development builds.

Using a version number means the cache gets nuked even when there's no change and on the other hand it's not invalidated if the file changes but the version doesn't, for whatever reason. It's just more efficient and more reliable to use a hash.

marc1706:
Then let's do it with a hash right away

3D-I:
> you currently updated this to have `?v=3.3.1`? I

The first proposal was to update the version number (see discussion on the ticket) then @JoshyPHP provided a script, I didn't use hashes because obviously the version number would have been changed during the build anyway (see running the script).

build/update_stylesheet_querystrings.php
```diff
@@ -0,0 +1,54 @@
...
+	array_map('patch_file', glob($glob));
+}
+
+function patch_file(string $filepath): void
+{
+	$file	= file_get_contents($filepath);
+	$old	= $file;
+	$new	= preg_replace_callback(
+		'(^@import\\s+url\\([\'"](?<basename>\\w++\\.css)\\?\\K(?:hash|v)=[^\'"]++)m',
+		function ($m) use ($filepath)
```
marc1706, line 40:
```suggestion
		function ($match) use ($filepath)
```

PR #6031, opened 2020-07-28
[ticket/16558] Improve viewonline page

v12mike:
Pages that are not handled by viewonline now shown as "Unrecognised Page"
Admins see the URI of session table page

PHPBB3-16558

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16558

### DIFF
--- a/phpBB/language/en/common.php
+++ b/phpBB/language/en/common.php
@@ -835,6 +835,7 @@ $lang = array_merge($lang, array(
 	'UNREAD_MESSAGES'		=> 'Unread messages',
 	'UNREAD_POST'			=> 'Unread post',
 	'UNREAD_POSTS'			=> 'Unread posts',
+	'UNRECOGNISED_PAGE'		=> 'Unrecognised page',
 	'UNWATCH_FORUM_CONFIRM'		=> 'Are you sure you wish to unsubscribe from this forum?',
 	'UNWATCH_FORUM_DETAILED'	=> 'Are you sure you wish to unsubscribe from the forum “%s”?',
 	'UNWATCH_TOPIC_CONFIRM'		=> 'Are you sure you wish to unsubscribe from this topic?',
--- a/phpBB/viewonline.php
+++ b/phpBB/viewonline.php
@@ -443,7 +443,7 @@ while ($row = $db->sql_fetchrow($result))
 	}
 	if ($auth->acl_get('a_'))
 	{
-		$location .= '<br>' . substr($row['session_page'], 0, 99);
+		$location .= nl2br("\n" . substr($row['session_page'], 0, 99));
 	}
 
 	/**
--- a/phpBB/viewonline.php
+++ b/phpBB/viewonline.php
@@ -285,9 +285,9 @@ while ($row = $db->sql_fetchrow($result))
 						$sql = 'SELECT forum_id
 							FROM ' . TOPICS_TABLE . "
 							WHERE topic_id = $topic_id";
-						$result = $db->sql_query($sql);
+						$result1 = $db->sql_query($sql);
 						$forum_id = (int) $db->sql_fetchfield('forum_id');
-						$db->sql_freeresult($result);
+						$db->sql_freeresult($result1);
 					}
 					else
 					{
@@ -302,9 +302,9 @@ while ($row = $db->sql_fetchrow($result))
 								FROM ' . TOPICS_TABLE . ' t, ' . POSTS_TABLE . ' p
 								WHERE p.post_id = ' . $post_id . '
 								AND t.topic_id = p.topic_id';
-							$result = $db->sql_query($sql);
+							$result1 = $db->sql_query($sql);
 							$forum_id = (int) $db->sql_fetchfield('forum_id');
-							$db->sql_freeresult($result);
+							$db->sql_freeresult($result1);
 						}
 						else
 						{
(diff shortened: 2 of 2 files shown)


phpBB/viewonline.php
```diff
@@ -263,6 +263,55 @@
...
+
+						$sql = 'SELECT forum_id
+							FROM ' . TOPICS_TABLE . "
+							WHERE topic_id = $topic_id";
+						$db->sql_query($sql);
+						$forum_id = (int) $db->sql_fetchfield('forum_id');
+					}
+					else
+					{
+						preg_match($pattern . 'p=(\d+)#', $row['session_page'], $matches, PREG_UNMATCHED_AS_NULL);
```
rxu, line 293: Same as above.

phpBB/language/en/common.php
```diff
@@ -835,6 +835,7 @@
 	'UNREAD_MESSAGES'		=> 'Unread messages',
 	'UNREAD_POST'			=> 'Unread post',
 	'UNREAD_POSTS'			=> 'Unread posts',
+	'UNRECOGNISED_PAGE'		=> 'Unrecognised page',
```
rxu, line 838: I'm pretty sure that will confuse a lot of users and will generate lots of support topics like `What does the Unrecognized page mean?`.

phpBB/viewonline.php
```diff
@@ -263,6 +263,55 @@
 		case 'viewtopic':
 			$forum_id = $row['session_forum_id'];
 
+			if (!$forum_id)
+			{
+				$array_match_index = 5;
```
rxu, line 268: There probably should be a cleaner way than match exactly index of 5.

phpBB/viewonline.php
```diff
@@ -263,6 +263,55 @@
...
+					preg_match($pattern . 't=(\d+)#', $row['session_page'], $matches, PREG_UNMATCHED_AS_NULL);
+					if ($matches[$array_match_index])
+					{
+						$topic_id = $matches[$array_match_index];
+
+						$sql = 'SELECT forum_id
+							FROM ' . TOPICS_TABLE . "
+							WHERE topic_id = $topic_id";
+						$db->sql_query($sql);
+						$forum_id = (int) $db->sql_fetchfield('forum_id');
```
rxu, line 289: This will probably produce a bit of memory leak as the query result is not freed. Same below.

rxu (changes requested):
The overall point can be also reached with an extension, just for example like it's done with the `Pages` here https://github.com/phpbb-extensions/pages/blob/17a0438177c84a6d0be8a5b242bfa5abdc40cb28/event/listener.php#L134 or with this one https://github.com/rxu/detailed_viewonline (this one is by me, sorry).

And fingers crossed it won't break extensions as well :)

3D-I:
Personally, I don't see the need for this.

phpBB/viewonline.php
```diff
@@ -263,6 +263,55 @@
...
 			$forum_id = $row['session_forum_id'];
 
+			if (!$forum_id)
+			{
+				$array_match_index = 5;
+				// find the forum_id from the page URI
+				$matches = array();
+
+				$pattern = '#(viewtopic|posting)\.' . $phpEx . '((\?|&)\w+=\w+)*(\?|&)';
+				preg_match($pattern . 'f=(\d+)#', $row['session_page'], $matches, PREG_UNMATCHED_AS_NULL);
```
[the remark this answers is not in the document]
v12mike: The flag was not needed, so now removed.

phpBB/viewonline.php
v12mike: fixed

phpBB/viewonline.php
```diff
@@ -263,6 +263,55 @@
...
+				if ($matches[$array_match_index])
+				{
+					$forum_id = (int) $matches[$array_match_index];
+				}
+				else
+				{
+					preg_match($pattern . 't=(\d+)#', $row['session_page'], $matches, PREG_UNMATCHED_AS_NULL);
```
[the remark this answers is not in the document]
v12mike: fixed

phpBB/language/en/common.php
v12mike: If that were a problem, would it be more sensible to omit unrecognised page reports to users other than admins?

phpBB/viewonline.php
v12mike: I put the constant there because it applies to several following regexs and it seemed easier to support if the regex gets changed in the future.  alternatives welcome.

phpBB/viewonline.php
v12mike: Sorry about that.  A fix has been pushed.

phpBB/language/en/common.php
rxu: I don't think so as that would lead to even more confusing: like having 10 users online in stats you'll get f.e. 1 user on viewonline page for regular user.

rxu:
> Personally, I don't see the need for this.

Tend to agree.

phpBB/language/en/common.php
v12mike: I don't understand why you prefer to display incorrect information rather than correct information?  
Am I misunderstanding the purpose of the viewonline page?

v12mike:
This patch has nothing to do with extensions.  It is fixing a deficiency in core phpBB that means that a range of core pages are not correctly identified on the view users page.

PR #6070, opened 2020-11-02
[ticket/16629] Fix ACP get_database_size() for MySql 8

3D-I:
PHPBB3-16629

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16629

### DIFF
--- a/phpBB/includes/functions_admin.php
+++ b/phpBB/includes/functions_admin.php
@@ -2832,15 +2832,13 @@ function view_warned_users(&$users, &$user_count, $limit = 0, $offset = 0, $limi
 
 /**
 * Get database size
-* Currently only mysql and mssql are supported
 */
 function get_database_size()
 {
-	global $db, $user, $table_prefix;
+	global $db, $user;
 
 	$database_size = false;
 
-	// This code is heavily influenced by a similar routine in phpMyAdmin 2.2.0
 	switch ($db->get_sql_layer())
 	{
 		case 'mysqli':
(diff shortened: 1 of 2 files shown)


phpBB/includes/functions_admin.php
```diff
@@ -2853,35 +2851,25 @@ function get_database_size()
...
+				$sql = 'SHOW TABLE STATUS
+					FROM ' . $db_name;
+				$result = $db->sql_query($sql, 7200);
 
-					$database_size = 0;
-					while ($row = $db->sql_fetchrow($result))
+				while ($row = $db->sql_fetchrow($result))
+				{
+					if (isset($row['Engine'])
+						&& ($row['Engine'] == ('MyISAM' || 'InnoDB' || 'Aria'))
```
MioVisman, line 2863: Very interesting comparison for php

3D-I: Got confused by `arbitrary expressions in empty`, my bad.

marc1706:
I don't have the described issue in my MySQL8 instance. Can you elaborate a bit more on how this issue can be reproduced and how you're solving this here?

3D-I:
The old 
`if (preg_match('#(3\.23|[45]\.|10\.[0-9]\.[0-9]{1,2}-+Maria)#', $version))` 
simply does not recognise the `$row['mysql_version']` which in my case is `string '8.0.20'` AKA **8.0.20**. 

You can test it here https://www.regexpal.com/?fam=118322
Moreover it is very old and useless at this stage _(who needs to check for versions 3 and 4 nowadays?)_, same applies for the second preg_match.

It is also useless the IF_ELSE statement below `if ($table_prefix != '')` since right or wrong it returns the same sum of vars:
 `$database_size += $row['Data_length'] + $row['Index_length'];`

BEFORE
immagine

AFTER
immagine

marc1706 (changes requested):

marc1706 (approved):

PR #6082, opened 2020-11-25
[ticket/16641] Fix config.php validation in ACP - PHP 8

3D-I:
PHPBB3-16641

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16641


phpBB/phpbb/filesystem/filesystem.php
```diff
@@ -613,11 +613,8 @@ protected function phpbb_is_writable($file)
 			}
 			else
 			{
-				$handle = @fopen($file, 'c');
-
-				if (is_resource($handle))
+				if (is_writable($file))
```
rxu, line 616: The problem here is that `phpbb_is_writable()` is to be used either in case of Windows OS or if `is_writable()` doesn't exist, see https://github.com/phpbb/phpbb/blob/5a55891d060ecd4e9011146b68bc1366ac663d90/phpBB/phpbb/filesystem/filesystem.php#L262

So, in theory there's a case when `is_writable()` being used here while doesn't exist .

3D-I:
Let us therefore directly fix the origin of the error described in the ticket.

phpBB/phpbb/filesystem/filesystem.php
3D-I: I have some doubts about the function itself, anyway.

rxu:
What if use `SplFileInfo` for that in `phpbb_is_writable`?

```php
$handle = new \SplFileInfo($file);
return $handle->isWritable();
```

3D-I:
@rxu Great, yesterday I looked for a long time for a solution like this but I couldn't find it. :+1:

rxu:
I guess `acp_main.php` changes shouldn't be needed anymore?

3D-I:
> 
> 
> I guess `acp_main.php` changes shouldn't be needed anymore?

They are not. They do not affect the resolution of the error but at the same time I have found that a little cleaning does not hurt. Since I was over there, I cleaned the house and made coffee.

rxu:
> BTW, there in ACP it seems to me that the ` && $phpbb_filesystem->is_writable(.......)` is redundant?

Yeah, that is what I was talking about actually.

3D-I:
> 
> 
> > BTW, there in ACP it seems to me that the ` && $phpbb_filesystem->is_writable(.......)` is redundant?
> 
> Yeah, that is what I was talking about actually.

I removed my previous changes there, so that's the original code, just I added that common variable.
Let me remove that bit, tested.
Changes in filesystem should stay, anyway. For the sake of other potential unknow errors.

rxu:
I've edited my previous comment :) 
I'm not sure how correctly `fileperms()` does operate in Windows.

3D-I:
Let's see what @marc1706 will say then.

3D-I:
I left a comment on Discord anyway.

3D-I:
Moreover, I am not sure about the reason of checking if `&& file_exists(config.php)` in ACP :smiley_cat: 
I hardly believe you can even run phpBB or access the ACP if `config.php` is not there, uh?

rxu:
Makes sense. If `config.php` doesn't exist you'll never get to that line of code.

3D-I:
Did a bit of housekeeping.

phpBB/includes/acp/acp_main.php
```diff
@@ -679,17 +679,17 @@ function main($id, $mode)
 			}
 		}
 
-		if (!defined('PHPBB_DISABLE_CONFIG_CHECK') && file_exists($phpbb_root_path . 'config.' . $phpEx) && $phpbb_filesystem->is_writable($phpbb_root_path . 'config.' . $phpEx))
+		if (!defined('PHPBB_DISABLE_CONFIG_CHECK'))
 		{
-			// World-Writable? (000x)
-			$template->assign_var('S_WRITABLE_CONFIG', (bool) (@fileperms($phpbb_root_path . 'config.' . $phpEx) & 0x0002));
+			$template->assign_var('S_WRITABLE_CONFIG', (bool) $phpbb_filesystem->is_writable($phpbb_root_path . 'config.' . $phpEx));
```
marc1706, line 684: The previous code checked whether the config file is world writable while your new code checks whether it's writable which would also apply when it's not actually world writable?

3D-I: Now it deals with Home PC servers, FTP etc., we had this doubt in effect.

marc1706: Now it actually still says config.php is world writable when it isn't. ;)

3D-I: The issue is with the switch `S_WRITABLE_CONFIG` and Home servers on windows, I think the actual should code cover all cases. Home servers on windows can not use the FTP facilities (chmod) but rely on the system ones (read-only).

Are you saying you tested it on linux and it doesn't works as expected?

phpBB/includes/acp/acp_main.php
```diff
@@ -679,17 +679,19 @@ function main($id, $mode)
...
 		}
 
-		if (!defined('PHPBB_DISABLE_CONFIG_CHECK') && file_exists($phpbb_root_path . 'config.' . $phpEx) && $phpbb_filesystem->is_writable($phpbb_root_path . 'config.' . $phpEx))
+		if (!defined('PHPBB_DISABLE_CONFIG_CHECK'))
 		{
-			// World-Writable? (000x)
-			$template->assign_var('S_WRITABLE_CONFIG', (bool) (@fileperms($phpbb_root_path . 'config.' . $phpEx) & 0x0002));
+			$phpbb_config_file = $phpbb_root_path . 'config.' . $phpEx;
+
+			$template->assign_var('S_WRITABLE_CONFIG', (bool) $phpbb_filesystem->is_writable($phpbb_config_file) || (@fileperms($phpbb_config_file) & 0x0002));
```
marc1706, line 686: The file permission mask is defined in binary as:
- first 3 bits: world (w)
- second 3 bits: group (g)
- third 3 bits: user (u)

So essentially (first bit is right): uuu ggg www
Each of those 3 bit groups are defining read, write, execute. Therefore 000 000 010 will define that it's world writable, no matter what the other bytes are defining. This evaluates to 0x0002 and hence the previous check did in fact confirm whether the file is world writable. Your change will already evaluate to true no matter if it's writable to the current user, the current group, or world writable. As a result, this complains about config.php being world writable even though it has a permission mask of 600.

marc1706: Wouldn't it make more sense to check if the file is readable before checking its file permissions? Judging by the ticket that could already solve this issue?

3D-I: What would be your proposal then, example?

marc1706: Let's maybe take a step back. The initial issue is that the `@fileperm()` fails cause the file is not readable? At least that's what I'm deducting from the ticket. Therefore, you opted to instead check if the file is writable which unfortunately also says it's writable when the current user is able to write to it.
However, this check should only check for file permissions for "world". Therefore, I think it makes sense to first check if the file can be read before using `fileperm()`. That way we'll ensure that it's both readable and that `fileperm()` can be checked, i.e.:
```suggestion
			$template->assign_var('S_WRITABLE_CONFIG', (bool) $phpbb_filesystem->is_readable($phpbb_config_file) && (@fileperms($phpbb_config_file) & 0x0002));
```

3D-I: IMHO there is no need to check if the `config.php` file exists (and is readable) unless I am missing something.
If it didn't you could never get to the code in question because the forum would probably not work, for sure.

The origin of the issue as per the ticket it was in the function `phpbb_is_writable` of the `phpbb/filesystem`
immagine

```
				$handle = @fopen($file, 'c');
				if (is_resource($handle))
				{
				return $handle->isWritable();
					fclose($handle);
					return true;
				}
```
That is being fixed / rewritten in this PR.

The cleanup of the ACP / tpl switch is a bonus, after a couple of more tests I can now push a new commit which should satisfy the logic in its whole, tested.

PR #6084, opened 2020-11-29
[ticket/16645] Simplified text_reparser.poll_option

JoshyPHP:
It's a minor change that could be merged into 3.3.x but it changes the constructor's signature, that's why it's against master.

PHPBB3-16645

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16645

### DIFF
--- a/phpBB/config/default/container/services_text_reparser.yml
+++ b/phpBB/config/default/container/services_text_reparser.yml
@@ -73,6 +73,7 @@ services:
         class: phpbb\textreparser\plugins\poll_option
         arguments:
             - '@dbal.conn'
+            - '%tables.poll_options%'
         calls:
             - [set_name, [poll_option]]
         tags:
--- a/phpBB/phpbb/textreparser/plugins/poll_option.php
+++ b/phpBB/phpbb/textreparser/plugins/poll_option.php
@@ -13,51 +13,31 @@
 
 namespace phpbb\textreparser\plugins;
 
-class poll_option extends \phpbb\textreparser\base
+class poll_option extends \phpbb\textreparser\row_based_plugin
 {
 	/**
-	* @var \phpbb\db\driver\driver_interface
+	* {@inheritdoc}
 	*/
-	protected $db;
-
-	/**
-	* Constructor
-	*
-	* @param \phpbb\db\driver\driver_interface $db Database connection
-	*/
-	public function __construct(\phpbb\db\driver\driver_interface $db)
+	public function get_columns()
 	{
-		$this->db = $db;
+		return [
+			'id'   => 'topic_id',
+			'text' => 'poll_option_text',
+		];
 	}
 
 	/**
 	* {@inheritdoc}
 	*/
-	public function get_max_id()
-	{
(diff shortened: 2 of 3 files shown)


marc1706:
I don't think this service is used externally so I think it would be fine to add this to 3.3.x.

JoshyPHP:
I agree, I don't think it matters which branch this is merged into. Not only this isn't something that's typically used by any extension but the service's API remains the same and only the constructor's signature changes. Realistically, the only effect is that you need to regenerate the service container.

I'm changing this PR's branch to 3.3.x to see if there's any conflict. Feel free to merge into either branches.

phpBB/phpbb/textreparser/plugins/poll_option.php
```diff
@@ -13,51 +13,31 @@
...
-	public function get_max_id()
+	public function get_columns()
 	{
-		$sql = 'SELECT MAX(topic_id) AS max_id FROM ' . POLL_OPTIONS_TABLE;
-		$result = $this->db->sql_query($sql);
-		$max_id = (int) $this->db->sql_fetchfield('max_id');
-		$this->db->sql_freeresult($result);
-
-		return $max_id;
+		return array(
```
3D-I, line 23: I think it is preferable to use the short array syntax here.

phpBB/config/default/container/services_text_reparser.yml
```diff
@@ -73,6 +73,7 @@ services:
         class: phpbb\textreparser\plugins\poll_option
         arguments:
             - '@dbal.conn'
+            - '%tables.poll_options%'
```
3D-I, line 76: May I ask why add this table here if apparently is not used in the constructor, where still the POLL_OPTIONS_TABLE constant is instead?

phpBB/phpbb/textreparser/plugins/poll_option.php
JoshyPHP: I agree, and I never use the old syntax when I write code but that part was copy/pasted and I didn't notice it before I pushed it. And because every push triggers like 8 hours of Travis time, I don't want to trigger a new build for a small change. I'll modify it locally and it'll get committed if I push an update.

phpBB/config/default/container/services_text_reparser.yml
JoshyPHP: The situation wrt table names is a bit of a mess. The reason some plugins use constants in their code is because the table names would have to be passed to the constructor and I didn't want to have as many different constructor signatures as there are plugins. I don't know whether it's better or worse than using constants but I figure that since those constants are used in main files, they're guaranteed to remain for at least a while.

Technically, it would be possible to keep the same constructor everywhere and pass the table names via a service method call. I don't like the idea of having essential data set outside of the constructor. It only matters when instances are created manually (in tests for example), obviously Symfony's container won't randomly forget to call the relevant method before using that object.

TL;DR: there are other ways to do the same thing. They may be better, but I'm not sure and in that situation I default to whatever requires less code/complexity.

marc1706 (approved):

PR #6092, opened 2020-12-13
[ticket/16655] Give ReCaptcha-v3 a custom max login attempts message

3D-I:
PHPBB3-16655

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16655

### DIFF
--- a/phpBB/language/en/common.php
+++ b/phpBB/language/en/common.php
@@ -397,6 +397,7 @@ $lang = array_merge($lang, array(
 	'LOGIN_CONFIRMATION'				=> 'Confirmation of login',
 	'LOGIN_CONFIRM_EXPLAIN'				=> 'To prevent brute forcing accounts the board requires you to enter a confirmation code after a maximum amount of failed logins. The code is displayed in the image you should see below. If you are visually impaired or cannot otherwise read this code please contact the %sBoard Administrator%s.', // unused
 	'LOGIN_ERROR_ATTEMPTS'				=> 'You exceeded the maximum allowed number of login attempts. In addition to your username and password you now also have to pass the CAPTCHA test.',
+	'LOGIN_ERROR_ATTEMPTS_RECAPTCHA_V3'	=> 'You have exceeded the maximum number of login attempts allowed.<br>In addition to your username and password the invisible reCAPTCHA v3 will be used to authenticate your session.',
 	'LOGIN_ERROR_EXTERNAL_AUTH_APACHE'	=> 'You have not been authenticated by Apache.',
 	'LOGIN_ERROR_OAUTH_SERVICE_DOES_NOT_EXIST'	=> 'A non-existent OAuth service has been requested.',
 	'LOGIN_ERROR_PASSWORD'				=> 'You have specified an incorrect password. Please check your password and try again. If you continue to have problems please contact the %sBoard Administrator%s.',
--- a/phpBB/phpbb/auth/provider/db.php
+++ b/phpBB/phpbb/auth/provider/db.php
@@ -151,22 +151,25 @@ class db extends base
 			$attempts = 0;
 		}
 
+		// Display a custom message for reCaptcha v3
+		$login_error_attempts = $this->config['captcha_plugin'] === 'core.captcha.plugins.recaptcha_v3' ? 'LOGIN_ERROR_ATTEMPTS_RECAPTCHA_V3' : 'LOGIN_ERROR_ATTEMPTS';
+
 		if (!$row)
 		{
 			if ($this->config['ip_login_limit_max'] && $attempts >= $this->config['ip_login_limit_max'])
 			{
-				return array(
+				return [
 					'status'		=> LOGIN_ERROR_ATTEMPTS,
-					'error_msg'		=> 'LOGIN_ERROR_ATTEMPTS',
-					'user_row'		=> array('user_id' => ANONYMOUS),
-				);
+					'error_msg'		=> $login_error_attempts,
+					'user_row'		=> ['user_id' => ANONYMOUS],
+				];
 			}
 
-			return array(
-				'status'	=> LOGIN_ERROR_USERNAME,
-				'error_msg'	=> 'LOGIN_ERROR_USERNAME',
-				'user_row'	=> array('user_id' => ANONYMOUS),
-			);
+			return [
+				'status'		=> LOGIN_ERROR_USERNAME,
+				'error_msg'		=> 'LOGIN_ERROR_USERNAME',
+				'user_row'		=> ['user_id' => ANONYMOUS],
+			];
 		}
 
 		$show_captcha = ($this->config['max_login_attempts'] && $row['user_login_attempts'] >= $this->config['max_login_attempts']) ||
@@ -181,11 +184,11 @@ class db extends base
 			$vc_response = $captcha->validate($row);
 			if ($vc_response)
 			{
-				return array(
+				return [
 					'status'		=> LOGIN_ERROR_ATTEMPTS,
-					'error_msg'		=> 'LOGIN_ERROR_ATTEMPTS',
+					'error_msg'		=> $login_error_attempts,
 					'user_row'		=> $row,
-				);
+				];
 			}
 			else
 			{


3D-I:
!set 3.3

phpBB/phpbb/auth/provider/db.php
```diff
@@ -151,22 +151,25 @@ public function login($username, $password)
 			$attempts = 0;
 		}
 
+		// Display a custom message for reCaptcha v3
+		$login_error_attempts = $this->config['captcha_plugin'] === 'core.captcha.plugins.recaptcha_v3' ? 'LOGIN_ERROR_ATTEMPTS_RECAPTCHA_V3' : 'LOGIN_ERROR_ATTEMPTS';
```
marc1706, line 155: I'm not sure if it wouldn't maybe make more sense to have this customizable per captcha plugin.

3D-I: It is not possible via code to know if you are using ReCaptcha v2 invisible but only Recaptcha v3. As for all others the usual message is fine.

phpBB/phpbb/auth/provider/db.php
```diff
@@ -151,22 +151,25 @@ public function login($username, $password)
...
+				];
 			}
 
-			return array(
-				'status'	=> LOGIN_ERROR_USERNAME,
-				'error_msg'	=> 'LOGIN_ERROR_USERNAME',
-				'user_row'	=> array('user_id' => ANONYMOUS),
-			);
+			return [
+				'status'		=> LOGIN_ERROR_USERNAME,
```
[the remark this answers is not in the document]
3D-I: I replaced the array syntax and automatically established a more logical evaluation of spaces. I don't find it unreadable or difficult to read in the differences, which are exhaustively pointed out by github itself. In this case simply an increase in tabs, no changes other than those concerning the syntax.

phpBB/phpbb/auth/provider/db.php
3D-I: Proposals?

phpBB/language/en/common.php
```diff
@@ -397,6 +397,7 @@
 	'LOGIN_CONFIRMATION'				=> 'Confirmation of login',
 	'LOGIN_CONFIRM_EXPLAIN'				=> 'To prevent brute forcing accounts the board requires you to enter a confirmation code after a maximum amount of failed logins. The code is displayed in the image you should see below. If you are visually impaired or cannot otherwise read this code please contact the %sBoard Administrator%s.', // unused
 	'LOGIN_ERROR_ATTEMPTS'				=> 'You exceeded the maximum allowed number of login attempts. In addition to your username and password you now also have to pass the CAPTCHA test.',
+	'LOGIN_ERROR_ATTEMPTS_RECAPTCHA_V3'	=> 'You have exceeded the maximum number of login attempts allowed.<br>In addition to your username and password the invisible reCAPTCHA v3 will be used to authenticate your session.',
```
CHItA, line 400: I'm not sure what the proper placement of this comment, but it reads strange for me to say that you should look out for something that you may not even see. Wouldn't it make more sense not to display this message at all with this captcha plugin?

3D-I: There is the reCaptcha logo at the bottom-right side of the screen.

CHItA: Sure. But you might not have to pick out the cars or buses. I'm not even sure when this message would be required at all, as for the other captchas are included in the form. Or at least the ones we ship with core. So I'm not sure if this message really helps/is useful, especially for recaptcha. If you fail the probability test, and pick out some images it would pop up in your face hopefully. If you use another type of captcha, it would be on the form. I'm not sure if this is good for accessability at all, but in my mind that would be the only argument to keep this message at all.

3D-I: We can even suppress it for recaptcha V3 but still there there will be one for recaptcha V2, at this very point I opted for consistency.

CHItA: Yeah, I think this is a good catch, I'm just not sure what the right fix is: changing the message, removing it or maybe something else.

3D-I: Like leaving this as it is could be an option 😉

CHItA: I prefer rephrasing it to not changing it.

3D-I: Okay by my side, no probs. Just give me the new wording.

CHItA: I mean your change to closing the PR. I thought you meant that by leaving it as it is. Sorry for the confusion there.

3D-I: Not sure what I should do?

3D-I: I meant leaving this PR as it is and merge it.

phpBB/phpbb/auth/provider/db.php
marc1706: I meant leave it up to the captcha implementation to define a custom message for login error attempts. I guess the captcha plugin should be able to determine itself if it's recaptcha v2 or v3?

3D-I: Well, not really. At least AFAIK for the reCatchaV2 thee are diffeent implementation that are not easily if not at all recognisable.
Unless I am missing something?

3D-I:
The true intention of this PR is to `give ReCaptcha-v3 a custom max login attempts message` as per the title, that's what is doing ... instead of suppress it as per the ticket in the tracker.

phpBB/phpbb/auth/provider/db.php
marc1706: The captcha plugin should know if it's `'core.captcha.plugins.recaptcha_v3'`, right? We already have the captcha factory in here. How about adding a method to `captcha_abstract` that adds a function `get_login_error_attempts()`? Afterwards you can use that inside the captcha plugins to return the language variable for this message. Therefore other plugins can use the same method as opposed to hardcoding this just for recaptcha v3.

3D-I: Recaptcha V2 has an invisible mode plus the normal standard, here it is where the logic fails due to the fact that the invisible one (V2) is not easily recognisable AFAIK.

DavidIQ: I agree this doesn't go here. The plugin should provide the key so any logic for determining the key should reside within the plugin. Otherwise what are you going to do when there is a v4 or a v5, etc.

3D-I: This are the vailable captchas
```
core.captcha.plugins.gd
core.captcha.plugins.gd_wave
core.captcha.plugins.nogd
core.captcha.plugins.qa
core.captcha.plugins.recaptcha
core.captcha.plugins.recaptcha_v3

```

marc1706: Did you plan on showing a different message for `core.captcha.plugins.recaptcha`? At least in this PR that does not seem to be intended. Why do you have to know this then? Even then, v2 could define a different error message with my proposal. Same goes for v3, gd, gd_wave, nogd, and qa. Or they could just display the default cause the abstract will default to the current message.

3D-I: > Did you plan on showing a different message for core.captcha.plugins.recaptcha? 

Nope... as I said:  The true intention of this PR is to give ReCaptcha-v3 a custom max login attempts message as per the title, that's what is doing ... instead of suppress it as per the ticket in the tracker.

> Why do you have to know this then?
Sorry? I knew that prior to push that PR... If I correctly understod.

> Please... let me see how or gimme some hint.

3D-I:
I am closing the PR here as it was not my intention to give each plugin a dedicated message, as per the title of the PR in question. After a long discussion I still don't understand how to give a dedicated message to the captcha V2 plugin, unable and unwilling to fulfill a request that is totally unrelated to me I leave the task to you, in case you want to take charge. Thank you all for your participation.

phpBB/phpbb/auth/provider/db.php
marc1706: @3D-I see this PR: https://github.com/phpbb/phpbb/pull/6155

PR #6104, opened 2020-12-31
Fix "Switching to TLS" by connection

vladviolenty:
Option verify_peer, verify_peer_name, allow_self_signed ignored by connection without encryption and next switch to TLS

### DIFF
--- a/phpBB/includes/functions_jabber.php
+++ b/phpBB/includes/functions_jabber.php
@@ -263,14 +263,13 @@ class jabber
 		if ($use_ssl)
 		{
 			$remote_socket = 'ssl://' . $server . ':' . $port;
-
-			// Set ssl context options, see http://php.net/manual/en/context.ssl.php
-			$options['ssl'] = array('verify_peer' => $verify_peer, 'verify_peer_name' => $verify_peer_name, 'allow_self_signed' => $allow_self_signed);
 		}
 		else
 		{
 			$remote_socket = $server . ':' . $port;
 		}
+		// Set ssl context options, see http://php.net/manual/en/context.ssl.php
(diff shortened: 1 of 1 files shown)


3D-I:
May I ask you why are you moving the array of SSL options outside of its scope?
On a side note, I do not see any issue opened in our bug-tracker and linked here.

See: https://area51.phpbb.com/docs/dev/3.3.x/development/git.html#commit-messages

vladviolenty:
Modern Xmpp protocol requires tls connection on 5222 port. if you xmpp server use are selfsigned SSL certificate after save settings in phpbb config you get error Switching to TLS.
this fix change scope of SSL config, that ssl option apply for default connection to port 5222.

i think,  worth removing the setting "Use secure connection", because port 5223 is not used

3D-I:
Ok, could you somehow demonstrate how to replicate the `Switching to TLS` error in the tracker issue/bug report that you should in any case open and link to this PR?

CHItA:
This doesn't seem like a full patch. If I understand correctly, you want to force using tls, but the option to select what protocol to use is not being removed.

3D-I:
@rubencm what do you think about this PR?

marc1706:
I looked a bit at this. Jabber will try to switch to TLS, even if SSL is not enabled, and I think the idea is that these options should be passed anyway. I don't think this will prevent non TLS connection from working.

marc1706:
How about actually disabling the forced switching to port 5223 when SSL is enabled? Currently it's not really possible to use port 5222 with SSL.

vladviolenty:
> How about actually disabling the forced switching to port 5223 when SSL is enabled? Currently it's not really possible to use port 5222 with SSL.

My patch solve this problem

marc1706:
No, your patch enables SSL when this is actually disabled in the ACP to circumvent this restriction of automatically switching to 5223.

marc1706:
Closing the pull request as there hasn't been any activity for a while.

PR #6115, opened 2021-01-07
installing phpbb on ibm cloud

Sumithkr:
[phpbb.docx](https://github.com/phpbb/phpbb/files/5781579/phpbb.docx)
Add this document on documentation form  it has all instruction how to install phpbb on IBM cloud


CHItA:
Hi, this is probably something better suited to be documented elsewhere. We don't want to maintain this documentation, and probably any user who wants to use a cloud solution can figure out how to install phpBB there as that would require a certain level of technical knowledge anyway.

Sumithkr:
we update our document 
[phpbb.docx](https://github.com/phpbb/phpbb/files/5857457/phpbb.docx)

PR #6116, opened 2021-01-07
[ticket/16679] Respect "allow password reset" setting on login

CHItA:
PHPBB3-16679

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16679

### DIFF
--- a/phpBB/includes/functions.php
+++ b/phpBB/includes/functions.php
@@ -2484,7 +2484,7 @@ function login_box($redirect = '', $l_explain = '', $l_success = '', $admin = fa
 		'LOGIN_ERROR'		=> $err,
 		'LOGIN_EXPLAIN'		=> $l_explain,
 
-		'U_SEND_PASSWORD' 		=> ($config['email_enable']) ? $controller_helper->route('phpbb_ucp_forgot_password_controller') : '',
+		'U_SEND_PASSWORD' 		=> ($config['email_enable'] && $config['allow_password_reset']) ? $controller_helper->route('phpbb_ucp_forgot_password_controller') : '',
 		'U_RESEND_ACTIVATION'	=> ($config['require_activation'] == USER_ACTIVATION_SELF && $config['email_enable']) ? append_sid("{$phpbb_root_path}ucp.$phpEx", 'mode=resend_act') : '',
 		'U_TERMS_USE'			=> append_sid("{$phpbb_root_path}ucp.$phpEx", 'mode=terms'),
 		'U_PRIVACY'				=> append_sid("{$phpbb_root_path}ucp.$phpEx", 'mode=privacy'),
--- a/phpBB/index.php
+++ b/phpBB/index.php
@@ -211,6 +211,7 @@ if ($show_birthdays)
 	$template->assign_block_vars_array('birthdays', $birthdays);
 }
 
+$controller_helper = $phpbb_container->get('controller.helper');
 // Assign index specific vars
 $template->assign_vars(array(
 	'TOTAL_POSTS'	=> $user->lang('TOTAL_POSTS_COUNT', (int) $config['num_posts']),
@@ -222,7 +223,7 @@ $template->assign_vars(array(
 	'BIRTHDAY_LIST'	=> (empty($birthday_list)) ? '' : implode($user->lang['COMMA_SEPARATOR'], $birthday_list),
 
 	'S_LOGIN_ACTION'			=> append_sid("{$phpbb_root_path}ucp.$phpEx", 'mode=login'),
-	'U_SEND_PASSWORD'           => ($config['email_enable']) ? append_sid("{$phpbb_root_path}ucp.$phpEx", 'mode=sendpassword') : '',
+	'U_SEND_PASSWORD'           => ($config['email_enable'] && $config['allow_password_reset']) ? $controller_helper->route('phpbb_ucp_forgot_password_controller') : '',
 	'S_DISPLAY_BIRTHDAY_LIST'	=> $show_birthdays,
 	'S_INDEX'					=> true,
 


3D-I:
This PR seems wrong to me? Please see: https://tracker.phpbb.com/browse/PHPBB3-16658

DavidIQ:
Good point. Looks like the same edit is needed here as well: https://github.com/phpbb/phpbb/blob/3.3.x/phpBB/index.php#L225

3D-I:
I recall I tested the whole stuff prior to reply at the tracker, as usual. It could be I am missing something though.

CHItA:
I'm not sure I'm following, what did you test?

3D-I:
What has been reported in the issue I above posted, the logic seemed clear to me, also tested.

billynoah:
> This PR seems wrong to me? Please see: https://tracker.phpbb.com/browse/PHPBB3-16658

From https://tracker.phpbb.com/browse/PHPBB3-16658
> The logic is correct, if you click on the link that is not hidden, an exception is triggered that clearly explains that the option has been disabled. It is of course based on the fact that emails are enabled otherwise they could not be sent to the administrator. To correctly hide that link you have to disable board-wide emails.

Sorry if I'm misunderstanding you, but are you saying that when an admin sets `Allow password reset` to `FALSE` that the forgotten password link should still be displayed?  That seems like it result in a poor user experience.

In my opinion, if the functionality has been disabled, there's no good reason to show the link. Can you explain your reasoning on this?

3D-I:
https://tracker.phpbb.com/browse/PHPBB3-16658?focusedCommentId=63733&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-63733

3D-I:
This is the exception thrown by the controller, which clearly explains the reason why is there, to have the chance to contact and Admin ;)

immagine

billynoah:
@3D-I - Ahh.. thanks for elaborating.  Although I don't like this behavior, I understand that way of looking at it. The exception could still be useful if a user had the forgotten password link bookmarked and landed there anyway, but I fail to see why we'd be advertising it as a place they should be going. If the admin wants to be contacted by users who've forgotten their passwords I suggest they either:

1.  Modify their theme with a contact link in such places as it would be useful
2. Leave the forgotten password functionality turned on

3D-I:
Sorry @billynoah but I do not agree.
I stay with my opinion and the code speaks books to me, if the user can't recover its PW then we just have to give him/her a chance to contact the administrator in order to solve the issue.

DavidIQ:
If the administrator disabled the feature then he probably doesn't want to be contacted. ;)
Also for something like LDAP or third party auth integration a password reset link is in most cases not useful.

3D-I:
@DavidIQ Then another solution / code should be adopted instead, not this.

CHItA:
As I see it, if the feature is disabled, there should be no link. It is kind of a no brainer, that you would try to contact the board admin if you have some technical problems. That error message will be displayed to anyone who wonders to the disabled password reset page.

billynoah:
One possible solution might be to introduce a third option.

Screen Shot 2021-01-07 at 1 40 40 PM

3D-I:
> 
> 
> As I see it, if the feature is disabled, there should be no link. It is kind of a no brainer, that you would try to contact the board admin if you have some technical problems. That error message will be displayed to anyone who wonders to the disabled password reset page.

Not really, the contact should be there if an option is checked by the admin. AKA new code.

3D-I:
> 
> 
> As I see it, if the feature is disabled, there should be no link.

Why not simply change the `protected function init_controller()` and related `UCP_PASSWORD_RESET_DISABLED` language string?

```php

	/**
	 * Init controller
	 */
	protected function init_controller()
	{
		$this->language->add_lang('ucp');

		if (!$this->config['allow_password_reset'])
		{
			throw new http_exception(Response::HTTP_OK, 'UCP_PASSWORD_RESET_DISABLED', [
				'<a href="mailto:' . htmlspecialchars($this->config['board_contact']) . '">',
				'</a>'
			]);
		}
	}

```

and

```php
'UCP_PASSWORD_RESET_DISABLED'	=> 'The password reset functionality has been disabled. If you need help accessing your account, please contact the %sBoard Administrator%s',
```

billynoah:
> > As I see it, if the feature is disabled, there should be no link.
> 
> Why not simply change then the `protected function init_controller()` and related `UCP_PASSWORD_RESET_DISABLED` language string?
> 

I think he meant "no link on the login page", not "no link on the defunct password reset page".

3D-I:
> 
> 
> > > As I see it, if the feature is disabled, there should be no link.
> > 
> > 
> > Why not simply change then the `protected function init_controller()` and related `UCP_PASSWORD_RESET_DISABLED` language string?
> 
> I think he meant "no link on the login page", not "no link on the defunct password reset page".

The newly proposed code in `index.php` is wrong since itself leads _(again)_ to the same function I spotted above

```php
	case 'sendpassword':
		/** @var \phpbb\controller\helper $controller_helper */
		$controller_helper = $phpbb_container->get('controller.helper');

		redirect($controller_helper->route('phpbb_ucp_forgot_password_controller'));
	break;

```

3D-I:
> 
> 
> Good point. Looks like the same edit is needed here as well: https://github.com/phpbb/phpbb/blob/3.3.x/phpBB/index.php#L225

https://github.com/phpbb/phpbb/pull/6116#issuecomment-756340090

3D-I:
> 
> 
> If the administrator disabled the feature then he probably doesn't want to be contacted. ;)
> Also for something like LDAP or third party auth integration a password reset link is in most cases not useful.

Don't forget that, as far as resetting passwords is concerned, some may want only server administrators to reset passwords for users, so a contact or at least some verbal message (no email link) is necessary.

marc1706 (approved):

3D-I:
@marc1706 can I ask you why my notes were wrong?

marc1706:
I think there are two topics in this discussion. One about not showing the password reset link when this feature is not enabled, and one about displaying a message that password reset is disabled on the actual page. I opted for merging this to cover topic 1, we can of course create another PR for topic 2.

3D-I:
All this could have been avoided, there is clearly some dead code left to begin with, at least in my humble opinion.

PR #6117, opened 2021-01-10
[ticket/16675] Adjust checking commit messages

rxu:
Check commit messages on both 'push' and 'pull_request' events.

Checklist:

- [x] Tests pass

<a href="https://tracker.phpbb.com/browse/PHPBB3-16675">PHPBB3-16675</a>.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -77,9 +77,10 @@ jobs:
                   .github/check-executable-files.sh ./
 
             - name: Check commit message
+              env:
+                  BRANCH: ${{ github.event.pull_request.base.sha }}
+              if: github.event_name == 'pull_request'
               run: |
-                  # Get previous merge commit to get commit range from
-                  export BRANCH=$(git rev-list --merges -n 1 HEAD^1)
                   git-tools/commit-msg-hook-range.sh $BRANCH..$GITHUB_SHA
 
     # Tests for MySQL and MariaDB
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -77,12 +77,10 @@ jobs:
                   .github/check-executable-files.sh ./
 
             - name: Check commit message
-              env:
-                  BRANCH: ${{ github.event.pull_request.base.sha }}
-                  PR_SHA: ${{ github.event.pull_request.head.sha }}
-              if: github.event.pull_request.head.sha != '' && github.event.pull_request.base.sha != ''
               run: |
-                  git-tools/commit-msg-hook-range.sh $BRANCH..$PR_SHA
+                  # Get previous merge commit to get commit range from
+                  export BRANCH=$(git rev-list --merges -n 1 HEAD^1)
+                  git-tools/commit-msg-hook-range.sh $BRANCH..$GITHUB_SHA
 
     # Tests for MySQL and MariaDB
     mysql-tests:


CHItA:
Is there an example where the current version didn't work?

rxu:
Sure, any merged pull requests in 3.3.x/master commits range on `push` event. F.e. https://github.com/phpbb/phpbb/runs/1664625037 where commit message check didn't run, and so on.

CHItA:
I don't think we want to run the checks on push as those should only contain merge commits. One scenario where the changes in run might be useful though is on old PRs with updates without a rebase I think. As it is possible that the base commit is not in the 100 commits we check out.

rxu:
I'm just following the current workflow where all other checks run on both `pull_request` and `push` events. Then probably duplicating any checks on `push` is redundant too as checks on `pull request` run on top of the base branch anyway.

CHItA:
Pushes can have can have merge conflict resolutions etc. Thus our unit and functional tests are a bit different in that regard. So IMO we could keep the `if` block, but the other changes look good to me. I added Marc, so he can figure this out.

.github/workflows/tests.yml
```diff
@@ -77,12 +77,10 @@ jobs:
...
             - name: Check commit message
-              env:
-                  BRANCH: ${{ github.event.pull_request.base.sha }}
-                  PR_SHA: ${{ github.event.pull_request.head.sha }}
-              if: github.event.pull_request.head.sha != '' && github.event.pull_request.base.sha != ''
               run: |
-                  git-tools/commit-msg-hook-range.sh $BRANCH..$PR_SHA
+                  # Get previous merge commit to get commit range from
+                  export BRANCH=$(git rev-list --merges -n 1 HEAD^1)
+                  git-tools/commit-msg-hook-range.sh $BRANCH..$GITHUB_SHA
```
CHItA, line 83: Where does `$GITHUB_SHA` comes from?

.github/workflows/tests.yml
```diff
@@ -77,12 +77,10 @@ jobs:
                   .github/check-executable-files.sh ./
 
             - name: Check commit message
-              env:
-                  BRANCH: ${{ github.event.pull_request.base.sha }}
-                  PR_SHA: ${{ github.event.pull_request.head.sha }}
-              if: github.event.pull_request.head.sha != '' && github.event.pull_request.base.sha != ''
```
CHItA, line 83: I would go with `if: github.event.pull_request.head.sha != ''`

.github/workflows/tests.yml
rxu: It's default GA env var https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables

.github/workflows/tests.yml
rxu: Or probably just `if: env.GITHUB_EVENT_NAME == 'pull_request'`

CHItA: That might be more explicit.

rxu:
Changed the code, or do we want to keep `export BRANCH=$(git rev-list --merges -n 1 HEAD^1)` definition? I'm confused a bit.

CHItA:
> Changed the code, or do we want to keep `export BRANCH=$(git rev-list --merges -n 1 HEAD^1)` definition? I'm confused a bit.

That change looks good to me.

marc1706 (approved):

PR #6130, opened 2021-01-20
Avoid notice if array key is not set

m-ober:
I'm integrating phpBB with an existing website in such a way that the website updates user data in phpBB (nickname, e-mail, password...)

After changing passwords, I'm calling `reset_login_keys()` with the phpBB user_id. This works except for two places where `$this->data['user_id']` is expected to be set.

### DIFF
--- a/phpBB/phpbb/session.php
+++ b/phpBB/phpbb/session.php
@@ -1555,7 +1555,7 @@ class session
 		// Let's also clear any current sessions for the specified user_id
 		// If it's the current user then we'll leave this session intact
 		$sql_where = 'session_user_id = ' . (int) $user_id;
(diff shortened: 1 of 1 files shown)


DavidIQ:
Isn't the fix for this to just set `$your_session_class_instance->data['user_id'] = 0;` in your code?

m-ober:
> Isn't the fix for this to just set `$your_session_class_instance->data['user_id'] = 0;` in your code?

It's indeed another possible fix - as long as this property is public. Is it guaranteed to stay public?

marc1706:
There is currently no plan to deprecate or remove this property from public access.

3D-I:
OK though `isset($this->data['user_id']` doesnt'arm at all ..? 🤔

PR #6138, opened 2021-02-06
[ticket/16698] Enhance check for default charset

3D-I:
PHPBB3-16698

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16698

### DIFF
--- a/phpBB/includes/acp/acp_main.php
+++ b/phpBB/includes/acp/acp_main.php
@@ -690,7 +690,7 @@ class acp_main
 		$encoding_translation	= $this->php_ini->getString('mbstring.encoding_translation');
 		$http_input				= $this->php_ini->getString('mbstring.http_input');
 		$http_output			= $this->php_ini->getString('mbstring.http_output');
-		$default_charset		= $this->php_ini->getString('default_charset');
+		$default_charset		= strtolower($this->php_ini->getString('default_charset'));
 
 		if (extension_loaded('mbstring'))
 		{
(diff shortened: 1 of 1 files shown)


EA117:
Not something that "must" be considered in this change.  But is this a good opportunity to also account for the fact that having default_charset "not set at all" (meaning it doesn't exist in the PHP.INI file at all) also means "PHP APIs will default to UTF-8"?

That wasn't always true, but is true in the versions of PHP that phpBB 3.3.0 and later support: https://www.php.net/manual/en/ini.core.php#ini.default-charset

The reason to mention it now is because it causes the same situation for phpBB that ticket 16698 is in the middle of fixing: Nothing is actually "wrong" because the PHP APIs are defaulting to UTF-8 when the PHP.INI setting doesn't exist; but phpBB's test of "what is the PHP.INI setting set to" is still failing, because we don't find that it's set to the value "UTF-8".

It also seems worth at least consider, since this default_charset test is new to phpBB 3.3.3, and this can be the right time to make sure these edge cases aren't triggering false positives of the new message.

3D-I:
Setting default_charset to an empty value is not recommended.

EA117:
Agreed.  But that's not the same thing as "does not exist in the PHP.INI file at all."

rxu:
Tend to agree with @EA117 as we have support requests about the warning from users who has no `default_charset` parameter in `php.ini` but their boards operate well.

phpBB/includes/acp/acp_main.php
```diff
@@ -704,7 +704,7 @@ function main($id, $mode)
 				'S_MBSTRING_ENCODING_TRANSLATION_FAIL'	=> $encoding_translation && ($encoding_translation != 0),
 				'S_MBSTRING_HTTP_INPUT_FAIL'			=> !empty($http_input),
 				'S_MBSTRING_HTTP_OUTPUT_FAIL'			=> !empty($http_output),
-				'S_DEFAULT_CHARSET_FAIL'				=> $default_charset !== 'UTF-8',
+				'S_DEFAULT_CHARSET_FAIL'				=> $default_charset !== 'utf-8',
```
rxu, line 707:
```suggestion
				'S_DEFAULT_CHARSET_FAIL'				=> $default_charset !== null && $default_charset !== 'utf-8',
```

3D-I: Did it differently since `$this->php_ini->getString()` can return NULL but `strtolower()` returns a NULL value as an empty string. As per https://www.php.net/manual/en/function.strtolower.php#116230

CHItA (approved):

PR #6141, opened 2021-02-09
[ticket/16705] Fix check_disk_space function - PHP 8

3D-I:
PHPBB3-16705

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16705

### DIFF
--- a/phpBB/phpbb/attachment/upload.php
+++ b/phpBB/phpbb/attachment/upload.php
@@ -317,16 +317,6 @@ class upload
 				return false;
 			}
 		}
-		else
-		{
-			$this->file_data['error'][] = $this->language->lang('ATTACH_DISK_FREE_SPACE');
-
-			$this->file_data['post_attach'] = false;
-
-			$this->file->remove();
-
-			return false;
-		}
 
 		return true;
 	}
--- a/phpBB/phpbb/attachment/upload.php
+++ b/phpBB/phpbb/attachment/upload.php
@@ -298,7 +298,7 @@ class upload
 	{
 		if (function_exists('disk_free_space'))
 		{
-			$free_space = disk_free_space($this->phpbb_root_path);
+			$free_space = @disk_free_space($this->phpbb_root_path);
 
 			if ($free_space <= $this->file->get('filesize'))
 			{
--- a/phpBB/phpbb/attachment/upload.php
+++ b/phpBB/phpbb/attachment/upload.php
@@ -296,8 +296,10 @@ class upload
 	 */
 	protected function check_disk_space()
 	{
-		if ($free_space = @disk_free_space($this->phpbb_root_path . $this->config['upload_path']))
+		if (function_exists('disk_free_space'))
 		{
+			$free_space = disk_free_space($this->phpbb_root_path);
+
 			if ($free_space <= $this->file->get('filesize'))
 			{
 				if ($this->auth->acl_get('a_'))
@@ -314,6 +316,16 @@ class upload
 
 				return false;
 			}
+		}
+		else
+		{
+			$this->file_data['error'][] = $this->language->lang('ATTACH_DISK_FREE_SPACE');
+
+			$this->file_data['post_attach'] = false;
+
+			$this->file->remove();
+
+			return false;
 		}
 
 		return true;
--- a/phpBB/language/en/posting.php
+++ b/phpBB/language/en/posting.php
@@ -46,7 +46,6 @@ $lang = array_merge($lang, array(
 	'ATTACH_COMMENT_NO_EMOJIS'	=> 'The attachment comment contains forbidden characters (Emoji).',
 	'ATTACH_DISK_FULL'			=> 'There is not enough free disk space to post this attachment.',
 	'ATTACH_QUOTA_REACHED'		=> 'Sorry, the board attachment quota has been reached.',
-	'ATTACH_DISK_FREE_SPACE'	=> 'Sorry, the PHP <samp>disk_free_space</samp> function is disabled in this system.',
 	'ATTACH_SIG'				=> 'Attach a signature (signatures can be altered via the UCP)',
 
 	'BBCODE_A_HELP'				=> 'Inline uploaded attachment: [attachment=]filename.ext[/attachment]',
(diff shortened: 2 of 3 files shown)


3D-I:
immagine

rubencm:
This has to be fixed in master too

3D-I:
> 
> 
> This has to be fixed in master too

Will the PR against 3.3.x be merged also in master?

3D-I:
Not a test expert here. I will leave that to you in case.

phpBB/phpbb/attachment/upload.php
```diff
@@ -296,8 +296,10 @@ protected function check_attach_quota()
 	 */
 	protected function check_disk_space()
 	{
-		if ($free_space = @disk_free_space($this->phpbb_root_path . $this->config['upload_path']))
+		if (function_exists('disk_free_space'))
 		{
+			$free_space = disk_free_space($this->phpbb_root_path);
```
marc1706, line 301:
```suggestion
			$free_space = disk_free_space($this->phpbb_root_path . $this->config['upload_path']);
```

3D-I: This is where the function fails there. 
Not sure why we are trying to check such directory?
The space is the space inside the root of the board as `./`

3D-I: @marc1706 this function is `only` being used here https://github.com/phpbb/phpbb/blob/91a5a73dca110aad9f44a90ebfc089517ae351b6/phpBB/phpbb/attachment/upload.php#L195 - I am trying to explain that the check here fails with your proposed change and it is not necessary at all IMHO. The attachment quota is checked by another function as a matter of fact.

marc1706:
```suggestion
			$free_space = disk_free_space($this->phpbb_root_path . '/');
```

phpBB/phpbb/attachment/upload.php
```diff
@@ -315,6 +317,16 @@ protected function check_disk_space()
 				return false;
 			}
 		}
+		else
+		{
+			$this->file_data['error'][] = $this->language->lang('ATTACH_DISK_FREE_SPACE');
```
marc1706, line 322: If it's not disabled we only give out system specific info for admins. Therefore I don't see the point of telling anyone that this function is disabled.

marc1706: Previous behavior was to silently fail if we can't check the free space. This can be accepted behavior and changing this to prevent all uploads is IMHO not correct.

3D-I: And let the storage folder increase in size since the TMP files are not deleted?

phpBB/phpbb/attachment/upload.php
3D-I: A double slash you mean?

phpBB/phpbb/attachment/upload.php
3D-I: > If it's not disabled we only give out system specific info for admins. Therefore I don't see the point of telling anyone that this function is disabled.

At least provides a meaningful error message, what would it be your purpose?

phpBB/phpbb/attachment/upload.php
marc1706: Resolved the test failure with changes to the test

phpBB/phpbb/attachment/upload.php
marc1706: This error message can only be meaningful to admins and would just confuse users. Also, the previous behavior was different and I don't see a need to change it. You're aborting upload because you don't know if there is free space as opposed to aborting because there is not enough space. This change will essentially break uploads with boards that do not support calling disk_free_space().

3D-I: Options:

1. add an extra language string and present the error message different for admins/users via $auth->acl_get()
2. add a new configuration in ACP to check if the disk space should be checked or not (NO as defaut)
3. remove the else statement (which I do not like)

marc1706: I think the big question is how one would end up at this else branch. In this case it's only if the function does not exist, hence a state when you are not able to check the disk space. I think the decision to stop any and all uploads because I'm not able to check this is wrong. Yes, you might show an admin that this is disabled. Then again, if this is disabled it might be an issue but most of the time it would be perfectly fine. Breaking uploads because we're not able to check something just seems wrong to me.

phpBB/phpbb/attachment/upload.php
marc1706, line 301: I would still keep the `@` in case we might run into some weird things like open_basedir restrictions
```suggestion
			$free_space = @disk_free_space($this->phpbb_root_path);
```

phpBB/phpbb/attachment/upload.php
3D-I: This has in fact been the behaviour of the previous code so far as it has been silenced. Nowadays I don't think there are any space limitations and so I could agree with you, but since there are still some sites that offer free web space limited to e.g. 5GB and which many people use I think there could be problems in the future. Let's decide together which is the best solution and I'll be happy to push any changes.

phpBB/phpbb/attachment/upload.php
3D-I: No need to silent it again, we are first checking if the function exists.

3D-I: > I would still keep the `@` in case we might run into some weird things like open_basedir restrictions

Sorry... I missed your statement, should I check if open_basedir is in effect as well?

marc1706: Silencing it with `@` is not about whether this function exists, but whether to suppress (at least some) warning messages that might be shown by this. If open basedir restrictions are in effect, supressing this with `@` is valid and works also on PHP 8.x.

phpBB/phpbb/attachment/upload.php
marc1706: I think in general it would probably be nice to have a system where admins gets informed about something like that, e.g. via notifications. But in general I think it does not make sense to abort the actual uploading of attachments and therefore the posting for this. Something like a notification "hey, your system is running low on free space" but that will exceed the scope of this issue / PR.

marc1706 (approved):

PR #6153, opened 2021-03-04
[ticket/16718] Make sure phpBB's autoloader is used before extensions

marc1706:
This resolves compatibility issues with extensions that use composer v2.

PHPBB3-16718

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16718

### DIFF
--- a/phpBB/phpbb/di/container_builder.php
+++ b/phpBB/phpbb/di/container_builder.php
@@ -478,6 +478,12 @@ class container_builder
 				}
 			}
 
+			// Make sure phpBB's autoloader is used before the extension autoloaders
+			if (count($extensions))
+			{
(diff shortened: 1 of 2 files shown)


marc1706:
Composer autoloaders typically load with `$prepend = true;` which results in every autoloader that gets registered to load before the previous one. The extension autoloaders will handle the class `Composer\InstalledVersions` which will not know about a lot of libraries that are needed in phpBB and cause phpBB to fail. By including phpBB's vendor autoloader before this, phpBB's version of `Composer\InstalledVersions` will be used again.

Downside of this: If an extension relies `Composer\InstalledVersions` and that class does not exist in phpBB, it might fail if it does not handle the exception if no version is returned.

marc1706:
Closing as this does not seem to work as hoped.

PR #6169, opened 2021-03-23
[ticket/16735] Fix access array offset on value of type bool on login

3D-I:
PHPBB3-16735

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16735

### DIFF
--- a/phpBB/phpbb/auth/provider/db.php
+++ b/phpBB/phpbb/auth/provider/db.php
@@ -154,8 +154,10 @@ class db extends base
 
 		$login_error_attempts = 'LOGIN_ERROR_ATTEMPTS';
 
-		$show_captcha = ($row) ? ($this->config['max_login_attempts'] && $row['user_login_attempts'] >= $this->config['max_login_attempts']) ||
-			($this->config['ip_login_limit_max'] && $attempts >= $this->config['ip_login_limit_max']) : false;
+		$user_login_attempts	= $row ? ($this->config['max_login_attempts'] && $row['user_login_attempts'] >= $this->config['max_login_attempts']) : false;
+		$ip_login_attempts		= (bool) ($this->config['ip_login_limit_max'] && $attempts >= $this->config['ip_login_limit_max']);
+
+		$show_captcha = $user_login_attempts || $ip_login_attempts;
 
 		if ($show_captcha)
 		{
--- a/phpBB/phpbb/auth/provider/db.php
+++ b/phpBB/phpbb/auth/provider/db.php
@@ -154,7 +154,7 @@ class db extends base
 
 		$login_error_attempts = 'LOGIN_ERROR_ATTEMPTS';
 
-		$user_login_attempts	= $row ? ($this->config['max_login_attempts'] && $row['user_login_attempts'] >= $this->config['max_login_attempts']) : false;
+		$user_login_attempts	= (is_array($row) && $this->config['max_login_attempts'] && $row['user_login_attempts'] >= $this->config['max_login_attempts']);
 		$ip_login_attempts		= ($this->config['ip_login_limit_max'] && $attempts >= $this->config['ip_login_limit_max']);
 
 		$show_captcha = $user_login_attempts || $ip_login_attempts;
--- a/phpBB/phpbb/auth/provider/db.php
+++ b/phpBB/phpbb/auth/provider/db.php
@@ -153,8 +153,10 @@ class db extends base
 		}
 
 		$login_error_attempts = 'LOGIN_ERROR_ATTEMPTS';
-		$show_captcha = ($this->config['max_login_attempts'] && $row['user_login_attempts'] >= $this->config['max_login_attempts']) ||
-			($this->config['ip_login_limit_max'] && $attempts >= $this->config['ip_login_limit_max']);
+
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/auth/provider/db.php
```diff
@@ -153,8 +153,12 @@ public function login($username, $password)
 		}
 
 		$login_error_attempts = 'LOGIN_ERROR_ATTEMPTS';
-		$show_captcha = ($this->config['max_login_attempts'] && $row['user_login_attempts'] >= $this->config['max_login_attempts']) ||
-			($this->config['ip_login_limit_max'] && $attempts >= $this->config['ip_login_limit_max']);
+
+		$user_login_attempts	= $row ? ($this->config['max_login_attempts'] && $row['user_login_attempts'] >= $this->config['max_login_attempts']) : false;
```
CHItA, line 157: `$user_login_attempts = (is_array($row) && $this->config['max_login_attempts'] && $row['user_login_attempts'] >= $this->config['max_login_attempts']);`

phpBB/phpbb/auth/provider/db.php
```diff
@@ -153,8 +153,12 @@ public function login($username, $password)
...
+		$ip_login_attempts		= (bool) ($this->config['ip_login_limit_max'] && $attempts >= $this->config['ip_login_limit_max']);
```
CHItA, line 158: I don't think you need the bool cast.

CHItA (changes requested):

phpBB/phpbb/auth/provider/db.php
3D-I: I think the way I proposed it is faster? Also used some lines above this changes.

`'user_id'				=> ($row) ? (int) $row['user_id'] : 0,`

phpBB/phpbb/auth/provider/db.php
3D-I: Probably not but I do not see anything against it, I can remove it if so you wish though.

phpBB/phpbb/auth/provider/db.php
CHItA: It shouldn't be faster, that would make zero sense, PHP needs to know the type either way, so type juggling cannot possibly be faster, but we are talking a couple of CPU cycles anyway, so it doesn't matter. `is_array` is explicit as to what you want to check.

3D-I: > We do not want to find whether the given variable is an array but a TRUE bool since a FALSE bool is already returned by the SQL query in case of false username.

CHItA: You have `$row['user_login_attempts']` in the code. This shouldn't work if `$row === true`. What am I missing?

phpBB/phpbb/auth/provider/db.php
CHItA: The `&&` operator produces a `bool`. So the cast is redundant.

phpBB/phpbb/auth/provider/db.php
3D-I: It is a ternary operatior `$row ? etc : false;` if **$row** is TRUE then evaluate the logic else return **false*

3D-I: This query
```php
		$sql = 'SELECT *
			FROM ' . USERS_TABLE . "
			WHERE username_clean = '" . $this->db->sql_escape($username_clean) . "'";
		$row = $this->db->sql_fetchrow($result);
```

Returns falkse in case, as I said. But returns everything we need otherwise, as an array already.

phpBB/phpbb/auth/provider/db.php
3D-I: Okay.

phpBB/phpbb/auth/provider/db.php
3D-I: image

CHItA: I don't really want to keep arguing, so here are my thoughts:
* the conditional operator cannot possibly be faster, as you first implicitly cast a variable to `bool`, which must require type look up, otherwise it cannot work, and thus `is_array` cannot possibly have any overhead;
* `is_array` is the most explicit way of checking that `$row` is an array, which is the thing you want to check, so you can index into it;
* your change is unnecessarily verbose, if it was `(1 == 1) ? true : false;` we surely could agree, that `(1 == 1)` is good enough;
* In this case, there is no need to branch, as you have nothing in the `else` branch, but `false`, hence, you want to compute a boolean expression;
* There is a shorter way to write the expression above, that is what I gave you;
* using `&&` is enough, the rest of the expression won't be evaluated - see `var_dump(false && 2 / 0);`;
* the way any code is written in this file or project is a poor argument to justify your changes, as you are changing the code. Surely, if all the code was awesome, you wouldn't need to make any changes

3D-I: Thanks for your time.

CHItA (approved):

marc1706 (approved):

PR #6176, opened 2021-03-26
[ticket/16740] Add method for ensuring resource is not passed to cache

marc1706:
Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16740

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -123,8 +123,8 @@ jobs:
                       db: "mysql:8.0"
                     - php: '8.0'
                       db: "mysql:5.7"
-                    - php: '8.1'
-                      db: "mysql:5.7"
+                    #- php: '8.1'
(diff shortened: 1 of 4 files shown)


marc1706:
Small explanation for the changes: The query id can be of type string or resource. Previously this was implicitly casted to int and hence worked for retrieving the data from the cache. In PHP 8.x this will no longer work and results in debug output. In general, the connect resource was used as key for arrays which is not properly supported by PHP.

Other fixed issues include sql_close() being called twice on the same resource and sql_table_exists() throwing PHP debug notices due to failing queries that output debug messages. Unfortunately, the @ operator does not silence that in PHP 8.0 anymore and there does not seem to be an easy to prevent this for pgsql apart from making sure queries can't have errors.

CHItA (approved):

PR #6179, opened 2021-04-02
Change {L_MARK_READ} to {L_MARK}

eshagh79:
To solve the problem of translating the variable "mark read" in the user control panel => notifications list

### DIFF
--- a/phpBB/styles/prosilver/template/ucp_notifications.html
+++ b/phpBB/styles/prosilver/template/ucp_notifications.html
@@ -57,7 +57,7 @@
 					<li class="header">
 						<dl>
 							<dt><div class="list-inner">{L_NOTIFICATIONS}</div></dt>
-							<dd class="mark">{L_MARK_READ}</dd>
+							<dd class="mark">{L_MARK}</dd>
 						</dl>
 					</li>
 				</ul>


DavidIQ:
What problem with `MARK_READ` is this addressing? Also the template you removed when putting in your description has links that explain in detail what is required for a pull request. Not doing those things means the checks won't pass.

eshagh79:
This was my first pull request phpbb. 
I thought they were not needed.
About the problem: In Persian, the translation "mark read" is very loud And it does not fit in 80 pixels
If this variable changes to "mark", the problem will be solved like other parts

eshagh79:
I am using Google Translator . If I made a mistake in answering, please forgive me

DavidIQ:
Sorry, but that then causes a problem for other languages. In English "Mark" and "Mark read" are two very different actions.

eshagh79:
this problem in : 
Board index => User Control Panel => Overview => Manage notifications : List header => Checkbox label

eshagh79:
My explanation is very awful.
I think you did not understand correctly

DavidIQ:
I understood perfectly. The correct text is being produced. `MARK` and `MARK_READ` are not interchangeable so you can either do a direct translation or find a different translation.

PR #6187, opened 2021-04-09
[ticket/16005] Remove phpBB2 converter

rubencm:
PHPBB3-16005

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16005


phpBB/config/installer/routing/installer.yml
```diff
@@ -36,31 +36,6 @@ phpbb_installer_update_conflict_download:
     defaults:
         _controller: phpbb.installer.controller.file_downloader:conflict_archive
 
-phpbb_convert_intro:
```
paul999, line 39: Won't removing the whole framework break all other existing converters as well? Is that something we really want?

3D-I: They can convert to 3.3.x then upgrade to 4.0.x?
Btw the old convertors are really very old.

rubencm: Well, it can be added later again, anyway there is no other converters.
The code can be added later again if there is need

Crizz0: And what about the custom user made converters from board software "abc" to phpBB? Are they not depending on this framework?

paul999: > And what about the custom user made converters from board software "abc" to phpBB? Are they not depending on this framework?

Yes, they are, and that is where my comment is from. By removing all of this we are removing support for all existing converters, and while they might be old, this might not be something we really want.

3D-I: > 
> 
> They can convert to 3.3.x then upgrade to 4.0.x?
> Btw the old convertors are really very old.

3D-I: The converters as a new and more modern framework it was a part of some GSOC in the past but the result was equal to null. No need IMHO to have all of this code for nothing.

rubencm: I can remove only the phpbb2 converter if you want, but remove everything is less code to maintain

phpBB/install/convertors/index.htm
```diff
@@ -0,0 +1,10 @@
+<html>
+<head>
+<title></title>
+<meta http-equiv="Content-Type" content="text/html; charset=iso-8859-1">
+</head>
+
+<body bgcolor="#FFFFFF" text="#000000">
+
```
3D-I, line 8: Deprecated.
```suggestion
<body style="background-color: #ffffff; color: #000000;">

```

rubencm: Is a copy/paste of the index.htm in /ext

3D-I: Yeah, all of those .htm files are old and deprecated, including the file extension.

phpBB/config/installer/routing/installer.yml
marc1706: Probably better like this

rubencm:
For me I would remove the entire converter system, that there is only one converter and is going to be removed now

PR #6206, opened 2021-05-03
[ticket/13713] User Mentions

marc1706:
Reworked version of https://github.com/phpbb/phpbb/pull/5225

Behavior is the same in the previous PR. The biggest difference is that the [zurb tribute js library](https://github.com/zurb/tribute) is being used instead of the abandoned at.js.

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-13713


phpBB/phpbb/textformatter/s9e/factory.php
```diff
@@ -126,6 +132,20 @@ class factory implements \phpbb\textformatter\cache_interface
 			</xsl:attribute>
 			<xsl:apply-templates/>
 		</a>',
+		'mention' => '<xsl:text>@</xsl:text>
+		<xsl:choose>
+			<xsl:when test="@profile_url">
+				<a class="mention">
```
JoshyPHP, line 138: Not that it makes a practical difference but you can set the `href` attribute directly on the element:
```xsl
<a class="mention" href="{@profile_url}">
```

phpBB/styles/prosilver/theme/colours.css
```diff
@@ -369,6 +369,41 @@ p.post-notice {
...
+}
+
+.mention-item:hover,
+.mention-item.cur {
+	background-color: #eeeeee;
+	color: #2d80d2;
+}
+
+.mention-item:hover .mention-media-avatar,
+.mention-item.cur .mention-media-avatar {
```
hanakin, line 399: please change all uses of .cur class to .is-active for consistency with standards

hanakin (changes requested):

phpBB/adm/style/admin.css
```diff
@@ -1670,6 +1670,112 @@ fieldset.submit-buttons legend {
...
+	overflow: auto; /* placed here for list to scroll with arrow key press */
+	max-height: 200px;
+	transition: all 0.2s ease;
+}
+
+.rtl .mention-container { /* mention-container */
+	text-align: right;
+}
+
+.mention-container ul { /* mention-list */
```
hanakin, line 1698: this needs to be its own class of mention-list rather than targeting an HTML selector to prevent proper inheritance. It should be failing styleint. This will probably require modifying the JS to add the class dynamically. something like `$('.mention-container').children('ul').toggleClass('mention-list')`

phpBB/adm/style/admin.css
```diff
@@ -1670,6 +1670,112 @@ fieldset.submit-buttons legend {
...
+	margin-right: 0;
+	margin-left: 16px;
+}
+
+.mention-media-avatar {
+	width: 40px;
+	height: 40px;
+}
+
+.mention-media-avatar img { /* TODO: reconsider for the future design with 1:1 ratio images */
```
hanakin, line 1724: we do not need this the avatar should be using the global avatar phpbb class which should already be doing this...

phpBB/adm/style/admin.css
```diff
@@ -1670,6 +1670,112 @@ fieldset.submit-buttons legend {
...
+	max-width: 100%;
+	max-height: 100%;
+}
+
+svg { /* TODO: remove it after general normalization */
```
hanakin, line 1729: this is not necessary here it should be at the top of the CSS file as its a global class suggest relocating it if its not there already

phpBB/styles/prosilver/theme/mentions.css
```diff
@@ -0,0 +1,92 @@
...
+	text-align: left;
+	border-radius: 2px;
+	position: absolute;
+	z-index: 999;
+	overflow: auto; /* placed here for list to scroll with arrow key press */
+	max-height: 200px;
+	transition: all 0.2s ease;
+}
+
+.mention-container ul { /* mention-list */
```
hanakin, line 26: again I think this should be mention-list class. may need to tweak the js for it though.

phpBB/styles/prosilver/theme/mentions.css
```diff
@@ -0,0 +1,92 @@
...
+	margin-right: 8px;
+	margin-left: 0;
+}
+
+.mention-media-avatar {
+	width: 40px;
+	height: 40px;
+}
+
+.mention-media-avatar img { /* TODO: reconsider for the future design with 1:1 ratio images */
```
hanakin, line 46: we do not need this the avatar should be using the global avatar phpbb class which should already be doing this...

phpBB/styles/prosilver/theme/mentions.css
```diff
@@ -0,0 +1,92 @@
...
+	max-width: 100%;
+	max-height: 100%;
+}
+
+svg { /* TODO: remove it after general normalization */
```
hanakin, line 51: this can be removed we should already have this in base.css I think. If not then it needs to go there...

hanakin (changes requested):

phpBB/adm/style/admin.css
marc1706: I don't think we have such a clear definition yet. There are only specific declarations for `.postprofile .avatar`

phpBB/adm/style/admin.css
marc1706: I don't currently see an option in zurb tribute to suppor this and I also don't see an option on how this could be overridden.

hanakin: It's not in zurb but you should be able to add the functionality via js like I mentioned above on call parse and add it essentially

phpBB/adm/style/admin.css
hanakin: What is the height even doing though?

marc1706: Ensuring the vatar is being displayed ;)

phpBB/adm/style/admin.css
marc1706: Done

phpBB/adm/style/admin.css
hanakin: yes prosilver is not using it yet but it can and should. We reworked all avatars to always have the class we just never reworked all the random css targeting them. The container should be setting the size for all of them which is why this is needed. otherwise, every avatar used everywhere would need specified HTML attributes for the height/width. I would suggest removing it here for adding `.avatar` in common.css after `.rowbg`

hanakin (changes requested):

hanakin:
@marc1706 looking at editor.js the only html in there is the svgs for default avatars (not sure why there are two?) which could just be included in the editor and hidden, then moved in and shown as needed. Also there is the span wrapped img avatars which could be handled the same way and update the content via the js. The rest of the html is contained in tribute.js which would be far more complex to circumvent but doable using the same method if you want to hate your life 😄 the complicated piece would be positioning it at the cursor...

phpBB/styles/prosilver/template/mentions_templates.html
```diff
@@ -0,0 +1 @@
+<template data-id="mentions-avatar-span"><span class="mention-media-avatar"><img class="avatar" src="" width="" height="" alt="" /></span></template>
```
3D-I, line 1: Looks odd if I can

Crizz0: `src=""` and `alt=""` are mandatory attributes. So they have to be there. But the closing `/` is only need in XHTML1.0 not in HTML 5 anymore.

hanakin: class names seem repetitive. I need to relook at the overall html structure and naming again... it was supposed to match https://codepen.io/hanakin/pen/NzwxWa?editors=1000 which would mean this should be changed to

`<template data-id="mention-media-span"><span class="mention-media"></span></template>`

seems like there is an extra wrapper in there for some reason...

no need for width and height thats what the max properties in the styles are for.

also should add the svgs in the same manner

hanakin (changes requested):

phpBB/styles/prosilver/template/mentions_templates.html
3D-I: Also the IMG tag is a void element.

3D-I: > 
> 
> `src=""` and `alt=""` are mandatory attributes. 

The `alt` attribute holds a text description of the image, which `isn't mandatory` but is incredibly useful for accessibility...

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Img

marc1706:
@hanakin HTML output should now align with the codepen

phpBB/styles/prosilver/theme/mentions.css
hanakin: given the changes to the html this should not require the img tag anymore

phpBB/styles/prosilver/theme/mentions.css
```diff
@@ -0,0 +1,88 @@
...
+.mention-media {
+	display: inline-flex;
+	flex-shrink: 0;
+	justify-content: center;
+	align-items: center;
+	margin-right: 8px;
+	margin-left: 0;
+}
+
+.mention-media-avatar {
```
hanakin, line 41: given the changes to the html I think this should be merged into the class above

phpBB/adm/style/admin.css
hanakin: given the changes to the html this should not require the img tag anymore

phpBB/adm/style/admin.css
```diff
@@ -1670,6 +1670,108 @@ fieldset.submit-buttons legend {
...
+	margin-right: 8px;
+	margin-left: 0;
+}
+
+.rtl .mention-media {
+	margin-right: 0;
+	margin-left: 16px;
+}
+
+.mention-media-avatar {
```
hanakin, line 1719: given the changes to the html this should be merged with the class above

hanakin (changes requested):

hanakin:
@marc1706 apart from the cleanup above looks good

marc1706:
And updated again.

hanakin (approved):
Looks good on my end

phpBB/language/en/ucp.php
```diff
@@ -332,6 +332,7 @@
 	'NOTIFICATION_TYPE_GROUP_REQUEST'					=> 'Someone requests to join a group you lead',
 	'NOTIFICATION_TYPE_FORUM'							=> 'Someone replies to a topic in a forum to which you are subscribed',
 	'NOTIFICATION_TYPE_IN_MODERATION_QUEUE'				=> 'A post or topic needs approval',
+	'NOTIFICATION_TYPE_MENTION'							=> 'Someone mentions you in a post',
```
CHItA, line 335: Shouldn't this be in past tense?

phpBB/phpbb/mention/controller/mention.php
```diff
@@ -0,0 +1,78 @@
...
+	 * @return JsonResponse|RedirectResponse
+	 */
+	public function handle()
+	{
+		if (!$this->request->is_ajax())
+		{
+			return new RedirectResponse(append_sid($this->phpbb_root_path . 'index.' . $this->php_ext));
+		}
+
+		$keyword = $this->request->variable('keyword', '', true);
```
CHItA, line 63: As far as I can tell we never check if the keyword is empty, it might make sense to require some amounts of characters before running queries.

CHItA (approved):

CHItA:
Other than the notes above, it looks good to me.

phpBB/language/en/ucp.php
marc1706: I think this is actually in line with the other descriptions:
image

phpBB/phpbb/mention/controller/mention.php
marc1706: This is actually by design, e.g. this would suggest friends directly when not having entered any input after the `@` yet.

neuropass:
Is it safe to assume there is a "enable/disable" option for this? included not loading all the JS and CSS related library when its disabled? Or this wasn't thought out at all?

DavidIQ:
> Is it safe to assume there is a "enable/disable" option for this? included not loading all the JS and CSS related library when its disabled? Or this wasn't thought out at all?

I think you could answer your own question by looking at the code. Otherwise you can wait until this is released as part of the next version.

PR #6207, opened 2021-05-03
[ticket/13700] Add method for initializing config after cache purge

marc1706:
PHPBB3-13700

The goal of this is to be able to re-initialize the config entries after a cache purge, e.g. when running db migrations via CLI.

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-13700

### DIFF
--- a/phpBB/phpbb/config/db.php
+++ b/phpBB/phpbb/config/db.php
@@ -13,20 +13,23 @@
 
 namespace phpbb\config;
 
+use phpbb\cache\driver\driver_interface as cache_interface;
+use phpbb\db\driver\driver_interface as db_interface;
+
 /**
 * Configuration container class
 */
-class db extends \phpbb\config\config
+class db extends config
 {
 	/**
 	* Cache instance
-	* @var \phpbb\cache\driver\driver_interface
+	* @var cache_interface
 	*/
 	protected $cache;
 
 	/**
 	* Database connection
-	* @var \phpbb\db\driver\driver_interface
+	* @var db_interface
 	*/
 	protected $db;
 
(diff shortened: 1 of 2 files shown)


phpBB/phpbb/config/db.php
```diff
@@ -39,16 +42,28 @@ class db extends \phpbb\config\config
...
+	/**
+	 * Initialise config with database and/or cached entries
+	 *
+	 * @param cache_interface $cache
+	 */
+	public function initialise(cache_interface $cache)
+	{
 		if (($config = $cache->get('config')) !== false)
 		{
 			$sql = 'SELECT config_name, config_value
```
rxu, line 69: During installation process, `phpbb_config` table does not exist until some point, so this SQL query causes hidden (not reported) SQL error.
Probably needs a check for `_config` table existence and initializing `$config` as an empty array if the table doesn't exist.

marc1706: I take it this was already an issue before this PR then?

rxu: Yeah, but looks like a good opportunity to fix it while changing the same code snippet :)

3D-I:
Is PHPBB3-13700 the right ticket?

marc1706:
I think so based on https://tracker.phpbb.com/browse/PHPBB3-13700?focusedCommentId=59504&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-59504 and the fact that I was able to reproduce issues in a similar fashion that resulted in the problems described in that ticket.

phpBB/phpbb/config/db.php
marc1706: Created a separate ticket for this: https://tracker.phpbb.com/browse/PHPBB3-16806

PR #6222, opened 2021-05-29
[ticket/16787] Properly check if fsockopen() is enabled - PHP 8

3D-I:
PHPBB3-16787

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16787

### DIFF
--- a/phpBB/phpbb/file_downloader.php
+++ b/phpBB/phpbb/file_downloader.php
@@ -43,7 +43,7 @@ class file_downloader
 		$this->error_string = '';
 
 		if (function_exists('fsockopen') &&
-			$socket = fsockopen(($port == 443 ? 'ssl://' : '') . $host, $port, $this->error_number, $this->error_string, $timeout)
+			$socket = @fsockopen(($port == 443 ? 'ssl://' : '') . $host, $port, $this->error_number, $this->error_string, $timeout)
 		)
 		{
 			@fputs($socket, "GET $directory/$filename HTTP/1.0\r\n");
--- a/phpBB/phpbb/file_downloader.php
+++ b/phpBB/phpbb/file_downloader.php
@@ -42,7 +42,9 @@ class file_downloader
 		$this->error_number = 0;
 		$this->error_string = '';
 
-		if ($socket = @fsockopen(($port == 443 ? 'ssl://' : '') . $host, $port, $this->error_number, $this->error_string, $timeout))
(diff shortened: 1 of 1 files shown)


3D-I:
I'm not sure I understand the problem with the test, as usual.

phpBB/phpbb/file_downloader.php
```diff
@@ -42,7 +42,9 @@ public function get($host, $directory, $filename, $port = 80, $timeout = 6)
 		$this->error_number = 0;
 		$this->error_string = '';
 
-		if ($socket = @fsockopen(($port == 443 ? 'ssl://' : '') . $host, $port, $this->error_number, $this->error_string, $timeout))
+		if (function_exists('fsockopen') &&
+			$socket = fsockopen(($port == 443 ? 'ssl://' : '') . $host, $port, $this->error_number, $this->error_string, $timeout)
```
marc1706, line 46: Add this back:
```suggestion
			$socket = @fsockopen(($port == 443 ? 'ssl://' : '') . $host, $port, $this->error_number, $this->error_string, $timeout)
```

marc1706 (changes requested):

rubencm:
Can that be just replaced by file_get_contents?, is much simpler

3D-I:
> 
> 
> Can that be just replaced by file_get_contents?, is much simpler
> Edit: Do we really need a class for this?

In a perfect world I would say that the use of different file reading techniques should be used, using code that automatically knows which one to use for each server configuration (cURL, file_get_contents, fsockopen etc...) or something like Guzzle in case. This before having to give a negative response and show an error to the end user. This PR proposes the solution for the problem highlighted by the ticket itself, only. As for the class, I would say that this is a question @marc1706 can answer as he is the one who wrote it 7 years ago.

marc1706:
This class replaces the original get_remote_file() function we had. So yes, moving this to a class was a viable step. Back then Guzzle was not an option yet. It could make sense to move this to use something like Guzzle, however that is outside the scope of this ticket and hence will require a new one.

marc1706 (approved):

3D-I:
> 
> 
> ... however that is outside the scope of this ticket and hence will require a new one.

Perhaps this ticket could be used on purpose.
https://tracker.phpbb.com/browse/PHPBB3-16738

PR #6225, opened 2021-06-04
[ticket/16790] Remove unused code

rubencm:
PHPBB3-16790

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16790

### DIFF
--- a/phpBB/includes/acp/acp_attachments.php
+++ b/phpBB/includes/acp/acp_attachments.php
@@ -272,15 +272,13 @@ class acp_attachments
 				$result = $db->sql_query($sql);
 
 				$defined_ips = '';
-				$ips = array();
 
 				while ($row = $db->sql_fetchrow($result))
 				{
-					$value = ($row['site_ip']) ? $row['site_ip'] : $row['site_hostname'];
+					$value = $row['site_ip'] ?: $row['site_hostname'];
 					if ($value)
 					{
 						$defined_ips .= '<option' . (($row['ip_exclude']) ? ' class="sep"' : '') . ' value="' . $row['site_id'] . '">' . $value . '</option>';
-						$ips[$row['site_id']] = $value;
 					}
 				}
 				$db->sql_freeresult($result);
@@ -353,7 +351,6 @@ class acp_attachments
 			break;
 
 			case 'extensions':
-
 				if ($submit || isset($_POST['add_extension_check']))
 				{
 					if ($submit)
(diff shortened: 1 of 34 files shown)


3D-I:
Are the `*CP` modules still to be kept as they are or planned to be ported to controllers?

rubencm:
They should be converted to controllers, but if I dont remember wrong mrgoldy had an issue doing it

phpBB/includes/acp/acp_board.php
```diff
@@ -477,7 +477,6 @@ function main($id, $mode)
 
 			default:
 				trigger_error('NO_MODE', E_USER_ERROR);
-			break;
```
CHItA, line 480: That's not in line with our coding guidelines.

CHItA:
This is really hard to review and I'm not sure how much benefit it has, as most of these files should be rewritten anyway at some point. We also require `break`s pretty much everywhere, which is arguably silly, but that should be updated first, before this is merged.

IMO, these changes while probably do no harm, do not provide much benefit either. To be clear, the code you touched is pretty horrible, my problem is that it will still be horrible after we merge this. I don't think these changes address any of the real issues, or make the code good.

phpBB/includes/acp/acp_attachments.php
```diff
@@ -226,7 +225,7 @@ function main($id, $mode)
 				}
 
 				// We go through the display_vars to make sure no one is trying to set variables he/she is not allowed to...
-				foreach ($display_vars['vars'] as $config_name => $null)
+				foreach (array_keys($display_vars['vars']) as $config_name)
```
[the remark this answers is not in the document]
CHItA: Giving up any performance or extra memory so that my IDE doesn't color a variable grey doesn't seem like a sound tradeoff. Also, this exact pattern is in the PHP docs: https://www.php.net/manual/en/function.array-key-first.php#refsect1-function.array-key-first-notes

marc1706 (changes requested):

tests/console/config/config_test.php
```diff
@@ -14,16 +14,14 @@
 use Symfony\Component\Console\Application;
 use Symfony\Component\Console\Tester\CommandTester;
 
-class phpbb_console_command_config_test extends phpbb_test_case
+class config_test extends phpbb_test_case
```
marc1706, line 17: Class name shouldn't be changed. (we should probably move tests to namespaces at some point ;) )

marc1706 (changes requested):

tests/console/config/config_test.php
rubencm: I changed it to be able to run tests individually without having to run all test suite. File names in tests are not consistent with the class names.

rubencm:
Dont merge, i am going to clean functions_compress and keep only the code that is used
A decent component for compress/decompress can be done later for the one click updater

marc1706 (approved):

PR #6227, opened 2021-06-09
[ticket/16793] Fix topic/post attachments visibility - PHP 8

3D-I:
PHPBB3-16793

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16793

### DIFF
--- a/phpBB/phpbb/content_visibility.php
+++ b/phpBB/phpbb/content_visibility.php
@@ -148,7 +148,7 @@ class content_visibility
 
 		$is_visible = ($visibility == ITEM_APPROVED) ||
 			($this->config['display_unapproved_posts']
-				&& ($this->user->data['user_id'] != ANONYMOUS)#
+				&& ($this->user->data['user_id'] != ANONYMOUS)
 				&& ($visibility == ITEM_UNAPPROVED || $visibility == ITEM_REAPPROVE)
 			) ||
 			$this->auth->acl_get('m_approve', $forum_id);
--- a/phpBB/phpbb/content_visibility.php
+++ b/phpBB/phpbb/content_visibility.php
@@ -145,13 +145,13 @@ class content_visibility
 	public function is_visible($mode, $forum_id, $data)
 	{
 		$visibility = $data[$mode . '_visibility'];
-
+		$poster_key = ($mode === 'topic') ? 'topic_poster' : 'poster_id';
 		$is_visible = ($visibility == ITEM_APPROVED) ||
-			($this->config['display_unapproved_posts']
-				&& ($this->user->data['user_id'] != ANONYMOUS)
-				&& ($visibility == ITEM_UNAPPROVED || $visibility == ITEM_REAPPROVE)
-			) ||
-			$this->auth->acl_get('m_approve', $forum_id);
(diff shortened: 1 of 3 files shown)


phpBB/phpbb/content_visibility.php
```diff
@@ -145,13 +145,13 @@ public function get_count($mode, $data, $forum_id)
 	public function is_visible($mode, $forum_id, $data)
 	{
 		$visibility = $data[$mode . '_visibility'];
-		$poster_key = ($mode === 'topic') ? 'topic_poster' : 'poster_id';
+
 		$is_visible = ($visibility == ITEM_APPROVED) ||
-			($this->config['display_unapproved_posts'] &&
-				($this->user->data['user_id'] != ANONYMOUS) &&
-				($visibility == ITEM_UNAPPROVED || $visibility == ITEM_REAPPROVE) &&
-				($this->user->data['user_id'] == $data[$poster_key])) ||
```
CHItA, line 153: This is needed to show unapproved posts to their authors. Hence this code incorrect.

CHItA (changes requested):

phpBB/phpbb/content_visibility.php
CHItA: What I meant to say is to the post's author only (or users who can approve the post). This code looks like it will show the post to anyone who is logged in.

3D-I: Not really, try it.

I tried it with me as the founder, with the poster itself and with another tester.

3D-I: For the "another tester" a "topic does not exist" it is shown, BTW.

3D-I: BTW, we are talking about attachments here.

CHItA: Sorry for the delay, so this seem to work because the query is already filtering the user ID. However, the real issue is not here. There are two errors in the ticket:
1) `... on line 153: Undefined array key "topic_poster"`
2) `... on line 153: Undefined array key "poster_id"`

1) is comming from [here](https://github.com/phpbb/phpbb/blob/5169a947f4caef876cd2f38c558d042aef0a24ae/phpBB/includes/functions_download.php#L659-L676), note that `topic_poster` is missing from the query.
2) is comming from [here](https://github.com/phpbb/phpbb/blob/3.3.x/phpBB/download/file.php#L219-L236), note that `poster_id` is missing from the query.

3D-I: I didn't have those 2 errors, also not in the logs.

CHItA: those should be notices. I'm going by the ticket, the user said they saw those when clicking download, which makes sense. So this should error out when you go via `file.php` which should result in broken images, as the content served is text with the notices the reporter posted.

3D-I: Yeah sure, btw the attachments are those I posted there. Just I haven't had any NOTICE nor error_logs.
Wich is quite strange in my environment, I am double checking the PHP 8.0.7 configuration.

marc1706 (approved):

PR #6233, opened 2021-06-15
[ticket/16799] Fix OAuth external account linking PHP fatal error

rxu:
Checklist:

- [x] Tests pass

<a href="https://tracker.phpbb.com/browse/PHPBB3-16799">PHPBB3-16799</a>.

Replacement for #6232.

### DIFF
--- a/phpBB/phpbb/auth/provider/oauth/oauth.php
+++ b/phpBB/phpbb/auth/provider/oauth/oauth.php
@@ -163,7 +163,7 @@ class oauth extends base
 		$provider = $this->request->variable('oauth_service', '', false);
 		$service_name = $this->get_service_name($provider);
 
-		if ($provider === '' || !array_key_exists($service_name, (array) $this->service_providers))
+		if ($provider === '' || !$this->service_providers->offsetExists($service_name))
 		{
 			return [
 				'status'		=> LOGIN_ERROR_EXTERNAL_AUTH,
@@ -452,7 +452,7 @@ class oauth extends base
 
 		$service_name = $this->get_service_name($link_data['oauth_service']);
 
-		if (!array_key_exists($service_name, (array) $this->service_providers))
+		if (!$this->service_providers->offsetExists($service_name))
 		{
 			return 'LOGIN_ERROR_OAUTH_SERVICE_DOES_NOT_EXIST';
 		}
--- a/phpBB/phpbb/auth/provider/oauth/oauth.php
+++ b/phpBB/phpbb/auth/provider/oauth/oauth.php
@@ -163,7 +163,7 @@ class oauth extends base
 		$provider = $this->request->variable('oauth_service', '', false);
 		$service_name = $this->get_service_name($provider);
 
-		if ($provider === '' || !array_key_exists($service_name, $this->service_providers))
+		if ($provider === '' || !array_key_exists($service_name, (array) $this->service_providers))
 		{
 			return [
 				'status'		=> LOGIN_ERROR_EXTERNAL_AUTH,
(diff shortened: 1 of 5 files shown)


3D-I:
About the lusitanian's error, if you are trying Oauth with a fake google account you can't seeit, since it is thrown when you get back to UCP after having linked your account. You can try with this one if you like

https://phpbbstudio.com/extensions/discord-oauth2-light

And you will see it
image

rxu:
Now I see.
https://www.php.net/manual/en/function.array-walk.php says
> Note:
If callback needs to be working with the actual values of the array, specify the first parameter of callback as a reference. Then, any changes made to those elements will be made in the original array itself.

But in PHPoAuthLib both parameters are references:
`            function (&$val, &$key): void {`

https://github.com/Lusitanian/PHPoAuthLib/blob/f992e3913c15d322c392d9041166b065e73027fb/src/OAuth/Common/Http/Client/AbstractClient.php#L67

So the library source needs to be adjusted in according to PHP docs.

3D-I:
> So the library source needs to be adjusted in according to PHP docs to look like
`function (&$val, $key): void {`

Yeah, that's what I said here:

https://tracker.phpbb.com/browse/PHPBB3-16799?focusedCommentId=64403&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-64403

rxu:
Well, it's worth sending PR to the library source repository then as we can't fix it on our own :)

3D-I:
It is abandoned. I opened a ticket to dimiss lusitanian.

Anyway, it is better such an error than a complete crash.

rxu:
There's a PR though https://github.com/Lusitanian/PHPoAuthLib/pull/575

3D-I:
Oh, it is new. But that repo is abandoned IMHO.

3D-I:
@rxu Is there any way to override that faulty function that you are aware of?
I have some idea but not sure is applicable.

rxu:
I'd rather vote for migrating to some alternative library, like https://github.com/laravel/socialite or similar.

3D-I:
> 
> 
> I'd rather vote for migrating to some alternative library, like https://github.com/laravel/socialite or similar.

I agree. For phpBB 4 or greater though.
 I do not see that applicable to the current 3.3 serie, IMHO.

phpBB/phpbb/auth/provider/oauth/oauth.php
```diff
@@ -163,7 +163,7 @@ public function login($username, $password)
 		$provider = $this->request->variable('oauth_service', '', false);
 		$service_name = $this->get_service_name($provider);
 
-		if ($provider === '' || !array_key_exists($service_name, $this->service_providers))
+		if ($provider === '' || !array_key_exists($service_name, (array) $this->service_providers))
```
CHItA, line 166: My preference would probably be `offsetExists($service_name)` here.

3D-I: > 
> 
> My preference would probably be `offsetExists($service_name)` here.

Why?

marc1706: Because the service collection implement ArrayAccess and therefore the default goto function should be https://www.php.net/manual/en/arrayaccess.offsetexists.php

CHItA: You also make the dependency on ArrayAccess explicit. Your code depends on it by using the array cast (if the class wouldn't derive from ArrayAccess, the cast would return class properties). Essentially, by making the dependency explicit, this should probably break tests if the service collection class were to change. Which is probably what we would want to happen.

CHItA (approved):

PR #6235, opened 2021-06-17
[ticket/16796] Fixing misalignment of header items in category on index.

MannixMD:
Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16796

Discussion: https://area51.phpbb.com/phpBB/viewtopic.php?f=141&t=77976&p=344901#p344901

Ticket: PHPBB-16796


DavidIQ:
This won't pass merge checks. See last point in the checklist in description as that is a requirement.

MannixMD:
I think I fixed the issue? I'm sorry doing this for the first time :)

3D-I:
Have allok at how the other PRs are made, I am sure you will get the point.

https://github.com/phpbb/phpbb/pulls

hanakin:
I fixed the easy stuff for you, the title and the PR message. You need to rebase on master and redo your commit messages to match the required commit message format

3D-I:
Never seen a title so long. 😃

3D-I:
Back to the topic, there are questions left there I think. Like the rean wahy some of those headers are now included into 'span' and something else not. Which suggest that the above change probably is not the right way to deal with this issue, specifically. @hanakin ?

https://area51.phpbb.com/phpBB/viewtopic.php?f=141&t=77976&p=344701#p344701

hanakin:
Have not looked at it as I’m out of capacity for the next couple weeks, but
ideally all should be handled the same and properly

3D-I:
I give up here. 😉

MannixMD:
Sorry for late response my pc died. I think I did it ?

MannixMD:
Well it was to long so I had to shorten it :D

marc1706:
Thanks for your PR. I've taken the liberty of picking your first commit, cleaning up all the other commits, and fixing the commit message.

hanakin:
@MannixMD sorry for the late response. First thanks for taking the time to submit a fix this it really helps us out alot. Keep in mind that this is the master branch which means 4.0 and allows us to make breaking changes. so we can as @3D-I pointed out fix the structure of the html here as well to be more standard and proper. I will do a more detailed review of the issue tonight as my PC at work will not let me for some reason right now. But looking at the source code does reveal that for some reason LAST POST header in fourmlist_body.html is not following the same standard as the rest of the headers it has a span around it. https://github.com/phpbb/phpbb/blame/master/phpBB/styles/prosilver/template/forumlist_body.html#L21 This is possibly not needed and should be removed as well. Also with any changes to the front-end however minor it may be we like to encourage the inclusion of before and after screenshots to speed up the approval and review process for PRs here on github. This allows for the transition from the tracker to github for all the discussion going forward as well.

hanakin:
@MannixMD  after testing yes the span highlighted above is the actual issue here.

Prosk8er:
@hanakin 

it not just the span though you would still have to do his css edit also otherwise lines 19-21 won't line up with line 18

hanakin:
Not based on my testing, but I did not test windows or Firefox so might be there. But either way it's due to a lack of consistent styling then on all the headers if that's the case.

Prosk8er:
i tried it on latest github version and through browser on area51
win 10 latest chrome and firefox

MannixMD:
Removing the span from last post aligns it with the previous two (posts and topics) but you still need to align it with the category/forum name on the left side which has line-height of 16px. So you either need to adjust the line-height for the titles or for the posts/topics/lastpost.

3D-I:
That span should stay after all got a CSS rule (2) for it.

If you `roll-back the present change` and apply this 2 then it should be ok, always in `content.css`

---
apply this 2 then it should be ok, always in `content.css`

line 169

```css
li.header dl.row-item dt .list-inner {
	padding-right: 50px;
	padding-left: 0;
	line-height: 10px;
}
```

line 256

```css
dd.lastpost > span {
	line-height: 10px;
}
```

---

image

hanakin:
I responded to the discussion on area51 to long to post here https://area51.phpbb.com/phpBB/viewtopic.php?f=141&t=77976&p=344901#p344901

hanakin:
@CHItA it needs more work discussion see my post on area51 above

Prosk8er:
> ```css
> line-height: 10px;
> ```

i agree with this fix because it brings the headers size back down to what it should be

3D-I:
Still the changes are not perfect, more @ area51's devoted topic.

hanakin:
Ok I think the change will work however we may want to consider adding some padding to the title going forward. @MannixMD as this is not a breaking change it can go into 3.x can you please rebase it onto that branch?

hanakin (changes requested):
Rebranch on 3.x

marc1706:
Rebased the changes onto 3.3.x and created a follow up PR for master as well:
#6457 
#6458 

Closing this PR in favor of those.

PR #6236, opened 2021-06-27
[ticket/16804] Extend bbcode help line tooltip

3D-I:
PHPBB3-16804

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16804

### DIFF
--- a/phpBB/phpbb/db/migration/data/v33x/extend_bbcode_tooltip.php
+++ b/phpBB/phpbb/db/migration/data/v33x/extend_bbcode_tooltip.php
@@ -1,0 +1,51 @@
+<?php
+/**
+*
+* This file is part of the phpBB Forum Software package.
+*
+* @copyright (c) phpBB Limited <https://www.phpbb.com>
+* @license GNU General Public License, version 2 (GPL-2.0)
+*
+* For full copyright and license information, please see
+* the docs/CREDITS.txt file.
+*
+*/
+
+namespace phpbb\db\migration\data\v33x;
+
+class extend_bbcode_tooltip extends \phpbb\db\migration\migration
+{
+	public function effectively_installed()
+	{
+		return $this->db_tools->sql_table_exists($this->table_prefix . 'bbcodes');
+	}
+
+	static public function depends_on()
+	{
+		return [
+			'\phpbb\db\migration\data\v33x\v335rc1'
+		];
+	}
+
+	public function update_schema()
+	{
+		return [
+			'change_columns'	=> [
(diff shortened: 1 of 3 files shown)


3D-I:
I will push the new migration file soon.

FlyThePlanet:
Thanks @3D-I !
The idea is to allow this kind of help/tooltip...

image

phpBB/phpbb/db/migration/data/v33x/extend_bbcode_tooltip.php
```diff
@@ -0,0 +1,51 @@
...
+*
+*/
+
+namespace phpbb\db\migration\data\v33x;
+
+class extend_bbcode_tooltip extends \phpbb\db\migration\migration
+{
+	public function effectively_installed()
+	{
+		return $this->db_tools->sql_table_exists($this->table_prefix . 'bbcodes');
```
marc1706, line 20: If I see this correctly, this migration will not be applied when the bbcodes table already exists. This should however be the case for all 3.x boards.

3D-I:
@marc1706 Is there something else I can do?

phpBB/phpbb/db/migration/data/v33x/extend_bbcode_tooltip.php
```diff
@@ -0,0 +1,51 @@
...
+*
+*/
+
+namespace phpbb\db\migration\data\v33x;
+
+class extend_bbcode_tooltip extends \phpbb\db\migration\migration
+{
+	public function effectively_installed()
+	{
+		return !$this->db_tools->sql_column_exists($this->table_prefix . 'bbcodes', 'bbcode_helpline');
```
marc1706, line 20: I actually wonder if there would be a way to find out which type the column is. In that case, this would then also return true of the helpline is of type `TEXT_UNI`

3D-I: I do not think there is something ready to be used for that, IIRC.
If somehow that column is already of that type nothing changes, just it gets changed one more time to the same type. This only could happen if someone manually changed that type though.

3D-I: Probably using a query which uses `SELECT DATA_TYPE, COLUMN_NAME  FROM etc..`, but as I said there is nothing AFAIR in the db tools for that.

3D-I:
I didn't get the response I was due so I wonder what the point of this PR is, so I'll close. Thx.

PR #6248, opened 2021-07-18
[ticket/16207] Require cookies for sessions

marc1706:
This will change the way phpBB behaves with regards to appending the session ID to URLs. The new and default behavior is to not append the session ID. Re-authentication for the ACP still works the same as before.
If the cookie settings are invalid, users can enable the `session.force_sid` in the config.yml to end up with the same behavior as before where the `sid` parameter will be appended if needed.

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16207


phpBB/includes/acp/acp_prune.php
```diff
@@ -327,7 +327,7 @@ function prune_users($id, $mode)
 						'USERNAME'			=> $usernames[$user_id],
 						'USER_ID'           => $user_id,
 						'U_PROFILE'			=> get_username_string('profile', $user_id, $usernames[$user_id]),
-						'U_USER_ADMIN'		=> ($auth->acl_get('a_user')) ? append_sid("{$phpbb_admin_path}index.$phpEx", 'i=users&amp;mode=overview&amp;u=' . $user_id, true, $user->session_id) : '',
+						'U_USER_ADMIN'		=> ($auth->acl_get('a_user')) ? append_sid("{$phpbb_admin_path}index.$phpEx", 'i=users&amp;mode=overview&amp;u=' . $user_id) : '',
```
CHItA, line 330: Do we still need `append_sid` at all?

marc1706: With the current change we'll still have the fallback to force the old behavior with the mentioned config setting. Maybe it could also be possible to move the cookie settings to a config file to allow fixing this manually of needed? I would however see this in an additional ticket and is IMHO outside the scope of this one.

CHItA (approved):

Derky (approved):

PR #6261, opened 2021-08-05
[ticket/16831] fix tabs missing delimiters and remove commented class

Prosk8er:
this is just to fix tabs, missing delimiters, and remove old commented classes in css files

PHPBB3-16831

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16831

### DIFF
--- a/phpBB/styles/prosilver/theme/buttons.css
+++ b/phpBB/styles/prosilver/theme/buttons.css
@@ -155,7 +155,7 @@
 /* Browser-specific tweaks */
 button::-moz-focus-inner {
 	padding: 0;
-	border: 0
+	border: 0;
 }
 
 /* Deprecated as of version 3.2
--- a/phpBB/styles/prosilver/theme/cp.css
+++ b/phpBB/styles/prosilver/theme/cp.css
@@ -5,7 +5,7 @@
 /* Main CP box
 ----------------------------------------*/
 .cp-menu {
-	float:left;
+	float: left;
 	width: 19%;
 	margin-top: 1em;
 	margin-bottom: 5px;
@@ -299,13 +299,6 @@ ol.def-rules li {
 .pmlist li.bg2 {
 	padding: 0 3px;
 }
-
-/* DEPRECATED 3.2.6
-.pmlist li.pm_message_reported_colour, .pm_message_reported_colour {
-	border-left-color: transparent;
-	border-right-color: transparent;
-}
-*/
 
 .pmlist li.pm_message_reported_colour, .pm_message_reported_colour,
 .pmlist li.pm_marked_colour, .pm_marked_colour,
--- a/phpBB/styles/prosilver/theme/base.css
+++ b/phpBB/styles/prosilver/theme/base.css
@@ -36,8 +36,8 @@ textarea {
 	line-height: inherit;
 }
 
-figure { margin: 0 }
-img { vertical-align: middle }
+figure { margin: 0; }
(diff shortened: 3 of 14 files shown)


phpBB/styles/prosilver/theme/buttons.css
```diff
@@ -155,39 +155,5 @@
...
-}
-
-.rtl .small-icon {
-	background-position: 100% 50%;
-}
-
-.rtl .small-icon > a {
-	padding-left: 0;
-	padding-right: 19px;
+	border: 0;
```
hanakin, line 158: Not sure we can remove these. This requires a minor version change ie 3.4 which is not currently planned. You can make this change in master however

hanakin (changes requested):

phpBB/styles/prosilver/theme/buttons.css
Prosk8er: i searched in the html files and couldn't find a call to it for 3.3.x so its not being used and was Deprecated as of version 3.2?
why cant it be removed in 3.3.x?

hanakin: It's a policy to deprecate 1 version then remove it next so If it was deprecated in 3.2 actually we should of removed it when we released 3.3 but I do not like doing it after a release as there could be some still using and we would be removing it breaking the code with out a big release. That's why I suggested to submit its removal on the master branch. It's not hurting anything be here

hanakin: Fix versions should have little to no changes to styling and html except for extreme bugs or unobtrusive visual fixes

hanakin: This is something I aim to change going forward with the new theme however by decoupling it's development from the main project and releasing as a style. But to do that we need to increase the front end dev team

Prosk8er: ok what about the ones that were commented out are they fine to be removed in this pr?

phpBB/styles/prosilver/theme/cp.css
```diff
@@ -300,13 +300,6 @@ ol.def-rules li {
 	padding: 0 3px;
 }
 
-/* DEPRECATED 3.2.6
-.pmlist li.pm_message_reported_colour, .pm_message_reported_colour {
-	border-left-color: transparent;
-	border-right-color: transparent;
-}
-*/
-
```
Prosk8er, line 309: @hanakin this is the one im referring about in my last message

3D-I: this can be removed IMHO, it is akready commented code so unusable, the same class exists elsewhere but not used, still
`"ROOT/3211\styles\prosilver\theme\colours.css"(905,2)`

Prosk8er: yeah i noticed a few that are blank i'll make a PR for the changes i just reverted and the ones that are blank

hanakin: Yes it's fine here cause it was commented out

hanakin (approved):
Look good now, minus the removal of the comment section I mentioned above

phpBB/styles/prosilver/theme/cp.css
Prosk8er: ok cool ill make a new ticket to remove the uncommented not used 3.2 classes for master

PR #6296, opened 2021-09-19
[ticket/16879] Add events to phpbb_mail()

3D-I:
PHPBB3-16879

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16879

### DIFF
--- a/phpBB/includes/functions_messenger.php
+++ b/phpBB/includes/functions_messenger.php
@@ -1945,7 +1945,7 @@ function phpbb_mail($to, $subject, $msg, $headers, $eol, &$err_msg)
 	 * @var string	headers					The email headers
 	 * @var string	eol						The endline character
 	 * @var string	additional_parameters	The additional parameters
-	 * @since 3.3.5-RC1
+	 * @since 3.3.6-RC1
 	 */
 	$vars = [
 		'to',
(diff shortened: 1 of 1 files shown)


marc1706:
Events will not be added between RC and stable versions for a patch level release.

phpBB/includes/functions_messenger.php
```diff
@@ -1935,8 +1935,54 @@ function phpbb_mail($to, $subject, $msg, $headers, $eol, &$err_msg)
 	 */
 	$additional_parameters = $config['email_force_sender'] ? '-f' . $config['board_email'] : '';
 
+	/**
+	 * Modify data before sending out emails with the PHP's mail function
```
marc1706, line 1939:
```suggestion
	 * Modify data before sending out emails with PHP's mail function
```

phpBB/includes/functions_messenger.php
```diff
@@ -1935,8 +1935,54 @@ function phpbb_mail($to, $subject, $msg, $headers, $eol, &$err_msg)
...
+		'headers',
+		'eol',
+		'additional_parameters',
+	];
+	extract($phpbb_dispatcher->trigger_event('core.phpbb_mail_before', compact($vars)));
+
 	$result = mail($to, mail_encode($subject, ''), wordwrap(utf8_wordwrap($msg), 997, "\n", true), $headers, $additional_parameters);
 
+	/**
+	 * Execute code after sending out emails with the PHP's mail function
```
marc1706, line 1963:
```suggestion
	 * Execute code after sending out emails with PHP's mail function
```

CHItA (approved):

rubencm:
Why is needed one for before and another for after?
Only before event dont cover all use cases?
@3D-I 

Also wouldn't be better to just dispatch one event before sending email (independently if is with smtp or mail)?

3D-I:
@rubencm The `after` is needed because of the `result`. If false (mail not sent) we can use the same event to try and send the email using another transport or store it locally (for testing purposes, as an example). Useful if testing extensions locally, as I do.

PR #6298, opened 2021-09-23
[ticket/16881] Fix ACP Statistic Table

Dark1z:
PHPBB3-16881

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16881

### DIFF
--- a/phpBB/adm/style/acp_main.html
+++ b/phpBB/adm/style/acp_main.html
@@ -139,8 +139,8 @@
 						<td class="tabled">{{ lang('PHP_VERSION') ~ lang('COLON') }}</td>
 						<td class="tabled"><strong>{{ PHP_VERSION_INFO }}</strong></td>
 					</tr>
-					<tr>
 					{% if S_TOTAL_ORPHAN %}
+					<tr>
 						<td class="tabled">{{ lang('NUMBER_ORPHAN') ~ lang('COLON') }}</td>
 						<td class="tabled">
 						{% if TOTAL_ORPHAN > 0 %}
@@ -150,8 +150,8 @@
 						{% endif %}
 						</td>
 						{% else %}
-					{% endif %}
 					</tr>
+					{% endif %}
 					{% if S_VERSIONCHECK %}
 					<tr>
 						<td class="tabled">{{ lang('BOARD_VERSION') ~ lang('COLON') }}</td>
@@ -204,10 +204,12 @@
 						<td class="tabled">{{ lang('FILES_PER_DAY') ~ lang('COLON') }}</td>
 						<td class="tabled"><strong>{{ FILES_PER_DAY }}</strong></td>
 					</tr>
+					{% if S_TOTAL_ORPHAN and S_VERSIONCHECK %}
 					<tr>
 						<td class="tabled">&nbsp;</td>
 						<td class="tabled">&nbsp;</td>
 					</tr>
+					{% endif %}
 				</tbody>
 			</table>
 		</div>


3D-I:
I left a comment in the tracker

Here is the full statistics block
image

Dark1z:
Ahh!!! I forgot to reserve the 2nd comment for Explanation.

So here is it:
Only use empty table row if required.

Before: 

&


After:

&


Best regards 👍

Dark1z:
> I left a comment in the tracker
> 
> Here is the full statistics block - so you are removing it if zeroed?

I left a comment for you on tracker too.
But let me reply the same here:
>No!
>the table on RHS.
>
>If "orphan attachments" row is not used in LHS table then remove the BLANK row from RHS table.
>Before due to the BLANK row from RHS table, there was a Blank row on both the tables as shown in Image.

Best regards 👍

3D-I:
I don't see such a discrepance here, example of your second IMG

image

Dark1z:
> I don't see such a discrepance here, example of your second IMG
> 
> So again, you are willing to hide the orphan attachments row if zeroed and as such you need to modify the template?

@3D-I ,

Try disabling the attachments.

Also do check the changed code properly,
I am not hiding "orphan attachments row" at all,
I am hiding the BLANK row.

3D-I:
> 
> Try disabling the attachments.
> 

I can see it now. 
Why didn't you also indicate how to reproduce this problem?
I would never have imagined disabling attachments in a forum. 
I would have avoided commenting at least. ;-)

Dark1z:
can somebody restart the following jobs:
1. https://github.com/phpbb/phpbb/pull/6298/checks?check_run_id=3691466604
2. https://github.com/phpbb/phpbb/pull/6298/checks?check_run_id=3691466658
It happened due to internal SQL or PHP error, and not related to the code changes.

Thanking you, Best regards 👍

Dark1z:
> > Try disabling the attachments.
> 
> I can see it now.
> Why didn't you also indicate how to reproduce this problem?
> I would never have imagined disabling attachments in a forum.
> I would have avoided commenting at least. ;-)

No problem, it happens , an itch to comment.
I have been there ;-)

3D-I:
Does the "orphan attachments" row disappear also if positive, when disabling attachments?
I mean the row disappears only if zeroed?

Dark1z:
> Does the "orphan attachments" row disappear also if positive, when disabling attachments?
> I mean that row disappears only if zeroed?

Yes!
coz of this:
https://github.com/phpbb/phpbb/pull/6298/commits/119e6882cc1c36a567b32d1febe542338ec00ba0#diff-2ea3f5f13d7a3be05b231a4502fd49a24ca26da726b892abc7abc70bf07de214R142

3D-I:
I think that the bug is elsewhere, in ACP 
https://github.com/phpbb/phpbb/blob/3d76942cdfe1ca86adf39a6ca1d8b125e73f16b1/phpBB/includes/acp/acp_main.php#L547-L560

All the attachment related rows are shown per default also if zeroed, same way the "orphans" should act.
This way there is no need to modify the template and everything will be more consistent, IMHO.

Dark1z:
> I think that the bug is elsewhere, in ACP
> https://github.com/phpbb/phpbb/blob/3d76942cdfe1ca86adf39a6ca1d8b125e73f16b1/phpBB/includes/acp/acp_main.php#L547-L560
> 
> All the attachment related rows are shown per default also if zeroed, same way the "orphans" should act.
> This way there is no need to modify the template and everything will be more consistent, IMHO.

Yes! sure you can make it as your new PR.

My goal is to handle the Extra BLANK table row on RHS table.
Hence this PR.

3D-I:
Ok, i pushed my own PR in case https://github.com/phpbb/phpbb/pull/6299

Dark1z:
> Ok, i pushed my own PR in case #6299

Ok. closing this issue due to PR #6299 

Best regards 👍

PR #6300, opened 2021-10-04
[ticket/16882] Add file_exists() function to Twig

3D-I:
PHPBB3-16882

Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16882

### DIFF
--- a/phpBB/phpbb/template/twig/extension.php
+++ b/phpBB/phpbb/template/twig/extension.php
@@ -206,12 +206,13 @@ class extension extends \Twig\Extension\AbstractExtension
 	/**
 	 * Check if a file exists
 	 *
-	 * Example: <samp>{% if file_exists('./ext/vendor/name/styles/all/theme/images/blah.png') %}</samp>
+	 * Example:
+	 * <samp>{% if file_exists('./ext/vendor/name/styles/all/theme/images/blah.png') %}</samp>
 	 *
-	 * @param string	$path	The path to be checked
+	 * @param string	$path	The relative path to be checked
 	 * @return bool				True if exists, false otherwise
 	 */
-	public function on_disk($path)
+	public function on_disk($path): bool
 	{
 		return call_user_func_array('file_exists', [$path]);
 	}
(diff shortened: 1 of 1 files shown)


3D-I:
The problems with Windows tests are not related to the code, so my task ends here. 

As for a possible test file, I don't think I can write one unless someone gives me links to some related documentation so I can study it.

phpBB/phpbb/template/twig/extension.php
```diff
@@ -93,6 +93,7 @@ public function getFunctions()
 			new \Twig\TwigFunction('lang', array($this, 'lang')),
 			new \Twig\TwigFunction('lang_defined', array($this, 'lang_defined')),
 			new \Twig\TwigFunction('lang_js', [$this, 'lang_js']),
+			new \Twig\TwigFunction('file_exists', [$this, 'on_disk']),
```
rubencm, line 96: If I don't remember wrong, you can just do
```php
new \Twig\TwigFunction('file_exists', 'file_exists'),
```

3D-I: Yeah but I want to use `if file_exists()` not `if on_disk()` in the markup.

rubencm: Then do
```php
new \Twig\TwigFunction('file_exists', 'file_exists'),
```
just like i did in with get_class

3D-I: ^^ see above, yup

iMattPro:
agree with @CHItA question. we really should avoid doing much of any PHP/backend work in the templates. aside from just the usual stuff needed to logically render out template content.

CHItA:
not quite sure how your example is simpler then doing it in PHP.

marc1706:
You're essentially adding PHP logic to HTML which is then converted back to PHP logic by twig. While the aim to have HTML not in PHP is correct, this is not what we end up with implementing this. I therefore agree with the others that this twig function will not help and will only result in people more or less adding PHP logic to HTML.

PR #6313, opened 2021-10-21
[ticket/16891] Do not change an extension status in the midle of a request

Nicofuma:
When enabling an extension, it should be considered as not being enabled for the entire request as if the "enabled" flag is switch furing the request, the extension will stay not loaded (same when disabling an extension).

Example when it can be an issue today : if the router is called for the first time after enabling the extension and if the extension uses a custom route loader (like phpbb/pages) then the router will fail because the custom route will be found but the custom loader will not be loaded and available.

PHPBB3-16891

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16891

### DIFF
--- a/phpBB/config/default/container/services.yml
+++ b/phpBB/config/default/container/services.yml
@@ -116,6 +116,7 @@ services:
             - '@dbal.conn'
             - '@config'
             - '@filesystem'
+            - '@router'
             - '%tables.ext%'
             - '%core.root_path%'
             - '%core.php_ext%'
--- a/phpBB/phpbb/extension/manager.php
+++ b/phpBB/phpbb/extension/manager.php
@@ -30,9 +30,11 @@ class manager
 	protected $cache;
 	protected $php_ext;
 	protected $extensions;
+	protected $recently_changed_ext_status;
 	protected $extension_table;
 	protected $phpbb_root_path;
 	protected $cache_name;
+	protected $router;
 
 	/**
 	* Creates a manager and loads information from database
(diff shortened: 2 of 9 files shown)


Nicofuma:
The issue is that when you enable an extension you end up with an invalid router in cache or a failure in success page in the acp.

This is happening because when we enable an extension, we enable it, then we clear the cache and in the end we render the page. But when we render the page we might need the router for the first time, which means that the router will build the url generator and at this point, 2 scenarios:
 1. The current status: it detects the new enabled extension and uses it to construct the generator; Issue with pages: to load phpbb/pages routes, we need to have the extension completely loaded and accessible through the container
 2. The first version of this MR: we correctly tell that the extension is not enabeld for the current request. But what happens is that the router build and cache an url generator without phpbb/pages routes. This generator is valid and stays around for the following requests, breaking the board everytime a phpbb/pages url is needed

Now, we have another, maybe better solution for that: implements in the router a `cache_can_be_disabled` interface and/or trait (please find me a name xD), register all implementations of this interface in a collection and use it in the cache drivers (`phpbb/cache/driver/base`) to disable the cache / dumepd cache in all relevant services (right now the router maybe the textformater too in the absolute)

iMattPro:
> Now, we have another, maybe better solution for that: implements in the router a `cache_can_be_disabled` interface and/or trait (please find me a name xD), 

`suspend_cache` ??23

marc1706 (approved):

PR #6321, opened 2021-10-29
[ticket/16902] Improve search results count - 3.3.x

rxu:
`3.3.x` version.

Checklist:

- [x] Tests pass

<a href="https://tracker.phpbb.com/browse/PHPBB3-16902">PHPBB3-16902</a>.

### DIFF
--- a/phpBB/phpbb/search/fulltext_mysql.php
+++ b/phpBB/phpbb/search/fulltext_mysql.php
@@ -569,6 +569,7 @@ class fulltext_mysql extends \phpbb\search\base
 		extract($this->phpbb_dispatcher->trigger_event('core.search_mysql_keywords_main_query_before', compact($vars)));
 
 		$sql_select			= ($type == 'posts') ? 'p.post_id' : 'DISTINCT t.topic_id';
+		$sql_select			.= $sort_by_sql[$sort_key] ? ", {$sort_by_sql[$sort_key]}" : '';
 		$sql_from			= ($join_topic) ? TOPICS_TABLE . ' t, ' : '';
 		$field				= ($type == 'posts') ? 'post_id' : 'topic_id';
 		if (count($author_ary) && $author_name)
@@ -609,11 +610,10 @@ class fulltext_mysql extends \phpbb\search\base
 		$this->db->sql_freeresult($result);
 
 		$id_ary = array_unique($id_ary);
-
 		// if the total result count is not cached yet, retrieve it from the db
 		if (!$result_count && count($id_ary))
 		{
-			$sql_found_rows = str_replace("SELECT $sql_select", "SELECT COUNT($sql_select) as result_count", $sql);
+			$sql_found_rows = str_replace("SELECT $sql_select", "SELECT COUNT(*) as result_count", $sql);
 			$result = $this->db->sql_query($sql_found_rows);
 			$result_count = (int) $this->db->sql_fetchfield('result_count');
 			$this->db->sql_freeresult($result);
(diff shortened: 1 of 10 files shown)


phpBB/phpbb/search/fulltext_mysql.php
```diff
@@ -828,12 +827,12 @@ public function author_search($type, $firstpost_only, $sort_by_sql, $sort_key, $
 		extract($this->phpbb_dispatcher->trigger_event('core.search_mysql_author_query_before', compact($vars)));
 
 		// If the cache was completely empty count the results
-		$calc_results = ($result_count) ? '' : 'SQL_CALC_FOUND_ROWS ';
+		$sql_select	= ($type == 'posts') ? 'p.post_id' : 't.topic_id';
```
Nicofuma, line 830: you could remove this variable to simplify the code (like before)

rxu: Well, it's used later for wrapping into `COUNT()`, so if I remove it here that would be the trick.

tests/test_framework/phpbb_functional_test_case.php
```diff
@@ -1433,4 +1433,112 @@ protected function get_hidden_fields($crawler, $url)
...
+	/**
+	 * Get HTML of the crawler
+	 * See https://symfony.com/doc/current/components/dom_crawler.html#component-dom-crawler-dumping
+	 *
+	 * @param Symfony\Component\DomCrawler\Crawler $crawler Crawler instance
+	 * @param string $url Request URL
+	 *
+	 * @return array Hidden form fields array
+	 */
+	protected function dump_crawler($crawler)
```
marc1706, line 1446: Is this needed?

tests/test_framework/phpbb_functional_test_case.php
```diff
@@ -1433,4 +1433,112 @@ protected function get_hidden_fields($crawler, $url)
...
+		{
+			$username_logged_in = $crawler->filter('li[id="username_logged_in"] > div > a > span')->text();
+		}
+		return $username_logged_in;
+	}
+
+	/**
+	 * Disable posting flood control
+	 */
+	protected function disable_flood_interval()
```
marc1706, line 1482: Currently this is only used in one place. Maybe replace it with a static function that can be called from other tests cases if needed at some point? Also, could you also add a function to enable the flood interval again? Ideally we should aim to try and keep side effects of tests to a minimum.

marc1706 (changes requested):

tests/test_framework/phpbb_functional_test_case.php
rxu: Was handy for debugging tests but removed.

tests/test_framework/phpbb_functional_test_case.php
rxu: Not sure about static but fixed as I could.

tests/test_framework/phpbb_functional_test_case.php
```diff
@@ -1437,4 +1437,87 @@ protected function get_hidden_fields($crawler, $url)
...
+		{
+			$this->logout();
+			$this->login($logged_in_username);
+		}
+	}
+
+	/**
+	* Check if a user exists by username(s) or user_id(s)
+	*
+	* @param array &$user_id_ary The user ids to check or empty if usernames used
```
marc1706, line 1501: Param descriptions are not in line with function

marc1706 (changes requested):
Apart from minor issue with docblocks it looks good.

tests/test_framework/phpbb_functional_test_case.php
```diff
@@ -1496,10 +1496,10 @@ protected function set_flood_interval($flood_interval)
 	}
 
 	/**
-	* Check if a user exists by username(s) or user_id(s)
+	* Check if a user exists by username or user_id
 	*
-	* @param array &$user_id_ary The user ids to check or empty if usernames used
-	* @param array &$username_ary The usernames to check or empty if user ids used
+	* @param string $username The username to check or empty if user_id used
+	* @param int $user_id The use id to check or empty if username used
```
3D-I, line 1502: `The user id to check or empty if username used` ?

marc1706: id instead of identifier is the way to go, should say user and not "use" though :)

marc1706 (approved):

PR #6324, opened 2021-11-13
[ticket/16908] Fix PHP warning on non-existent post id requests

rxu:
Checklist:

- [x] Tests pass

<a href="https://tracker.phpbb.com/browse/PHPBB3-16908">PHPBB3-16908</a>.

### DIFF
--- a/phpBB/posting.php
+++ b/phpBB/posting.php
@@ -87,24 +87,25 @@ switch ($mode)
 		$post_id = $request->variable('p', 0);
 		if ($post_id)
 		{
-			$topic_forum = array();
+			$topic_forum = [];
 
 			$sql = 'SELECT t.topic_id, t.forum_id
 				FROM ' . TOPICS_TABLE . ' t, ' . POSTS_TABLE . ' p
 				WHERE p.post_id = ' . $post_id . '
 				AND t.topic_id = p.topic_id';
 			$result = $db->sql_query($sql);
-			$topic_forum = $db->sql_fetchrow();
-			$topic_id = (int) $topic_forum['topic_id'];
-			$forum_id = (int) $topic_forum['forum_id'];
+			$topic_forum = $db->sql_fetchrow($result);
 			$db->sql_freeresult($result);
 		}
 
-		if (!$post_id || !$topic_id || !$forum_id)
+		if (!$post_id || !$topic_forum)
 		{
 			$user->setup('posting');
 			trigger_error('NO_POST');
 		}
+
+		$topic_id = (int) $topic_forum['topic_id'];
+		$forum_id = (int) $topic_forum['forum_id'];
 	break;
 }
 
--- a/tests/functional/posting_test.php
+++ b/tests/functional/posting_test.php
@@ -300,4 +300,26 @@ class phpbb_functional_posting_test extends phpbb_functional_test_case
 			$crawler->filter('#preview .content')->html()
 		);
 	}
+
+	public function nonexistent_post_id_data()
+	{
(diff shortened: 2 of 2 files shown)


phpBB/posting.php
```diff
@@ -87,24 +87,25 @@
...
 			$result = $db->sql_query($sql);
-			$topic_forum = $db->sql_fetchrow();
-			$topic_id = (int) $topic_forum['topic_id'];
-			$forum_id = (int) $topic_forum['forum_id'];
+			$topic_forum = $db->sql_fetchrow($result);
 			$db->sql_freeresult($result);
 		}
 
-		if (!$post_id || !$topic_id || !$forum_id)
+		if (!$post_id || !$topic_forum)
```
3D-I, line 101: May I suggest what I wrote to solve this ticket, which is slightly different?

rxu: Sure, what code do you have in mind?

3D-I: Well, I did delete it. Here is what I wrote now and briefly tested.

```php
	case 'edit':
	case 'delete':
	case 'quote':
	case 'soft_delete':
		$post_id = $request->variable('p', 0);

		$sql = 'SELECT post_id
			FROM ' . POSTS_TABLE . '
			WHERE post_id = ' . $post_id;
		$result = $db->sql_query_limit($sql, 1);
		$row = $db->sql_fetchrow($result);
		$db->sql_freeresult($result);

		$post_id = $row['post_id'] ?? 0;

		if ($post_id)
		{
			$topic_forum = [];

			$sql = 'SELECT t.topic_id, t.forum_id
				FROM ' . TOPICS_TABLE . ' t, ' . POSTS_TABLE . ' p
				WHERE p.post_id = ' . $post_id . '
				AND t.topic_id = p.topic_id';
			$result = $db->sql_query($sql);
			$topic_forum = $db->sql_fetchrow($result);
			$db->sql_freeresult($result);

			$topic_id = (int) $topic_forum['topic_id'];
			$forum_id = (int) $topic_forum['forum_id'];
		}
		else
		{
			$user->setup('posting');
			trigger_error('NO_POST');
		}
	break;
}
```

rxu: Well, that additional query actually duplicates the current query (the only difference is the current one joins topics table, but this doesn't change the result).
I'd think about adding a limit of 1 to the query, not sure if it brings any benefits though.

3D-I: There is no need to go ahead if the post does not exists and a simple query which asserts it does or not simpliefies the workflow. The same method should also be used above in the same swittch/cases for the forums (IMHO).

3D-I: On a side note, is this is a regression due to the same PR which took care of removing the forum queries from links?

rxu: Sorry, I'm still on the point those queries do the same thing :)

> is this is a regression due to the same PR which took care of removing the forum queries from links?

Yes, the whole code was added here
https://github.com/phpbb/phpbb/commit/adacf21edb0bdd3e8aaa649ca037586a68ccf5e2#diff-2bd852c3b9a70bfaabb479ef5b67b657056c1e74de09e433e52a644092ec28c9

3D-I: Two different points of view, indeed. :)

3D-I: Again, ther is no need to run a query which joins two tables if the post ID does not exist to begin with, you can not rely on the request var to assert it.

rxu: Post existence is checked with that query. There's no need in additional one. There's no relying on single request value.

PR #6328, opened 2021-11-21
[ticket/16913] Add Search Index Progress Bar with Stats

Dark1z:
Checklist:

- [x] Tests pass

ScreenShot:
.
1st run (On Submit) : 

.
2nd & consecutive run (On Refresh) : 

.

https://tracker.phpbb.com/browse/PHPBB3-16913

### DIFF
--- a/phpBB/language/en/acp/search.php
+++ b/phpBB/language/en/acp/search.php
@@ -125,7 +125,7 @@ $lang = array_merge($lang, array(
 	),
 	'SEARCH_INDEX_CREATED'					=> 'Successfully indexed all posts in the board database.',
 	'SEARCH_INDEX_REMOVED'					=> 'Successfully deleted the search index for this backend.',
-	'SEARCH_INDEX_PROGRESS'					=> '<br><progress value="%1$d" max="%3$d"></progress><br> %4$.2f %% <br>Done: %1$d | Pending: %2$d | Total: %3$d',
+	'SEARCH_INDEX_PROGRESS'					=> 'Done: %1$d | Pending: %2$d | Total: %3$d',
 	'SEARCH_INTERVAL'						=> 'User search flood interval',
 	'SEARCH_INTERVAL_EXPLAIN'				=> 'Number of seconds users must wait between searches. This interval is checked independently for each user.',
 	'SEARCH_STORE_RESULTS'					=> 'Search result cache length',
--- a/phpBB/language/en/acp/search.php
+++ b/phpBB/language/en/acp/search.php
@@ -52,7 +52,7 @@ $lang = array_merge($lang, array(
 	'DEFAULT_SEARCH_RETURN_CHARS'			=> 'Default number of returned characters',
 	'DEFAULT_SEARCH_RETURN_CHARS_EXPLAIN'	=> 'The default number of characters that will be returned while searching. A value of 0 will return the entire post.',
 	'DELETE_INDEX'							=> 'Delete index',
-	'DELETING_INDEX_IN_PROGRESS'			=> 'Deleting the index in progress',
+	'DELETING_INDEX_IN_PROGRESS'			=> 'Deletion of index is in progress…',
 	'DELETING_INDEX_IN_PROGRESS_EXPLAIN'	=> 'The search backend is currently cleaning its index. This can take a few minutes.',
 
 	'FULLTEXT_MYSQL_INCOMPATIBLE_DATABASE'	=> 'The MySQL fulltext backend can only be used with MySQL4 and above.',
@@ -92,7 +92,7 @@ $lang = array_merge($lang, array(
 	'GO_TO_SEARCH_INDEX'					=> 'Go to search index page',
 
 	'INDEX_STATS'							=> 'Index statistics',
-	'INDEXING_IN_PROGRESS'					=> 'Indexing in progress',
+	'INDEXING_IN_PROGRESS'					=> 'Indexing in progress…',
 	'INDEXING_IN_PROGRESS_EXPLAIN'			=> 'The search backend is currently indexing all posts on the board. This can take from a few minutes to a few hours depending on your board’s size.',
 
 	'LIMIT_SEARCH_LOAD'						=> 'Search page system load limit',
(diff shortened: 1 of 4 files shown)


rxu:
What about already existing progress bar?
https://github.com/phpbb/phpbb/blob/3.3.x/phpBB/adm/style/acp_search.html#L157

Dark1z:
> What about already existing progress bar? https://github.com/phpbb/phpbb/blob/3.3.x/phpBB/adm/style/acp_search.html#L157

ScreenShots :
1.

2.

.
Best regards 👍

rxu:
And? :) That doesn't do anything with the current progress bar so phpBB will get 2 for some reason.

phpBB/language/en/acp/search.php
```diff
@@ -125,6 +125,7 @@
 	),
 	'SEARCH_INDEX_CREATED'					=> 'Successfully indexed all posts in the board database.',
 	'SEARCH_INDEX_REMOVED'					=> 'Successfully deleted the search index for this backend.',
+	'SEARCH_INDEX_PROGRESS'					=> '<br><progress value="%1$d" max="%3$d"></progress><br> %4$.2f %% <br>Done: %1$d | Pending: %2$d | Total: %3$d',
```
rxu, line 128: It'd be better to have all HTML markup in template rather than on PHP.

Dark1z:
> And? :) That doesn't do anything with the current progress bar so phpBB will get 2 for some reason.

the one which you are mentioning can be called "Loading Bar",
which loads this template file:
https://github.com/phpbb/phpbb/blob/3.3.x/phpBB/adm/style/progress_bar.html#L30
which in turn loads this GIF:
https://github.com/phpbb/phpbb/blob/3.3.x/phpBB/adm/images/progress_bar.gif

phpBB/language/en/acp/search.php
Dark1z: can not do that in `trigger_error(...)` function.

Dark1z: do you have a way to do that ???

rxu: Well, in theory that cab be done via adding relevant code to `\adm\style\message_body.html` and assigning related template vars before `trigger_error()` call.

Dark1z: since it's a small thing,
and also the `%1$d` & `%3$d` are repetitive,
hence I added it in the lang file.

DavidIQ:
We are moving away from adding/including html in language variables so adding more of it is going to be a non-starter for this PR.

Dark1z:
> We are moving away from adding/including html in language variables so adding more of it is going to be a non-starter for this PR.

Then is it ok to move it to php file ???

rubencm:
I like the progress bar, but I would do it when the data you already have: `max_post_id` and `post_counter` to avoid add extra queries

DavidIQ:
> > We are moving away from adding/including html in language variables so adding more of it is going to be a non-starter for this PR.
> 
> Then is it ok to move it to php file ???

In general HTML should go in template files.

Dark1z:
> I like the progress bar, but I would do it when the data you already have: `max_post_id` and `post_counter` to avoid add extra queries

I did not get what you mean?
how am I going to get the `$done_count`, `$remain_count` & `$total_count` from `max_post_id` and `post_counter` ?

Dark1z:
> In general HTML should go in template files.

Yes! But that's not possible via `trigger_error(...)`,
so I have to inject similar to this:
https://github.com/phpbb/phpbb/pull/6328/files#diff-2f31352ab73095253f68c8cd43b1065cd218084bbadea4deacd66deec5d3e0f2R556-R563

3D-I:
I am not sure that this adds to or improves the current statistics, I find it at least just a matter of aesthetics not entirely necessary. Rather I would be in favour of having a slide that updates ajaxed and certainly not on page refresh (popup). Unless I am missing something?

Dark1z:
> I am not sure that this adds to or improves the current statistics, I find it at least just a matter of aesthetics not entirely necessary. Rather I would be in favour of having a bar that updates ajaxed and certainly not on page refresh (popup). Unless I am missing something? If the goal is to increase the size of the progress bar by 2x, then IMHO it is very trivial. Not to mention the HTML inclusion, which is somehow a "no-no".

Have you never done the search index on 100K post or more board ?

It's default behavior is to page refresh, I am not changing that.
I only add a progress bar and some stats.

This isn't a "trivial" on large boards.

If you have better solution to "HTML inclusion", then kindly make a PR for the same.

Best regards 👍

3D-I:
> Have you never done the search index on board with 100K or more post ?

In the last 15 years I think at least a couple (dozens) of times or more. I don't see the connection anyway. 

If a PR in my opinion is almost superfluous, I say so expressing my opinion, then I understand that it can be annoying but I don't think I should limit my thoughts.

Dark1z:
> In the last 15 years I think at least a couple (dozens) of times or more. I don't see the connection anyway.
> 
> If a PR in my opinion is almost superfluous, I say so expressing my opinion, then I understand that it can be annoying but I don't think I should limit my thoughts.


Also, I am not touching that Pop-Up loading progress bar,
IMHO, I do not know the use of that, or why is it used for, so did not touch it.

Best regards 👍

3D-I:
> IMHO, I do not know the use of that, or why is it used for, so did not touch it.

Enough said 😃

Dark1z:
> Enough said 😃

I think you are referring to this part of code : 
https://github.com/phpbb/phpbb/blob/ceca69d121fbb5b5952a9f65eb567e54030aaca7/phpBB/includes/acp/acp_search.php#L536-L554
Which generates the Pop-Up loading progress bar,
for which (again) I have not touched.

Best regards 👍

3D-I:
Again,  I find this PR not necessary. Period. Thanks.

Dark1z:
> Again, I find this PR not necessary. Period. Thanks.

But I find this PR necessary. Period. Thanks.

Dark1z:
!set wip

GanstaZ:
Don't get me wrong.. I'm not against progress bar update, but adding more of those trigger errors into modules (controllers in the future) is not good as those will be removed.

Dark1z:
> Don't get me wrong.. I'm not against progress bar update, but adding more of those trigger errors into modules (controllers in the future) is not good as those will be removed.

Is there any alternatives ?

Best regards 👍

GanstaZ:
Hard to say.. master branch compared to 3.3.x differs a lot.. In master version i removed all triggers from acp search with a solution i use in other files, but it wont work on 3.3.x.

Dark1z:
!unset WIP

Dark1z:
https://github.com/phpbb/phpbb/pull/6328#issuecomment-1036459835
> Hard to say.. master branch compared to 3.3.x differs a lot.. In master version i removed all triggers from acp search with a solution i use in other files, but it wont work on 3.3.x.

@GanstaZ ,

can I get references to those commits or PR of master version,
I currently need them for master version of this PR.

Best regards 👍

GanstaZ:
> [#6328 (comment)](https://github.com/phpbb/phpbb/pull/6328#issuecomment-1036459835)
> 
> can I get references to those commits or PR of master version, I currently need them for master version of this PR.
> 
> Best regards 👍

My PR progress is frozen because of health issues.. Best thing you can do, is to discuss your changes with phpBB team.

PR #6348, opened 2021-12-12
[ticket/16940] Optimize phpBB Native Search [master]

Dark1z:
- Use `sql_query_limit` instead of `sql_query`
- Update SQL query to reflect the above change
- Assign proper last `post_id` to `$post_counter`

> As-per previous code, between 2 post_ids there may or may not be 100/100 posts,
> sometimes or most of the times some posts might be deleted
> and hence no of post can be less than 100, eg. 80/100.
> 
> I optimized such that there will be always be 100/100 posts to index it in one `while(...)` loop.

> Example (Approx, since it's system depended):
>  
> Step is no of refresh, inside each step there are no of `while(...)` loops.
>  
> In Previous code:
> Step 1: 80 | 60 | 100 = 240 - (1 min)
> Step 2: 40 | 90 | 70 = 200 - (1 min)
> Step 3: 50 | 70 | 90 = 210 - (1 min)
>  
> PR code:
> Step 1: 100 | 100 | 100 = 300 - (1 min)
> Step 2: 100 | 100 | 100 = 300 - (1 min)
> Step 3: 50 = 50 - (few secs)
>  
> also on larger board, even no of Steps will reduce and hence also the time.
> 
> I hope you understand what I mean, this is the best explanation I can come with.

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16940

### DIFF
--- a/phpBB/phpbb/search/backend/base.php
+++ b/phpBB/phpbb/search/backend/base.php
@@ -331,7 +331,7 @@ abstract class base implements search_backend_interface
 
 		while (still_on_time() && $post_counter <= $max_post_id)
 		{
-			$rows = $this->get_posts_between($post_counter + 1, $post_counter + self::BATCH_SIZE);
+			$rows = $this->get_posts_batch_after($post_counter);
 
 			if ($this->db->sql_buffer_nested_transactions())
 			{
@@ -346,9 +346,8 @@ abstract class base implements search_backend_interface
 					$this->index('post', (int) $row['post_id'], $row['post_text'], $row['post_subject'], (int) $row['poster_id'], (int) $row['forum_id']);
 				}
 				$row_count++;
+				$post_counter = $row['post_id'];
 			}
-
-			$post_counter += self::BATCH_SIZE;
 		}
 
 		// pretend the number of posts was as big as the number of ids we indexed so far
@@ -385,7 +384,7 @@ abstract class base implements search_backend_interface
 		$row_count = 0;
 		while (still_on_time() && $post_counter <= $max_post_id)
 		{
-			$rows = $this->get_posts_between($post_counter + 1, $post_counter + self::BATCH_SIZE);
+			$rows = $this->get_posts_batch_after($post_counter);
 			$ids = $posters = $forum_ids = array();
 			foreach ($rows as $row)
 			{
@@ -400,7 +399,7 @@ abstract class base implements search_backend_interface
 				$this->index_remove($ids, $posters, $forum_ids);
 			}
 
-			$post_counter += self::BATCH_SIZE;
+			$post_counter = end($ids);
 		}
 
 		if ($post_counter <= $max_post_id)
@@ -445,19 +444,17 @@ abstract class base implements search_backend_interface
 	}
 
 	/**
-	 * Get posts between 2 ids
+	 * Get batch of posts after id
 	 *
-	 * @param int $initial_id
-	 * @param int $final_id
+	 * @param int $post_id
 	 * @return \Generator
 	 */
-	protected function get_posts_between(int $initial_id, int $final_id): \Generator
+	protected function get_posts_batch_after(int $post_id): \Generator
 	{
 		$sql = 'SELECT post_id, post_subject, post_text, poster_id, forum_id
 			FROM ' . POSTS_TABLE . '
-			WHERE post_id >= ' . $initial_id . '
-				AND post_id <= ' . $final_id;
-		$result = $this->db->sql_query($sql);
+			WHERE post_id > ' . $post_id;
+		$result = $this->db->sql_query_limit($sql, self::BATCH_SIZE);
 
 		while ($row = $this->db->sql_fetchrow($result))
 		{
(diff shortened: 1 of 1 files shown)


rubencm:
If there are a lot of posts, this will probably cause a timeout

Dark1z:
> If there are a lot of posts, this will probably cause a timeout

But in previous code if there are 100/100 post then it will also cause timeout

rubencm:
no, because it check still_on_time() function every 100 posts, with your solution, if there are a lot of posts, that function will never be called

phpBB/phpbb/search/backend/base.php
```diff
@@ -331,7 +331,7 @@ public function create_index(int &$post_counter = 0): ?array
 
 		while (still_on_time() && $post_counter <= $max_post_id)
 		{
-			$rows = $this->get_posts_between($post_counter + 1, $post_counter + self::BATCH_SIZE);
+			$rows = $this->get_posts_batch_after($post_counter);
```
rubencm, line 334: Is not +1? or I think you are indexing some posts twice

phpBB/phpbb/search/backend/base.php
```diff
@@ -445,19 +444,17 @@ protected function forum_ids_with_indexing_enabled(): array
...
 	 */
-	protected function get_posts_between(int $initial_id, int $final_id): \Generator
+	protected function get_posts_batch_after(int $post_id): \Generator
 	{
 		$sql = 'SELECT post_id, post_subject, post_text, poster_id, forum_id
 			FROM ' . POSTS_TABLE . '
-			WHERE post_id >= ' . $initial_id . '
-				AND post_id <= ' . $final_id;
-		$result = $this->db->sql_query($sql);
+			WHERE post_id > ' . $post_id;
```
rubencm, line 456: if you do it in this way, i think you have to specify the order in the query

Dark1z: in that case query builder will be advisable, Right ?

rubencm: i dont understand. what i mean is that you should make sure that posts are ordered by post_id

Dark1z: Yes ! I meant via query builder `sql_build_query(...)`

Dark1z: Never mind, I'll use order_by directly.
commit will be made soon.

CHItA: If this is an optimization could you share some numbers on how is this faster? `post_id` should be the primary key and thus an index, therefore I don't really see what the benefit here is. Although I could be just plain wrong.

Dark1z: > If this is an optimization could you share some numbers on how is this faster? `post_id` should be the primary key and thus an index, therefore I don't really see what the benefit here is. Although I could be just plain wrong.

primary key is it's self an index,
but it can not be used to search post text.

also as-per previous code, between 2 post_ids there may or may not be 100/100 posts,
sometimes or most of the times some posts might be deleted and hence no of post can be less than 100, eg. 80/100.

I optimized such that there will be always be 100/100 posts to index it in one `while(...)` loop.

CHItA: yeah, I get that part, but is that an optimization in terms of performance? Or do you mean optimizing the behaviour?

Dark1z: I am not an expert, but as-per me, it should be both.

Dark1z: Example (Approx, since it's system depended):

Step is no of refresh,
inside each step there are no of `while(...)` loops.

In Previous code:
Step 1: 80 | 60 | 100 = 240
Step 2: 40 | 90 | 70 = 200
Step 3: 50 | 70 | 90 = 210

PR code:
Step 1: 100 | 100 | 100 = 300
Step 2: 100 | 100 | 100 = 300
Step 3: 50 = 50

also on larger board, even no of Steps will reduce.

I hope you understand what I mean, this is the best explanation I can come with.

phpBB/phpbb/search/backend/base.php
```diff
@@ -400,10 +399,10 @@ public function delete_index(int &$post_counter = null): ?array
 				$this->index_remove($ids, $posters, $forum_ids);
 			}
 
-			$post_counter += self::BATCH_SIZE;
+			$post_counter = $ids[count($ids) - 1];
```
CHItA, line 402: You need to move this into the if now.

CHItA (changes requested):
Otherwise looks good.

rubencm (approved):

phpBB/phpbb/search/backend/base.php
Dark1z: Ok.
Should I also move `$row_count += count($ids);` to `if(...)` ?

CHItA: you don't have to, the point here is if `$ids` is empty, `count($ids) - 1` is `-1`, and not a valid index.

Dark1z: Ahh! ok.
will do this changes in few mins.

CHItA (approved):

PR #6351, opened 2021-12-24
added a way to run the app instantly?

fangdajiang:
Hello, we are TeamCode. We have created a Tin application for this app, which help users to quickly run and try the app without installing and configuring the environment. Users don't need to pay for this service. Hope it can help you better promote the app. Thx.

[TeamCode try-it-now](https://www.teamcode.com/tin/clone?applicationId=273460556549484544)
Guidance: https://www.teamcode.com/docs/en-US/tin/clone-tin

This is the entire usage process：

Users click the badge Run in Cloud (teamcode demo).
Sign up first if they have not logged in.
After that, they will be directed to the Clone page and start to build & run this app.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ To run an installation from the repo (and not from a pre-built package) on a loc
   cd phpBB
   php ../composer.phar install
   ```
+- Run without installation [![TeamCode try-it-now](https://static01.teamcode.com/badge/demo.svg)](https://www.teamcode.com/tin/clone?applicationId=273460556549484544)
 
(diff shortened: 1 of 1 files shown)


rubencm:
Thank you very much, but we already have an internal tool for that https://www.phpbb.com/demo/

fangdajiang:
> 

Thanks. The product we provide is aim to let user use the app separately, or customize it at will. So what we hope is our product could save user's time.

3D-I:
I would like to add a note of my own. Offering paid services in a free software seems to me to be at least inadequate, let alone wanting to include them in its documentation. 

It also seems to me to be clearly `SPAM`, if only because your page does not exist https://github.com/fangdajiang.

PR #6353, opened 2022-01-03
[ticket/16944] Use icon bundle to load Iconify locally

rxu:
Checklist:

- [ ] Tests pass

<a href="https://tracker.phpbb.com/browse/PHPBB3-16944">PHPBB3-16944</a>.


iMattPro:
I have 1 issue with this...extensions! they wouldn't be able to use any other icons except the ones bundled here.

rxu:
> I have 1 big issue with this...extensions! they wouldn't be able to use any other icons except the ones bundled here.

Well, as per the https://docs.iconify.design/icon-components/bundles/#offline
> if icon data is not available, component attempts to load data for icon from Iconify API.

No idea how does it work but this should solve the issue :) Also, extensions should be able to add their own bundles if needed.

rxu:
@cyberalien Could you please evaluate if this approach looks correct at all, if you have a time for this. Thanks.

cyberalien:
> @cyberalien Could you please evaluate if this approach looks correct at all, if you have a time for this. Thanks.
> 
> P.S. also I have no idea how to fix ESLint issues as bundle script is autogenerated by `$collection->scriptify()` method.

Sure

cyberalien:
About ESLint errors: I guess it is from `phpBB/assets/iconify/*.js`? If so, those are notices about code style. The only solution is to add that directory to `ignoreFiles` in `.stylelintrc` because those files are automatically generated and should not be linted.

Code in `create_iconify_bundle.php` looks good. This line is not needed:
```php
use Iconify\IconsJSON\Finder;
```
because you are using `loadIconifyCollection()` that automatically imports `Finder`.


Also, I see you are using FontAwesome 4 icons. Please dump them as fast as possible. FontAwesome 4 is intended to be used as icon font only, many icons are designed to be rendered outside of bounding box and they are not aligned to any logical grid. There are plenty of excellent open source icon sets, FontAwesome is not one of them. I suggest to take a look here: https://icon-sets.iconify.design and select one of good icon sets. Good sets have icons designed for specific grid, which you can see on left of icon set (number of icons, followed by icon set grid), bad ones are at the bottom of list. For phpBB I think these icon sets will fit nicely: Bootstrap Icons, Carbon, Tabler. Icon sets designed for 24x24 grid won't look good because phpBB uses small font, so it is better to use 16x16 icon sets.

iMattPro:
> Well, as per the https://docs.iconify.design/icon-components/bundles/#offline
> 
> > if icon data is not available, component attempts to load data for icon from Iconify API.

That actually seems to work!

rxu:
@cyberalien Thanks a lot. But is it ok to use a local copy of `iconify.min.js` and will in this case not bundled icons data be loaded from Iconify API?

iMattPro:
> Also, I see you are using FontAwesome 4 icons. Please dump them as fast as possible. FontAwesome 4 is intended to be used as icon font only, many icons are designed to be rendered outside of bounding box and they are not aligned to any logical grid. There are plenty of excellent open source icon sets, FontAwesome is not one of them. I suggest to take a look here: https://icon-sets.iconify.design and select one of good icon sets. Good sets have icons designed for specific grid, which you can see on left of icon set (number of icons, followed by icon set grid), bad ones are at the bottom of list. For phpBB I think these icon sets will fit nicely: Bootstrap Icons, Carbon, Tabler. Icon sets designed for 24x24 grid won't look good because phpBB uses small font, so it is better to use 16x16 icon sets.

I think Icon redesign choices should be left up to the Chameleon style designers, not this PR

cyberalien:
> @cyberalien Thanks a lot. But is it ok to use a local copy of `iconify.min.js` and will in this case not-bundled icons data be loaded from Iconify API?

Sure, it doesn't make any difference. It is even better for reliability, since everything is under control of forum host.

I see that bundle is loaded before Iconify, so everything should be fine. This guarantees that icon data is available when Iconify loads, so it won't accidentally send redundant API queries.

> I think Icon redesign choices should be left up to the Chameleon style designers, not this PR

Of course. But this issue is related, since they'll be changing icons list in `create_iconify_bundle.php` that is added in this PR, so I thought I'd mention it here.

hanakin:
Not sure what this pr is attempting to achieve here? But this pr seems to go our policies. Separation of concerns, you are defining icons in php and js that are to be used in html? Not to mention the fact that when you want to use an icon you have to define to in three places for no inherent upside?

What is the intent here?

DavidIQ:
> this pr seems to go our policies.

It doesn't. If you check the file that was added it's in the develop folder and is meant to generate the iconify files in the assets directory:

```
/**
* This file creates Iconify bundle JS file in assets/iconify directory.
* See https://docs.iconify.design/icon-components/bundles/examples/svg-framework-custom.html
* iconify/json-tools and iconify/json dev requirements should be installed for the script to work.
*/
```

iMattPro:
> What is the intent here?

The intent is to load Iconify locally.

phpBB/develop/create_iconify_bundle.php
```diff
@@ -0,0 +1,277 @@
+<?php
```
rubencm, line 1: What do you think about moving this to a symfony command?
I see all the scripts in this folder as something to be removed in the future

cyberalien:
> What is the intent here?

File `create_iconify_bundle.php` creates cache for icon data.

Why is it needed? It isn't, but it is the best optimisation option and here is why:

When SVG framework finds an icon to render, it requires icon data. If icon data is provided using this preloading method (which generates static script that is also cached in browser, so it is loaded only once), icon data is available immediately and SVG framework renders icons instantly.

So what happens if icon data is not available? SVG framework loads icon data from API. That is not instant, though usually should be just a fraction of a second. API is hosted on multiple servers in different regions, uses Route53 latency routing to make sure visitor is always redirected to closest server with minimal latency, there is also redundancy feature to connect to backup API in case main one fails. However even that might sometimes fail, like if user tries to run phpBB on local network without internet access. So providing bundle for data for icons used by UI is much better solution than relying on third party servers.

Alternatives? Actually there is one great alternative: UnoCSS: https://unocss.antfu.me (GitHub repo: https://github.com/antfu/unocss) It is like Tailwind CSS, but supports syntax from Tailwind, Windi, Twind and best of all, offers all Iconify icons. If you will use UnoCSS in UI development, no need for SVG framework or bundles.

hanakin:
Thanks for all the clarification. So as I understand this the intent is to run the php file on install to build the js file that can be used as a fallback include for icons?

The issue i really have is the hardcoding in the php file of the nessecary icons? This is a front end choice and not a backend.

@cyberalien if this is what you were refereeing to not being nessecary and instead being more dynamic I would say it needs to be dynamic

rxu:
@hanakin The PHP file is development script and is not used in normal phpBB operation but only on development stage to produce Iconify bundle.
As for the Iconify bundle itself, it will be used first and Iconify API will be used as a fallback for icons not defined in a bundle at the moment.

hanakin:
I understand that. The issues are still valid. As 1. Not all icons will be generated via iconify currently not to say they can not be added but thats a different topic. 2. Not all icons are converted over yet. We still have to rework the topic icons, smiles, and ranks systems. This should then need to account for these. No matter if its intended for development or not in order to be able to maintain it. You are adding a tie between the front-end and the backend which regardless of its use is not something we do. A better approach would be to externalize the icon list to a json file that can be auto generated from the markup or some other way.

rxu:
Both issues can be invalidated by the mentioned above Iconify feature:

> Well, as per the https://docs.iconify.design/icon-components/bundles/#offline
> 
> > if icon data is not available, component attempts to load data for icon from Iconify API.
>

iMattPro:
> That makes no sense as an argument. Your essentially saying do not maintain it? Just externalize it into a json file located in the icons folder maybe.

I don't think that what he's saying at all. He's saying that, all the icons currently used in phpBB are being listed in that PHP script so they can be bundled into a local copy of Iconify. Any requested icons that are not found in the local copy will be pulled down from Iconify online.

The point is so phpBB does not need to be online or dependent on an external site for all its icons. Which might actually come in handy when you think about it for users who may be in some of those countries that tend to block content.

marc1706:
I think the general idea behind this PR makes sense. To be honest, I'd rather package too many icons than forcing people to reload icons from other sites by default. And yes, maybe this could be done also on a site level so that phpBB regenerates these icons after installing extensions, styles, or similar changes.

hanakin:
What i am saying is that its better to just include the svgs for fallback for any icons we are using. As the icon function can fallback if iconify is not available or disabled. You can also easily change it in the template. We are not using any of the more advanced features of iconify to my knowledge that requires it.

3D-I:
I agree with this PR, at the present time it is the best proposal there can be in this respect. Postponing the release of phpBB 4 seems pointless to me, also considering the very scarce injection of news in the style sector. My opinion is always to leave the much-acclaimed new style features to phpBB 4.1, also because I doubt that rewriting *CPs is feasible anytime soon, this applies to all icons such as topic icons, smiles, and ranks. 

Thanks @rxu for your contribution, I'm very sorry to see it closed but I understand.

PR #6361, opened 2022-01-30
[ticket/16958] Add exception handler for handling uncaught exceptions

marc1706:
This should improve the user experience a bit when it comes to exceptions being thrown while using phpBB. In addition to certainly having a nicer touch to them visually, they should prevent users of phpBB from ending up at a white page (error 500 style white page).

Checklist:

- [x] Tests pass

ToDo:
- [x] Check behavior on normal requests
- [x] Check behavior on AJAX requests
- [x] ~Check potential use in installer~

Screenshots:
- Debug disabled: image
- Debug enabled: image
- Debug enabled and display of arguments enabled in php.ini: 
image

https://tracker.phpbb.com/browse/PHPBB3-16958


marc1706:
I've marked this as WIP for now as I still want to test the behavior on other pages.

marc1706:
Working on proper errors during ajax requests as well:
image

3D-I:
> Working on proper errors during ajax requests as well

Also with debug enabled?

marc1706:
> > Working on proper errors during ajax requests as well
> 
> Also with debug enabled?

In ajax requests I have opted to not add a backtrace.

Derky (approved):

Derky (changes requested):
It's looking good, but I've some testing results:

1. SQL errors are HTML escaped twice in the gray box:
double_escaped_sql_error

2. When using an incorrect database password I get a white page instead of an error message in this styling. 

3. If a PHP error occurs during an AJAX request (for example a variable doesn't exist on an object) the error is hidden with a "something went wrong" in the AJAX pop-up. But you can no longer see the actual error message if you open up the network tab to see the AJAX response. The original error should also be returned from this endpoint and ideally be shown in the pop-up as well.
Original behaviour: (able to find the actual error message) 
original_behaviour_ajax_backend_error
Current (this PR) behaviour: (impossible to find the actual error message) 
this_pr_behaviour_ajax_backend_error png

rubencm:
The exceptions that are thrown from phpbb (runtime_exception for example), are multilanguage, since the message is a language variable and they have an additional constructor parameter for the language parameters. Maybe you could try translating them if the exception is an instance of runtime_exception or implements exception_interface.

One more thing, as I report here https://tracker.phpbb.com/browse/PHPBB3-16198 the component we are using to handle errors is discontinued, I don't know if it is worth developing on top of it, or better update it first

phpBB/phpbb/debug/debug.php
```diff
@@ -13,9 +13,12 @@
 
 namespace phpbb\debug;
 
-use Symfony\Component\Debug\BufferingLogger;
```
rubencm, line 16: Then you can remove `symfony/debug` in composer.json

rubencm:
Update `symfony/error-component` to `^6.3` (https://github.com/phpbb/phpbb/pull/6528)

PR #6368, opened 2022-02-20
[ticket/15028] Rename automatic to advanced update

marc1706:
Also made it clear that it's only for expert users.

PHPBB3-15028

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-15028

### DIFF
--- a/phpBB/docs/INSTALL.html
+++ b/phpBB/docs/INSTALL.html
@@ -303,7 +303,7 @@
 
 	<p>This update method should only be used for installations with modifications to core phpBB files. If you simply use Extensions or custom Styles and have not modified core files, please use the Full Package update.</p>
 
-	<p>This package detects changed files and merges in changes if needed. Since this type of update has a potential to cause issues while upgrading, it is not recommended being used for updates and/or upgrades.</p>
+	<p>This package detects changed files and merges in changes if needed. Since this type of update has a potential to cause issues while upgrading, it should only be used by expert users.</p>
 
 	<p>The automatic update package will update the board from a given version to the latest version. A number of automatic update files are available, and you should choose the one that corresponds to the version of the board that you are currently running. For example, if your current version is <strong>3.3.0</strong>, you need the <code>phpBB-3.3.0_to_3.3.1.zip/tar.bz2</code> file.</p>
 
--- a/phpBB/docs/INSTALL.html
+++ b/phpBB/docs/INSTALL.html
@@ -291,7 +291,7 @@
 
 	<p>The patch file package is for those wanting to update through the patch application, and should only be used by those who are comfortable with it.</p>
 
-	<p>The patch file is one solution for those with changes in to the phpBB core files and do not want to re-add them back to all the changed files. To use this you will need command line access to a standard UNIX type <strong>patch</strong> application. If you do not have access to such an application, but still want to use this update approach, we strongly recommend the <a href="#update_advanced">Advanced update package</a> explained below. It is also the recommended update method.</p>
+	<p>The patch file is one solution for those with changes in to the phpBB core files and do not want to re-add them back to all the changed files. To use this you will need command line access to a standard UNIX type <strong>patch</strong> application. If you do not have access to such an application, but still want to use this update approach, we strongly recommend the <a href="#update_full">Full package update</a> explained above. It is also the recommended update method.</p>
 
 	<p>A number of patch files are provided to allow you to update from previous stable releases. Select the correct patch, e.g. if your current version is <strong>3.3.0</strong>, you need the <code>phpBB-3.3.1-patch.zip/tar.bz2</code> file. Place the correct patch in the parent directory containing the phpBB core files (i.e. index.php, viewforum.php, etc.). With this done you should run the following command: <code>patch -cl -d [PHPBB DIRECTORY] -p1 &lt; [PATCH NAME]</code> (where PHPBB DIRECTORY is the directory name your phpBB Installation resides in, for example phpBB, and where PATCH NAME is the relevant filename of the selected patch file). This should complete quickly, hopefully without any HUNK FAILED comments.</p>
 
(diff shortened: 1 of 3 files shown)


phpBB/docs/INSTALL.html
```diff
@@ -291,17 +291,19 @@ <h4>Advanced settings</h4>
...
 
-<a name="update_auto"></a><h3>4.iv. Automatic update package</h3>
+<a name="update_advanced"></a><h3>4.iv. Advanced update package (Expert users)</h3>
 
-	<p>This update method is only recommended for installations with modifications to core phpBB files. This package detects changed files automatically and merges in changes if needed.</p>
+	<p>This update method should only be used for installations with modifications to core phpBB files. If you simply use Extensions or custom Styles and have not modified core files, please use the Full Package update.</p>
+
+	<p>This package detects changed files and merges in changes if needed. Since this type of update has a potential to cause issues while upgrading, it is not recommended being used for updates and/or upgrades.</p>
```
3D-I, line 306: `Since this type of update has a potential to cause issues while upgrading, it is not recommended being used for updates and/or upgrades.`

This seems a bit convoluted to me.

david63: I was about to say the same 

"it is not recommended being used for updates and/or upgrades." then what is it to be used for?

marc1706: It is not recommended to be used, instead the full package update should be used.

3D-I: I would suggest to re-phrase what I quoted above, in the whole.

3D-I: Besides, who is going to explain it to all those who support phpBB in the support forums, many of them are still tied to old versions of phpBB. It's just confusing, it's common knowledge that it shouldn't be used, it's been explained and all `ad nauseam`. I don't find it important.

rxu: Probably also to describe that it is not a replacement for the full package updates/upgrades and can be used if:
- full package update/upgrade can not be used for some reasons, &&
- user has advanced coding skills and phpBB codebase knowledge, &&
- no support will be provided for such kind of updates/upgrades (optionally).

3D-I: I have never had issue with it, in any case I do not see the urge to put this on the 3.3.6 prep-release, IMHO.

3D-I: The best solution would be to remove it completely, it should have been done with the advent of phpBB 3.1.0

marc1706: It will be removed at some point in the future but for now this change should make it more clear that it's only to be used in the rarest of cases.

david63: The last sentence of this change still does not make sense

3D-I: The problem is that it does not make it clearer but more convoluted. As above. If then the criticism is not constructive then I give up, the strong powers always win in the non-ideal world. 😴

3D-I: Maybe:
`Since this type of update has a potential to cause issues, it is not recommended being used on production sites.`

rxu: Now that sounds like this kind of update is deprecated and will be removed (f.e. in phpBB 4.x), then probably just to point this out in the description.

3D-I: `Since this type of update is only for expert users/coders, it is not recommended being used on production sites`

GanstaZ: I do agree about wording, but let's be honest, those people who shouldn't use so called automatic update don't read descriptions anyway. Proof for that is support forum. Issue will be solved, when this update method will be removed for good.

3D-I: ^^ I do agree. This PR makes no sense at this very point in time. Too late.

I would avoid this and remove that method in the next 3.3.7.

marc1706: I have now rephrased this sentence a bit.

3D-I: I personally have thought a lot about this PR. I confirm that it does not make sense in my opinion and I would avoid it. In the end it doesn't matter how we want to rephrase the description, regardless of the fact that it was convoluted from the very beginning. So I don't feel like approving the changes.

DavidIQ (approved):

phpBB/docs/INSTALL.html
```diff
@@ -291,17 +291,19 @@ <h4>Advanced settings</h4>
...
 
-<a name="update_auto"></a><h3>4.iv. Automatic update package</h3>
+<a name="update_advanced"></a><h3>4.iv. Advanced update package (Expert users)</h3>
 
-	<p>This update method is only recommended for installations with modifications to core phpBB files. This package detects changed files automatically and merges in changes if needed.</p>
+	<p>This update method should only be used for installations with modifications to core phpBB files. If you simply use Extensions or custom Styles and have not modified core files, please use the Full Package update.</p>
+
+	<p>This package detects changed files and merges in changes if needed. Since this type of update has a potential to cause issues while upgrading, it should only be used by expert users.</p>
 
 	<p>The automatic update package will update the board from a given version to the latest version. A number of automatic update files are available, and you should choose the one that corresponds to the version of the board that you are currently running. For example, if your current version is <strong>3.3.0</strong>, you need the <code>phpBB-3.3.0_to_3.3.1.zip/tar.bz2</code> file.</p>
```
3D-I, line 308: `automatic update` still is present in this paragraph, 2 times, BTW.

phpBB/docs/INSTALL.html
```diff
@@ -291,17 +291,19 @@ <h4>Advanced settings</h4>
...
```
3D-I, line 309: `automatic update` still is present in this paragraph, 2 times, BTW.

Derky (approved):

PR #6371, opened 2022-03-07
[ticket/19969] Hide flash status when post settings disallow [FLASH] …

nomind60s:
[ticket/19969] Hide flash status when post settings disallow FLASH BBCode

Add check for S_BBCODE_FLASH around FLASH_STATUS

PHPBB3-16969

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16969

### DIFF
--- a/phpBB/styles/prosilver/template/posting_editor.html
+++ b/phpBB/styles/prosilver/template/posting_editor.html
@@ -65,7 +65,9 @@
 			{BBCODE_STATUS}<br />
 			<!-- IF S_BBCODE_ALLOWED -->
 				{IMG_STATUS}<br />
+				<!-- IF S_BBCODE_FLASH -->
 				{FLASH_STATUS}<br />
+				<!-- ENDIF -->
 				{URL_STATUS}<br />
 			<!-- ENDIF -->
 			{SMILIES_STATUS}


lionel-rowe:
Shouldn't flash support just be be phased out entirely? Given the long-standing security issues and the fact that it doesn't even run in browsers any more

nomind60s:
> Shouldn't flash support just be be phased out entirely, even in 3.3.x? Given the long-standing security issues and the fact that it doesn't even run in browsers any more

https://tracker.phpbb.com/browse/PHPBB3-16574 is removing it entirely in 4.0. This is a stopgap solution, per the tracker ticket associated with the commit.

lionel-rowe:
> This is a stopgap solution, per the tracker ticket associated with the commit.

In that case, on the principle of "don't let the perfect be the enemy of the good", I'm all for it.

Style-wise I believe all new PRs should use [twig syntax](https://www.phpbb.com/community/viewtopic.php?f=461&t=2424606), though:

```diff
-				<!-- IF S_BBCODE_FLASH -->
-				{FLASH_STATUS}<br />
-				<!-- ENDIF -->
+				{% if S_BBCODE_FLASH %}
+					{{ FLASH_STATUS }}
+					<br>
+				{% endif %}
```

3D-I:
Apart from the fact that when you modify a template it is also a due act to bring it to the current higher technical levels, such as Twig, HTML5 etc. . This PR makes no logical sense to me, in fact I voted against it anyway. If we were to make this indication invisible then we would have to do it with all the others, in case they all disappear there would still be an empty space in the layout... which is not a nice thing to see. Anyway it doesn't add anything new but it creates discrepancy. Perfectly useless, if I can say.

lionel-rowe:
> @3D-I Apart from the fact that when you modify a template it is also a due act to bring it to the current higher technical levels, such as Twig, HTML5 etc. . This PR makes no logical sense to me, in fact I voted against it anyway. If we were to make this indication invisible then we would have to do it with all the others, in case they all disappear there would still be an empty space in the layout... which is not a nice thing to see. Anyway it doesn't add anything new but it creates discrepancy. Perfectly useless, if I can say.

The other options are still meaningful, though. "[img] is disabled" makes sense, because a user might reasonably expect `[img]` to be enabled; whereas "[flash] is disabled" is like saying "sorry, this site doesn't support Netscape Navigator 9.0". Who cares? _No_ modern website supports Netscape Navigator 9.0!

3D-I:
> The other options are still meaningful, though. "[img] is disabled" makes sense, because a user might reasonably expect `[img]` to be enabled; whereas "[flash] is disabled" is like saying "sorry, this site doesn't support Netscape Navigator 9.0". Who cares? _No_ modern website supports Netscape Navigator 9.0!

Still 3.3 supports Flash with BBCodes, Flash is being removed totally in 4.0. If you have a good replacement for flash in your browser you still can play with it.

lionel-rowe:
> Still 3.3 supports Flash with BBCodes, Flash is being removed totally in 4.0. If you have a good replacement for flash in your browser you still can play with it.

Flash is dead. Adobe dropped support for it at the end of 2020. Zero modern browsers support it; even IE11 doesn't support it any more. Even if an admin of a phpBB board enables it and a user tries to use it, the BBCode will display nothing more than a blank image saying "Adobe Flash Player is no longer supported".

That's why my number 1 choice would be dropping support entirely in 3.3.x, but again... a quick-fix is better than nothing. It's better than continuing to display obsolete and confusing information to users just because the immediate solution wasn't 100% perfect.

3D-I:
Again, if you have a substitute for Flash in your browser you can play with it.
https://www.techpout.com/flash-player-alternatives/
And I perfectly know the history of adobe etc, we all know that... indeed. 😃 
I tried to convince the developers to remove flash completely in 3.3.x but failed.

lionel-rowe:
> @3D-I I tried to convince the developers to remove flash completely in 3.3.x but failed

Again I'd say... "don't let the perfect be the enemy of the good". Why vote down a better-than-nothing solution if the perfect one has already been rejected?

I don't think I have anything more constructive to add here, so unsubscribing... I've made my vote clear 😉

nomind60s:
From my perspective it is about consistency. If Flash is disabled in ACP then it shouldn't exist within the interface to the members. As it stands now, if you disable Flash in ACP, when posting the Flash button is hidden, so why shouldn't it also be in legend? That's inconsistent to me, hence the fix proposed. If a board has Flash enabled in ACP, then the posting button for Flash still appears, as does the legend item. 

The correct solution is coming with the full removal in 4.0, which is still a long ways off.

hanakin:
As @DavidIQ suggested in the tracker. Rework the list and language vars to all only list things that are enabled! Then be done with all this arguing about nothing 😉

3D-I:
> Then be done with all this arguing about nothing 😉
Free discussion is a right, not to mention the fact that there are probably people who know a lot about this code from its inception and know both the front- and back-end very well. 😉

hanakin:
Discussion is moderated and meant to be on-topic so not entirely free! Please watch the tone! I have also been around since the front-ends inception and am 💯 on it (having installed and customized my first forum in college in 2001-02 time frame). Not sure who you think you are and are talking too. But suggest taking a step back. This is not a place for snide comments. Keep responses to the topic!

3D-I:
To say indirectly that the discussion is about nothing seems to me to be a bit too pressupon, like your way of putting it. Instead of bragging, prove the facts and then you can ask me who you think I am. Fly low.

marc1706:
It's certainly not super consistent but I think it's ok to not prominently show that fact that flash is not available (which it should be for pretty much everyone, hence nobody needs to see this).

PR #6376, opened 2022-04-01
[ticket/16977] Fix cron-job img tag layout and accessibility

lionel-rowe:
The cron-job image, as set in `viewforum.php` and `controller/helper.php`:

```php
$template->assign_var('RUN_CRON_TASK', '<img src="' . $url . '" width="1" height="1" alt="cron" />');
```

Renders like this:

```html
<img src="/forum/app.php/cron/cron.task.text_reparser.poll_option" width="1" height="1" alt="cron">
```

2 issues:

1. CSS `position` is not set, defaulting to `static`, therefore still affects layout (breaks some styles)
2. For some reason, `alt` is set to `"cron"`, when it should be `""`; `alt` attribute is supposed to be an accessible alternative representation to the user agent, so should be set to empty string if the element is not supposed to be visible

---

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16977


3D-I:
> CSS position is not set, defaulting to static, therefore still affects layout (breaks some styles)

Do you have some reference (link) to those styles?

> For some reason, alt is set to "cron", when it should be ""; alt attribute is supposed to be an accessible alternative representation to the user agent, so should be set to empty string if the element is not supposed to be visible

It is a good habit to completely remove any not used HTML tag, instead... in case.

phpBB/phpbb/controller/helper.php
```diff
@@ -364,7 +364,7 @@ protected function set_cron_task()
 			if ($task)
 			{
 				$url = $task->get_url();
-				$this->template->assign_var('RUN_CRON_TASK', '<img src="' . $url . '" width="1" height="1" alt="cron" />');
+				$this->template->assign_var('RUN_CRON_TASK', '<img src="' . $url . '" width="1" height="1" alt="" style="position: fixed; top: -10000px" />');
```
3D-I, line 367: If you are bound to modify this you shold know that the inline styling  is to be avoided nowadays, not to mention that the IMG tag does not not need a closing ` />` but `>`

lionel-rowe:
> Do you have some reference (link) to those styles?

[ProLight](https://www.phpbb.com/customise/db/style/prolight/) is one of them. But the issue is with core, not the style — invisible elements should never affect layout.

> The alt attribute holds a text description of the image, which isn't mandatory but is incredibly useful for accessibility — screen readers read this description out to their users so they know what the image means. Alt text is also displayed on the page if the image can't be loaded for some reason: for example, network errors, content blocking, or linkrot.

Yes, _setting_ the `alt` image is important; but it should only be set to a _non-empty_ string when the image has content that's meaningful to the user. A lot of ink has already been spilled on this topic — see e.g. any of the following:
* https://osric.com/chris/accidental-developer/2012/01/when-should-alt-text-be-blank/
* https://accessibility.oit.ncsu.edu/it-accessibility-at-nc-state/developers/accessibility-handbook/alternative-text/#2
* https://www.w3.org/WAI/tutorials/images/decorative/
* https://stackoverflow.com/questions/23526791/when-is-it-acceptable-to-leave-the-alt-text-empty-on-an-html-image-if-ever

3D-I:
The problem is where the image is 1px Xx 1px, which has been made for a good reason. Invisible.
The problem lies probably in that style... IMHO.

HTML attributes _(if empty)_ can not be in the core. Like `alt=""`. BTW.

rubencm:
I think this was a hack to run the cron jobs asynchronously so it doesn't interfere with page load.
Would there be a problem if the image was removed and simply changed with a `fetch` call from javascript?

lionel-rowe:
> HTML attributes (if empty) can not be in the core. Like alt="". BTW.

In that case, the style rules for core are directly violating W3C WAI accessibility criteria in this specific case.

Or do you mean the HTML should be `` instead of ``? If so, happy to make that change, as the two are semantically equivalent.

lionel-rowe:
> I think this was a hack to run the cron jobs asynchronously so it doesn't interfere with page load.
Would there be a problem if the image was removed and simply changed with a fetch call from javascript?

Yeah, that's what I figured. Don't see why that wouldn't do the trick, other than that `fetch` isn't supported in IE11, and IIRC phpBB 3.3 officially supports as low as IE9 (🤯🤯🤯).

phpBB/phpbb/controller/helper.php
lionel-rowe: > you shold know that the inline styling is to be avoided nowadays

In general, yes; I left it inline because it seems all external CSS files are only loaded for the user's currently selected style, e.g. ProSilver, whereas this should be set for all styles. Any suggestions for where it should go?

> IMG tag does not not need a closing `/>` but `>`

I left it as-is for consistency with elsewhere in core, but I'll change if we're moving away from that style 👍

3D-I: > I left it as-is for consistency with elsewhere in core, but I'll change if we're moving away from that style 👍

Right or wrong, when changing a line of code that belongs to the core, as it is old code, it is a good habit to improve it.

3D-I:
> I think this was a hack to run the cron jobs asynchronously so it doesn't interfere with page load. Would there be a problem if the image was removed and simply changed with a `fetch` call from javascript?

What if someone disabled JS?

3D-I:
> > HTML attributes (if empty) can not be in the core. Like alt="". BTW.
> 
> In that case, the style rules for core are directly violating W3C WAI accessibility criteria in this specific case.
> 
> Or do you mean the HTML should be `` instead of ``? If so, happy to make that change, as the two are semantically equivalent.

It should do, I'll leave the answer to the Devs though.

phpBB/phpbb/controller/helper.php
lionel-rowe: I notice all the HTML examples in https://area51.phpbb.com/docs/33x/coding-guidelines.html still use trailing slashes for void elements (`<br />` etc), and also that https://www.phpbb.com/community/ still proudly claims "Valid XHTML 1.0 Strict" in its footer, which couldn't possibly be true if there were void elements without the closing slash. Are there some more comprehensive and up-to-date style guidelines somewhere that clarify this stuff?

DavidIQ: It is valid: https://www.w3.org/TR/xhtml1/#h-4.6

Where do you get that it isn't?

3D-I: Nope, there are not. All of that is very old code and guidelines.

lionel-rowe: > It is valid: https://www.w3.org/TR/xhtml1/#h-4.6
> 
> Where do you get that it isn't?

Yes, _with_ slash is compliant for XHTML; _without_ slash isn't. But as 3D-I says, XHTML is obsolete anyway these days, and compliance it was never necessary even to run in old browsers. Modern HTML5 prefers _without_ slashes (but as before, using the with-slash style will still run fine in any modern browser).

HTML is very forgiving... until you forget to properly escape some user input, that is 😉

lionel-rowe:
> > I think this was a hack to run the cron jobs asynchronously so it doesn't interfere with page load. Would there be a problem if the image was removed and simply changed with a `fetch` call from javascript?
> 
> What if someone disabled JS?

Upon testing, the worst case scenario here is that the cron job doesn't run on that specific page load; it'll still run the next time a user _not_ using IE or disabling JS visits. In other words, the worst case is as if (for the purpose of the cron job) that individual site visit hadn't taken place.

Is this a worthwhile tradeoff for the reduced hackiness? My gut says yes, but I'd like to get others' thoughts.

There's also the option of `XMLHttpRequest`, which is IE compatible but would still require JS being enabled.

hanakin:
Not true our guidelines have been updated https://area51.phpbb.com/docs/dev/master/development/html/guidance.html

No slashes and " are required. We follow idiomatic standards for html.

We also no longer support IE so not sure thats an issue. As 3di pointed out all html should be avoided in core.

Wether the cron not always running is acceptable, I will let others weigh in on. But my vote is on a js solution.

3D-I:
@hanakin this is a special case. That cron should always running. And JS is not acceptable.

SiteSplat:
Yup agree with @3D-I on this. 

> @hanakin this is a special case. That cron should always be running (ready to). And JS is not an acceptable solution, IMHO.
> 
> And I pointed out that _INLINE STYLING_ should be avoided in core in this case, please read the proposed code.

Yup agree with @3D-I on this. Please pursue a non JS solution.

lionel-rowe:
OK, next question: I agree _in principle_ that all HTML should be kept in templates, so an ideal solution would look something like this*:

`phpBB/viewforum.php` and `phpBB/controller/helper.php`:
```php
$template->assign_var('CRON_TASK_URL', $url);
```

`phpBB/styles/prosilver/template/overall_footer.html`:
```twig
{% if not S_IS_BOT %}
	{#
		Runs the cron task by triggering server request to the URL on load.
		`img` is preferred over JS to ensure maximum compatibility.
		`sr-only` hides visually, `aria-hidden` hides from screen-readers;
		`hidden` or `display: none` would prevent the task from running.
	#}
	<img src="{CRON_TASK_URL | e('html_attr')}" alt="" width="1" height="1" class="sr-only" aria-hidden="true">
{% endif %}
```

_However_, this would leave `RUN_CRON_TASK` undefined, so any style that used a different `overall_footer.html` that expected `RUN_CRON_TASK` but not `CRON_TASK_URL` would fail to run any cron tasks at all!

This seems like a much more far-reaching consequence than the lack of support for IE or people turning off JS, both of which are tiny demographics.

That leaves 3 alternatives that I can think of:

1. Assign _both_ `CRON_TASK_URL` (for use by new styles) _and_ `RUN_CRON_TASK` (for legacy use).
2. Continue to assign only `RUN_CRON_TASK`, with the old HTML-in-PHP solution, leaving a comment in the code to explain why this is necessary in this case.
3. Add a partial `cron.html.twig` under `phpBB/styles/all/template/` (or somewhere else?) that contains the `img` tag alone, then assigning the _output_ of that template to `RUN_CRON_TASK`. Perhaps the rendering logic of this could even be done inside `phpBB/phpbb/cron/task/wrapper.php`, so that it'd be easily accessible from both `phpBB/viewforum.php` and `phpBB/controller/helper.php`; if not, I'm not really sure where it should go.

My vote goes to #3, as it seems the most "correct" while also being robust. Thoughts?

---

\* I kept `alt=""` instead of simply `alt` as that's the style used in [an example in the style guide](https://area51.phpbb.com/docs/dev/master/development/html/guidance.html#format) @hanakin linked to.

3D-I:
Just a note (again) _the alt text is also displayed on the page if the image can't be loaded for some reason: for example, network errors, content blocking, or linkrot._

lionel-rowe:
> Just a note (again) _the alt text is also displayed on the page if the image can't be loaded for some reason: for example, network errors, content blocking, or linkrot._

You've mentioned this before, but I'm not sure what its significance is? The tag is styled invisibly, so it won't be visible either way; setting `alt` to empty string simply makes the semantics/accessibility characteristics consistent with the visual styling.

In no case should an `img` tag that's simply used to make a request to a server endpoint have a contentful `alt` attribute; it's specifically intended to be invisible to the user agent (just as it would be if a `fetch` request was substituted).

3D-I:
It is very know that you can see the word "cron" in case of issues.

lionel-rowe:
> It is very known that you can see the word "**cron**" in case of issues. The tag is not styled invisibly, the IMG is 1px X 1px which makes it invisible to the human eye.

Displaying the text "cron" via `alt` attribute in the site footer is a really bad method of error reporting, because
1. It's confusing ("wtf is a cron?");
2. It doesn't mention anything about it being an error ("huh, I guess they just decided to display a random word there");
3. It's not localized ("cron 到底什么鬼？"); and
4. It's inaccessible, because it abuses the `alt` attribute for something it was never intended for. `alt` displays to screen readers whether or not the image has loaded correctly.

If we want to add user-visible error reporting, the proper way would be via `onerror` or `addEventListener('error', ...)`, along with a `phpbb.alert(localizedTitle, localizedMessage)`, then adding the relevant translations. However, it's not clear to me that this kind of thing should _ever_ be visible to ordinary forum users, because it's not their concern; rather, it should be handled in the back end, perhaps by logging when a request is made that renders the cron tag but fails to hit the corresponding endpoint within a couple of seconds.

I'd humbly suggest that the question of error handling stays out of scope for this particular PR, which related to styling and accessibility. Perhaps someone else can create another PR for the logging issue if that isn't already handled satisfactorily.

DavidIQ:
Agreed. The `alt` tag is for accessibility and should not be used for the cron job image. I wonder if the (eventual?) solution should to just make the cron task serve a blank image when used in this way.

3D-I:
The display of the raw text of the ALT attribute is perfectly fine in those situations as per 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Img

That's it by design btw, IMHO. In this special case.

DavidIQ:
> The display of the raw text of the ALT attribute is perfectly fine in those situations as per https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Img
> 
> That's it by design btw, IMHO. In this special case.

Perfect example of why to leave it blank. From that page:

> The `alt` attribute holds a text description of the image, which isn't mandatory but is incredibly useful for accessibility — screen readers read this description out to their users so they know what the image means.

If you're using a text reader and it reads "cron" as part of the page it would be very weird. If it's left blank it would not adversely affect anything so there should be no reason to not make the change.

3D-I:
So, let's go to the origin of it and blame it. Let's see what the devs will say.
 In any case it can not be left blank, unused code.

DavidIQ:
> it can not be left blank

This is patently false. It can be left blank. I'll quote from the link you provided:

> Setting this attribute to an empty string (alt="") indicates that this image is not a key part of the content (it's decoration or a tracking pixel), and that non-visual browsers may omit it from [rendering](https://developer.mozilla.org/en-US/docs/Glossary/Rendering_engine).

3D-I:
That's not a part of the guidelines though, I recall also extensions being denied for such a "mistake".
You guys should really make peace with yourselves, not that one day one says something and then it turns out not to be true. Small parenthesis.

lionel-rowe:
In the rare event that there _is_ something in the phpBB style rules or conventions that directly contradicts W3C's Web Content Accessibility Guidelines, I'd suggest that it'll generally be phpBB that needs to change and not the WCAG... phpBB's great and all, but it's not as authoritative as the standards organization for the web itself 😉

DavidIQ (approved):

3D-I:
> In the rare event that there _is_ something in the phpBB style rules or conventions that directly contradicts W3C's Web Content Accessibility Guidelines, I'd suggest that it'll generally be phpBB that needs to change and not the WCAG... phpBB's great and all, but it's not as authoritative as the standards organization for the web itself 😉

There is a front-end developer team leader in charge, on the graphs.

lionel-rowe:
Changes made based on what I outlined in option #‍3 of https://github.com/phpbb/phpbb/pull/6376#issuecomment-1086654462

3D-I:
@lionel-rowe I love your last commit. 💯 I have no time to review it now, I saw something though that can be improved, anyway... I will post here my thoughts maybe tonight or tomorrow.

3D-I:
Given the evolution of the discussion and the code of the above PR I have removed my donwvote.

phpBB/phpbb/cron/task/wrapper.php
```diff
@@ -105,6 +105,27 @@ public function get_url()
...
 
+	/**
+	 * Returns HTML for an invisible `img` tag that can be displayed on page
+	 * load to trigger a request to the relevant cron task endpoint.
+	 *
+	 * @return string       HTML to render to trigger cron task
+	 */
+	public function get_html_tag()
+	{
+		global $template;
```
3D-I, line 116: This should be removed. No globals.

phpBB/styles/all/template/cron.html.twig
```diff
@@ -0,0 +1,7 @@
+{#
+	Runs the cron task by triggering server request to the URL on load. `img` is
+	preferred over JS to ensure maximum compatibility. We use `class="sr-only"`
+	to hide visually and `aria-hidden="true"` to hide from screen-readers; using
+	`hidden` or `display: none` would prevent the task from running.
+#}
+<img class="sr-only" aria-hidden="true" src="{CRON_TASK_URL|e('html_attr')}" width="1" height="1" alt="">
```
3D-I, line 7: "{{ CRON_TASK_URL|e('html_attr') })"

phpBB/styles/prosilver/template/overall_footer.html
```diff
@@ -60,7 +60,7 @@ <h3 class="alert_title">&nbsp;</h3><p class="alert_text"></p>
 
 <div>
 	<a id="bottom" class="anchor" accesskey="z"></a>
-	<!-- IF not S_IS_BOT -->{RUN_CRON_TASK}<!-- ENDIF -->
+	{% if not S_IS_BOT %}{RUN_CRON_TASK}{% endif %}
```
3D-I, line 63: `{{ RUN_CRON_TASK }}` ??

phpBB/styles/all/template/cron.html.twig
3D-I: Please remove the `.twig` extension.

Like `cron.html` is ok.

phpBB/phpbb/cron/task/wrapper.php
lionel-rowe: > This should be removed. No globals.

What's the correct way of accessing `$template` here then? I noticed a common pattern is this:

```php
global $phpbb_container;

/** @var \phpbb\template\template $template Template object */
$template = $phpbb_container->get('template');
```

However, that simply pushes the problem one step down the line, as you still need the `global $phpbb_container;`. Or is `$phpbb_container` allowed as a global?

3D-I: `use phpbb\template\template;`

_then_

`protected $template`

_then in the constructor_

`public function __construct(task $task, helper $routing_helper, $phpbb_root_path, $php_ext, template $template)`

_then_

`$this->template	= $template;`

**NOT TESTED.**

lionel-rowe: Fixed with dependency injection (doesn't need `use phpbb\template\template;`).

3D-I: > Fixed with dependency injection (doesn't need `use phpbb\template\template;`).

I wrote on the fly just to give you a starter, well done.

phpBB/styles/all/template/cron.html
```diff
@@ -0,0 +1,7 @@
+{#
+	Runs the cron task by triggering server request to the URL on load. `img` is
+	preferred over JS to ensure maximum compatibility. We use `class="sr-only"`
+	to hide visually and `aria-hidden="true"` to hide from screen-readers; using
+	`hidden` or `display: none` would prevent the task from running.
+#}
+<img class="sr-only" aria-hidden="true" src="{{ CRON_TASK_URL | e('html_attr') }}" width="1" height="1" alt="">
```
3D-I, line 7: `"{{ CRON_TASK_URL|e('html_attr') }}"` - no spaces in between.

3D-I (approved):
Looks good to me at a first glance.

phpBB/phpbb/cron/task/wrapper.php
```diff
@@ -105,6 +112,25 @@ public function get_url()
...
+	 *
+	 * @return string       HTML to render to trigger cron task
+	 */
+	public function get_html_tag()
+	{
+		$this->template->set_filenames([
+			'cron_html_tag' => 'cron.html',
+		]);
+
+		$this->template->assign_vars([
```
3D-I, line 127: Oops, here we should use `assign_var` because only one var is being instantiated, sorry,

3D-I: `$this->template->assign_var('CRON_TASK_URL', $this->get_url());` should do.

hanakin (approved):
Looks good 👍

marc1706 (approved):

marc1706:
Just checked myself and it seems like tests will still need to be updated.

rubencm:
@lionel-rowe Your commits need to have this format:
```[ticket/16977] Fix cron-job img tag layout and accessibility

PHPBB3-16977```

You can do it [like this](https://coderwall.com/p/_thg9a/reword-a-git-commit-message) or if you want i can update it for you

lionel-rowe:
Tests and commit messages should now be fixed.

marc1706:
I've created a master version of your PR that will ease merging the PR: https://github.com/phpbb/phpbb/pull/6381
Once the builds have ran through I'll proceed with merging everything.

PR #6377, opened 2022-04-05
[ticket/16981] Fix HTML-encoded emojis in email subject line

lionel-rowe:
Fixes [PHPBB-16981 — HTML-encoded emojis `&#128396;️ &#128208;` in email subject line](https://tracker.phpbb.com/browse/PHPBB3-16981)

This fix is a straight find-and-replace of all instances of `htmlspecialchars_decode` with `html_entity_decode` (excluding `phpBB/install/convert` and `phpBB/install/convertors`).

Background: For compatibility with buggy legacy MySQL encodings (`utf8mb3` and `utf8`, the latter of which is an alias for the former rather than true UTF-8), emojis and other non-BMP Unicode characters are being stored in many places as the equivalent HTML entities, rather than as raw UTF-8.

Due to this, `htmlspecialchars_decode` is insufficient to normalize strings into a non-HTML format, because it converts `&amp;`, `&quot;`, `&#039;`, `&lt;`, `&gt;` but nothing else.

Replacing a few specific instances of `htmlspecialchars_decode` would fix the observed bug, but leave the door wide open to a class of related bugs popping up in other places (different email templates, different email fields, forum names, usernames, etc.)

Blanket replacement of `htmlspecialchars_decode` with `html_entity_decode` is safe, because `htmlspecialchars_decode` is only ever used in instances where the result is expected to be non-HTML-encoded, and the replacement has no effect where the source text contains no emoji HTML entities.

With that said, it'd be good to get more eyes on this, in case there are some corner cases I haven't thought about.

A [performance comparison between the two methods](https://replit.com/@lionel_rowe/html-entity-decode-perf#main.php) generally gives near-identical results. Typical results as follows:

```
htmlspecialchars_decode:    0.9544191360s (1,000,000 iterations)
html_entity_decode:         0.9495959282s (1,000,000 iterations)
```

---

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16981

### DIFF
--- a/phpBB/includes/acp/acp_bbcodes.php
+++ b/phpBB/includes/acp/acp_bbcodes.php
@@ -86,7 +86,7 @@ class acp_bbcodes
 				$display_on_posting = $request->variable('display_on_posting', 0);
 
 				$bbcode_match = $request->variable('bbcode_match', '');
-				$bbcode_tpl = htmlspecialchars_decode($request->variable('bbcode_tpl', '', true), ENT_COMPAT);
+				$bbcode_tpl = html_entity_decode($request->variable('bbcode_tpl', '', true), ENT_COMPAT);
 				$bbcode_helpline = $request->variable('bbcode_helpline', '', true);
 			break;
 		}
--- a/phpBB/includes/acp/acp_board.php
+++ b/phpBB/includes/acp/acp_board.php
@@ -720,8 +720,8 @@ class acp_board
 				$messenger->set_addresses($user->data);
 				$messenger->anti_abuse_headers($config, $user);
 				$messenger->assign_vars(array(
-					'USERNAME'	=> htmlspecialchars_decode($user->data['username'], ENT_COMPAT),
-					'MESSAGE'	=> htmlspecialchars_decode($request->variable('send_test_email_text', '', true), ENT_COMPAT),
+					'USERNAME'	=> html_entity_decode($user->data['username'], ENT_COMPAT),
+					'MESSAGE'	=> html_entity_decode($request->variable('send_test_email_text', '', true), ENT_COMPAT),
 				));
 				$messenger->send(NOTIFY_EMAIL);
 
--- a/phpBB/includes/acp/acp_email.php
+++ b/phpBB/includes/acp/acp_email.php
@@ -205,7 +205,7 @@ class acp_email
 				$email_template = 'admin_send_email';
 				$template_data = array(
 					'CONTACT_EMAIL' => phpbb_get_board_contact($config, $phpEx),
-					'MESSAGE'		=> htmlspecialchars_decode($message, ENT_COMPAT),
+					'MESSAGE'		=> html_entity_decode($message, ENT_COMPAT),
 				);
 				$generate_log_entry = true;
 
@@ -252,7 +252,7 @@ class acp_email
 
 					$messenger->anti_abuse_headers($config, $user);
 
-					$messenger->subject(htmlspecialchars_decode($subject, ENT_COMPAT));
+					$messenger->subject(html_entity_decode($subject, ENT_COMPAT));
 					$messenger->set_mail_priority($priority);
 
 					$messenger->assign_vars($template_data);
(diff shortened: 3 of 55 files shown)


3D-I:
No need to change 55 files for to fix `HTML-encoded emojis in email subject`,  moreover ... as I said in the tracker, your isue is not replicable.

marc1706:
I am able to reproduce this issue, but I would also go for a different approach. Instead of applying this to all places where we do htmlspecialchars_decode, I'd rather opt for applying it where needed right now. In this case I'd say let's apply it to the email variables in the notification types. You could even just use ut8_decode_ncr().

lionel-rowe:
> @3D-I as I said in the tracker, your issue is not replicable by my side.

Did you try the replication steps I specified when I replied to that tracker comment more than a week ago? You never responded.

> @marc1706 Instead of applying this to all places where we do htmlspecialchars_decode, I'd rather opt for applying it where needed right now. In this case I'd say let's apply it to the email variables in the notification types. You could even just use `ut8_decode_ncr()`.

Per my comments in the PR:

> Replacing a few specific instances of `htmlspecialchars_decode` would fix the observed bug, but leave the door wide open to a class of related bugs popping up in other places (different email templates, different email fields, forum names, usernames, etc.)
>
> Blanket replacement of `htmlspecialchars_decode` with `html_entity_decode` is safe, because `htmlspecialchars_decode` is only ever used in instances where the result is expected to be non-HTML-encoded, and the replacement has no effect where the source text contains no emoji HTML entities.
>
> A [performance comparison between the two methods](https://replit.com/@lionel_rowe/html-entity-decode-perf#main.php) generally gives near-identical results.

Fixing the problem in 1 place but not others, even with no bugs elsewhere yet observed, simply means it is likely to recur in other places. Given that replacing all instances with `html_entity_decode` has no effect on unaffected strings, not even a performance effect, what's the argument in favor of the whack-a-mole approach?

3D-I:
I've had my say.

lionel-rowe:
How do I move this forward? I'm willing to change it to the ad-hoc version if that's the only way it'll get approved, though I've laid out my case for why that's the wrong approach and would prefer to hear what the counterarguments are first.

3D-I:
Use `ut8_decode_ncr()` in the right places. Forget all the rest.

lionel-rowe:
> Use `ut8_decode_ncr()`
> ( which is the HTML counterpart of `utf8_encode_ucr()` )
> in the right places.
>
> Forget all the rest.

OK, so what's the reasoned argument for why that's the best approach? What are the benefits of that approach, and what are the drawbacks of the one I'm proposing?

3D-I:
I had my say. /me back to my duties.

marc1706 (approved):

marc1706:
> > @3D-I as I said in the tracker, your issue is not replicable by my side.
> 
> Did you try the replication steps I specified when I replied to that tracker comment more than a week ago? You never responded.
> 
> > @marc1706 Instead of applying this to all places where we do htmlspecialchars_decode, I'd rather opt for applying it where needed right now. In this case I'd say let's apply it to the email variables in the notification types. You could even just use `ut8_decode_ncr()`.
> 
> Per my comments in the PR:
> 
> > Replacing a few specific instances of `htmlspecialchars_decode` would fix the observed bug, but leave the door wide open to a class of related bugs popping up in other places (different email templates, different email fields, forum names, usernames, etc.)
> > Blanket replacement of `htmlspecialchars_decode` with `html_entity_decode` is safe, because `htmlspecialchars_decode` is only ever used in instances where the result is expected to be non-HTML-encoded, and the replacement has no effect where the source text contains no emoji HTML entities.
> > A [performance comparison between the two methods](https://replit.com/@lionel_rowe/html-entity-decode-perf#main.php) generally gives near-identical results.
> 
> Fixing the problem in 1 place but not others, even with no bugs elsewhere yet observed, simply means it is likely to recur in other places. Given that replacing all instances with `html_entity_decode` has no effect on unaffected strings, not even a performance effect, what's the argument in favor of the ad-hoc approach?

I have checked the code and deemed the changes to be fine.

PR #6379, opened 2022-04-06
[ticket/16984] Accessibility - fix tab order for keyboard navigation

lionel-rowe:
Fixes [PHPBB3-16984](https://tracker.phpbb.com/browse/PHPBB3-16984): Accessibility — confusing tab order for keyboard navigation

This is a blanket regex find-and-replace of `tabindex=(['"]?)[1-9]\d*\1` for `tabindex=$10$1` to bring the HTML in-line with the Web AIM guidance:

> **Important**
> `tabindex` values of 1+ must be avoided. These elements will receive keyboard focus before elements with no `tabindex` value (or `tabindex="0"`) resulting in a navigation order that is different from the visual and/or screen reader order. Instead of using `tabindex`, simply adjust the page's source code order to support a logical navigation and reading order, then use CSS to adjust the visual presentation.

The source code already uses a logical navigation and reading order, so the fix is simple 😊

---

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16984


marc1706:
If I read this correctly, the cleaner approach would probably be to completely remove the tabindex? In that case I'd actually then opt for removing it instead of setting it to 0.

lionel-rowe:
@marc1706 Lack of `tabindex` isn't always equivalent to `tabindex=0`:

```html
<!-- `tabindex` can safely be removed, as `input` is already focusable by default -->
<input type="submit" tabindex="0">

<!-- `div` is not usually focusable, so removing `tabindex` would prevent element from being focused -->
<div class="fake-button" tabindex="0"></div>
```

As most of these are `input`s etc though, should be OK to remove — I'll do some testing and see if anything breaks.

PR #6383, opened 2022-04-12
Improve handling of DDoS/brute force attacks on login form

rob006:
Checklist:

- [ ] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-

----

Background:

I had a DDoS attack on login form. It quickly filled the sessions table with tens of thousands of sessions for the anonymous user. A lot of sessions for this user dramatically decreased the performance of `SELECT COUNT(session_id) AS sessions FROM  SESSIONS_TABLE  ...` query, since the efficiency of `session_user_id` index was almost non-existent (99% of sessions was for anonymous user, so the query needed to count almost all rows in table). I excluded this query for the anonymous user, since it was not needed anyway. This significantly improved performance of forum, but...

...sessions were still created and the sessions table was growing. I ignored this, but after a few days I realized that cron does not work correctly. It turns out that task for clearing old sessions tried to load hundreds of thousands of sessions in one query, which exceeded the memory limit, and the whole process failed (and this was repeated over and over since old sessions were never removed). I added a limit to this query, so sessions could be cleaned up in smaller batches (I'm still not sure if this limit is optimal, but it worked for me).

### DIFF
--- a/phpBB/phpbb/captcha/plugins/captcha_abstract.php
+++ b/phpBB/phpbb/captcha/plugins/captcha_abstract.php
@@ -157,7 +157,8 @@ abstract class captcha_abstract
 			FROM ' . CONFIRM_TABLE . ' c
 			LEFT JOIN ' . SESSIONS_TABLE . ' s ON (c.session_id = s.session_id)
 			WHERE s.session_id IS NULL' .
-				((empty($type)) ? '' : ' AND c.confirm_type = ' . (int) $type);
+				((empty($type)) ? '' : ' AND c.confirm_type = ' . (int) $type)
+			. ' LIMIT 100000';
 		$result = $db->sql_query($sql);
 
 		if ($row = $db->sql_fetchrow($result))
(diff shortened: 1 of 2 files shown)


3D-I:
Plese open a ticket in the tracker, first. Then link it to your PR.

phpBB/phpbb/captcha/plugins/captcha_abstract.php
```diff
@@ -157,7 +157,8 @@ function garbage_collect($type)
 			FROM ' . CONFIRM_TABLE . ' c
 			LEFT JOIN ' . SESSIONS_TABLE . ' s ON (c.session_id = s.session_id)
 			WHERE s.session_id IS NULL' .
-				((empty($type)) ? '' : ' AND c.confirm_type = ' . (int) $type);
+				((empty($type)) ? '' : ' AND c.confirm_type = ' . (int) $type)
+			. ' LIMIT 100000';
 		$result = $db->sql_query($sql);
```
marc1706, line 162: Should be using sql_query_limit() instead.

PR #6392, opened 2022-04-25
[ticket/8071] Add sql_last_inserted_id alias for sql_nextid

lionel-rowe:
Fixes [DBAL function sql_nextid - name is misleading](https://tracker.phpbb.com/browse/PHPBB3-8071).

Per the tracker issue:
> This function gets the ID of the last inserted row, not the ID of the next row to be inserted (which is difficult due to different autoincrement steps and also lends itself nicely to race conditions).

@iwisdom replied that the naming is consistent with how PHP itself names this functionality, but the link they posted doesn't support that (I guess this is referring to a much older version of PHP? The comment is from 2009).

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-8071

### DIFF
--- a/phpBB/phpbb/db/driver/driver_interface.php
+++ b/phpBB/phpbb/db/driver/driver_interface.php
@@ -299,8 +299,9 @@ interface driver_interface
 	 * The returned value can be used for selecting the item that has just been
 	 * inserted or for updating another table with an ID pointing to that item.
 	 *
-	 * Will be deprecated in a future version of phpBB in favor of
-	 * `sql_last_inserted_id`.
+	 * Alias of `sql_last_inserted_id`.
+	 *
+	 * @deprecated 3.3.11-RC1 Replaced by sql_last_inserted_id(), to be removed in 4.1.0-a1
 	 *
 	 * @return	string|false	Auto-incremented value of the last inserted row
 	 */
@@ -312,11 +313,7 @@ interface driver_interface
 	 * just been inserted or for updating another table with an ID pointing to
 	 * that item.
 	 *
-	 * Alias of `sql_nextid`.
-	 *
 	 * @return	string|false	Auto-incremented value of the last inserted row
-	 *
-	 * @since 3.3.8-RC1
 	 */
 	public function sql_last_inserted_id();
 
--- a/phpBB/phpbb/db/driver/driver_interface.php
+++ b/phpBB/phpbb/db/driver/driver_interface.php
@@ -289,11 +289,36 @@ interface driver_interface
 	public function cast_expr_to_bigint($expression);
 
 	/**
-	* Get last inserted id after insert statement
-	*
-	* @return	string		Autoincrement value of the last inserted row
-	*/
+	 * Gets the ID of the **last** inserted row immediately after an INSERT
+	 * statement.
+	 *
+	 * **Note**: Despite the name, the returned ID refers to the row that has
+	 * just been inserted, rather than the hypothetical ID of the next row if a
+	 * new one was to be inserted.
+	 *
+	 * The returned value can be used for selecting the item that has just been
+	 * inserted or for updating another table with an ID pointing to that item.
+	 *
+	 * Will be deprecated in a future version of phpBB in favor of
+	 * `sql_last_inserted_id`.
+	 *
+	 * @return	string|false	Auto-incremented value of the last inserted row
+	 */
 	public function sql_nextid();
+
+	/**
+	 * Gets the ID of the last inserted row immediately after an INSERT
+	 * statement. The returned value can be used for selecting the item that has
+	 * just been inserted or for updating another table with an ID pointing to
+	 * that item.
+	 *
+	 * Alias of `sql_nextid`.
+	 *
+	 * @return	string|false	Auto-incremented value of the last inserted row
+	 *
+	 * @since 3.3.8-RC1
+	 */
+	public function sql_last_inserted_id();
 
 	/**
 	* Add to query count
--- a/phpBB/phpbb/db/driver/mssql_odbc.php
+++ b/phpBB/phpbb/db/driver/mssql_odbc.php
@@ -269,8 +269,8 @@ class mssql_odbc extends \phpbb\db\driver\mssql_base
 	}
 
 	/**
-	* {@inheritDoc}
-	*/
+	 * {@inheritDoc}
+	 */
 	function sql_nextid()
 	{
 		$result_id = @odbc_exec($this->db_connect_id, 'SELECT @@IDENTITY');
@@ -287,6 +287,14 @@ class mssql_odbc extends \phpbb\db\driver\mssql_base
 		}
 
 		return false;
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function sql_last_inserted_id()
+	{
+		return $this->sql_nextid();
 	}
 
 	/**
--- a/phpBB/phpbb/db/driver/mssql_odbc.php
+++ b/phpBB/phpbb/db/driver/mssql_odbc.php
@@ -269,9 +269,9 @@ class mssql_odbc extends \phpbb\db\driver\mssql_base
 	}
 
 	/**
-	 * {@inheritDoc}
+	 * {@inheritdoc}
 	 */
-	function sql_nextid()
+	public function sql_last_inserted_id()
 	{
 		$result_id = @odbc_exec($this->db_connect_id, 'SELECT @@IDENTITY');
 
@@ -287,14 +287,6 @@ class mssql_odbc extends \phpbb\db\driver\mssql_base
 		}
 
 		return false;
-	}
-
-	/**
-	 * {@inheritdoc}
-	 */
-	public function sql_last_inserted_id()
-	{
-		return $this->sql_nextid();
 	}
 
 	/**
--- a/phpBB/phpbb/db/driver/driver.php
+++ b/phpBB/phpbb/db/driver/driver.php
@@ -635,6 +635,14 @@ abstract class driver implements driver_interface
 	}
 
 	/**
+	 * {@inheritDoc}
+	 */
+	public function sql_nextid()
+	{
+		return $this->sql_last_inserted_id();
+	}
+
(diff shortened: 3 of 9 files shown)


lionel-rowe:
> The interface explains very well what this function does.

If there was a function that sent flowers to your mother that was named `fire_nukes`, with a comment above it simply saying "sends flowers to your mother", would you say that was explained very well?

At a bare minimum, the documentation should explicitly state that the function is misleadingly named and that "next" means "last" (i.e. the exact opposite meaning from what it should).

Also, the wording of the `@return` annotation, "Autoincrement value of the last inserted row", implies that calling the function itself will have the side-effect of autoincrementing the value, which is incorrect.

You can see my changes in `phpBB/phpbb/db/driver/driver_interface.php` for an example of how I'd prefer the documentation to look:

```
	/**
	 * @deprecated
	 *
	 * Gets the ID of the **last** inserted row immediately after an INSERT
	 * statement.
	 *
	 * **Note**: Despite the name, the returned ID refers to the row that has
	 * just been inserted, rather than the hypothetical ID of the next row if a
	 * new one was to be inserted. In other words, the returned value can be
	 * safely used for selecting the item that has just been inserted or for
	 * updating another table with an ID pointing to that item.
	 *
	 * @return	string		Auto-incremented value of the last inserted row
	*/
	public function sql_nextid();
```

lionel-rowe:
> @3D-I The interface explains very well what this function does.

If there was a function that sent flowers to your mother that was named `fire_nukes`, with a comment above it simply saying "sends flowers to your mother", would you say that was explained very well?

At a bare minimum, the documentation should explicitly state that the function is misleadingly named and that "next" means "last" (i.e. the exact opposite meaning from what it should).

Also, the wording of the `@return` annotation, "Autoincrement value of the last inserted row", implies that calling the function itself will have the side-effect of autoincrementing the value, which is incorrect.

You can see my changes in `phpBB/phpbb/db/driver/driver_interface.php` for an example of how I'd prefer the documentation to look:

```
	/**
	 * @deprecated
	 *
	 * Gets the ID of the **last** inserted row immediately after an INSERT
	 * statement.
	 *
	 * **Note**: Despite the name, the returned ID refers to the row that has
	 * just been inserted, rather than the hypothetical ID of the next row if a
	 * new one was to be inserted. In other words, the returned value can be
	 * safely used for selecting the item that has just been inserted or for
	 * updating another table with an ID pointing to that item.
	 *
	 * @return	string		Auto-incremented value of the last inserted row
	*/
	public function sql_nextid();
```

3D-I:
Points of view. Just fix the docblocks and that'it.

lionel-rowe:
> Side note: do not name my mother as an example, in my land this is considered outrageous. Thank you. 🇮🇹 😄

Haha sorry, I thought sending flowers to one's mother would be a nice thing to do! I just wanted an example of "useful and beneficial function that you might want to call" where the function is named as something completely unrelated, no offense meant 😆🌷

3D-I:
Ok, so back to the topic:

> Just fix the docblocks and that's it.

Everybody knows what that functions does, right or wrong since ages.
No specification obliges us to submit to what PHP uses as the name of the methods used by a framework, free will... in this case consolidated.

As an additional note I would like to point out that changing 37 files only happens in case of extreme remedies and established bugs, certainly not to create an alias. It doesn't fix anything but it's just cosmetic and can confuse users.

lionel-rowe:
> Everybody knows what that functions does

I certainly didn't. Well, I do now, after wasting quite a bit of time trying to figure this out. But it's puzzling to me why you assume people would know that "up" actually means "down" when they're new to the codebase... unless the assumption is that no new developers will ever start working with phpBB? In any case, the tracker issue that this resolves was opened a full 13 years ago by someone who I'm guessing has long since moved on, so it's certainly not only phpBB newbies circa 2022 who are having this issue...

> As an additional note I would like to point out that changing 37 files only happens in case of extreme remedies and established bugs, certainly not to create an alias

How would you feel about an intermediate solution — create the alias, deprecate the original method, but leave the current usages (despite being deprecated) as they are? That way, only a few files need to change.

3D-I:
> How would you feel about an intermediate solution — create the alias, deprecate the original method, but leave the current usages (despite being deprecated) as they are? That way, only a few files need to change.

Definitely not allocating this PR to 3.3.x branch but master. Although I think you'll find some tough positions there too. I would advise you to refine the docblocks and close your contribution happily. This seems to me to be the best solution in this particular case.

3D-I:
> But it's puzzling to me why you assume people would know that "up" actually means "down" when they're new to the codebase

Coders should read the docs and escpecially the code (interfaces)... that's 101.

lionel-rowe:
> Coders should read the docs and especially the code (interfaces)... that's 101.

I feel like this is going in circles... the very first thing I did was to read the inline documentation. If the docs say "goes down", but the function name is `goes_up`, would you not be at least a bit doubtful about using that function?

3D-I:
> Just fix the docblocks and that's it.

And the Docs if necessary.

iMattPro:
Rather than deprecating the old function, and replacing all its uses with a newly renamed version, I would prefer the opposite and much simpler change:

Leave everything as it was, and just create an alias in the driver for the `sql_nextid` function with a name like `sql_last_inserted_id` (or `sql_new_id` or `sql_inserted_id` or `sql_returned_id` or `sql_lastid` or whatever makes the most sense without being too wordy)

battye:
I think this function name change makes sense - because a lot of files are modified here, could a solution that's agreeable to everyone be that this change applies to 4.0 instead of happening in 3.3.x?

3D-I:
> Definitely not allocating this PR to 3.3.x branch but master.

3D-I:
> Rather than deprecating the old function, and replacing all its uses with a newly renamed version, I would prefer the opposite and much simpler change:
> 
> Leave everything as it was, and just create an alias in the driver for the `sql_nextid` function with a name like `sql_last_inserted_id` (or `sql_new_id` or `sql_inserted_id` or `sql_returned_id` or `sql_lastid` or whatever makes the most sense without being too wordy)

This sounds like a good idea. Although I don't see the need for all this, just editing the documents would be enough IMHO.  And yes, less verbosely indeed.

phpBB/phpbb/db/driver/driver_interface.php
```diff
@@ -289,12 +289,37 @@ public function sql_fetchrow($query_id = false);
...
+	 *
+	 * **Note**: Despite the name, the returned ID refers to the row that has
+	 * just been inserted, rather than the hypothetical ID of the next row if a
+	 * new one was to be inserted.
+	 *
+	 * The returned value can be used for selecting the item that has just been
+	 * inserted or for updating another table with an ID pointing to that item.
+	 *
+	 * Will be deprecated in a future version of phpBB in favor of
+	 * `sql_last_inserted_id`.
```
lionel-rowe, line 303: Is this OK to mention "Will be deprecated"? Signals that there's an eventual intention to deprecate, but without doing so immediately.

phpBB/phpbb/db/driver/driver_interface.php
```diff
@@ -289,12 +289,37 @@ public function sql_fetchrow($query_id = false);
...
+	 * Gets the ID of the last inserted row immediately after an INSERT
+	 * statement. The returned value can be used for selecting the item that has
+	 * just been inserted or for updating another table with an ID pointing to
+	 * that item.
+	 *
+	 * Alias of `sql_nextid`.
+	 *
+	 * @return	string|false	Auto-incremented value of the last inserted row
+	 *
+	 * @since 3.3.8-RC1
```
lionel-rowe, line 319: Is this the right version number to put?

phpBB/phpbb/db/driver/driver_interface.php
```diff
@@ -289,12 +289,37 @@ public function sql_fetchrow($query_id = false);
...
+	 * just been inserted, rather than the hypothetical ID of the next row if a
+	 * new one was to be inserted.
+	 *
+	 * The returned value can be used for selecting the item that has just been
+	 * inserted or for updating another table with an ID pointing to that item.
+	 *
+	 * Will be deprecated in a future version of phpBB in favor of
+	 * `sql_last_inserted_id`.
+	 *
+	 * @return	string|false	Auto-incremented value of the last inserted row
```
lionel-rowe, line 305: Return value changed from `string` to `string|false` as the implementations return `false` on failure.

phpBB/phpbb/db/driver/driver_interface.php
marc1706: Generally the right version number but we typically do not put this info for newly added methods.

phpBB/phpbb/db/driver/driver_interface.php
marc1706: I would actually add `@deprecated 3.3.8-RC1 Replaced by sql_last_inserted_id(), to be removed in 4.0.0-a1`

phpBB/phpbb/db/driver/mssql_odbc.php
```diff
@@ -289,6 +289,14 @@ function sql_nextid()
 		return false;
 	}
 
+	/**
+	 * {@inheritdoc}
+	 */
+	public function sql_last_inserted_id()
+	{
+		return $this->sql_nextid();
```
marc1706, line 297: I would actually move the logic from `sql_nextid()` to this and call `sql_last_inserted_id()` from `sql_nextid()`. This will make it way easier to later remove `sql_nextid()`.

marc1706 (changes requested):
Generally I would deprecate `sql_nextid()`. It'll still work fine but users might get warned about potential changes. This should also be mentioned in the release announcement for 3.3.8 then.

marc1706:
@lionel-rowe have you had time to look at my comments?

marc1706:
@lionel-rowe Hi, I was wondering if you could have a look at my comments. Would love to get this PR going again. Any help needed from our side to support you?

Derky (approved):

marc1706 (approved):

PR #6396, opened 2022-04-29
[ticket/13859] Allow up-to-date format for Youtube profile field URLs

lionel-rowe:
Per [the tracker issue](https://tracker.phpbb.com/browse/PHPBB3-13859) from May 2015:

> In 3.1 a youtube profilefield was added in which we can insert our youtube username to link to it.... There is just one major problem though...
>
> New youtube members no longer get an username. So I think we should decide on providing different ways to access their profile.

This PR will allow users to use any valid YouTube-domain URL, which allows for
all 3 formats (`/channel/...`, `/c/...`, and the now-legacy `/user/...`).

Per [YouTube's docs](https://support.google.com/youtube/answer/6180214?hl=en):

> ## Channel URL (ID-based)
> Example: youtube.com/channel/UCUZHFZ9jIKrLroW8LcyJEQQ
>
> This is the standard URL that YouTube channels use.
>
> ## Custom URL
> Example: youtube.com/c/YouTubeCreators
>
> A custom URL is a shorter, easy-to-remember URL that you can share with your
> audience.
>
> ## Legacy username URL
> Example: youtube.com/user/YouTube
>
> Depending on when your channel was created, it may have a username. Usernames are no longer required for channels today, but you can still use this URL to direct to your channel — even if your channel name has changed since you chose your username. Existing usernames can't be changed.

PHPBB3-13859

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-13859

### DIFF
--- a/phpBB/phpbb/db/migration/data/v33x/profilefield_youtube_update.php
+++ b/phpBB/phpbb/db/migration/data/v33x/profilefield_youtube_update.php
@@ -1,0 +1,82 @@
+<?php
+/**
+*
+* This file is part of the phpBB Forum Software package.
+*
+* @copyright (c) phpBB Limited <https://www.phpbb.com>
+* @license GNU General Public License, version 2 (GPL-2.0)
+*
+* For full copyright and license information, please see
+* the docs/CREDITS.txt file.
+*
+*/
+
+namespace phpbb\db\migration\data\v33x;
+
+class profilefield_youtube_update extends \phpbb\db\migration\migration
+{
+	protected $youtube_url_matcher = 'https:\\/\\/(www\\.)?youtube\\.com\\/.+';
+
+	public function effectively_installed()
+	{
+		$profile_fields = $this->table_prefix . 'profile_fields';
+
+		$result = $this->db->sql_query(
+			"SELECT field_validation
+				FROM $profile_fields
+				WHERE field_name = 'phpbb_youtube'"
+		);
+
+		$row = $this->db->sql_fetchrow($result);
+		$this->db->sql_freeresult($result);
+
+		return $row['field_validation'] === $this->youtube_url_matcher;
+	}
+
+	static public function depends_on()
+	{
+		return ['\phpbb\db\migration\data\v33x\v337'];
+	}
+
+	public function update_data()
+	{
+		return [['custom', [[$this, 'update_youtube_profile_field']]]];
+	}
+
+	public function update_youtube_profile_field()
+	{
+		$profile_fields = $this->table_prefix . 'profile_fields';
+		$profile_fields_data = $this->table_prefix . 'profile_fields_data';
+
+		$field_validation = $this->db->sql_escape($this->youtube_url_matcher);
+
+		$min_length = strlen('https://youtube.com/c/') + 1;
+
+		$this->db->sql_query(
+			"UPDATE $profile_fields SET
+				field_length = '40',
+				field_minlen = '$min_length',
+				field_maxlen = '255',
+				field_validation = '$field_validation',
+				field_contact_url = '%s'
+				WHERE field_name = 'phpbb_youtube'"
+		);
+
+		$yt_profile_field = 'pf_phpbb_youtube';
+		$prepend_legacy_youtube_url = $this->db->sql_concatenate(
+			"'https://youtube.com/user/'", $yt_profile_field
+		);
+		$is_not_already_youtube_url = $this->db->sql_not_like_expression(
+			$this->db->get_any_char()
+				. 'youtube.com/'
+				. $this->db->get_any_char()
+		);
+
+		$this->db->sql_query(
+			"UPDATE $profile_fields_data SET
+				$yt_profile_field = $prepend_legacy_youtube_url
+				WHERE $yt_profile_field <> ''
+				AND $yt_profile_field $is_not_already_youtube_url"
+		);
+	}
+}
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/db/migration/data/v33x/profilefield_youtube_update.php
```diff
@@ -0,0 +1,69 @@
...
+		$row = $this->db->sql_fetchrow($result);
+		$this->db->sql_freeresult($result);
+
+		return $row['field_validation'] === $this->youtube_url_matcher;
+	}
+
+	static public function depends_on()
+	{
+		return [
+			'\phpbb\db\migration\data\v310\profilefield_youtube',
```
rubencm, line 39: I think it should depend on `\phpbb\db\migration\data\v330\v330` too

rubencm (approved):

rubencm:
Just one think, I am not sure if is completely safe in the regexp to put `.+` to allow all kind of characters.
I have been checking and is being escaped correctly, so there shouldn't be any problem

rubencm:
I have reported one bug related with this, but not caused by this pr https://tracker.phpbb.com/browse/PHPBB3-16992

phpBB/phpbb/db/migration/data/v33x/profilefield_youtube_update.php
```diff
@@ -0,0 +1,75 @@
...
```
3D-I, line 39: I think this is not necessary.
`\phpbb\db\migration\data\v310\profilefield_youtube`

phpBB/phpbb/db/migration/data/v33x/profilefield_youtube_update.php
```diff
@@ -0,0 +1,75 @@
...
+				field_minlen = '$min_length',
+				field_maxlen = '255',
+				field_validation = '$field_validation',
+				field_contact_url = '%s'
+				WHERE field_name = 'phpbb_youtube'"
+		);
+
+		$this->db->sql_query(
+			"UPDATE $profile_fields_data SET pf_phpbb_youtube
+				= CONCAT('https://youtube.com/user/', pf_phpbb_youtube)
```
3D-I, line 70: `sql_concatenate` can be used here, instead ?

phpBB/phpbb/db/migration/data/v33x/profilefield_youtube_update.php
lionel-rowe: Do you mean `v310\profilefield_youtube` isn't necessary because `v330\v330` is already included? If so, I'll remove it.

3D-I: Exactly, also `v330\v330` should be amended as `v330\v337`

3D-I: `return ['\phpbb\db\migration\data\v33x\v337'];`

phpBB/phpbb/db/migration/data/v33x/profilefield_youtube_update.php
```diff
@@ -0,0 +1,76 @@
...
+
+		$prepend_youtube_url = $this->db->sql_concatenate(
+			"'https://youtube.com/user/'", 'pf_phpbb_youtube'
+		);
+
+		$this->db->sql_query(
+			"UPDATE $profile_fields_data SET
+				pf_phpbb_youtube = $prepend_youtube_url
+				WHERE pf_phpbb_youtube != ''
+				AND pf_phpbb_youtube NOT LIKE '%youtube.com/%'"
```
3D-I, line 73: Can `sql_not_like_expression()` be used here? Instead.

phpBB/phpbb/db/migration/data/v33x/profilefield_youtube_update.php
```diff
@@ -0,0 +1,76 @@
...
+		);
+
+		$prepend_youtube_url = $this->db->sql_concatenate(
+			"'https://youtube.com/user/'", 'pf_phpbb_youtube'
+		);
+
+		$this->db->sql_query(
+			"UPDATE $profile_fields_data SET
+				pf_phpbb_youtube = $prepend_youtube_url
+				WHERE pf_phpbb_youtube != ''
```
3D-I, line 72: Please use `<>` instead of `!=`

https://area51.phpbb.com/docs/dev/master/development/php/layout_guidelines.html#avoid-db-specific-sql

3D-I: You should be able to use `WHERE pf_phpbb_youtube IS NOT NULL` I think, instead. Not tested though.

lionel-rowe: > You should be able to use `WHERE pf_phpbb_youtube IS NOT NULL` I think, instead. Not tested though.

I don't think so, the column is already non-nullable with a default value of empty string, so that'd be a semantic change. I'll use `<>` though, thanks for the tip!

phpBB/phpbb/db/migration/data/v33x/profilefield_youtube_update.php
```diff
@@ -0,0 +1,82 @@
...
+				WHERE field_name = 'phpbb_youtube'"
+		);
+
+		$row = $this->db->sql_fetchrow($result);
+		$this->db->sql_freeresult($result);
+
+		return $row['field_validation'] === $this->youtube_url_matcher;
+	}
+
+	static public function depends_on()
```
marc1706, line 36:
```suggestion
	public static function depends_on()
```

marc1706 (changes requested):
Rest looks fine.

marc1706:
The CI is now just complaining about the commit message format ;)

marc1706 (approved):

PR #6412, opened 2022-08-12
[ticket/17026] Session viewonline not defined in Memberlist

Dark1z:
In `memberlist.php` the `session_viewonline` is not defined,
hence the function `phpbb_show_profile(...)` does not have the required data to produce online status of user.
This is helpful for extensions and styles.

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-17026

### DIFF
--- a/phpBB/memberlist.php
+++ b/phpBB/memberlist.php
@@ -1634,17 +1634,20 @@ switch ($mode)
 		if (count($user_list))
 		{
 			// Session time?! Session time...
-			$sql = 'SELECT session_user_id, MAX(session_time) AS session_time
+			$sql = 'SELECT session_user_id, MAX(session_time) AS session_time, MIN(session_viewonline) AS session_viewonline
 				FROM ' . SESSIONS_TABLE . '
 				WHERE session_time >= ' . (time() - $config['session_length']) . '
 					AND ' . $db->sql_in_set('session_user_id', $user_list) . '
 				GROUP BY session_user_id';
 			$result = $db->sql_query($sql);
 
-			$session_times = array();
+			$session_ary = [];
 			while ($row = $db->sql_fetchrow($result))
 			{
-				$session_times[$row['session_user_id']] = $row['session_time'];
+				$session_ary[$row['session_user_id']] = [
+					'session_time' => $row['session_time'],
+					'session_viewonline' => $row['session_viewonline'],
+				];
 			}
 			$db->sql_freeresult($result);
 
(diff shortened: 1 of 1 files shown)


GanstaZ:
If it's for extensions, then it should use events. In your ticket, you marked it as Bug priority major, but is it a bug or was it originally meant to work like that in memberlist?

Dark1z:
> If it's for extensions, then it should use events. In your ticket, you marked it as Bug priority major, but is it a bug or was it originally meant to work like that in memberlist?

check `phpbb_show_profile(...)` there is a need for `session_viewonline`.

GanstaZ:
> > If it's for extensions, then it should use events. In your ticket, you marked it as Bug priority major, but is it a bug or was it originally meant to work like that in memberlist?
> 
> Also for **styles** as mentioned above. check `phpbb_show_profile(...)` there is a need for `session_viewonline`.

As i see it, it works normally as it should (if not, then there is no point in having who is online) & i don't think that viewing memberlist should be showing online status for all users, that's why i mentioned events.

Dark1z:
> As i see it, it works normally as it should (if not, then there is no point in having who is online) & i don't think that viewing memberlist should be showing online status for all users, that's why i mentioned events.

It already has `memberrow.S_ONLINE` defined,
but only works when logged in user is Admin or user with permission `u_viewonline`.

So styles can use this to display it.

phpBB/memberlist.php
```diff
@@ -1708,7 +1711,7 @@
 			$id_cache = array();
 			while ($row = $db->sql_fetchrow($result))
 			{
-				$row['session_time'] = (!empty($session_times[$row['user_id']])) ? $session_times[$row['user_id']] : 0;
+				$row = array_merge($row, !empty($session_ary[$row['user_id']]) ? $session_ary[$row['user_id']] : ['session_time' => 0, 'session_viewonline' => 0]);
```
marc1706, line 1714: The readability of this line is not that great and running that many array_merge() calls is also not recommended. I'd go with this instead:
```
				$row['session_time'] = $session_ary[$row['user_id']]['session_time'] ?? 0;
				$row['session_viewonline'] = $session_ary[$row['user_id']]['session_viewonline'] ?? 0;
```

Dark1z: Done.

Thanks 👍

marc1706 (approved):

PR #6416, opened 2022-09-19
[ticket/17010] Implement notification method for web push

marc1706:
Checklist:

- [x] Tests pass

Todos:
- [x] Run notification tests for actual deliveries only on slow tests or check against web-push-testing existing
- [x] Add subscribing to push to UCP
- [x] Allow setting webpush per notification type in UCP
- [x] Have actual push delivery / display working
- [x] Add cronjob for deleting notification data

https://tracker.phpbb.com/browse/PHPBB3-17010


phpBB/phpbb/notification/method/webpush.php
```diff
@@ -0,0 +1,297 @@
...
+		}
+
+		// Load all users we want to notify (we need their email address)
+		$user_ids = [];
+		foreach ($this->queue as $notification)
+		{
+			$user_ids[] = $notification->user_id;
+		}
+
+		// We do not send emails to banned users
```
Nicofuma, line 148: You might want say notification rather than email :)

phpBB/phpbb/notification/method/webpush.php
```diff
@@ -0,0 +1,297 @@
...
+
+		// Get subscriptions for users
+		$user_subscription_map = [];
+		$sql = 'SELECT user_id, endpoint, p256dh, auth, encoding
+			FROM ' . $this->push_subscriptions_table . '
+			WHERE ' . $this->db->sql_in_set('user_id', $notify_users);
+		$result = $this->db->sql_query($sql);
+		while ($row = $this->db->sql_fetchrow($result))
+		{
+			if (isset($user_subscriptions[$row['user_id']]))
```
Nicofuma, line 167: Should be `$user_subscription_map`

phpBB/phpbb/notification/method/webpush.php
```diff
@@ -0,0 +1,297 @@
...
+				'subject' => generate_board_url(false),
+				'publicKey' => $this->config['webpush_vapid_public'],
+				'privateKey' => $this->config['webpush_vapid_private'],
+			],
+		];
+
+		$web_push = new \Minishlink\WebPush\WebPush($auth);
+
+		$number_of_notifications = 0;
+		// Time to go through the queue and send emails
```
Nicofuma, line 188: Email again :)

marc1706:
First notifications locally working with Google Chrome on Linux:
image

marc1706:
@phpbb/development-team PR can now be reviewed :)

Crizz0:
When I tried to activate it in the ACP, I get this error:

ticket-17010


`Fatal error: Uncaught RuntimeException: Unable to create the key in \phpBB\vendor\web-token\jwt-core\Util\ECKey.php:160 Stack trace: #0 \phpBB\vendor\web-token\jwt-core\Util\ECKey.php(119): Jose\Component\Core\Util\ECKey::createECKeyUsingOpenSSL('P-256') #1 \phpBB\vendor\web-token\jwt-key-mgmt\JWKFactory.php(78): Jose\Component\Core\Util\ECKey::createECKey('P-256', Array) #2 \phpBB\vendor\minishlink\web-push\src\VAPID.php(165): Jose\Component\KeyManagement\JWKFactory::createECKey('P-256') #3 \phpBB\includes\acp\acp_board.php(537): Minishlink\WebPush\VAPID::createVapidKeys() #4 \phpBB\includes\functions_module.php(676): acp_board->main('board', 'webpush') #5 \phpBB\adm\index.php(80): p_master->load_active() #6 {main} thrown in \phpBB\vendor\web-token\jwt-core\Util\ECKey.php on line 160`

marc1706:
VAPID keys can now be created in the ACP by clicking a separate button (the explain part of this setting still needs updating):
image

The JS code uses subtle crypto:
https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto

For most people, this should be fine to use and most likely has a better support than the previous code (which failed for Crizz0 already):
https://caniuse.com/?search=SubtleCrypto

phpBB/adm/style/ajax.js
```diff
@@ -157,6 +157,55 @@ phpbb.addAjaxCallback('row_delete', function(res) {
...
+
+			const publicKeyBuffer = await crypto.subtle.exportKey('raw', keyPair.publicKey);
+			const publicKeyString = phpbb.base64UrlEncode(phpbb.rawKeyToBase64(publicKeyBuffer));
+
+			return {
+				privateKey: privateKeyString,
+				publicKey: publicKeyString
+			};
+		} catch (error) {
+			console.error('Error generating keys with SubtleCrypto:', error);
```
marc1706, line 193: Should we show an error pop up instead (of course with proper translation)?

marc1706: This will only show if an admin is using a browser that does not support SubtleCrypto and tries generating the VAPID keys.

Crizz0:
Missing lang var: `{LOG CONFIG WEBPUSH}`

phpBB/adm/style/ajax.js
rxu: Makes sense.

phpBB/styles/all/js/push_worker.js.twig
```diff
@@ -0,0 +1,59 @@
...
+			body: formData
+		})
+		.then((response) => response.json())
+		.then((response) => {
+			const responseBody = response.title + "\n" + response.text;
+			const options = {
+				body: responseBody,
+				data: response,
+				icon: response.avatar.src
+				// foo: ''
```
CHItA, line 40: these could be removed i assume

phpBB/styles/all/js/webpush.js.twig
```diff
@@ -0,0 +1,287 @@
+/* global phpbb */
```
CHItA, line 1: is it really the best way to make the js twig? couldn't we just inject links?

marc1706: When writing it like this I kind of liked the simplicity of not needing this in two separate places. One downside of this current approach is however that linting won't properly work and I guess this will load slower than a plain JS file plus the injection of links.
For the push worker I do however think that this twig approach is the best way of doing it.

phpBB/language/en/acp/common.php
```diff
@@ -219,6 +219,7 @@
 	'ACP_VIEW_GLOBAL_MOD_PERMISSIONS'	=> 'View global moderation permissions',
 	'ACP_VIEW_USER_PERMISSIONS'			=> 'View user-based permissions',
 
+	'ACP_WEBPUSH_SETTINGS'		=> 'Webpush settings',
```
Derky, line 222: I see some different ways of writing "Webpush", "WebPush" and "web push". I think the correct way would be "Web Push" but while this is technically correct, maybe just referring to "Push Notifications" would be more clear for our end users. What do you think?

phpBB/phpbb/ucp/controller/webpush.php
```diff
@@ -0,0 +1,296 @@
...
+		}
+
+		$item_id = $this->request->variable('item_id', 0);
+		$type_id = $this->request->variable('type_id', 0);
+
+		$sql = 'SELECT push_data
+			FROM ' . $this->notification_webpush_table . '
+			WHERE user_id = ' . $this->user->id() . '
+				AND notification_type_id = ' . $type_id . '
+				AND item_id = ' . $item_id;
```
Derky, line 118: Could you add an `(int)` cast for all 3 variables used in the query?

phpBB/phpbb/ucp/controller/webpush.php
```diff
@@ -0,0 +1,296 @@
...
+	public function unsubscribe(symfony_request $symfony_request): JsonResponse
+	{
+		$this->check_subscribe_requests();
+
+		$data = json_sanitizer::decode($symfony_request->get('data', ''));
+
+		$endpoint = $data['endpoint'];
+
+		$sql = 'DELETE FROM ' . $this->push_subscriptions_table . '
+			WHERE user_id = ' . $this->user->id() . "
```
Derky, line 265: Could you please add an `(int)` here as well?

phpBB/phpbb/ucp/controller/webpush.php
```diff
@@ -0,0 +1,296 @@
...
+	 *
+	 * @return array Subscriptions for user
+	 */
+	protected function get_subscriptions(): array
+	{
+		$subscriptions = [];
+
+		$sql = 'SELECT endpoint, expiration_time
+			FROM ' . $this->push_subscriptions_table . '
+			WHERE user_id = ' . $this->user->id();
```
Derky, line 286: Please add an `(int)` cast here as well.

phpBB/styles/prosilver/template/ucp_notifications.html
```diff
@@ -7,38 +7,6 @@ <h2 class="cp-title">{TITLE}</h2>
 	<div class="inner">
 
 		<p class="cp-desc">{TITLE_EXPLAIN}</p>
-
-		<!-- IF MODE == 'notification_options' -->
```
Derky, line 11: Why did you remove this block?

marc1706: Has been moved to ucp_notifications_options.html which is used instead of this template file

phpBB/styles/prosilver/theme/colours.css
```diff
@@ -697,6 +697,15 @@ dd.profile-warnings {
...
 
+.button[disabled],
+.button[disabled]:hover,
+.button.disabled,
+.button.disabled:hover {
+	background: #e0e0e0;
+	border-color: #9e9e9e;
+	color: #9e9e9e;
+}
+
```
rxu, line 708: Probably dumb question but most of this is a duplication of `buttons.css` so the latter will be overriden with these ones as in `stylesheet.css` it goes later than `buttons.css`.

phpBB/language/en/acp/board.php
```diff
@@ -594,6 +594,17 @@
...
+	'WEBPUSH_ENABLE'				=> 'Enable Web Push',
+	'WEBPUSH_ENABLE_EXPLAIN'		=> 'Allow receiving notifications via Web Push. It is required to enter or generate valid VAPID identification keys to be able to use Web Push.',
+	'WEBPUSH_GENERATE_VAPID_KEYS'	=> 'Generate Identification keys',
+	'WEBPUSH_VAPID_PUBLIC'			=> 'Server identification public key',
+	'WEBPUSH_VAPID_PUBLIC_EXPLAIN'	=> 'The Voluntary Application Server Identification (VAPID) public key will be shared to authenticate push messages sent by your site.<br><em><strong>Warning:</strong> Changing the VAPID public key will automatically invalidate all Webpush subscriptions.</em>',
+	'WEBPUSH_VAPID_PRIVATE'			=> 'Server identification private key',
+	'WEBPUSH_VAPID_PRIVATE_EXPLAIN'	=> 'The Voluntary Application Server Identification (VAPID) private key will be used to create authenticated push messages sent by your site. The VAPID private key <strong>must</strong> be a valid public-private key pair with the VAPID public key.<br><em><strong>Warning:</strong> Changing the VAPID private key will automatically invalidate all Webpush subscriptions.</em>',
+]);
+
```
iMattPro, line 607: If these re-written lines make a little better sense can they be integrated here? This being a new feature I think it's important to be really clear explaining this to the board admins.
```
'ACP_WEBPUSH_SETTINGS_EXPLAIN'	=> 'Here you can enable Web Push for board notifications. Web Push is a protocol for the real-time delivery of events to user agents, commonly referred to as push messages. It is compatible with the majority of modern browsers on both desktop and mobile devices. Users can opt to receive Web Push alerts in their browser by subscribing and enabling their preferred notifications in the UCP.',
'WEBPUSH_ENABLE'				=> 'Enable Web Push',
'WEBPUSH_ENABLE_EXPLAIN'		=> 'Allow users to receive notifications in their browser or device via Web Push. To utilize Web Push, you must input or generate valid VAPID identification keys.',
'WEBPUSH_GENERATE_VAPID_KEYS'	=> 'Generate Identification keys',
'WEBPUSH_VAPID_PUBLIC'			=> 'Server identification public key',
'WEBPUSH_VAPID_PUBLIC_EXPLAIN'	=> 'The Voluntary Application Server Identification (VAPID) public key is shared to authenticate push messages from your site.<br><em><strong>Caution:</strong> Modifying the VAPID public key will automatically render all Web Push subscriptions invalid.</em>',
'WEBPUSH_VAPID_PRIVATE'			=> 'Server identification private key',
'WEBPUSH_VAPID_PRIVATE_EXPLAIN'	=> 'The Voluntary Application Server Identification (VAPID) private key is used to generate authenticated push messages dispatched from your site. The VAPID private key <strong>must</strong> form a valid public-private key pair alongside the VAPID public key.<br><em><strong>Caution:</strong> Modifying the VAPID private key will automatically render all Web Push subscriptions invalid.</em>',
```

rxu:
After some more testing ACP error log records have been noticed which look like:
```
Web Push message could not be sent: Client error: `POST https://fcm.googleapis.com/fcm/send/fEqTxyze5jA:APA91bGamtwzxIFCAkL4WbzS2pZ_iEaG1VvaWiTqvY_VJqMg0NxbxOfqsLrKSiowd3aaLzoBh5l6b3es0_tbSAfTgKQv3GjsrRZzZP4ooYWJF3CdJ3ZM7tI-otzcKfWq2MM_BCv1Ijal` resulted in a `410 Gone` response:
push subscription has unsubscribed or expired.
```

Same issue for Webpush extension (in development) which mostly shares the same codebase: https://github.com/phpbb-extensions/webpushnotifications/issues/23.
Additional info: https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerGlobalScope/pushsubscriptionchange_event

marc1706:
> After some more testing ACP error log records have been noticed which look like:
> 
> ```
> Web Push message could not be sent: Client error: `POST https://fcm.googleapis.com/fcm/send/fEqTxyze5jA:APA91bGamtwzxIFCAkL4WbzS2pZ_iEaG1VvaWiTqvY_VJqMg0NxbxOfqsLrKSiowd3aaLzoBh5l6b3es0_tbSAfTgKQv3GjsrRZzZP4ooYWJF3CdJ3ZM7tI-otzcKfWq2MM_BCv1Ijal` resulted in a `410 Gone` response:
> push subscription has unsubscribed or expired.
> ```
> 
> Same issue for Webpush extension (in development) which mostly shares the same codebase: [phpbb-extensions/webpushnotifications#23](https://github.com/phpbb-extensions/webpushnotifications/issues/23). Additional info: https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerGlobalScope/pushsubscriptionchange_event

I think we'll have to remove these subscriptions then when they return the mentioned error

rxu:
That sounds like a user will lost subscription at all. For me, it seems rather like it should be "refreshed"/re-subscribed silently.

marc1706:
According to the documentation at https://web.dev/articles/push-notifications-web-push-protocol#response_from_push_service it should however be removed, there doesn't seem to be a refresh token system:410Gone. The subscription is no longer valid and should be removed from application server. This can be reproduced by calling `unsubscribe()` on a `PushSubscription`.On Feb 1, 2024 09:38, rxu ***@***.***> wrote:
That sounds like a user will lost subscription at all. For me, it seems rather like it should be "refreshed"/re-subscribed silently.

phpBB/language/en/acp/board.php
marc1706: Thank you, I have integrated your suggestion.

phpBB/styles/prosilver/theme/colours.css
marc1706: I have removed the duplicated css now

marc1706:
@rxu I have now added handling that will remove subscriptions that have expired.

rxu:
@marc1706 Looks good, there're no `Web Push message could not be sent: Client error`s anymore so far.

rxu:
Currently webpush notification popup looks typically:
image

Can/should it be customized in some manner? F.e. by adding [optional] icon etc.

marc1706:
@rxu Yes, we could have a look at that. I'd however aim at first having the base functionality added and then we can continue with adding more features. :)

phpBB/phpbb/ucp/controller/webpush.php
```diff
@@ -0,0 +1,241 @@
...
+	{
+		$this->check_subscribe_requests();
+
+		$data = json_sanitizer::decode($symfony_request->get('data', ''));
+
+		$endpoint = $data['endpoint'];
+
+		$sql = 'DELETE FROM ' . $this->push_subscriptions_table . '
+			WHERE user_id = ' . (int) $this->user->id() . "
+				AND endpoint = '" . (int) $this->db->sql_escape($endpoint) . "'";
```
Derky, line 233: You're sql_escaping an int-casted variable here. Is `$endpoint` a string?

Derky (changes requested):
Only one comment for [phpBB/phpbb/ucp/controller/webpush.php](https://github.com/phpbb/phpbb/pull/6416/files/fcfed793858d44473cc242d73d0ba044cf0786ad#diff-33f42c6f4405e386e7e85ad49973031555ed3e84bf7f29907c32888da2900c80) . Apart from that all looks good!

phpBB/phpbb/ucp/controller/webpush.php
marc1706: Endpoint is a string and hence the `(int)` shouldn't be there :)

Derky (approved):
Looks good now, thanks!

PR #6428, opened 2022-11-12
[ticket/16105] Use Google reCAPTCHA globally

wordlesswind:
PHPBB3-16105

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-16105

### DIFF
--- a/phpBB/adm/style/captcha_recaptcha_acp.html
+++ b/phpBB/adm/style/captcha_recaptcha_acp.html
@@ -21,6 +21,20 @@
 	<dd><input id="recaptcha_privkey" name="recaptcha_privkey" value="{RECAPTCHA_PRIVKEY}" size="50" type="text" /></dd>
 </dl>
 
+<dl>
+	<dt>
+		<label>{{ lang('RECAPTCHA_V3_DOMAIN') ~ lang('COLON') }}</label>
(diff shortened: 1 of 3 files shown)


wordlesswind:
It seems that the v3 version plugin provides an optional list, but I don't know PHP, so I don't know how to change v2 plugin.

https://github.com/phpbb/phpbb/blob/696ff06263839d3b85ae4dff492343677c7796a5/phpBB/phpbb/captcha/plugins/recaptcha_v3.php#L28-L32

As well, I would like to be able to add one more domain to the list if possible: `recaptcha.google.cn`.
Although it is a `.cn` domain, it also has a global CDN, and unlike other domains, `google.cn` has received the following license number from the Chinese government: `京ICP备13004732号-2`.

marc1706:
As you already pointed out, the recaptcha v3 does already support choosing this optional URL instead of google.com. However, if I understand you correctly, the best approach would actually be:
- Support selecting URL in recaptcha v2 the same way as already possible in v3 (in ACP)
- Add recaptcha.google.cn to possible domains

wordlesswind:
Yes, but I don't know how to modify it and maybe someone needs to take over.

rubencm (approved):

PR #6431, opened 2022-11-26
[ticket/17062] Call search service using dependency injection

rubencm:
PHPBB3-17062

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-17062

### DIFF
--- a/phpBB/phpbb/cron/task/core/tidy_search.php
+++ b/phpBB/phpbb/cron/task/core/tidy_search.php
@@ -13,75 +13,39 @@
 
 namespace phpbb\cron\task\core;
 
+use phpbb\config\config;
+use phpbb\cron\task\base;
+use phpbb\search\search_backend_factory;
+
 /**
 * Tidy search cron task.
 *
 * Will only run when the currently selected search backend supports tidying.
 */
-class tidy_search extends \phpbb\cron\task\base
+class tidy_search extends base
 {
 	/**
-	* phpBB root path
-	* @var string
-	*/
-	protected $phpbb_root_path;
-
-	/**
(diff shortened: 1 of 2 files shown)


phpBB/phpbb/cron/task/core/tidy_search.php
```diff
@@ -114,7 +70,16 @@ public function run()
 	*/
 	public function is_runnable()
 	{
-		return class_exists($this->config['search_type']);
+		try
+		{
+			$this->search_backend_factory->get_active();
```
rxu, line 75: What if you add `$this->search` protected property, then here `$this->search = $this->search_backend_factory->get_active();` and then just `$this->search->tidy()` in the `function run()` above.
Just to avoid `$this->search_backend_factory->get_active();` call duplication.

rubencm: Yes, I've thought about that, but for that it would have to be guaranteed that is_runnable is always called before run.
Also, I personally prefer the service to be stateless.

rxu: Well, it's always should be called here https://github.com/phpbb/phpbb/blob/master/phpBB/phpbb/cron/event/cron_runner_listener.php#L84 within `is_ready()` function, but ok.

rubencm: I have changed it a little with a solution we both like. Methods don't need to be called in specific order (it works as stateless), and also the call to get_active is done only once.

marc1706 (approved):

PR #6470, opened 2023-02-04
[ticket/17100] Introduce twig macros for commonly used form elements

marc1706:
See the ticket on thoughts behind these changes and why it certainly does not cover all of the HTML in PHP files.

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-17100


hanakin:
Of note we opted for functions for the icons to avoid all the imports. We might want to consider the same here as there will be quite a lot of them

marc1706:
@hanakin updated the PR in that way

phpBB/adm/index.php
```diff
@@ -53,12 +53,19 @@
...
 
 // Set custom style for admin area
-$template->set_custom_style(array(
-	array(
-		'name' 		=> 'adm',
-		'ext_path' 	=> 'adm/style/',
-	),
-), $phpbb_admin_path . 'style');
+/** @var \phpbb\template\base $template */
+$template->set_custom_style(
```
marc1706, line 57: This was changed to support a single form macro file in both adm and prosilver ... so might not be needed

phpBB/phpbb/template/twig/environment.php
```diff
@@ -75,7 +75,7 @@ public function __construct(\phpbb\config\config $phpbb_config, \phpbb\filesyste
 
 		$options = array_merge(array(
 			'cache'			=> (defined('IN_INSTALL')) ? false : $cache_path,
-			'debug'			=> false,
+			'debug'			=> true,
```
CHItA, line 78: Why is this changed?

CHItA (commented):
Otherwise looks good to me.

phpBB/phpbb/template/twig/environment.php
marc1706: I think I changed this for some more debugging and then forgot to change it back. Will revert this line.

phpBB/phpbb/template/twig/twig.php
```diff
@@ -90,6 +90,7 @@ public function __construct(
 		{
 			$this->twig->addExtension($extension);
 		}
+		$this->twig->addExtension(new \Twig\Extension\DebugExtension());
```
rubencm, line 93: This line makes the development environment to fail

phpBB/includes/acp/acp_board.php
```diff
@@ -28,16 +33,32 @@ class acp_board
...
+
 	function main($id, $mode)
 	{
 		global $user, $template, $request, $language;
 		global $config, $phpbb_root_path, $phpEx;
 		global $cache, $phpbb_container, $phpbb_dispatcher, $phpbb_log;
 
-		/** @var \phpbb\language\language $language Language object */
-		$language = $phpbb_container->get('language');
+		if (!$language)
```
rubencm, line 51: I think this condition can never happen

phpBB/includes/acp/acp_board.php
```diff
@@ -1019,11 +1071,15 @@ function quick_reply($value, $key)
 	*/
 	function timezone_select($value, $key)
 	{
-		global $template, $user;
+		global $user;
```
rubencm, line 1074: there is already a property for user in this class

phpBB/includes/functions_acp.php
```diff
@@ -294,34 +320,125 @@ function build_cfg_template($tpl_type, $key, &$new_ary, $config_key, $vars)
 			$min = isset($tpl_type[1]) ? (int) $tpl_type[1] : false;
 			$max = isset($tpl_type[2]) ? (int) $tpl_type[2] : false;
 
-			$tpl = '<input id="' . $key . '" type="' . $tpl_type[0] . '"' . (( $min !== false ) ? ' min="' . $min . '"' : '') . (( $max !== false ) ? ' max="' . $max . '"' : '') . ' name="' . $name . '" value="' . $new_ary[$config_key] . '" />';
+			$tpl = [
+				'tag'		=> 'input',
+				'id'		=> $key,
+				'type'		=> $tpl_type[0],
+				'name'		=> $name,
+				'min'		=> $min !== false ? $min : '',
```
rubencm, line 328: ```$min ?: ''```

shorter version, same for max

phpBB/includes/functions_acp.php
```diff
@@ -294,34 +320,125 @@ function build_cfg_template($tpl_type, $key, &$new_ary, $config_key, $vars)
...
-			$cols = (int) $tpl_type[2];
-
-			$tpl = '<textarea id="' . $key . '" name="' . $name . '" rows="' . $rows . '" cols="' . $cols . '">' . $new_ary[$config_key] . '</textarea>';
+			$tpl = [
+				'tag'		=> 'textarea',
+				'id'		=> $key,
+				'name'		=> $name,
+				'rows'		=> (int) $tpl_type[1],
+				'cols'		=> (int) $tpl_type[2],
+				'content'	=> $new_ary[$config_key],
```
rubencm, line 365: In the template this is uppercase, i am not sure if this matter, but sometimes you pass lowercase variables and sometime uppercase to the template

phpBB/includes/functions_acp.php
```diff
@@ -294,34 +320,125 @@ function build_cfg_template($tpl_type, $key, &$new_ary, $config_key, $vars)
...
+			];
 		break;
 
 		case 'radio':
-			$key_yes	= ($new_ary[$config_key]) ? ' checked="checked"' : '';
-			$key_no		= (!$new_ary[$config_key]) ? ' checked="checked"' : '';
-
 			$tpl_type_cond = explode('_', $tpl_type[1]);
-			$type_no = ($tpl_type_cond[0] == 'disabled' || $tpl_type_cond[0] == 'enabled') ? false : true;
+			$type_no = !(($tpl_type_cond[0] == 'disabled' || $tpl_type_cond[0] == 'enabled'));
```
rubencm, line 371: This don't matter much, but parenthesis here is redundant

rubencm:
What I did in acp_storage would look much better with all of this :sweat_smile:

rubencm (approved):

phpBB/includes/functions_acp.php
marc1706: The check is actually stricter than what you proposed and fully mirrors the previous check.

phpBB/includes/functions_acp.php
marc1706: I've actually opted for this for a better distinction between data supplied to the forms twig function and actual template data. The uppercase version is created in the twig forms extension where the input data (form_data) gets translated to the data that gets output to the template.

phpBB/includes/functions_acp.php
marc1706: Cleaned it up even a tiny bit better ;)

CHItA (approved):

PR #6475, opened 2023-03-31
[ticket/17125] use box-sizing for message box text areas

iMattPro:
PHPBB3-17125

Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-17125

### DIFF
--- a/phpBB/styles/prosilver/template/posting_editor.html
+++ b/phpBB/styles/prosilver/template/posting_editor.html
@@ -47,7 +47,6 @@
 
 	<!-- INCLUDE posting_buttons.html -->
 
-	<div class="smiley-message-box">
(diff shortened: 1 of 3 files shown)


iMattPro:
Actually it looks like border-box box-sizing on the message textbox will fix this too, and may be a better solution??

marc1706:
> @marc1706 Actually it looks like border-box box-sizing on the message textbox will fix this too, and may be a better solution??

yes, I think that would be the better option

iMattPro:
I just noticed this is also happening on iPhone's using the latest version of iOS (16.4) so suddenly it's an issue across Macs, iPhones and iPads. So a lot more than just me will notice this, lol!

marc1706 (approved):

PR #6478, opened 2023-04-16
[ticket/17097] Fix PHP 8.2 deprecation warnings

m-ober:
Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-17097

### DIFF
--- a/phpBB/phpbb/attachment/upload.php
+++ b/phpBB/phpbb/attachment/upload.php
@@ -48,6 +48,9 @@ class upload
 	/** @var dispatcher */
 	protected $phpbb_dispatcher;
 
+	/** @var string */
+	protected $phpbb_root_path;
+
 	/** @var plupload Plupload */
 	protected $plupload;
 
--- a/phpBB/phpbb/auth/auth.php
+++ b/phpBB/phpbb/auth/auth.php
@@ -524,19 +524,19 @@ class auth
 			ORDER BY role_id ASC';
 		$result = $db->sql_query($sql);
 
-		$this->role_cache = array();
+		$role_cache = array();
 		while ($row = $db->sql_fetchrow($result))
 		{
-			$this->role_cache[$row['role_id']][$row['auth_option_id']] = (int) $row['auth_setting'];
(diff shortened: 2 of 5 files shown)


m-ober:
Maybe it would be better to add `role_cache` as property to the `auth` class, but I don't know if this was left out intentional.

In case of `phpbb_root_path` it looks like it was simply forgotten.

m-ober:
We could also replace the `ReturnTypeWillChange` attributes with the actual return type, *except* for `offsetGet`, which has a `mixed` return type that is only supported by PHP >=8.0.

m-ober:
As this question came up in chat, Attributes are backwards compatible as long as they are on a separate line. Then, they are just parsed as comments by PHP <8.0.

marc1706:
> Maybe it would be better to add `role_cache` as property to the `auth` class, but I don't know if this was left out intentional.
> 
> In case of `phpbb_root_path` it looks like it was simply forgotten.

It actually does not need to be a class property as it's never used as such. In master, it was changed to no longer use this:
https://github.com/phpbb/phpbb/commit/774c609c4a33b3ff4d7673e23feab266689e96cb#diff-d3edffaad6045eb7577d57cbb613c8903eb4d239c9f41aeaabfd1218ebdc596c

marc1706 (approved):

m-ober:
@marc1706 Thanks for the review! I have cherry-picked the changes from master (for the changes regarding the `$role_cache` variable only). Also, I have added the `#[\ReturnTypeWillChange]` attribute to the same methods in the `config` class. I never got a deprecation notice from this class, but as it is also implementing the `ArrayAccess` interface, it could trigger the same notices.

m-ober:
Any idea why the "Check commit message" step is failing? The error message is not very helpful.

rubencm:
It should be

```
[ticket/17097] Fix PHP 8.2 deprecation warnings]

PHPBB3-17097
```

m-ober:
@rubencm Thanks, fixed.

marc1706 (approved):

marc1706:
Thanks for your contribution 👍

PR #6479, opened 2023-04-19
[ticket/17130] Change storage of enable_magic_url in Posts and PMs

iMattPro:
Checklist:

- [x] Tests pass

https://tracker.phpbb.com/browse/PHPBB3-17130

I don't know if we also need a migration to update these fields in post and PM tables, at least for boards that have Allow magic URLs disabled board wide. I've got a migration locked and loaded but have not put it here yet.


phpBB/includes/ucp/ucp_pm_compose.php
```diff
@@ -828,7 +828,7 @@ function compose_pm($id, $mode, $action, $user_folders = array())
 
 		$enable_bbcode 		= (!$bbcode_status || isset($_POST['disable_bbcode'])) ? false : true;
 		$enable_smilies		= (!$smilies_status || isset($_POST['disable_smilies'])) ? false : true;
-		$enable_urls 		= (isset($_POST['disable_magic_url'])) ? 0 : 1;
+		$enable_urls 		= (!$url_status || isset($_POST['disable_magic_url'])) ? 0 : 1;
```
rxu, line 831: Will it work like that?
```php
$enable_urls 		= ($url_status || !$request->is_set_post('disable_magic_url'));
```

Also, it probably requires `&&` as far as disallowing links in posts means disallowing them regardless of `'disable_magic_url'` option.

iMattPro:
I put up the migration. What it would do, and this would only apply to boards that have disallowed board-wide Magic URLs in posts and PMs:
- First sets all `enable_magic_url` in posts and PMs to `0`
- Then start looking at all the posts and PMs for any that actually contain rendered Magic URLs and change those `enable_magic_url` fields back to `1`
This may be a bit of a big hammer. But in reality if a board has disallowed Magic URLs then that field should have always been set to 0 anyways, so this migration fixes that.

If the admin later decides to start allowing magic URL's again, this would be the behavior:
- New posts will have magic URLs
- Old posts still won't be magic.
- Running the text reporter on posts or PMs will not turn old posts into magic URLs, they will remain as before.
- Editing an old post, the checkbox for Do not parse URLs will be checked by default, but could be un-checked to make the URLs magic after editing.

phpBB/includes/ucp/ucp_pm_compose.php
iMattPro: No it needs to be an OR actually, exactly like the smilies and bbcode ones above

marc1706:
Closing in favor of #6604

PR #6483, opened 2023-05-05
[ticket/17135] Introduce Symfony Mailer for emails backend

rxu:
Checklist:

- [x] Tests pass

<a href="https://tracker.phpbb.com/browse/PHPBB3-17135">PHPBB3-17135</a>.


rxu:
!set WIP

rxu:
This actually needs some review as it's not clear what's next. In current PR state emailing just works (at least for simple cases).
Also I couldn't find a good way to implement email automated tests as [default Symfony instruction](https://symfony.com/doc/current/mailer.html#write-a-functional-test) relies on extending `FrameworkBundle` / extending `KernelTestCase` which is not used in phpBB test framework.

rxu:
!unset WIP

phpBB/includes/acp/acp_email.php
```diff
@@ -229,31 +217,50 @@ function main($id, $mode)
...
 
+				$messenger = $phpbb_container->get('messenger.method_collection');
+				$messenger_collection_iterator = $messenger->getIterator();
 				for ($i = 0, $size = count($email_list); $i < $size; $i++)
 				{
 					$used_lang = $email_list[$i][0]['lang'];
 					$used_method = $email_list[$i][0]['method'];
 
-					for ($j = 0, $list_size = count($email_list[$i]); $j < $list_size; $j++)
+					while ($messenger_collection_iterator->valid())
```
marc1706, line 227: The iterator can be used with a foreach which is IMHO more readable

phpBB/includes/acp/acp_email.php
```diff
@@ -229,31 +217,50 @@ function main($id, $mode)
 				);
 				extract($phpbb_dispatcher->trigger_event('core.acp_email_send_before', compact($vars)));
 
+				$messenger = $phpbb_container->get('messenger.method_collection');
```
marc1706, line 220: I recommend adding a type hint for $messenger (\phpbb\di\service_collection)

phpBB/includes/acp/acp_inactive.php
```diff
@@ -114,29 +114,20 @@ function main($id, $mode)
 
 						if ($config['require_activation'] == USER_ACTIVATION_ADMIN && !empty($inactive_users))
 						{
-							if (!class_exists('messenger'))
-							{
-								include($phpbb_root_path . 'includes/functions_messenger.' . $phpEx);
-							}
-
-							$messenger = new messenger(false);
+							$messenger = $phpbb_container->get('messenger.method_collection');
```
marc1706, line 117: Same here, I think a type hint would help

phpBB/includes/functions_messenger.php
marc1706, line 1: So is this still used even though you added the new classes?

phpBB/includes/ucp/ucp_activate.php
```diff
@@ -131,21 +131,25 @@ function main($id, $mode)
...
-			$messenger->anti_abuse_headers($config, $user);
-
-			$messenger->assign_vars(array(
-				'USERNAME'	=> html_entity_decode($user_row['username'], ENT_COMPAT))
-			);
-
-			$messenger->send($user_row['user_notify_type']);
+			$messenger = $phpbb_container->get('messenger.method_collection');
+			$messenger_collection_iterator = $messenger->getIterator();
+			while ($messenger_collection_iterator->valid())
```
marc1706, line 136: Please use foreach for this

phpBB/includes/ucp/ucp_profile.php
```diff
@@ -170,27 +170,21 @@ function main($id, $mode)
...
-							include_once($phpbb_root_path . 'includes/functions_messenger.' . $phpEx);
-
 							$server_url = generate_board_url();
 
 							$user_actkey = gen_rand_string(mt_rand(6, 10));
 
-							$messenger = new messenger(false);
-
+							$messenger = $phpbb_container->get('messenger.method_collection');
+							$email = $messenger->offsetGet('messenger.method.email');
```
marc1706, line 178: Not sure why you need the method collection then?

phpBB/includes/ucp/ucp_register.php
```diff
@@ -457,30 +457,26 @@ function main($id, $mode)
...
-
-					$messenger->template($email_template, $data['lang']);
-
-					$messenger->to($data['email'], $data['username']);
-
-					$messenger->anti_abuse_headers($config, $user);
-
-					$messenger->assign_vars(array(
+					$messenger = $phpbb_container->get('messenger.method_collection');
+					$email = $messenger->offsetGet('messenger.method.email');
```
marc1706, line 461: I think there is more places like this and you shouldn't need to go through the method collection if you just get the email method.

marc1706: Also, I'd recommend going for clearer names, e.g. $email_method instead of just $email

phpBB/includes/ucp/ucp_resend.php
```diff
@@ -132,20 +133,31 @@ function main($id, $mode)
...
-					$messenger->anti_abuse_headers($config, $user);
-
-					$messenger->assign_vars(array(
-						'USERNAME'			=> html_entity_decode($user_row['username'], ENT_COMPAT),
-						'U_USER_DETAILS'	=> generate_board_url() . "/memberlist.$phpEx?mode=viewprofile&u={$user_row['user_id']}",
-						'U_ACTIVATE'		=> generate_board_url() . "/ucp.$phpEx?mode=activate&u={$user_row['user_id']}&k={$user_row['user_actkey']}")
-					);
-
-					$messenger->send($row['user_notify_type']);
+					while ($messenger_collection_iterator->valid())
```
marc1706, line 139: foreach please, I think you get the point

phpBB/phpbb/db/migration/data/v400/remove_smtp_auth_method.php
```diff
@@ -0,0 +1,38 @@
...
+		];
+	}
+
+	public function update_data()
+	{
+		return [
+			['config.remove', ['smtp_auth_method']],
+		];
+	}
+}
```
marc1706, line 38: Maybe also add a revert method

phpBB/phpbb/notification/method/messenger_base.php
```diff
@@ -112,22 +112,29 @@ protected function notify_using_messenger($notify_method, $template_dir_prefix =
...
-			$messenger->set_addresses($user);
-
-			$messenger->assign_vars(array_merge(array(
-				'USERNAME'						=> $user['username'],
-
-				'U_NOTIFICATION_SETTINGS'		=> generate_board_url() . '/ucp.' . $this->php_ext . '?i=ucp_notifications&mode=notification_options',
-			), $notification->get_email_template_variables()));
-
-			$messenger->send($notify_method);
+			/** @psalm-suppress InvalidTemplateParam */
```
marc1706, line 115: What are you suppressing here and why?

phpBB/phpbb/notification/method/messenger_base.php
```diff
@@ -112,22 +112,29 @@ protected function notify_using_messenger($notify_method, $template_dir_prefix =
...
+			$messenger_collection_iterator = $this->messenger->getIterator();
+			while ($messenger_collection_iterator->valid())
```
marc1706, line 117: foreach

phpBB/phpbb/template/twig/extension.php
```diff
@@ -163,6 +163,7 @@ public function loop_subset(\Twig\Environment $env, $item, $start, $end = null,
 		// We always include the last element (this was the past design)
 		$end = ($end == -1 || $end === null) ? null : $end + 1;
 
+		/** @psalm-suppress UndefinedFunction */
```
marc1706, line 166: Not sure why this is needed?

phpBB/phpbb/messenger/method/phpbb_jabber.php
```diff
@@ -1,117 +1,254 @@
...
-if (!defined('IN_PHPBB'))
+ *
+ * Based on Jabber class from Flyspray project
+ *
+ * @version class.jabber2.php 1595 2008-09-19 (0.9.9)
+ * @copyright 2006 Flyspray.org
+ * @author Florian Schmitz (floele)
+ *
+ * Slightly modified by Acyd Burn (2006)
+ * Refactored to a service (2023)
```
marc1706, line 25: Hey look, it's 2024 :grinning:

phpBB/phpbb/messenger/method/base.php
```diff
@@ -0,0 +1,518 @@
...
+		$this->twig_lexer = $twig_lexer;
+		$this->template_cache_path = $template_cache_path;
+		$this->root_path = $phpbb_root_path;
+
+		$this->set_use_queue();
+	}
+
+	/**
+	 * Get messenger method id
+	 * @return mixed
```
marc1706, line 146: What mixed return are you expecting?

phpBB/phpbb/messenger/method/base.php
```diff
@@ -0,0 +1,518 @@
...
+	 */
+	abstract public function get_id();
+
+	/**
+	 * Check if the messenger method is enabled
+	 * @return bool
+	 */
+	abstract public function is_enabled();
```
marc1706, line 154: In general please add return type hinting

phpBB/phpbb/messenger/method/base.php
```diff
@@ -0,0 +1,518 @@
...
+	{
+		$this->use_queue = $use_queue;
+	}
+
+	/**
+	 * Resets all the data (address, template file, etc etc) to default
+	 *
+	 * @return void
+	 */
+	abstract public function reset();
```
marc1706, line 171: Maybe it makes sense to define an interface and only include base boilerplate code in the abstract class base?

phpBB/phpbb/messenger/method/base.php
```diff
@@ -0,0 +1,518 @@
...
+	 * @param user $user
+	 * @param queue $queue
+	 * @param path_helper $path_helper
+	 * @param manager $ext_manager
+	 * @param service_collection $twig_extensions_collection
+	 * @param lexer $twig_lexer
+	 * @param string $template_cache_path
+	 * @param string $phpbb_root_path
+	 */
+	function __construct(
```
marc1706, line 109: Missing visibility

phpBB/phpbb/messenger/method/base.php
```diff
@@ -0,0 +1,518 @@
...
+	 */
+	abstract public function get_queue_object_name();
+
+	/**
+	 * Set up subject for mail
+	 *
+	 * @param string	$subject	Email subject
+	 * @return void
+	 */
+	public function subject($subject = '')
```
marc1706, line 193: Generally more type hinting

phpBB/phpbb/messenger/method/base.php
```diff
@@ -0,0 +1,518 @@
...
+	{
+	}
+
+	/**
+	 * Set the reply to address
+	 *
+	 * @param string	$address	Email "Reply to" address
+	 * @return void
+	 */
+	public function replyto($address)
```
marc1706, line 226: Naturally I'd call this reply_to, so why not use this refactoring opportunity?

phpBB/phpbb/messenger/method/base.php
```diff
@@ -0,0 +1,518 @@
...
+	 */
+	public function template($template_file, $template_lang = '', $template_path = '', $template_dir_prefix = '')
+	{
+		$template_dir_prefix = (!$template_dir_prefix || $template_dir_prefix[0] === '/') ? $template_dir_prefix : '/' . $template_dir_prefix;
+
+		$this->setup_template();
+
+		if (!trim($template_file))
+		{
+			trigger_error('No template file for emailing set.', E_USER_ERROR);
```
marc1706, line 262: Should this be hardcoded even though we have the language class in here?

phpBB/phpbb/messenger/method/phpbb_email.php
```diff
@@ -0,0 +1,637 @@
...
+use Symfony\Component\Mime\Header\PathHeader;
+use Symfony\Component\Mime\Header\UnstructuredHeader;
+
+/**
+ * Messenger class
+ */
+class phpbb_email extends base
+{
+	/** @var array */
+	private const PRIORITY_MAP = [
```
marc1706, line 34: Why are we duplicating this from symfony?

phpBB/phpbb/messenger/method/phpbb_email.php
```diff
@@ -0,0 +1,637 @@
...
+
+		if (empty($this->email->getFrom()))
+		{
+			$this->from($board_contact);
+			$headers['From'] = $this->from;
+		}
+
+		$headers['Return-Path'] = new Address($this->config['board_email']);
+		$headers['Sender'] = new Address($this->config['board_email']);
+		$headers['X-Priority'] = sprintf('%d (%s)', $this->mail_priority, self::PRIORITY_MAP[$this->mail_priority]);
```
marc1706, line 346: Can't believe this would will be needed with Symfony mailer and I don#t think it is?

phpBB/phpbb/messenger/method/phpbb_email.php
```diff
@@ -0,0 +1,637 @@
...
+	/**
+	 * Set up extra mail headers
+	 *
+	 * @param string	$header_name	Email header name
+	 * @param string	$header_value	Email header body
+	 * @return void
+	 */
+	public function header($header_name, $header_value)
+	{
+		$header_name = $header_name;
```
marc1706, line 256: This looks weird ...

phpBB/phpbb/messenger/method/phpbb_email.php
```diff
@@ -0,0 +1,637 @@
...
+
+		return $method;
+	}
+
+	/**
+	 * Set email headers
+	 *
+	 * @return bool
+	 */
+	protected function build_header()
```
marc1706, line 327: So the whole point of this is just to allow modification via the event?

phpBB/phpbb/messenger/method/phpbb_email.php
```diff
@@ -0,0 +1,637 @@
...
+	/**
+	 * Send out emails
+	 *
+	 * @return bool
+	 */
+	public function send()
+	{
+		$this->prepare_message();
+
+		$contact_name = html_entity_decode($this->config['board_contact_name'], ENT_COMPAT);
```
marc1706, line 521: unused

phpBB/phpbb/messenger/queue.php
```diff
@@ -0,0 +1,218 @@
...
+	protected $dispatcher;
+
+	/** @var \phpbb\filesystem\filesystem_interface */
+	protected $filesystem;
+
+	/** @var service_collection */
+	protected $messenger_method_collection;
+
+	/** @var int */
+	protected $package_size = 0;
```
marc1706, line 45: Seems unused?

phpBB/config/default/container/services_messenger.yml
```diff
@@ -0,0 +1,57 @@
...
+            - '@messenger.queue'
+            - '@path_helper'
+            - '@ext.manager'
+            - '@template.twig.extensions.collection'
+            - '@template.twig.lexer'
+            - '%core.template.cache_path%'
+            - '%core.php_ext%'
+
+    messenger.method.email:
+        class: phpbb\messenger\method\phpbb_email
```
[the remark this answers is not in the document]
rxu: Because Symfony class has the same name: `use Symfony\Component\Mime\Email`, so the same naming break things.

phpBB/config/default/container/services_messenger.yml
```diff
@@ -0,0 +1,57 @@
...
+        shared: false
+        parent: messenger.method.base
+        calls:
+            - [reset, []]
+            - [set_transport, []]
+        tags:
+            - { name: messenger.method }
+
+    messenger.method.jabber:
+        class: phpbb\messenger\method\phpbb_jabber
```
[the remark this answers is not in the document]
rxu: Just to be inline with above email class naming.

phpBB/phpbb/messenger/method/phpbb_email.php
rxu: Headers aren't set at this point yet. Using Symfony `Email()->priority()` here would mess things as other headers got set later.

phpBB/phpbb/messenger/method/phpbb_email.php
rxu: Yes but also to build a preset of headers.

phpBB/phpbb/messenger/queue.php
rxu: It's used in `phpbb_email` class in `queue->init()` call.

phpBB/config/default/container/services_messenger.yml
```diff
@@ -0,0 +1,57 @@
...
+            - '@template.twig.lexer'
+            - '%core.template.cache_path%'
+            - '%core.php_ext%'
+
+    messenger.method.email:
+        class: phpbb\messenger\method\phpbb_email
+        shared: false
+        parent: messenger.method.base
+        calls:
+            - [reset, []]
```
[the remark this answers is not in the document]
rxu: `reset` method has both init and reset functionality, it just inits things here.

phpBB/config/default/container/services_messenger.yml
```diff
@@ -0,0 +1,57 @@
...
+            - [set_transport, []]
```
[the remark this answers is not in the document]
rxu: It's used in several places, f.e. sending email and process queue. So I've decided just to initialize it from the start so that it would be the same object for different places.

phpBB/config/default/container/services_messenger.yml
```diff
@@ -0,0 +1,57 @@
...
+        tags:
+            - { name: messenger.method }
+
+    messenger.method.jabber:
+        class: phpbb\messenger\method\phpbb_jabber
+        shared: false
+        parent: messenger.method.base
+        calls:
+            - [reset, []]
```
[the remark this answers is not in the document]
rxu: This is something like I've answered above about set_transport. The point is initializing things so they'd be the same through the whole service (and other services). Otherwise you never know the result.

phpBB/includes/acp/acp_board.php
```diff
@@ -474,7 +474,6 @@ function main($id, $mode)
 						'smtp_delivery'			=> array('lang' => 'USE_SMTP',				'validate' => 'bool',	'type' => 'radio:yes_no', 'explain' => true),
 						'smtp_host'				=> array('lang' => 'SMTP_SERVER',			'validate' => 'string',	'type' => 'text:25:50', 'explain' => true),
 						'smtp_port'				=> array('lang' => 'SMTP_PORT',				'validate' => 'int:0:99999',	'type' => 'number:0:99999', 'explain' => true),
-						'smtp_auth_method'		=> array('lang' => 'SMTP_AUTH_METHOD',		'validate' => 'string',	'type' => 'select', 'method' => 'mail_auth_select', 'explain' => true),
```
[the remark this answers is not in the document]
rxu: Yes, Mailer processes this internally.

phpBB/includes/ucp/ucp_profile.php
rxu: To get the email service from it?

rxu: Or just `$phpbb_container->get('messenger.method.email');`?

phpBB/phpbb/notification/method/messenger_base.php
rxu: Because psalm throws error without this, and I don't already remember what error because the commits tests history is lost already.

phpBB/phpbb/template/twig/extension.php
rxu: Because psalm throws error about undefined `twig_slice` function.

phpBB/includes/ucp/ucp_profile.php
marc1706: Yes, the latter one should work just fine and IMHO makes more sense here

phpBB/phpbb/messenger/method/phpbb_email.php
rxu: It turned out it's just to set the custom `X-MSMail-Priority` which doesn't present in Symfony mailer.

phpBB/includes/acp/acp_email.php
rxu: I'm not quite sure where to add a type hint here.

rxu: Do you mean this?
```php
$messenger = (\phpbb\di\service_collection) $phpbb_container->get('messenger.method_collection');
```

rubencm:
```php
/** @var service_collection */
$phpbb_container->get('messenger.method_collection');
```

rxu:
Addressing review issues made this a complete mess which I have no idea what to do with.
So, I give up, sorry.

rxu:
Now couldn't figure out how to fix the Psalm error.

phpBB/phpbb/messenger/method/base.php
rxu: I don't know, I can do that but, to be honest, from my understanding having an interface while having at the same time an abstract class which implements only that one interface is pretty useless, as abstract class should implement all the interface declared methods anyway, so it won't make abstract class any lighter.
It would make sense if f.e. an [abstract] class would implement 2 or more different interfaces but this is not the case. So we'll end up in just having 1 class more (interface) for some reason.
May be just remove some `abstract` declarations which presumably shoudn't be mandatory for any messenger method?

phpBB/phpbb/messenger/method/jabber.php
```diff
@@ -1,118 +1,259 @@
...
+	/**
+	 * Set port to connect to server
+	 * use_ssl flag should be set first
+	 *
+	 * @param int $port Port to connect to server
+	 * @return self
+	 */
+	public function port(int $port = 5222): self
+	{
+		$this->port	= ($port) ? $port : 5222;
```
marc1706, line 194: I guess checking > 0 would make more sense than checking whether it evaluates to true.

phpBB/phpbb/messenger/method/jabber.php
```diff
@@ -1,118 +1,259 @@
...
+
+	/**
+	 * Set username to connect to server
+	 *
+	 * @param string $username Username to connect to server
+	 * @return self
+	 */
+	public function username(string $username = ''): self
+	{
+		if (strpos($username, '@') === false)
```
marc1706, line 213: Use [https://www.php.net/manual/en/function.str-contains.php](url)

phpBB/phpbb/messenger/method/jabber.php
```diff
@@ -1,118 +1,259 @@
...
+	/**
+	 * Set server to connect
+	 * Username should be set first
+	 *
+	 * @param string $server Server to connect
+	 * @return self
+	 */
+	public function server(string $server = ''): self
+	{
+		$this->connect_server = ($server) ? $server : 'localhost';
```
marc1706, line 235:
```suggestion
		$this->connect_server = ($server) ?: 'localhost';
```

phpBB/phpbb/messenger/method/jabber.php
```diff
@@ -228,12 +376,178 @@ function login()
...
+			return;
+		}
+
+		$pos = !empty($this->to) ? count($this->to) : 0;
+		$this->to[$pos]['uid'] = trim($address);
+		$this->to[$pos]['name'] = trim($realname);
+	}
+
+	/**
+	 * {@inheritDoc}
```
marc1706, line 410: what is its parent?

phpBB/phpbb/messenger/method/jabber.php
```diff
@@ -358,31 +663,31 @@ function listen($timeout = 10, $wait = false)
 	}
 
 	/**
-	* Initiates account registration (based on data used for contructor)
-	* @access public
-	* @return bool
-	*/
-	function register()
+	 * Initiates account registration (based on data used for contructor)
```
marc1706, line 666:
```suggestion
	 * Initiates account registration (based on data used for constructor)
```

phpBB/phpbb/messenger/method/jabber.php
```diff
@@ -398,16 +703,16 @@ function send_presence($message = '', $type = '', $unavailable = false)
...
-	*/
-	function response($xml)
+	 * This handles all the different XML elements
+	 *
+	 * @param array $xml
+	 * @return bool
+	 */
+	function response(array $xml): bool
 	{
 		if (!is_array($xml) || !count($xml))
```
marc1706, line 717: `is_array` shouldn't be needed thanks to type hint

phpBB/phpbb/notification/method/messenger_base.php
marc1706: I just tried without the suppress and couldn't reproduce a psalm error here so I'd recommend removing it

phpBB/phpbb/template/twig/extension.php
marc1706: Also can't reproduce that but you could also try using the underlying function:
`return CoreExtension::slice($env->getCharset(), $item, $start, $end, $preserveKeys);`

phpBB/phpbb/template/twig/extension.php
```diff
@@ -213,6 +214,7 @@ public function lang_js(): string
 	{
 		$args = func_get_args();
 
+		/** @psalm-suppress UndefinedFunction */
```
marc1706, line 217: Same here, maybe use the underlying call instead of the function?

marc1706 (changes requested):

phpBB/includes/acp/acp_inactive.php
```diff
@@ -207,37 +197,37 @@ function main($id, $mode)
...
+									$messenger_method->set_addresses($row);
+									$messenger_method->anti_abuse_headers($config, $user);
+									$messenger_method->assign_vars([
+										'USERNAME'		=> html_entity_decode($row['username'], ENT_COMPAT),
+										'REGISTER_DATE'	=> $user->format_date($row['user_regdate'], false, true),
+										'U_ACTIVATE'	=> generate_board_url() . "/ucp.$phpEx?mode=activate&u=" . $row['user_id'] . '&k=' . $row['user_actkey'],
+									]);
+
+									$messenger_method->send();
+									$messenger_method->save_queue();
```
marc1706, line 222: It seems very inefficient to have the save queue call inside the loops here instead of after the do while loop.

phpBB/includes/acp/acp_inactive.php
```diff
@@ -252,6 +241,11 @@ function main($id, $mode)
 					}
 					$db->sql_freeresult($result);
 
+					foreach ($messenger_collection_iterator as $messenger_method)
```
marc1706, line 244: I don't think it makes sense to call `save_queue` if the queue was not changed. Therefore I'd recommend placing it before the update query on line 231.

marc1706 (changes requested):

phpBB/includes/acp/acp_inactive.php
rxu: But then it wan't change anything in comparison with initial code as `save_queue()` will be called exactly the same times as before, for every used and for every messenger method. Currently it'll be called just as many times as many messenger methods exist.

marc1706: If the query returned an empty result, your code will still call save queue for every messenger method. However, if you place your code in the same location as the previous save_queue call, then it'll will only do that in case the query did not return an empty result. I'm not saying that you should remove the foreach.

marc1706:
> Also I don't like using `NOFIFY_BOTH = 2` constant as it limits possible messenger methods to the built-in email&jabber only. It would be better using something like `$row['user_notify_type'] = NOTIFY_EMAIL | NOTIFY_IM | NOTIFY_CUSTOM_N` and then check it like `if (($row['user_notify_type'] & NOTIFY_CUSTOM_N) == NOTIFY_CUSTOM_N)`, but that's not directly related to introducing Mailer and would be something for another PR.

I agree that we should have a look at how to better handle this and I also agree that this might be best to handle in a follow up ticket and PR.

marc1706 (approved):

PR #6530, opened 2023-09-09
[ticket/17151] Make macros available for extensions

rxu:
Also add tests and error reporting.

PHPBB3-17151

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-17151

### DIFF
--- a/tests/functional/fixtures/ext/foo/bar/mcp/main_info.php
+++ b/tests/functional/fixtures/ext/foo/bar/mcp/main_info.php
@@ -11,25 +11,17 @@
 *
 */
 
-namespace foo\bar\acp;
-
-/**
-* @ignore
-*/
-if (!defined('IN_PHPBB'))
-{
-	exit;
-}
+namespace foo\bar\mcp;
 
 class main_info
 {
 	function module()
 	{
 		return array(
-			'filename'	=> '\foo\bar\acp\main_module',
-			'title'		=> 'ACP_FOOBAR_TITLE',
+			'filename'	=> '\foo\bar\mcp\main_module',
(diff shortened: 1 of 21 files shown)


rxu:
Also currently forms macros is not used for building `select`s, but that can be resolved later in a separate PR.

tests/functional/fixtures/ext/foo/bar/styles/prosilver/template/foobar_mcp.html
```diff
@@ -0,0 +1,3 @@
+<!-- INCLUDE overall_header.html -->
```
marc1706, line 1: Should use twig syntax instead

tests/functional/fixtures/ext/foo/bar/mcp/main_info.php
```diff
@@ -0,0 +1,28 @@
...
+*
+*/
+
+namespace foo\bar\mcp;
+
+class main_info
+{
+	function module()
+	{
+		return array(
```
marc1706, line 20: Short array syntax in this file

tests/functional/fixtures/ext/foo/bar/ucp/main_info.php
```diff
@@ -19,9 +19,9 @@ function module()
 	{
 		return array(
```
marc1706, line 20: Would also opt for changing this file to short array syntax while you're changing it

marc1706 (changes requested):

marc1706 (approved):

PR #6544, opened 2023-09-23
[ticket/15851] Automatic update downloader

CHItA:
PHPBB3-15851

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-15851

### DIFF
--- a/phpBB/language/en/install.php
+++ b/phpBB/language/en/install.php
@@ -222,11 +222,11 @@ $lang = array_merge($lang, array(
 		<p>We noticed that the last update of your phpBB installation hasn’t been completed. Visit the <a href="%1$s" title="%1$s">database updater</a>, ensure <em>Update database only</em> is selected and click on <strong>Submit</strong>. Don\'t forget to delete the "install"-directory after you have updated the database successfully.</p>',
 
 	// Auto update
-	'COULD_NOT_DOWNLOAD_UPDATE_PACKAGE'		=> 'Failed to download the update package.',
-	'COULD_NOT_DOWNLOAD_UPDATE_SIGNATURE'	=> 'Failed to download the update package signature.',
+	'UPDATE_PACKAGE_DOWNLOAD_FAILURE'		=> 'Failed to download the update package.',
+	'UPDATE_SIGNATURE_DOWNLOAD_FAILURE'		=> 'Failed to download the update package signature.',
 	'UPDATE_SIGNATURE_INVALID'				=> 'The update package is corrupted.',
-	'COULD_NOT_EXTRACT_UPDATE'				=> 'Could not extract files from the update package.',
-	'COULD_NOT_WRITE_UPDATE_FILES'			=> 'Could not copy files from the update package.',
+	'UPDATE_PACKAGE_EXTRACT_FAILURE'		=> 'Could not extract files from the update package.',
+	'UPDATE_FILES_COPY_FAILURE'				=> 'Could not copy files from the update package.',
 
 	//
 	// Server data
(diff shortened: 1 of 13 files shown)


phpBB/phpbb/update/get_updates.php
```diff
@@ -0,0 +1,169 @@
...
+ */
+
+namespace phpbb\update;
+
+use GuzzleHttp\Client;
+use GuzzleHttp\Exception\GuzzleException;
+use phpbb\filesystem\exception\filesystem_exception;
+use phpbb\filesystem\filesystem_interface;
+use SodiumException;
+use ZipArchive;
```
rubencm, line 21: composer should include php-zip as requirement
There is a `functions_compress.php` in includes folder too that can be used if the server don't have zip installed (or just dont allow updates)

CHItA: i doubt that would actually work if this is not working.

phpBB/phpbb/update/get_updates.php
```diff
@@ -0,0 +1,169 @@
...
+
+	/**
+	 * Validate the downloaded file.
+	 *
+	 * @param string $file_path			Path to the download.
+	 * @param string $signature_path	The signature file.
+	 *
+	 * @return bool Whether the signature is correct or not.
+	 */
+	public function validate(string $file_path, string $signature_path): bool
```
Derky, line 92: Please return the reason why the signature is not valid. (Download doesn't exist, signature doesn't exist, unable to decode base64, what SodiumException occurred)

CHItA:
97% percent of those cannot even happen. The rest is not all that useful
IMHO.

phpBB/phpbb/update/get_updates.php
CHItA: that doesn't really make any sense to me, we only care if the signature is valid or not. Further more, how this is used, it is impossible to not have the file.

phpBB/language/en/install.php
```diff
@@ -223,6 +223,13 @@
 
 		<p>We noticed that the last update of your phpBB installation hasn’t been completed. Visit the <a href="%1$s" title="%1$s">database updater</a>, ensure <em>Update database only</em> is selected and click on <strong>Submit</strong>. Don\'t forget to delete the "install"-directory after you have updated the database successfully.</p>',
 
+	// Auto update
+	'COULD_NOT_DOWNLOAD_UPDATE_PACKAGE'		=> 'Failed to download the update package.',
+	'COULD_NOT_DOWNLOAD_UPDATE_SIGNATURE'	=> 'Failed to download the update package signature.',
+	'UPDATE_SIGNATURE_INVALID'				=> 'The update package is corrupted.',
+	'COULD_NOT_EXTRACT_UPDATE'				=> 'Could not extract files from the update package.',
+	'COULD_NOT_WRITE_UPDATE_FILES'			=> 'Could not copy files from the update package.',
```
marc1706, line 231: How about these more uniform ones?
```suggestion
	'UPDATE_PACKAGE_DOWNLOAD_FAILURE'		=> 'Failed to download the update package.',
	'UPDATE_SIGNATURE_DOWNLOAD_FAILURE'	=> 'Failed to download the update package signature.',
	'UPDATE_SIGNATURE_INVALID'				=> 'The update package is corrupted.',
	'UPDATE_PACKAGE_EXTRACT_FAILURE'				=> 'Could not extract files from the update package.',
	'UPDATE_FILES_COPY_FAILURE'			=> 'Could not copy files from the update package.',
```

marc1706 (changes requested):

phpBB/phpbb/update/get_updates.php
marc1706: It wouldn't, I did however add the composer requirement now.

marc1706 (approved):

PR #6571, opened 2023-12-03
[ticket/15214] Add event & functionality for assigning template event priority

toxyy:
Event added to allow template events to be assigned priority per extension, event location chosen so that it only fires once.
Twig node event class refactored to allow template event priority assignment, compile calls are deferred until all locations are processed per extension namespace.
Priority precedence mirrors Symfony priority, with higher numbers being placed at the beginning of the array.
Duplicate priority assignment will currently have the later events compiled before the others.

PHPBB3-15214

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-15214

### DIFF
--- a/phpBB/phpbb/template/twig/node/event.php
+++ b/phpBB/phpbb/template/twig/node/event.php
@@ -21,12 +21,16 @@ class event extends \Twig\Node\Node
 	*/
 	protected $listener_directory = 'event/';
 
-	/** @var \Twig\Environment */
+	/** @var \phpbb\template\twig\environment */
 	protected $environment;
 
-	public function __construct(\Twig\Node\Expression\AbstractExpression $expr, \phpbb\template\twig\environment $environment, $lineno, $tag = null)
+	/** @var array */
+	protected $template_event_priority_array;
+
+	public function __construct(\Twig\Node\Expression\AbstractExpression $expr, \phpbb\template\twig\environment $environment, $lineno, $tag = null, $template_event_priority_array = [])
 	{
 		$this->environment = $environment;
+		$this->template_event_priority_array = $template_event_priority_array;
 
 		parent::__construct(array('expr' => $expr), array(), $lineno, $tag);
 	}
@@ -42,9 +46,18 @@ class event extends \Twig\Node\Node
 
 		$location = $this->listener_directory . $this->getNode('expr')->getAttribute('name');
 
+		$compiler_steps = [];
+
 		foreach ($this->environment->get_phpbb_extensions() as $ext_namespace => $ext_path)
 		{
 			$ext_namespace = str_replace('/', '_', $ext_namespace);
+
+			if (isset($this->template_event_priority_array[$ext_namespace][$location]))
+			{
+				$priority_key = $this->template_event_priority_array[$ext_namespace][$location];
+			}
+
+			$compiler_calls = [];
 
 			if ($this->environment->isDebug())
 			{
@@ -52,7 +65,7 @@ class event extends \Twig\Node\Node
 				//  templates on page load rather than at compile. This is
 				//  slower, but makes developing extensions easier (no need to
 				//  purge the cache when a new event template file is added)
-				$compiler
+				$compiler_calls[] = fn() => $compiler
 					->write("if (\$this->env->getLoader()->exists('@{$ext_namespace}/{$location}.html')) {\n")
 					->indent()
 				;
@@ -60,7 +73,7 @@ class event extends \Twig\Node\Node
 
 			if ($this->environment->isDebug() || $this->environment->getLoader()->exists('@' . $ext_namespace . '/' . $location . '.html'))
 			{
-				$compiler
+				$compiler_calls[] = fn() => $compiler
 					->write("\$previous_look_up_order = \$this->env->getNamespaceLookUpOrder();\n")
 
 					// We set the namespace lookup order to be this extension first, then the main path
@@ -72,10 +85,38 @@ class event extends \Twig\Node\Node
 
 			if ($this->environment->isDebug())
 			{
-				$compiler
+				$compiler_calls[] = fn() => $compiler
 					->outdent()
 					->write("}\n\n")
 				;
+			}
+
+			if (!empty($compiler_calls))
+			{
+				if (isset($priority_key))
+				{
+					if (array_key_exists($priority_key, $compiler_steps))
+					{
+						array_splice($compiler_steps, $priority_key, 0, [$compiler_calls]);
+					}
+					else
+					{
+						$compiler_steps[$priority_key] = $compiler_calls;
+					}
+				}
+				else
+				{
+					array_unshift($compiler_steps, $compiler_calls);
+				}
+			}
+		}
+
+		krsort($compiler_steps);
+		foreach ($compiler_steps as $ext_namespace_steps)
+		{
+			foreach ($ext_namespace_steps as $step)
+			{
+				$step();
 			}
 		}
 	}
--- a/phpBB/phpbb/template/twig/node/event.php
+++ b/phpBB/phpbb/template/twig/node/event.php
@@ -51,9 +51,12 @@ class event extends \Twig\Node\Node
 		// Group and sort extension template events in according to their priority (0 by default if not set)
 		foreach ($this->environment->get_phpbb_extensions() as $ext_namespace => $ext_path)
 		{
-			$ext_namespace = str_replace('/', '_', $ext_namespace);
-			$priority_key = $this->template_event_priority_array[$ext_namespace][$location] ?? 0;
-			$template_events[$priority_key][] = $ext_namespace;
+			if ($this->environment->isDebug() || $this->environment->getLoader()->exists('@' . $ext_namespace . '/' . $location . '.html'))
+			{
+				$ext_namespace = str_replace('/', '_', $ext_namespace);
+				$priority_key = $this->template_event_priority_array[$ext_namespace][$location] ?? 0;
+				$template_events[$priority_key][] = $ext_namespace;
+			}
 		}
 		krsort($template_events);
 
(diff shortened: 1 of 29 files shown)


toxyy:
I don't know if the compiler object's write methods change the state of the compiler object itself. If so then `$compiler` should be passed as `&$compiler` in the `use` of the anonymous functions, but as it's currently written, it works fine.

phpBB/phpbb/template/twig/node/event.php
```diff
@@ -42,40 +46,80 @@ public function compile(\Twig\Compiler $compiler)
...
+						->write("}\n\n")
+					;
+				};
+			}
+
+			if (!empty($compiler_calls))
+			{
+				if (isset($this->template_event_priority_array[$ext_namespace][$location]))
+				{
+					$priority_key = $this->template_event_priority_array[$ext_namespace][$location];
```
rxu, line 99: Priority may not be set for the majority of template events, hence this assignment will cause `Undefined array key` PHP warnings. So it probably needs to be something like
```php
$priority_key = $this->template_event_priority_array[$ext_namespace][$location] ?? 0;
```

phpBB/phpbb/template/twig/tokenparser/event.php
```diff
@@ -18,14 +18,38 @@ class event extends \Twig\TokenParser\AbstractTokenParser
...
+		* @var	array	template_event_priority_array	Array with template event priority assignments per extension namespace
+		* @since 3.3.12-RC1
+		*/
+		if ($this->phpbb_dispatcher)
+		{
+			$vars = array('template_event_priority_array');
+			extract($this->phpbb_dispatcher->trigger_event('core.twig_tokenparser_constructor', compact($vars)));
+		}
+
+		$this->template_event_priority_array = $template_event_priority_array;
```
rxu, line 52: It probably makes sense to `unset($template_event_priority_array);` as it's not used anymore.

phpBB/phpbb/template/twig/node/event.php
toxyy: That's fine as is, if it's not set, it goes to [the else on line 112](https://github.com/phpbb/phpbb/blob/31fb27e8d88a1ad51ce1dbdeb5845054e14262bb/phpBB/phpbb/template/twig/node/event.php#L112):

```php
if (isset($this->template_event_priority_array[$ext_namespace][$location]))
{
	...
}
else
{
	array_unshift($compiler_steps, $compiler_calls);
}
```

If it's unset, `array_unshift` places the current `$compiler_calls` at the front of the `$compiler_steps` array. This ensures that the highest number key always takes precedence. In the screenshot I included in [my first comment on this pr](https://github.com/phpbb/phpbb/pull/6571#issuecomment-1837608225), you can see that the other extension's `overall_header_head_append` are indeed executed in their original order, and they do not make use of this event.

rxu: That's ok, but it'll PHP erroring on assignment itself.

toxyy:
All I did was change what's within a docblock comment... that test should not have failed

rxu:
That's randon network issue, unrelated.

toxyy:
No matter, I think the format of this docblock update is cleaner anyways

phpBB/phpbb/template/twig/node/event.php
```diff
@@ -42,40 +46,80 @@ public function compile(\Twig\Compiler $compiler)
...
+						array_splice($compiler_steps, $priority_key, 0, [$compiler_calls]);
+					}
+					else
+					{
+						$compiler_steps[$priority_key] = $compiler_calls;
+					}
+				}
+				else
+				{
+					array_unshift($compiler_steps, $compiler_calls);
```
rxu, line 112: Any priority values set in event listeners will be changed/dropped here as far as with the `array_unshift` all numerical array keys will be modified to start counting from zero. https://www.php.net/manual/en/function.array-unshift.php
So the logic shuld be adjusted somehow to avoid that.

phpBB/phpbb/template/twig/node/event.php
```diff
@@ -42,40 +46,80 @@ public function compile(\Twig\Compiler $compiler)
...
+				};
 			}
 
 			if ($this->environment->isDebug())
 			{
-				$compiler
-					->outdent()
-					->write("}\n\n")
-				;
+				$compiler_calls[] = function() use($compiler) {
```
rxu, line 87: It seems that this call is intended to be the last in the queue but will be messed up later with `krsort()`; so that won't kept the last call.

phpBB/phpbb/template/twig/node/event.php
```diff
@@ -42,40 +46,80 @@ public function compile(\Twig\Compiler $compiler)
...
 			{
 				// If debug mode is enabled, lets check for new/removed EVENT
 				//  templates on page load rather than at compile. This is
 				//  slower, but makes developing extensions easier (no need to
 				//  purge the cache when a new event template file is added)
-				$compiler
-					->write("if (\$this->env->getLoader()->exists('@{$ext_namespace}/{$location}.html')) {\n")
-					->indent()
-				;
+				$compiler_calls[] = function() use($compiler, $ext_namespace, $location) {
```
rxu, line 63: This call always should be the 1st one, hence always at the top of "krsorted" array.

rxu:
To sum my all above comments, I feel like simplified all the stuff and below code forks just fine.

```php
	public function compile(\Twig\Compiler $compiler)
	{
		$compiler->addDebugInfo($this);

		$location = $this->listener_directory . $this->getNode('expr')->getAttribute('name');

		$template_events = [];
		foreach ($this->environment->get_phpbb_extensions() as $ext_namespace => $ext_path)
		{
			$ext_namespace = str_replace('/', '_', $ext_namespace);
			$priority_key = $this->template_event_priority_array[$ext_namespace][$location] ?? 0;
			$template_events[$priority_key][] = $ext_namespace;
		}
		krsort($template_events);

		foreach ($template_events as $priority => $events)
		{
			foreach ($events as $ext_namespace)
			{
				if ($this->environment->isDebug() || $this->environment->getLoader()->exists('@' . $ext_namespace . '/' . $location . '.html'))
				{
					if ($this->environment->isDebug())
					{
						// If debug mode is enabled, lets check for new/removed EVENT
						//  templates on page load rather than at compile. This is
						//  slower, but makes developing extensions easier (no need to
						//  purge the cache when a new event template file is added)
						$compiler
							->write("if (\$this->env->getLoader()->exists('@{$ext_namespace}/{$location}.html')) {\n")
							->indent();
					}

					$compiler
						->write("\$previous_look_up_order = \$this->env->getNamespaceLookUpOrder();\n")

						// We set the namespace lookup order to be this extension first, then the main path
						->write("\$this->env->setNamespaceLookUpOrder(array('{$ext_namespace}', '__main__'));\n")
						->write("\$this->env->loadTemplate('@{$ext_namespace}/{$location}.html')->display(\$context);\n")
						->write("\$this->env->setNamespaceLookUpOrder(\$previous_look_up_order);\n");

					if ($this->environment->isDebug())
					{
						$compiler
							->outdent()
							->write("}\n\n");
					}
				}
			}
		}
	}
```

rxu:
I guess the issue with the commit messages here is an empty line needed between the explanation line and `PHPBB3-15214` line.

toxyy:
> I guess the issue with the commit messages here is an empty line needed between the explanation line and `PHPBB3-15214` line.

There are also some other failures like here:
https://github.com/phpbb/phpbb/actions/runs/7152664064/job/19478240191?pr=6571#step:10:1060

I'll do a force push to fix the messages in a bit.

toxyy:
Yeah looks like it doesn't like the added tests right now

toxyy:
@rxu Yeah looks like it doesn't like the added tests right now

rxu:
That's weird as everything works when installing both of the test extensions on a local test board and yaml for both of `services.yml` is valid.

toxyy:
@rxu That solved the errors we were seeing before, but we've got a new set of failures 😓

rxu:
I'm afraid Win+PHP 8.3 errors aren't related to this PR. Will dig into it a bit.

toxyy:
Tests passed earlier in this PR so I'm not sure what changed.

toxyy:
Also, I've been thinking.

Currently, all events are added to the same priority of 0 in the order that they would originally be run. I'm not sure how Symfony events work exactly, but would there be any scenario in which someone would want to place a template event in the middle of those? Technically, that would mean that they cannot all be at priority key 0. That is also installation dependent, as the priority keys would change based on what extensions you have enabled, assuming multiple extensions use the same events.

Would it be beneficial to somehow make it so you could say "if extension X is installed, make my template event compile before theirs"? That could be solved with a priority key, yes, but lets say you also want all other extensions to execute before yours as well - meaning you specifically want your template event to compile right before extension X's event of the same name (though anyone else doing the same would run into default behavior, of course).

Let's take navlinks for example. Maybe you want to place your link right before the other nav links of extension X, but you don't want it to be the first links added. With a priority key, the only way to get it before extension X would be to make it compile first, meaning your links would be first. That doesn't seem versatile.

Is what I'm explaining making sense, and is that a valid use case to try to add functionality for?

rxu:
I'm not sure I fully got the point in question.
With your PR you're able to set priority for any other extension template event (by other authors, unlike core events). So if you want to change, f.e., quick links extensions order with your extension you just set priority keys as you wish for another extension(s) template events.

toxyy:
Ah yes you could do that. I guess that case would be solved then.

rxu:
And the error is:
```
PHP FATAL ERROR:  UNCAUGHT TWIG\ERROR\LOADERERROR: UNABLE TO FIND TEMPLATE "@FOO_BAR/EVENT/NAVBAR_HEADER_QUICK_LINKS_AFTER.HTML" (LOOKED INTO: ) IN "NAVBAR_HEADER.HTML" AT LINE 75. IN D:\A\PHPBB3\PHPBB3\PHPBB\VENDOR\TWIG\TWIG\SRC\LOADER\FILESYSTEMLOADER.PHP

D:\a\phpbb3\phpbb3\tests\test_framework\phpbb_functional_test_case.php:983
D:\a\phpbb3\phpbb3\tests\test_framework\phpbb_functional_test_case.php:143
D:\a\phpbb3\phpbb3\tests\test_framework\phpbb_functional_test_case.php:807
D:\a\phpbb3\phpbb3\tests\functional\notification_test.php:46
```

That's weird again as this template event shouldn't be involved in notification tests.

toxyy:
If that said that it was for `@FOO_FOO` I'd say to also add `foo/foo/styles/prosilver/template/` to `$fixtures` but I'm at a loss currently.

rxu:
The issue revealed by this PR is the twig compiled cache folder was never purged in test environment. Will send a fix PR to this PR soon.

toxyy:
`Failed asserting that 'PHP FATAL ERROR:  UNCAUGHT TWIG\ERROR\LOADERERROR: THERE ARE NO REGISTERED PATHS FOR NAMESPACE "FOO_BAR" IN "NAVBAR_HEADER.HTML" AT LINE 75. IN D:\A\PHPBB\PHPBB\PHPBB\VENDOR\TWIG\TWIG\SRC\LOADER\FILESYSTEMLOADER.PHP:227`

😄

rxu:
Hmm. Tests were passing successfully here https://github.com/rxu/phpbb3/actions/runs/7157430146 on Win+PHP 8.3 with this code. One more weirdness :) Okay, will look into it furher.

rxu:
@toxyy Could you try adding these 3 lines to `.github/workflows/tests.yml` https://github.com/rxu/phpbb3/commit/0e2b34963d08889235c9754557bf57673c3e7106#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957fR171-R173
Tests pass here https://github.com/rxu/phpbb3/actions/runs/7158983020

toxyy:
@rxu same error 😄 
` Failed asserting that 'PHP FATAL ERROR:  UNCAUGHT TWIG\ERROR\LOADERERROR: THERE ARE NO REGISTERED PATHS FOR NAMESPACE "FOO_BAR" IN "NAVBAR_HEADER.HTML" AT LINE 75. IN D:\A\PHPBB\PHPBB\PHPBB\VENDOR\TWIG\TWIG\SRC\LOADER\FILESYSTEMLOADER.PHP:227`

rxu:
Nice.

toxyy:
@marc1706 maybe you have some ideas? Not sure who else has worked with the tests a lot

toxyy:
@rxu It worked this time 😄 thanks for the help

rxu:
Great. The issue is that after the `purge_cache()` call tests go forward fast so that the next one got executed before all cache files are finished to delete. Hence the error. `sleep()` solves it.

marc1706:
As noted on discord, the PR might result in BC breaks for some and adds quite a bit of changes to behavior and introduces new functionality so this shouldn't aim for 3.3.x but rather 4.0.

rxu:
@toxyy mind to rebase vs `master`?

toxyy:
@rxu there are a lot of conflicts that I'll need to go through, but sure. I also need to change the event's since version.

rxu:
@toxyy managed to resolve rebasing issues here https://github.com/rxu/phpbb3/tree/ticket/15214-toxxyy-4
That would probably require reconstructing your current PR branch (f.e. hard resetting to the latest core commit, pulling from `15214-toxxyy-4` branch and then force-pushing to your origin repo).

toxyy:
@rxu thanks, I'll try to get to that this week.

rxu:
> @rxu thanks, I'll try to get to that this week.
@toxyy It's been a long time since that :) I can send a replacement PR while preserving your commits authorship, if you're fine with it.

toxyy:
> @toxyy It's been a long time since that :) I can send a replacement PR while preserving your commits authorship, if you're fine with it.
@rxu Yes it's been a very long week :)
Go ahead, I think this is a great addition for phpBB, I'd rather see it in the codebase sooner by you than later by me when I can get to it.

rxu:
Done, so this PR can be closed.

PR #6595, opened 2024-04-10
Default avatar display

phpbbtr:
Default avatar display
my code changes


meis2m:
i have an idea for New Default Avatars

1- The default avatar for each user will be a generated image with the username's first letter on a colored background. If the user name changes, the avatar will change as well. These avatars support round and square rendering.
some thing like that
https://i.stack.imgur.com/ckB3G.png

2- displaying avatars in forumlist body and whoi is online in index too.

marc1706:
Thanks for your suggestion. This functionality was implemented as part of this ticket and merged into master for 4.0.0 alpha 2:
https://tracker.phpbb.com/browse/PHPBB-17566

PR #6616, opened 2024-05-17
[ticket/17317] Change button text for pm recipient to increase readability

danieltj27:
I've updated the text on the submit button used to remove recipients from a PM that is currently being composed. It used to just be a small `x` that was hard (for me and of course others) to see.

I've updated it to say `Remove` which is already an existing language string so no new translations are required for this change. I also added a 2px padding so the text isn't flush to the edges of the button.

It makes it more readable and actually looks styled as opposed to looking like an element that was forgotten about. I think this makes the interface much better to use.

PHPBB3-17317

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-17317

### DIFF
--- a/phpBB/styles/prosilver/template/posting_pm_header.html
+++ b/phpBB/styles/prosilver/template/posting_pm_header.html
@@ -77,8 +77,12 @@
 						<ul class="recipients">
 							<!-- BEGIN to_recipient -->
 							<li>
-								<!-- IF to_recipient.IS_GROUP --><a href="{to_recipient.U_VIEW}"><strong>{to_recipient.NAME}</strong></a><!-- ELSE -->{to_recipient.NAME_FULL}<!-- ENDIF -->&nbsp;
-								<!-- IF not S_EDIT_POST --><input type="submit" name="remove_{to_recipient.TYPE}[{to_recipient.UG_ID}]" value="x" class="button2" /><!-- ENDIF -->
+								<!-- IF not S_EDIT_POST -->
+									<button type="submit" name="remove_{to_recipient.TYPE}[{to_recipient.UG_ID}]">
+										<i class="icon fa-times icon-red" aria-hidden="true"></i><span class="sr-only">{L_REMOVE}</span>
+									</button>
+								<!-- ENDIF -->
+								<!-- IF to_recipient.IS_GROUP --><a href="{to_recipient.U_VIEW}"><strong>{to_recipient.NAME}</strong></a><!-- ELSE -->{to_recipient.NAME_FULL}<!-- ENDIF -->
 							</li>
 							<!-- END to_recipient -->
 						</ul>
--- a/phpBB/styles/prosilver/template/posting_pm_header.html
+++ b/phpBB/styles/prosilver/template/posting_pm_header.html
@@ -31,7 +31,11 @@
 					<ul class="recipients">
 						<!-- BEGIN to_recipient -->
 						<li>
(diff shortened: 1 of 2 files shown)


danieltj27:
I've added a second commit because I accidentally changed `padding: 0;` to `padding: 2px 0;` when it should have been changed to `padding: 0 2px;` instead.

Button need left and right padding, not top and bottom.

danieltj27:
I've updated the change to what @iMattPro has [suggested in the ticket](https://tracker.phpbb.com/browse/PHPBB3-17317?focusedCommentId=65780&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-65780) because I agree that it's a better change compared to simply having a default button with the word remove in it.

phpBB/styles/prosilver/template/posting_pm_header.html
```diff
@@ -48,7 +52,11 @@
 					<ul class="recipients">
 						<!-- BEGIN bcc_recipient -->
 						<li>
-							<!-- IF not S_EDIT_POST --><input type="submit" name="remove_{bcc_recipient.TYPE}[{bcc_recipient.UG_ID}]" value="x" class="button2" /><!-- ENDIF -->
+							<!-- IF not S_EDIT_POST -->
+								<button type="submit" name="remove_{bcc_recipient.TYPE}[{bcc_recipient.UG_ID}]">
+									<i class="icon fa-times icon-red" aria-hidden="true"></i><span class="sr-only">{L_REMOVE}</span>
+								</button>
+							<!-- ENDIF -->
 							<!-- IF bcc_recipient.IS_GROUP --><a href="{bcc_recipient.U_VIEW}" style="color: {{ bcc_recipient.COLOUR }}"><strong>{bcc_recipient.NAME}</strong></a><!-- ELSE -->{bcc_recipient.NAME_FULL}<!-- ENDIF -->
```
iMattPro, line 60: There's a third one of these down at line 73/81 that needs to be updated.

danieltj27: Totally missed that! Done.

danieltj27:
Latest commit addresses both issues in review by @iMattPro.

I'd say it's time to merge and have a brew 🙂

phpBB/styles/prosilver/theme/common.css
```diff
@@ -780,10 +780,10 @@ fieldset.fields1 dd.recipients {
 	margin-left: 1em;
 }
 
-fieldset.fields1 ul.recipients  input.button2 {
+fieldset.fields1 ul.recipients input.button2 {
```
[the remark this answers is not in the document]
iMattPro: None of this rule is needed anymore.

phpBB/styles/prosilver/theme/common.css
```diff
@@ -780,7 +780,7 @@ fieldset.fields1 dd.recipients {
...
```
iMattPro, line 783: None of this rule is needed anymore.

danieltj27: Removed in latest

iMattPro (approved):

marc1706:
The commit hook check found an issue again. You had for example:
```
[ticket/17317] Some change
PHPBB3-17317
```

However, the commit message hook requires there to be either a comment or an empty line before the PHPBB3-17317:
```
[ticket/17317] Some change

PHPBB3-17317
```

danieltj27:
Thanks, sorry about that. I've done that on a couple of commits but now I know for next time!

marc1706 (approved):

marc1706:
@danieltj27 On a side note, I have also adapted the changes in master. However, the only difference is that the icon does use a normal cursor for the button now. Might want to check this out and maybe adapt for it accordingly in master.

PR #6618, opened 2024-05-17
[ticket/17312] Add column for tracking user last active time

marc1706:
Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB3-17312

### DIFF
--- a/phpBB/phpbb/db/migration/data/v33x/add_user_last_active.php
+++ b/phpBB/phpbb/db/migration/data/v33x/add_user_last_active.php
@@ -29,7 +29,7 @@ class add_user_last_active extends migration
 		return [
 			'add_columns'	=> [
 				$this->table_prefix . 'users'	=> [
-					'user_last_active'		=> ['TIMESTAMP', 0],
+					'user_last_active'		=> ['TIMESTAMP', 0, 'after' => 'user_lastvisit'],
 				],
 			],
 		];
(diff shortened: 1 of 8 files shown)


rxu:
`function user_add()` needs to be adjusted as well (`$additional_vars` array).

rxu:
Tested on localhost.
Migration worked fine on update, updating of `user_last_active` field works correctly as well as users sorting in memberlist and new posts search.
As for the `user_lastvisit`: having account with user_lastvisit Fri Feb 10 2023 14:48:45 GMT+0000, logged in with that account, browsed through a test board around a bit and `user_lastvisit` didn't update. Logged out, logged in again and `user_lastvisit` got updated with that account (synced with `session_last_visit` column in the sessions table).
So I guess it seems to be working as intended: `user_last_active` updates during the current session while `user_lastvisit` updates once per session after logout/login.

marc1706:
@rxu I've added it to user_add as well as where it made sense and user_lastvisit was used for the last active time.

phpBB/phpbb/db/migration/data/v33x/add_user_last_active.php
```diff
@@ -0,0 +1,79 @@
...
+			'\phpbb\db\migration\data\v33x\v3311',
+		];
+	}
+
+	public function update_schema()
+	{
+		return [
+			'add_columns'	=> [
+				$this->table_prefix . 'users'	=> [
+					'user_last_active'		=> ['TIMESTAMP', 0, 'after' => 'user_lastvisit'],
```
Derky, line 32: I think this column can also use an index for query optimization.

rubencm: For which query?

rxu:
Will it make sense to replace `user_lasvisit` with `user_last_active` in acp_bots, acp_inactive, acp_prune?

phpBB/phpbb/db/migration/data/v33x/add_user_last_active.php
Derky: For the usages in variables `$sort_key_sql['l']` and `$sql_where` in `memberlist.php`

Derky:
> Will it make sense to replace `user_lasvisit` with `user_last_active` in acp_bots, acp_inactive, acp_prune?

If it does, maybe we can do that in a separate ticket for the next version so we don't risk breaking features just before the release.

phpBB/phpbb/db/migration/data/v33x/add_user_last_active.php
marc1706: An index doesn't actually add any improvement to the query as far as I can see. There is also no index for `user_lastvisit`.

marc1706:
I've used user last active in places where it mentions the user's last active time already (in the acp) but have left it out in places where it said last visit or might affect other functionality.

Derky (approved):

phpBB/phpbb/db/migration/data/v33x/add_user_last_active.php
Derky: Oh weird, thanks for checking.

PR #6637, opened 2024-06-11
[ticket/17337] Fix mysqli driver is missing transaction begin statement - 3.3.x

rxu:
For `3.3.x`.

Checklist:

- [x] Tests pass

Tracker ticket:

<a href="https://tracker.phpbb.com/browse/PHPBB-17337">PHPBB-17337</a>.

### DIFF
--- a/phpBB/phpbb/db/driver/mysqli.php
+++ b/phpBB/phpbb/db/driver/mysqli.php
@@ -155,7 +155,9 @@ class mysqli extends \phpbb\db\driver\mysql_base
 		switch ($status)
 		{
 			case 'begin':
-				return @mysqli_autocommit($this->db_connect_id, false);
+				@mysqli_autocommit($this->db_connect_id, false);
+				$result = @mysqli_begin_transaction($this->db_connect_id);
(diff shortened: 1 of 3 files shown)


marc1706:
I think one should note that autocommit seems like the way to go with MyISAM while begin_transaction seems like the correct way for InnoDB. Having both is kind of weird to be honest. It might make sense to switch to only supporting InnoDB and only apply this change in master.

rxu:
MyISAM doesn't support transactions at all as per MySQL docs.
The problem is many 3.3 boards use InnoDB, but what is even worse is that sometimes DB is messed up by using MyISAM and InnoDB tables at the same time.
So the transaction result is pretty unpredictable.
What if we add checking tables status in transaction then, so that if at least 1 of them is MyISAM transaction won't start at all? Like (roughly)
```php
foreach ($tables as $table)
{
	$table_status = $this->get_table_status($table_name);
	if $table_status['Engine'] === 'MyISAM'
	{
		return;
	}
}
```

marc1706 (approved):

PR #6651, opened 2024-06-20
[ticket/17347] Add console command to delete users by user IDs

marc1706:
Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-17347


rxu:
There can be a potential issue if someone will attempt to delete bots via this command. They'll be deleted in the users table but not in the bots table which will cause lots of PHP errors/warnings later on like
```
<b>[phpBB Debug] PHP Warning</b>: in file <b>[ROOT]/phpbb/session.php</b> on line <b>640</b>: <b>Trying to access array offset on value of type bool</b><br />
<b>[phpBB Debug] PHP Warning</b>: in file <b>[ROOT]/phpbb/session.php</b> on line <b>650</b>: <b>Undefined array key "user_id"</b><br />
<b>[phpBB Debug] PHP Warning</b>: in file <b>[ROOT]/phpbb/session.php</b> on line <b>1334</b>: <b>Undefined array key "user_type"</b><br />
```
and so on.

phpBB/phpbb/console/command/user/delete_id.php
```diff
@@ -170,4 +197,25 @@ protected function interact(InputInterface $input, OutputInterface $output): voi
...
+	 * Deletes a bot user
+	 *
+	 * @param array $user_row
+	 * @return void
+	 */
+	protected function delete_bot_user(array $user_row): void
+	{
+		$sql = 'DELETE FROM ' . $this->bots_table . '
+					WHERE user_id = ' . (int) $user_row['user_id'];
+		$this->db->sql_query($sql);
```
rxu, line 221: This looks identical to the query below so probably could be added to the loop.

marc1706: Good point, was based on acp_bots but since we're doing something slightly different here we don't need to separate this.

Derky (approved):

PR #6657, opened 2024-06-25
[ticket/17326] Add font icon field for custom BBCodes

danieltj27:
Adds a new field to the custom BBCode add/edit form that allows admins to define a Font Awesome icon to be displayed instead of the BBCodes name on the posting.php editor page.

PHPBB-17326

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-17326

### DIFF
--- a/phpBB/adm/style/admin.js
+++ b/phpBB/adm/style/admin.js
@@ -279,11 +279,24 @@ function parse_document(container)
 		});
 
 		// Live update BBCode font icon preview
-		$('#bbcode_font_icon').on('keyup', function(e) {
-			const iconName = $(this).val();
-			if (iconName.match(/^[\w-]+$/)) {
-				$('#bbcode_icon_preview').attr('class', "o-icon o-icon-font fa-fw fas icon fa-" + $(this).val());
-			}
+		const updateIconClass = (element, newClass) => {
+			element.classList.forEach(className => {
+				if (className.startsWith('fa-') && className !== 'fa-fw') {
+					element.classList.remove(className);
+				}
+			});
+
+			element.classList.add(`fa-${newClass}`);
+		};
+
+		const pageIconFont = document.getElementById('bbcode_font_icon');
+
+		pageIconFont.addEventListener('keyup', function() {
+			updateIconClass(this.nextElementSibling, this.value);
+		});
+
+		pageIconFont.addEventListener('blur', function() {
+			updateIconClass(this.nextElementSibling, this.value);
 		});
 	});
 })(jQuery);
--- a/phpBB/adm/style/admin.js
+++ b/phpBB/adm/style/admin.js
@@ -277,5 +277,10 @@ function parse_document(container)
 		$('.actions a:has(i.acp-icon)').mouseover(function () {
 			$(this).css("text-decoration", "none");
 		});
+
+		// Live update BBCode font icon preview
+		$('#bbcode_font_icon').on('keyup', function(e) {
+			$('#bbcode_icon_preview').attr('class', "o-icon o-icon-font fa-fw fas icon fa-" + $(this).val());
+		});
 	});
 })(jQuery);
--- a/phpBB/adm/style/admin.js
+++ b/phpBB/adm/style/admin.js
@@ -291,12 +291,14 @@ function parse_document(container)
 
 		const pageIconFont = document.getElementById('bbcode_font_icon');
 
-		pageIconFont.addEventListener('keyup', function() {
-			updateIconClass(this.nextElementSibling, this.value);
-		});
+		if (pageIconFont) {
+			pageIconFont.addEventListener('keyup', function () {
+				updateIconClass(this.nextElementSibling, this.value);
+			});
 
-		pageIconFont.addEventListener('blur', function() {
-			updateIconClass(this.nextElementSibling, this.value);
-		});
+			pageIconFont.addEventListener('blur', function () {
+				updateIconClass(this.nextElementSibling, this.value);
+			});
+		}
 	});
 })(jQuery);
--- a/phpBB/adm/style/admin.js
+++ b/phpBB/adm/style/admin.js
@@ -280,7 +280,10 @@ function parse_document(container)
 
 		// Live update BBCode font icon preview
 		$('#bbcode_font_icon').on('keyup', function(e) {
-			$('#bbcode_icon_preview').attr('class', "o-icon o-icon-font fa-fw fas icon fa-" + $(this).val());
+			const iconName = $(this).val();
+			if (iconName.match(/^[\w-]+$/)) {
+				$('#bbcode_icon_preview').attr('class', "o-icon o-icon-font fa-fw fas icon fa-" + $(this).val());
+			}
 		});
 	});
 })(jQuery);
--- a/phpBB/adm/style/admin.js
+++ b/phpBB/adm/style/admin.js
@@ -280,6 +280,12 @@ function parse_document(container)
 
 		// Live update BBCode font icon preview
 		const updateIconClass = (element, newClass) => {
+			// Ignore invalid class names
+			const faIconRegex = /^(?!-)(?!.*--)[a-z0-9-]+(?<!-)$/;
+			if (!faIconRegex.test(newClass)) {
+				return;
+			}
+
 			element.classList.forEach(className => {
 				if (className.startsWith('fa-') && className !== 'fa-fw') {
 					element.classList.remove(className);
--- a/phpBB/includes/acp/acp_bbcodes.php
+++ b/phpBB/includes/acp/acp_bbcodes.php
@@ -44,12 +44,12 @@ class acp_bbcodes
 		switch ($action)
 		{
 			case 'add':
-				$bbcode_match = $bbcode_tpl = $bbcode_helpline = '';
+				$bbcode_match = $bbcode_tpl = $bbcode_helpline = $bbcode_font_icon = '';
 				$display_on_posting = 0;
 			break;
 
 			case 'edit':
-				$sql = 'SELECT bbcode_match, bbcode_tpl, display_on_posting, bbcode_helpline
+				$sql = 'SELECT bbcode_match, bbcode_tpl, display_on_posting, bbcode_helpline, bbcode_font_icon
 					FROM ' . BBCODES_TABLE . '
 					WHERE bbcode_id = ' . $bbcode_id;
 				$result = $db->sql_query($sql);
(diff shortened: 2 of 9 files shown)


danieltj27:
Resubmission after #6648 was closed because I had to delete my old repository.

danieltj27:
Does anyone know what happened to the tests? The exact same code changes were made from the previous PR and everything passed without a hitch. I've checked the test logs and I'm super confused as to what the fails mean.

phpBB/styles/prosilver/template/posting_buttons.html
```diff
@@ -97,7 +97,11 @@
 
 	<!-- BEGIN custom_tags -->
 	<button type="button" class="button button-secondary bbcode-{custom_tags.BBCODE_TAG_CLEAN}" name="addbbcode{custom_tags.BBCODE_ID}" value="{custom_tags.BBCODE_TAG}" onclick="bbstyle({custom_tags.BBCODE_ID})" title="{{ custom_tags.BBCODE_HELPLINE }}">
-		{custom_tags.BBCODE_TAG}
+		{% if custom_tags.BBCODE_FONT_ICON != '' %}
+			{{ Icon('font', custom_tags.BBCODE_FONT_ICON, '', false, custom_tags.BBCODE_TAG_CLEAN) }}
+		{% else %}
+			{{ custom_tags.BBCODE_TAG }}
+		{%
```
rubencm, line 104:
```suggestion
		{% 
```
```suggestion
		{% endif %}
```

rubencm, line 104:
```suggestion
		{% endif %}
```

danieltj27: Thanks!

phpBB/includes/acp/acp_bbcodes.php
```diff
@@ -231,6 +236,11 @@ function main($id, $mode)
 						trigger_error($user->lang['BBCODE_HELPLINE_TOO_LONG'] . adm_back_link($this->u_action), E_USER_WARNING);
 					}
 
+					if (strlen($bbcode_font_icon) > 50)
```
marc1706, line 239: I'd also add a simple regex check

phpBB/language/en/acp/posting.php
```diff
@@ -53,6 +53,9 @@
 	'BBCODE_HELPLINE_EXPLAIN'	=> 'This field contains the mouse over text of the BBCode.',
 	'BBCODE_HELPLINE_TEXT'		=> 'Help line text',
 	'BBCODE_HELPLINE_TOO_LONG'	=> 'The help line you entered is too long.',
+	'BBCODE_FONT_ICON'			=> 'BBCode icon',
+	'BBCODE_FONT_ICON_EXPLAIN'	=> 'Enter the name of a Font Awesome icon (without the fa prefix) to display instead of the BBCode name appearing on the button. <a href="https://fontawesome.com/v6/icons/" target="_blank">Click here</a> to view the list of available icons.',
```
marc1706, line 57: HTML should be defined in the template file and ideally you should only link to the free icons. Also, maybe mention that only solid icons are supported?

phpBB/phpbb/db/migration/data/v400/add_bbcode_font_icon.php
```diff
@@ -0,0 +1,51 @@
...
+			'\phpbb\db\migration\data\v400\dev'
+		];
+	}
+
+	public function update_schema()
+	{
+		return [
+			'add_columns'	=> [
+				$this->table_prefix . 'bbcodes'	=> [
+					'bbcode_font_icon'	=> ['VCHAR:50', ''],
```
marc1706, line 35: I'd opt for something maybe slightly longer? How about 64?

marc1706 (changes requested):

phpBB/adm/style/admin.js
```diff
@@ -261,5 +261,10 @@ function parse_document(container)
 		$('.actions a:has(i.acp-icon)').mouseover(function () {
 			$(this).css("text-decoration", "none");
 		});
+
+		// Live update BBCode font icon preview
+		$('#bbcode_font_icon').on('keyup', function(e) {
+			$('#bbcode_icon_preview').attr('class', "o-icon o-icon-font fa-fw fas icon fa-" + $(this).val());
```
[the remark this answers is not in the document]
danieltj27: Sure thing, I'll add this!

phpBB/language/en/acp/posting.php
danieltj27: Sure thing, I'll tweak this.

phpBB/phpbb/db/migration/data/v400/add_bbcode_font_icon.php
danieltj27: Do you know what the longest icon name is? I managed to find `american-sign-language-interpreting` in the v4 set but it seems as though that name was shortened in v6. Happy to change to 64 though although I think 50 is probably unnecessarily big in hindsight... 😅

phpBB/includes/acp/acp_bbcodes.php
danieltj27: Updated string length check and added `preg_match`.

danieltj27:
Addressed all of Marc's comments in the above commits. Let me know if there's anything else.

phpBB/includes/acp/acp_bbcodes.php
```diff
@@ -232,6 +238,11 @@ function main($id, $mode)
 						trigger_error($user->lang['BBCODE_HELPLINE_TOO_LONG'] . adm_back_link($this->u_action), E_USER_WARNING);
 					}
 
+					if (strlen($bbcode_font_icon) > 64 && preg_match('/^[\w-]+$/', $bbcode_font_icon))
+					{
+						trigger_error($user->lang['BBCODE_FONT_ICON_TOO_LONG'] . adm_back_link($this->u_action), E_USER_WARNING);
```
marc1706, line 241: I guess something like BBCODE_FONT_ICON_INVALID would make more sense?

phpBB/includes/acp/acp_bbcodes.php
```diff
@@ -232,6 +238,11 @@ function main($id, $mode)
 						trigger_error($user->lang['BBCODE_HELPLINE_TOO_LONG'] . adm_back_link($this->u_action), E_USER_WARNING);
 					}
 
+					if (strlen($bbcode_font_icon) > 64 && preg_match('/^[\w-]+$/', $bbcode_font_icon))
```
marc1706, line 239:
```suggestion
					if (strlen($bbcode_font_icon) > 64 || preg_match('/^[\w-]+$/', $bbcode_font_icon))
```
Would make more sense with the suggested change.

phpBB/language/en/acp/posting.php
```diff
@@ -53,6 +53,9 @@
 	'BBCODE_HELPLINE_EXPLAIN'	=> 'This field contains the mouse over text of the BBCode.',
 	'BBCODE_HELPLINE_TEXT'		=> 'Help line text',
 	'BBCODE_HELPLINE_TOO_LONG'	=> 'The help line you entered is too long.',
+	'BBCODE_FONT_ICON'			=> 'BBCode icon',
+	'BBCODE_FONT_ICON_EXPLAIN'	=> 'Enter the name of a Font Awesome icon (without the fa prefix) to display instead of the BBCode name appearing on the button. %1$sClick here%2$s to view the list of available icons. Only solid style icons are supported.',
+	'BBCODE_FONT_ICON_TOO_LONG'	=> 'The icon name you have entered is too long.',
```
marc1706, line 58: maybe invalid instead?

phpBB/adm/style/admin.js
```diff
@@ -261,5 +261,13 @@ function parse_document(container)
 		$('.actions a:has(i.acp-icon)').mouseover(function () {
 			$(this).css("text-decoration", "none");
 		});
+
+		// Live update BBCode font icon preview
+		$('#bbcode_font_icon').on('keyup', function(e) {
+			const iconName = $(this).val();
+			if (iconName.match(/^[\w-]+$/)) {
+				$('#bbcode_icon_preview').attr('class', "o-icon o-icon-font fa-fw fas icon fa-" + $(this).val());
```
iMattPro, line 269: I wish that all those other classes don't have to be used here, because they could change at any time, and might get missed in an update. Would be better if it could just work on the one font class. 

Like...I don't know...maybe get the class list and look for all that include `fa-` and toggle that with the new font class name. Or loop through the class list, remove any that `startsWith()` `fa-`, then add back to the element the `fa-fw` and whatever the new `fa-icon` name is.

danieltj27: You love creating work for me don't you 😂

iMattPro: Yep! You can look at what I did in Pages for 4.0 for inspiration. It still relies on naming a bunch of those classes, so I'm not totally happy with it, but I do like it better than what Pages was doing before.
https://github.com/phpbb-extensions/pages/blob/3a44f8f398fd272fe87c6be260fcabc535f15639/adm/style/scripts.js#L45

danieltj27: I might have an idea for altering the class name without having to mess with the other class names. The issue with the linked solution is you're still defining the class names in a file so as you say, if they change (like in an FA update), you'll have to remember to update them there.

Perhaps you could get the last class in the list, check it starts with `fa-` but is not `fa-fw` and then remove it and then add the new class name. That should work.

iMattPro: Yeah, I'm working on this code right now as an update for pages:
```js
// Remove all classes starting with 'fa-'
Array.from(element.classList).forEach(className => {
	if (className.startsWith('fa-')) {
		element.classList.remove(className);
	}
});

// Add the new class with 'fa-' prefix and 'fa-fw'
element.classList.add('fa-fw', `fa-${newClass}`);
```

iMattPro: Just needs to add `and not fa-fw` logic to the conditional

iMattPro: Code updated https://github.com/phpbb-extensions/pages/blob/1deee23ef4465d70e6c2a1b9392ef794db6ee20c/adm/style/scripts.js#L45

marc1706:
@danieltj27 I've replaced your merges with a rebase. Try to use rebases instead of merging master as that usually prevents unwanted clutter and complicated version history.

danieltj27:
I've copied the js from the pages extension and fixed the issue with the icon no longer being displayed. This should be ready to merge now.

phpBB/adm/style/admin.js
```diff
@@ -261,5 +261,26 @@ function parse_document(container)
...
+					element.classList.remove(className);
+				}
+			});
+
+			element.classList.add(`fa-${newClass}`);
+		};
+
+		const pageIconFont = document.getElementById('bbcode_font_icon');
+
+		pageIconFont.addEventListener('keyup', function() {
```
marc1706, line 278: Should check if pageIconFont was found before accessing it, otherwise you produce JS errors on every page

PR #6669, opened 2024-07-03
[ticket/17363] Add web app manifest

iMattPro:
PHPBB-17363

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-17363


iMattPro:
The output of this manifest json contains only the required members of the web app manifest specification:
```
'name' : The name of the site
'short_name': Short name, 12 chars or less, used if the name above is too long
'display': Set to standalone to site looks like a fullscreen app
'orientation': Set to portrait as the default orientation, you can still flip device to landscape though
'dir': language direction, based on user's CONTENT DIR setting
'start_url': The path, relative to the manifest, of the home/landing page the app launches you in to.
'scope' The path, relative to the manifest, of where you can browse within, if you navigate outside the scope, it reverts to a normal web page inside a browser tab or window.
```

I've added a core event to this new code as well so the manifest could be extended. One optional attribute someone might want to add are device-specific resolution icons. These would be icons that would be used as your site's "App" icon on device. I obviously did not add the code for that here because we don't provide icons, just as we don't provide favicons. 

Through an extension, one could provide the icons, a place to store the icons, and add them to the manifest JSON. The minimum requirement when using icons is a 192x192 and a 512x512. But you could keep going, adding specific sized icons for every device's native resolution. 

That's why I think the optional icon support should be handled by extension rather than the in the core. Without icons specified, the default behavior is the site's App icon on device will use a site's favicon if there is one, otherwise just a black background with the first letter of the site name in white.

marc1706 (approved):

PR #6684, opened 2024-07-07
[ticket/17365] Prevent search limit being bypassed with operators

crowjake:
Provided `$countable_keywords` wherein the existing `$keywords` value is modified so that any `-`, `+` and `|` characters without preceding spaces is replaced with the same but with a space in front of each. 

These spaces allow the string to be more accurately split when used instead of $keywords inside the $num_keywords calculation. 

This prevents the word limit being bypassed in search by the use of operators without whitespace.

PHPBB-17365

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-17365

### DIFF
--- a/phpBB/phpbb/search/fulltext_native.php
+++ b/phpBB/phpbb/search/fulltext_native.php
@@ -299,7 +299,11 @@ class fulltext_native extends \phpbb\search\base
 		);
 
 		$keywords = preg_replace($match, $replace, $keywords);
-		$num_keywords = count(explode(' ', $keywords));
+
+		// Ensure a space exists before +, - and | to make the split and count work correctly
+		$countable_keywords = preg_replace('/(?<!\s)(\+|\-|\|)/', ' $1', $keywords);
+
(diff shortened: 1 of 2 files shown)


marc1706:
There is no need to close and open new pull requests every time you want to change something. You can either close and reopen the same PR or request a rerun of tests if something else failed.

crowjake:
> There is no need to close and open new pull requests every time you want to change something. You can either close and reopen the same PR or request a rerun of tests if something else failed.

Noted, and thanks for clarifying. I had a few false starts before I worked out how to do it correctly, I'll make sure I'm doing it the right way from here on out.

crowjake:
I had to force-push so that the previous commit message could be amended to have lines shorter than 80 characters - this was necessary because "Tests / PHP 7.2 - none (pull_request)" was failing due to this standard being unmet.

marc1706:
Thanks for the PR. I've added a test that should cover the different cases based on the ticket description.

PR #6685, opened 2024-07-08
[ticket/17369] Add forum id to approve action url for redirection

danieltj27:
Added the forum_id to the approve action link which ensures that when a topic is permanently deleted, the forum_id is present so that it can be used in the redirection link to go to the previous forum. This addresses a bug in which the forum_id is not provided and after deleting a topic, the user will be taken to a 404 page as the forum id fallbacks to 0.

PHPBB-17369

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-17369

### DIFF
--- a/phpBB/viewtopic.php
+++ b/phpBB/viewtopic.php
@@ -2071,7 +2071,7 @@ for ($i = 0, $end = count($post_list); $i < $end; ++$i)
 		'U_EMAIL'		=> $user_cache[$poster_id]['email'],
 		'U_JABBER'		=> $user_cache[$poster_id]['jabber'],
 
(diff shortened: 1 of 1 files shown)


danieltj27:
It's a super sneaky bug.

To reproduce it, soft-delete a topic and then go back to the `viewtopic.php` page. Click the **_Delete_** button and then on the next page confirm the deletion (click **_Submit_**).

You'll then get taken to a confirmation page stating that the topic was deleted successfully and there's a link you can click to **_Return to the forum last visited_**.

Clicking that link will take you to `viewforum.php?f=0` because the forum ID hasn't been passed through to the MCP during the topic deletion process so the link, or 'connection' to the forum is lost.

Adding the forum ID in the action link means the forum ID is preserved through the process and that the user can click the return link and properly go back to the forum the topic belonged to.

danieltj27:
What happened with the checks? 🤔

marc1706 (approved):

PR #6693, opened 2024-07-17
[ticket/17343] Display push notifications in user language

marc1706:
PHPBB-17343

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-17343

### DIFF
--- a/phpBB/phpbb/notification/method/webpush.php
+++ b/phpBB/phpbb/notification/method/webpush.php
@@ -140,13 +140,7 @@ class webpush extends messenger_base implements extended_method_interface
 		{
 			$data = $notification->get_insert_array();
 			$data += [
-				'push_data'		=> json_encode([
-					'heading'	=> $this->config['sitename'],
-					'title'		=> strip_tags($notification->get_title()),
-					'text'		=> strip_tags($notification->get_reference()),
-					'url'		=> htmlspecialchars_decode($notification->get_url()),
-					'avatar'	=> $notification->get_avatar(),
-				]),
+				'push_data'				=> json_encode(array_merge($notification->get_insert_array(), ['notification_type_name' => $notification->get_type()])),
 				'notification_time'		=> time(),
 				'push_token'			=> hash('sha256', random_bytes(32))
 			];
--- a/phpBB/phpbb/notification/method/webpush.php
+++ b/phpBB/phpbb/notification/method/webpush.php
@@ -140,7 +140,10 @@ class webpush extends messenger_base implements extended_method_interface
 		{
 			$data = $notification->get_insert_array();
 			$data += [
-				'push_data'				=> json_encode(array_merge($notification->get_insert_array(), ['notification_type_name' => $notification->get_type()])),
+				'push_data'				=> json_encode(array_merge(
+					$data,
+					['notification_type_name' => $notification->get_type()],
+				)),
 				'notification_time'		=> time(),
 				'push_token'			=> hash('sha256', random_bytes(32))
 			];
@@ -345,6 +348,14 @@ class webpush extends messenger_base implements extended_method_interface
 		return array_intersect_key($data, $row);
 	}
 
+	/**
+	 * Get template data for the UCP
+	 *
+	 * @param helper $controller_helper
+	 * @param form_helper $form_helper
+	 *
+	 * @return array
+	 */
 	public function get_ucp_template_data(helper $controller_helper, form_helper $form_helper): array
 	{
 		$subscription_map = $this->get_user_subscription_map([$this->user->id()]);
--- a/phpBB/phpbb/ucp/controller/webpush.php
+++ b/phpBB/phpbb/ucp/controller/webpush.php
@@ -220,6 +220,13 @@ class webpush
 		throw new http_exception(Response::HTTP_FORBIDDEN, 'NO_AUTH_OPERATION');
 	}
 
+	/**
+	 * Get notification data for output from json encoded data stored in database
+	 *
+	 * @param string $notification_data Encoded data stored in database
+	 *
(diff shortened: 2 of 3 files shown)


rxu:
For the extension we did it this way https://github.com/phpbb-extensions/webpushnotifications/pull/65 if it makes any sense.

phpBB/phpbb/ucp/controller/webpush.php
```diff
@@ -178,23 +190,54 @@ private function get_anonymous_notifications(): string
...
+		// Old notification data is pre-parsed and just needs to be returned
+		if (isset($row_data['heading']))
+		{
+			return $notification_data;
+		}
+
+		// Get notification from row_data
+		$notification = $this->notification_manager->get_item_type_class($row_data['notification_type_name'], $row_data);
+
+		$notification_data = [
```
iMattPro, line 230: Maybe just 

```php
return json_encode([
	'heading'	=> $this->config['sitename'],
	'title'		=> strip_tags($notification->get_title()),
	'text'		=> strip_tags($notification->get_reference()),
	'url'		=> htmlspecialchars_decode($notification->get_url()),
	'avatar'	=> $notification->get_avatar(),
]);
```

phpBB/phpbb/notification/method/webpush.php
```diff
@@ -140,13 +140,7 @@ public function notify()
...
 			$data = $notification->get_insert_array();
 			$data += [
-				'push_data'		=> json_encode([
-					'heading'	=> $this->config['sitename'],
-					'title'		=> strip_tags($notification->get_title()),
-					'text'		=> strip_tags($notification->get_reference()),
-					'url'		=> htmlspecialchars_decode($notification->get_url()),
-					'avatar'	=> $notification->get_avatar(),
-				]),
+				'push_data'				=> json_encode(array_merge($notification->get_insert_array(), ['notification_type_name' => $notification->get_type()])),
```
iMattPro, line 143: I'm confused by this. Where is all this push data going now? Looks like its gone? Is that supposed to be `$notification->get_insert_array()` again? This looks like it's an Ouroboros, lol. It looks sort of like it's:
```php
$data = $notification->get_insert_array();
$data += ['push_data' => $data]
```
like it's putting itself in iitself.

iMattPro: Ok maybe I'm starting to get this, lol. But if using using `$notification->get_insert_array()` is intended twice here, maybe just use `$data` the second time?

```php
'push_data' => json_encode(array_merge($data, ['notification_type_name' => $notification->get_type()])),
```

marc1706:
> For the extension we did it this way [phpbb-extensions/webpushnotifications#65](https://github.com/phpbb-extensions/webpushnotifications/pull/65) if it makes any sense.

One of the reasons want to do this in the controller is to prevent having to reload language multiple times while sending out notifications.

phpBB/phpbb/notification/method/webpush.php
marc1706: It might actually make sense to add a migration and adjust the table instead.

marc1706:
> One thing to maybe add to this, although not totally related, is Emoji support. Apparently if Emoji's are in the notification's body/message (such as from a PM or Topic title), they appear as something like `&#128518;` instead of 😆

Thanks for the info, that should also be fixed :)

iMattPro:
We did this: https://github.com/phpbb-extensions/webpushnotifications/pull/82

marc1706:
Also ended up with that:
image

marc1706:
Usernames are also correctly displayed now.

iMattPro:
Are there any pros/cons to storing the rendered emoji in the DB like this is doing now vs. storing it as hex code like before and decoding it when sending the push notification out?

phpBB/phpbb/ucp/controller/webpush.php
```diff
@@ -178,23 +196,55 @@ private function get_anonymous_notifications(): string
...
+		if (isset($row_data['heading']))
+		{
+			return $notification_data;
+		}
+
+		// Get notification from row_data
+		$notification = $this->notification_manager->get_item_type_class($row_data['notification_type_name'], $row_data);
+
+		// Load users for notification
+		$this->user_loader->load_users($notification->users_to_query());
```
iMattPro, line 237: Should you add `[USER_IGNORE]` as 2nd parameter here?

marc1706: I don't think that's needed and also not what is done in the board notifications.

phpBB/phpbb/notification/method/webpush.php
marc1706: I've cleaned this up a bit and it should be less confusing now

iMattPro:
> > Are there any pros/cons to storing the rendered emoji in the DB like this is doing now vs. storing it as hex code like before and decoding it when sending the push notification out?
> > Also it working for me too Screenshot 2024-10-02 at 7 15 09 AM
> 
> We're storing the notification data the same way we're storing it for board notifications so I think it's fine to leave it like this.

Are we? I remember looking in a DB entry for the _notifications table and seeing the emoji as hex codes in there.

phpBB/phpbb/ucp/controller/webpush.php
iMattPro: I was just asking because we use this same method in the other webpush class (the method class) and there we are using the 2nd parameter with USER_IGNORE

marc1706: Though that is before creating the notification. Here it's about displaying it correctly so there shouldn't be any of these left.

iMattPro (approved):

PR #6698, opened 2024-07-31
[ticket/17375] Ensure user last visit does not get updated for active sessions in session gc

marc1706:
Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-17375

### DIFF
--- a/phpBB/phpbb/session.php
+++ b/phpBB/phpbb/session.php
@@ -819,7 +819,7 @@ class session
 				// Update the form key
 				$sql = 'UPDATE ' . USERS_TABLE . '
 					SET user_form_salt = \'' . $db->sql_escape($this->data['user_form_salt']) . '\',
-						user_last_active = ' . (int) $this->data['session_time'] . '
+						user_last_active = ' . (int) $this->time_now . '
 					WHERE user_id = ' . (int) $this->data['user_id'];
 				$db->sql_query($sql);
 			}
--- a/phpBB/phpbb/session.php
+++ b/phpBB/phpbb/session.php
@@ -964,8 +964,8 @@ class session
 		}
 
 		/**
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/session.php
```diff
@@ -964,16 +964,17 @@ function session_gc()
...
+		 * Inner SELECT gets most recent expired sessions for unique session_user_id
 		 * Outer SELECT gets data for them
 		 */
 		$sql_select = 'SELECT s1.session_page, s1.session_user_id, s1.session_time AS recent_time
 			FROM ' . SESSIONS_TABLE . ' AS s1
 			INNER JOIN (
 				SELECT session_user_id, MAX(session_time) AS recent_time
 				FROM ' . SESSIONS_TABLE . '
-				WHERE session_user_id <> ' . ANONYMOUS . '
+				WHERE session_time < ' . ($this->time_now - (int) $config['session_length']) . '
```
rxu, line 976: This will select only expired sessions which is not what we need I think.

rxu: And possible proposal. For the index page, `session_last_visit` is used to display last visit time. It could be used for profiles possibly then.
https://github.com/phpbb/phpbb/blob/4038bb9dc3a49197046fe3ef0f1b807866250d3f/phpBB/includes/functions.php#L3856
image

Also, in this case we would get rid of all those huge SQL queries at all as they're only used to update user last visit.

marc1706: This change essentially reverts the last visit behavior back to what it was before 3.3.12. It has worked fine before and used the expired sessions as a way of determining the last visit time.

Derky (approved):

PR #6699, opened 2024-08-04
[ticket/17376] Add aria label to quote reference link

danieltj27:
Adds an aria label attribute to the quoted post reference link so that screen readers can inform the user of what the link is used for. Before this change the link was too vague and would not have been well described to users of its purpose if they couldn't see it.

PHPBB-17376

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-17376

### DIFF
--- a/phpBB/language/en/viewtopic.php
+++ b/phpBB/language/en/viewtopic.php
@@ -112,6 +112,7 @@ $lang = array_merge($lang, array(
 	'VIEW_INFO'				=> 'Post details',
 	'VIEW_NEXT_TOPIC'		=> 'Next topic',
 	'VIEW_PREVIOUS_TOPIC'	=> 'Previous topic',
+	'VIEW_QUOTED_POST'		=> 'View quoted post',
 	'VIEW_RESULTS'			=> 'View results',
 	'VIEW_TOPIC_POSTS'		=> array(
 		1	=> '%d post',
--- a/phpBB/styles/prosilver/template/bbcode.html
+++ b/phpBB/styles/prosilver/template/bbcode.html
@@ -38,11 +38,11 @@
 				<xsl:value-of select="$L_COLON"/>
 				<xsl:if test="@post_url">
 					<xsl:text> </xsl:text>
-					<a href="{@post_url}" aria-label="{L_VIEW_QUOTED_POST}" data-post-id="{@post_id}" onclick="if(document.getElementById(hash.substr(1)))href=hash">&#8593;</a>
+					<a href="{@post_url}" aria-label="{L_VIEW_QUOTED_POST}" data-post-id="{@post_id}" onclick="if(document.getElementById(hash.substr(1)))href=hash"><i class="icon fa-arrow-circle-up fa-fw" aria-hidden="true"></i></a>
 				</xsl:if>
 				<xsl:if test="@msg_url">
 					<xsl:text> </xsl:text>
-					<a href="{@msg_url}" data-msg-id="{@msg_id}">&#8593;</a>
+					<a href="{@msg_url}" aria-label="{L_VIEW_QUOTED_POST}" data-msg-id="{@msg_id}"><i class="icon fa-arrow-circle-up fa-fw" aria-hidden="true"></i></a>
 				</xsl:if>
 				<xsl:if test="@date">
 					<span class="responsive-hide"><xsl:value-of select="@date"/></span>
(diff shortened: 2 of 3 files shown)


danieltj27:
I used `{L_VIEW_QUOTED_POST}` instead of `{{ lang('VIEW_QUOTED_POST') }}` to match the rest of the file for conformity but I'll happily switch the the newer syntax if needed. It's a small change, and might not be 'prefect' as such but it is definitely a step in the right direction.

Perhaps this link could be changed from an arrow to just text in a future version. However that would be a matter of preference and not necessarily on-topic with the issue here.

`aria-label`: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label

GanstaZ:
I think it's best to use `{{ lang('VIEW_QUOTED_POST') }}`.

danieltj27:
> I think it's best to use `{{ lang('VIEW_QUOTED_POST') }}`.

I did try that but Twig wouldn't parse it so I had to match what was already there with the single brackets.

I've updated the template to fix both instances of the anchor that's added and also replace the unicode icon with a Font Awesome icon so it stands out a bit more. Not saying this is a perfect solution and will make everyone happy, but it's certainly more obvious and much better for accessibility.

What it looks like 👇

phpbb-quote-update-fa-icon-anchor

danieltj27:
Had to force push again because my commit message went just over the limit 😅

battye:
> Does anyone know why the php tests are failing? I can't understand the error messages.

Hi @danieltj27, it looks to be because of the test data in https://github.com/phpbb/phpbb/blob/c64055509ffd0916d4df9a6acfc0f9a4a4785ce7/tests/text_formatter/s9e/default_formatting_test.php#L286

The failure is: 

```
1) phpbb_textformatter_s9e_default_formatting_test::test_default_formatting with data set #55 ('[quote=Username post_id=123]....quote]', '<blockquote cite="phpBB/viewt...quote>')
Failed asserting that two strings are identical.

--- Expected
+++ Actual
@@ @@

-'<blockquote cite="phpBB/viewtopic.php?p=123#p123"><div><cite>Username wrote: <a href="phpBB/viewtopic.php?p=123#p123" data-post-id="123" onclick="if(document.getElementById(hash.substr(1)))href=hash">↑</a></cite>...</div></blockquote>'

+'<blockquote cite="phpBB/viewtopic.php?p=123#p123"><div><cite>Username wrote: <a href="phpBB/viewtopic.php?p=123#p123" aria-label="VIEW_QUOTED_POST" data-post-id="123" onclick="if(document.getElementById(hash.substr(1)))href=hash"><i class="icon fa-arrow-circle-up fa-fw" aria-hidden="true"></i></a></cite>...</div></blockquote>'
```

Interestingly, it looks like the language text isn't parsing either because it's printing VIEW_QUOTED_POST verbatim in the actual result, but the reason it's failing is because the test data hasn't been updated to include your change with the aria-label.

rxu:
`tests/text_formatter/s9e/default_formatting_test.php` quote tests should be adjusted to reflect new expected results in affected datasets.
As for the language entry, `viewtopic.php` language file has to be loaded to the test file to be parsed correctly.

danieltj27:
This looks like it's going to be a head scratcher for me then.

Do you know where I would include the `viewtopic` file in the test to ensure the language is brought across properly or do I even need to do that, should I insert the text as is without the language key?

For example on line 81, the word `wrote` doesn't rely on languages.
```
			array(
				'[quote="username"]quoted[/quote]',
				'<blockquote><div><cite>username wrote:</cite>quoted</div></blockquote>'
			),
```

rxu:
As for the test, I guess you just need to change erroring extected data from 
```html
<blockquote cite="phpBB/viewtopic.php?p=123#p123"><div><cite>Username wrote: <a href="phpBB/viewtopic.php?p=123#p123" data-post-id="123" onclick="if(document.getElementById(hash.substr(1)))href=hash">↑</a></cite>...</div></blockquote>
```
to
```html
<blockquote cite="phpBB/viewtopic.php?p=123#p123"><div><cite>Username wrote: <a href="phpBB/viewtopic.php?p=123#p123" aria-label="VIEW_QUOTED_POST" data-post-id="123" onclick="if(document.getElementById(hash.substr(1)))href=hash"><i class="icon fa-arrow-circle-up fa-fw" aria-hidden="true"></i></a></cite>...</div></blockquote>
```

as that's what you do expect to output actually.

As for the language file loading, looking into it deeper I think that won't work directly as far as unit tests don't use language class. You can mock loading language entries though in a similar way to how it's done in https://github.com/phpbb/phpbb/blob/3.3.x/tests/language/language_test.php but I'd leave it as is now (just a language key).

So as a bottom line, you could just try changing that line in the corresponding unit test file and see if it helps.

PR #6708, opened 2024-08-25
[ticket/17385] Fix for CDB version checks with missing SSL flag

LukeWCS:
Checklist:

- [ ] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-17385

### DIFF
--- a/phpBB/phpbb/extension/manager.php
+++ b/phpBB/phpbb/extension/manager.php
@@ -548,10 +548,11 @@ class manager
 		}
 
 		$version_check = $meta['extra']['version-check'];
+		$ssl = ($version_check['host'] == 'www.phpbb.com') ?: ($version_check['ssl'] ?? false);
 
 		$version_helper = new \phpbb\version_helper($this->cache, $this->config, new file_downloader());
 		$version_helper->set_current_version($meta['version']);
-		$version_helper->set_file_location($version_check['host'], $version_check['directory'], $version_check['filename'], isset($version_check['ssl']) ? $version_check['ssl'] : false);
+		$version_helper->set_file_location($version_check['host'], $version_check['directory'], $version_check['filename'], $ssl);
 		$version_helper->force_stability($stability);
 
 		return $version_helper->get_ext_update_on_branch($force_update, $force_cache);


rxu:
Personally, I'd make the fix more generic and unlink it from phpbb.com (that looks hackish a bit tbh). F.e. if `ssl` is not set, then try with use_ssl = true, if it returns error then try with use_ssl = false, if it returns error too then report it to admin (regardless of the host).

LukeWCS:
@rxu 
> I'd make the fix more generic and unlink it from phpbb.com (that looks hackish a bit tbh)
In general, I agree with you, it is not perfect. ;-) But since a version check of a CDB Ext only works if SSL is used anyway, this is less tragic. Especially because the fix only targets phpBB.com and does not affect other version servers at all.

Your proposed method has one disadvantage: we already have phpBB installations with an extremely large number of extensions and timeout problems. See this topic in the EMP support area:

[MySQL server has gone away [2006]](https://www.phpbb.com/customise/db/extension/extension_manager_plus/support/topic/247519)

If we now work with HTTPS first and then with HTTP again if SSL fails, the timeout problem gets even worse because phpBB unfortunately does not implement the versions in chunks, but all at once.

gn36:
In my eyes, the whole problem arises because the version checker doesn't follow 301 Redirects. phpBB.com redirects correctly from http:// addresses to https. But instead of following this like every web browser, phpBB installations simply give up like on a 404 error.
In my eyes the best solution would be to follow that redirect. This would also allow extension writers to restructure their servers in the future without loosing connection to all older extension installations, as long as they setup correct redirects.
If you want to ensure that we don't get into endless redirect loops, you could limit the number of redirects followed and check if you check the same URL multiple times. For sure, this is more complicated than the suggested solution, but in my eyes, it would be more correct than what we have right now.
Yes, the simpler solution would be to check with SSL if non-SSL fails, but that might not be where the new file is at all moved to.

rxu:
> So solution could be to backport the guzzle update from master

Indeed, on master there's no such an issue. Good catch, backporting it looks like a good idea.

marc1706:
I've created a PR that backports these changes to 3.3.x:
https://github.com/phpbb/phpbb/pull/6715

LukeWCS:
I think this PR can be closed, as it is obsolete.

marc1706:
Closing as such. See #6715.

PR #6712, opened 2024-08-31
[ticket/17387] Fix PHP warnings in search results

rxu:
Checklist:

- [x] Tests pass

Tracker ticket:

<a href="https://tracker.phpbb.com/browse/PHPBB-17387">PHPBB-17387</a>.

### DIFF
--- a/phpBB/includes/functions_content.php
+++ b/phpBB/includes/functions_content.php
@@ -329,10 +329,10 @@ function get_context(string $text, array $words, int $length = 400): string
 {
 	if ($length <= 0)
 	{
-		return '...';
+		return $text;
 	}
 
-	// we need to turn the entities back into their original form, to not cut the message in between them
+	// We need to turn the entities back into their original form, to not cut the message in between them
 	$text = html_entity_decode($text);
 
 	// Replace all spaces/invisible characters with single spaces
--- a/phpBB/includes/functions_content.php
+++ b/phpBB/includes/functions_content.php
@@ -333,7 +333,7 @@ function get_context(string $text, array $words, int $length = 400): string
 	}
 
 	// We need to turn the entities back into their original form, to not cut the message in between them
-	$text = html_entity_decode($text);
+	$text = htmlspecialchars_decode($text);
 
 	// Replace all spaces/invisible characters with single spaces
 	$text = preg_replace("/\s+/u", ' ', $text);
@@ -414,7 +414,7 @@ function get_context(string $text, array $words, int $length = 400): string
 		$output[] = utf8_substr($fragment, $fragment_start, $fragment_end - $fragment_start + 1);
 	}
 
-	return ($fragments[0][0] !== 0 ? '... ' : '') . htmlentities(implode(' ... ', $output)) . ($end < $text_length - 1 ? ' ...' : '');
+	return ($fragments[0][0] !== 0 ? '... ' : '') . htmlspecialchars(implode(' ... ', $output)) . ($end < $text_length - 1 ? ' ...' : '');
 }
 
 /**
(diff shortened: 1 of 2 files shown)


rxu:
Also those `preg_match` and `preg_match_all` only matches numers and latin alphabet chars  (but not UTF8 ones), not sure if it affects the issue.

rubencm:
If the spaces are common to other languages, they can be used too as a reference. I used the first alphanumeric character because of the `.`, `,`, `!`, `?` ... symbols

phpBB/includes/functions_content.php
```diff
@@ -327,18 +327,14 @@ function bump_topic_allowed($forum_id, $topic_bumped, $last_post_time, $topic_po
 */
 function get_context(string $text, array $words, int $length = 400): string
 {
-	if ($length <= 0)
-	{
-		return '...';
```
rubencm, line 332: Just return $text, is more simple i think

phpBB/includes/functions_content.php
```diff
@@ -400,16 +396,14 @@ function get_context(string $text, array $words, int $length = 400): string
 		$fragment_end = $end - $start + 1;
 
 		// Find the first valid alphanumeric character in the fragment to don't cut words
-		if ($start > 0)
+		if ($start > 0 && preg_match('/[^a-zA-Z0-9][a-zA-Z0-9]/u', $fragment, $matches, PREG_OFFSET_CAPTURE))
```
rubencm, line 399: This can be changed to the next character to a space, point, comma, etc to be valid in other languages

phpBB/includes/functions_content.php
rxu: Makes sense.

phpBB/includes/functions_content.php
rxu: I don't really know how would it look like, could you please suggest a code change for that?

rubencm: Something like
```php
if ($start > 0 && preg_match('/[\s!?.,]+(.)/u', $fragment, $matches, PREG_OFFSET_CAPTURE))
```

rxu: I'd probably better just convert current regex to unicode which should yield the same result but for any language, like `/[^\p{L}\p{N}][\p{L}\p{N}]/ui`.

rubencm: Much better, I didn't know that regexp pattern

rubencm (approved):

rubencm:
The `i` modifier is not needed

rxu:
At this point it still fails in some complex cases, f.e. this
```php
$text = 'Manzana,plátano,naranja,fresa,mango,uva,piña,pera,kiwi,cereza,sandía,melón,papaya,arándano,durazno';
$words = ['plátano', 'arándano'];
$length = 10;

var_dump(get_context($text, $words, $length));
```

will return `"... plátano ... ...` instead of `... plátano ... arándano ...` (the latter will be returned if `$length` got increased to 20).

rxu:
[Deleted my previous comment as incorrect (that was messed up localhost board code).]
Now this seems to work fine with unicode strings.

rubencm:
> [Deleted my previous comment as incorrect (that was messed up localhost board code).] Now this seems to work fine with unicode strings.
> 
> EDIT: not exactly. F.e.
> 
> ```
> $text = 'Manzana,plátano,naranja,fresa,mango,uva,piña,pera,kiwi,cereza,sandía,melón,papaya,arándano,durazno';
> $words = ['plátano', 'arándano'];
> $length = 3;
> ```
> 
> will return `... átan ... ánda ...`.

Well thats normal since fragment length is only 3, smaller than the word, so it gets all the fragment.

rxu:
So the fix is ready then. Thanks.

marc1706 (approved):

PR #6736, opened 2024-10-17
[ticket/17411] Add core event to search.php

rxu:
Checklist:

- [x] Tests pass

Tracker ticket:

<a href="https://tracker.phpbb.com/browse/PHPBB-17411">PHPBB-17411</a>.

### DIFF
--- a/phpBB/search.php
+++ b/phpBB/search.php
@@ -92,7 +92,7 @@ switch ($search_id)
 
 $search_auth_check_override = false;
 /**
-* This event allows you to override serach auth checks
+* This event allows you to override search auth checks
 *
 * @event core.search_auth_check_override
 * @var	bool	search_auth_check_override	Whether or not the search auth check overridden
--- a/phpBB/search.php
+++ b/phpBB/search.php
@@ -90,8 +90,21 @@ switch ($search_id)
 	break;
 }
 
+$search_auth_check_override = false;
+/**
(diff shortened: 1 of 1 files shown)


rxu:
!set Event

phpBB/search.php
```diff
@@ -90,8 +90,21 @@
 	break;
 }
 
+$search_auth_check_override = false;
+/**
+* This event allows you to override serach auth checks
```
rubencm, line 95:
```suggestion
* This event allows you to override search auth checks
```

rubencm:
If I don't understand wrong, if you override the variable, you can bypass this check even if search is disabled.
I think what the user wants to do, can be done by using `core.common` and doing the logic there.

rxu:
> If I don't understand wrong, if you override the variable, you can bypass this check even if search is disabled. I think what the user wants to do, can be done by using `core.common` and doing the logic there.

Not just bypass but f.e. perform additional checks which can't be done there.

rubencm:
I think it should be possible, if the file is `search.php` just do anything you want like redirect the user to login if is anonymous

rxu:
There's an example in the topic linked in the ticket.

rubencm:
I have talked with @rxu in private, and this can be done using user setup events too, but this way is more specific.
Both ways are valid.

marc1706 (approved):

PR #6752, opened 2024-11-22
[ticket/17437] Prosilver v2 design for Triton (4.0)

danieltj27:
The prosilver v2 redesign that focuses on updating the current style codebase with a modern design and cleaner HTML and CSS structure allowing style authors to more easily create custom styles whilst also giving phpBB 4.0 a fresh look for the future. _This is very much a work in progress!_

Checklist:

- [ ] Tests pass

Tracker ticket:

[PHPBB-17437](https://tracker.phpbb.com/browse/PHPBB-17437)


phpBB/styles/prosilver/composer.json
```diff
@@ -19,7 +19,7 @@
 		"irc": "irc://irc.freenode.org/phpbb"
 	},
 	"extra": {
-		"display-name": "prosilver",
+		"display-name": "prosilver v2",
```
iMattPro, line 22: Well. It's off to a good start so far 🤣

danieltj27: Technically ready for merging 😂

danieltj27:
I'm going to make a big decision right now and say that this new version of prosilver will not include any preprocessors like Sass or similar. There are a couple of reasons for this:

1. I don't want to pick a preprocessor because whilst it would be very handy to use one, picking one means deciding on something and it's easier in my opinion to decide nothing than picking something someone might not want to use for whatever reason.
2. The most important reason is it would mean style authors, but more importantly forum admins would have to reverse engineer the CSS that is processed (and likely minified) to make simple style changes. I think it's best to avoid future headaches by making the CSS as simple as possible.

Having said all of that, it doesn't mean we can't or shouldn't use new CSS features like variables and so on. But keeping the stylesheets plain and simple to work with is important.

I've also checked out @marc1706 branch `feature/prosilver_25` and like the start he's made. I'm going to try and work that initially stuff into this if I can.

danieltj27:
Undone all changes to the base prosilver style and created a new prosilver2 style as per [this comment](https://tracker.phpbb.com/browse/PHPBB-17437?focusedId=66508&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-66508).

meis2m:
i love this idea... i can help you too if you want.

vinny:
> I'm going to make a big decision right now and say that this new version of prosilver will not include any preprocessors like Sass or similar. There are a couple of reasons for this:
> 
> 1. I don't want to pick a preprocessor because whilst it would be very handy to use one, picking one means deciding on something and it's easier in my opinion to decide nothing than picking something someone might not want to use for whatever reason.
> 2. The most important reason is it would mean style authors, but more importantly forum admins would have to reverse engineer the CSS that is processed (and likely minified) to make simple style changes. I think it's best to avoid future headaches by making the CSS as simple as possible.
> 
> Having said all of that, it doesn't mean we can't or shouldn't use new CSS features like variables and so on. But keeping the stylesheets plain and simple to work with is important.
> 
> I've also checked out @marc1706 branch and like the start he's made. I'm going to try and work that initially stuff into this if I can.`feature/prosilver_25`

I agree with you. Sass would be complicated for administrators without knowledge, just check the style forum, most of them can't change the logo, banner, the color of a link with simple CSS.

meis2m:
i have some suggestions:
1- we need ACP style settings , in style settings admin can change logo, color, favicon ... and access to html and css files directly.
2- in settings admin can choose grid option or forumlist option.
3- have a dark mode - users can select dark mode manually. some where in footer or header.
4- add widget system and have a site builder , admin can customise style something like elementor in wordpress or site builder in vbulletin6.

DavidIQ:
> i have some suggestions: 1- we need ACP style settings , in style settings admin can change logo, color, favicon ... and access to html and css files directly. 2- in settings admin can choose grid option or forumlist option. 3- have a dark mode - users can select dark mode manually. some where in footer or header. 4- add widget system and have a site builder , admin can customise style something like elementor in wordpress or site builder in vbulletin6.

None of that is going to be done here. Only the style is being refreshed, as the title and description suggest. Feel free to create tickets for each so they can be considered for a future version.

danieltj27:
@meis2m just as @DavidIQ has said this is purely focusing on the style itself.

I do agree that having style settings, particularly a unified way of editing styles like logos and so on would be great, but that is out of scope of this work and would require lots of back-end work to get the settings ready to be used by said style. Good idea and I'd like to see that sort of thing added too, but not right now 😏

vinny:
> i have some suggestions: 1- we need ACP style settings , in style settings admin can change logo, color, favicon ... and access to html and css files directly. 2- in settings admin can choose grid option or forumlist option. 3- have a dark mode - users can select dark mode manually. some where in footer or header. 4- add widget system and have a site builder , admin can customise style something like elementor in wordpress or site builder in vbulletin6.

All of this can be done with an extension, if you look in the CDB, there is already an extension for each suggestion.

PR #6753, opened 2024-11-23
[ticket/17361] Improve storage

rubencm:
PHPBB-17361

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-17361

It can be tested with this test extensions:
 - https://github.com/rubencm/phpbb-ext-storage-db
 - https://github.com/rubencm/phpbb-ext-storage-subfolders
 - https://github.com/rubencm/phpbb-ext-storage-s3

todo:
 - [ ] s3 extension with hotlinks
 - [x] check that files are migrated correctly when updating from 3.3.x
 - [ ] tests
 - [x] language keys in storage/adapter
 - [x] after write, fclose is not needed
 - [x] make sure stream is always closed after read
 - [ ] event in controller for hotlink?
 - [x] Use storage in `phpbb/console/command/thumbnail
/generate.php` or in create_thumbnail?, and `phpBB/phpbb/console/command/thumbnail/delete.php`
 - [x] Maybe, try somehow to return a BinaryFileResponse when storage is local to restore partial downloads if there is a decent way
 - [x] Fix attachments controller
 - [ ] Check that validate_data and validate_path in acp_storage work as expected
 - [x] custom storage extension
 - [x] Untrack files on migrations revert
 - [x] Track files should receive an array to support more than 1 file
 - [ ] Add method in storage to allow to call custom adapter methods
 - [ ] Configuration to disable editing https://github.com/phpbb-extensions/pwakit/pull/33#issuecomment-2640487834


phpBB/phpbb/console/command/thumbnail/delete.php
```diff
@@ -38,16 +48,16 @@ class delete extends \phpbb\console\command\command
...
-	* @param \phpbb\config\config $config The config
-	* @param \phpbb\user $user The user object (used to get language information)
-	* @param \phpbb\db\driver\driver_interface $db Database connection
-	* @param string $phpbb_root_path Root path
+	* @param user $user The user object (used to get language information)
+	* @param driver_interface $db Database connection
+	* @param language $language Language
+	* @param storage $storage Storage
 	*/
-	public function __construct(\phpbb\config\config $config, \phpbb\user $user, \phpbb\db\driver\driver_interface $db, $phpbb_root_path)
```
marc1706, line 46: I would recommend ordering these a-z as none of them are optional

phpBB/includes/functions_user.php
```diff
@@ -2124,7 +2124,8 @@ function group_correct_avatar($group_id, $old_entry)
 
 	try
 	{
-		$storage->rename($old_filename, $new_filename);
+		$storage->write($new_filename, $storage->read($old_filename));
```
marc1706, line 2127: I actually like the old rename, any reason to remove it?

phpBB/phpbb/console/command/thumbnail/generate.php
```diff
@@ -134,11 +147,15 @@ protected function execute(InputInterface $input, OutputInterface $output)
 		{
 			if (isset($extensions[$row['extension']]['display_cat']) && $extensions[$row['extension']]['display_cat'] == \phpbb\attachment\attachment_category::IMAGE)
 			{
-				$source = $this->phpbb_root_path . $this->config['upload_path'] . '/' . $row['physical_filename'];
-				$destination = $this->phpbb_root_path . $this->config['upload_path'] . '/thumb_' . $row['physical_filename'];
+				$source = tempnam(sys_get_temp_dir(), 'thumbnail_source');
```
marc1706, line 150: Can we maybe use filesystem classes for this?

phpBB/phpbb/console/command/thumbnail/recreate.php
```diff
@@ -12,13 +12,32 @@
...
+	 * @var language
+	 */
+	protected $language;
+
+	/**
+	 * Constructor
+	 *
+	 * @param language $language Language
+	 */
+	public function __construct(user $user, language $language)
```
marc1706, line 34: Change to a-z ordering

phpBB/phpbb/db/migration/data/v400/storage_backup_data.php
```diff
@@ -0,0 +1,64 @@
...
+			['custom', [[$this, 'update_backup_data']]],
+		];
+	}
+
+	public function update_backup_data()
+	{
+		$methods = ['sql', 'sql.gz', 'sql.bz2'];
+
+		$dir = $this->phpbb_root_path . 'store/';
+		$dh = @opendir($dir);
```
marc1706, line 39: Please use a more useful name than dh :)

phpBB/phpbb/storage/file_tracker.php
```diff
@@ -0,0 +1,198 @@
...
+
+namespace phpbb\storage;
+
+use phpbb\cache\driver\driver_interface as cache;
+use phpbb\db\driver\driver_interface as db;
+use phpbb\storage\exception\storage_exception;
+
+class file_tracker
+{
+
```
marc1706, line 22: remove blank line

phpBB/phpbb/storage/file_tracker.php
```diff
@@ -0,0 +1,198 @@
...
+	 */
+	public function track_file(string $storage, string $path, int $size): void
+	{
+		$sql_ary = array(
+			'file_path'		=> $path,
+			'storage'		=> $storage,
+			'filesize'		=> $size,
+		);
+
+		$sql = 'INSERT INTO ' . $this->storage_table . $this->db->sql_build_array('INSERT', $sql_ary);
```
marc1706, line 68: Do we need to first check if the file is already tracked or do we rely on this being done outside this function?

phpBB/phpbb/storage/file_tracker.php
```diff
@@ -0,0 +1,198 @@
...
+	 * @return bool	True if file is tracked
+	 */
+	public function is_tracked(string $storage, string $path): bool
+	{
+		$sql_ary = array(
+			'file_path'		=> $path,
+			'storage'		=> $storage,
+		);
+
+		$sql = 'SELECT file_id FROM ' .  $this->storage_table . '
```
marc1706, line 111: FROM part on next line, happens more often in this file

phpBB/phpbb/storage/file_tracker.php
```diff
@@ -0,0 +1,198 @@
...
+	 *
+	 * @param string $path The file
+	 *
+	 * @return int Size in bytes.
+	 *
+	 * @throws storage_exception When unable to retrieve file size
+	 */
+	public function file_size(string $storage, string $path): int
+	{
+		$sql_ary = array(
```
marc1706, line 131: Use short array syntax with `[]`

marc1706 (changes requested):

phpBB/includes/functions_user.php
rubencm: Yes, because is not used anywhere, only here, and this is probably never executed.

phpBB/phpbb/storage/file_tracker.php
rubencm: Probably not, this is only executed when the file is not tracked.
Also, if it is, the insert should throw an error because of the unique index i added in the migration

rubencm: This is covered by the unique index key and the exists check in the write method in storage class

phpBB/phpbb/console/command/thumbnail/delete.php
rubencm: I have undone this to follow the format of the rest of the commands.
First goes user (that is from the parent class) and then the rest

iMattPro:
@rubencm I noticed a little issue in filespec_storage - not this version but the currently merged version - where failed uploaded files are not being removed. For example, I took a text file but gave it an image extension, so it looks like a PNG, but it's not really a PNG. So I try to upload it and get an error message that it was not a real image. 

However the logic in the `move_file()` method is such that, because there is an error message, we return out of the move_file method before setting `file_moved` to true, so in the end, the file never actually gets deleted. It stays on the filesystem and in the storage DB. 

Seems like when we get to the `remove()` method, since file_moved is still false even though the file was in fact moved, all that runs is the `unlink` but that is being run using what appears to be a temporary PHP stream of the file and not the destination file itself.

So in the end, the failed uploaded file still ends up being uploaded and being tracked in the DB.

phpBB/phpbb/files/filespec_storage.php
```diff
@@ -446,12 +446,7 @@ public function move_file($storage, $overwrite = false, $skip_image_check = fals
 		{
 			$fp = fopen($this->filename, 'rb');
 
-			$storage->write_stream($this->destination_file, $fp);
-
-			if (is_resource($fp))
-			{
-				fclose($fp);
-			}
+			$storage->write($this->destination_file, $fp);
```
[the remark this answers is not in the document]
marc1706: Can you explain which file does not get removed. The initially uploaded temp file or the the destination file? I think if the image check fails, this part still works fine and hence the file is correctly moved to the destination. It was like that before.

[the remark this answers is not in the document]
iMattPro: The temp file is removed but the destination/actual file is not removed.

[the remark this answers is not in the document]
marc1706: That's what I thought. I'd commend setting `$this->file_moved` to true when removing the temporary file on line 458 and before returning. If there is no exception we can assume that the file was moved.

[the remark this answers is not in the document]
iMattPro: If we do that though, if there was a legitimate storage_exception thrown and caught here at line 451, and file_moved is set to false, wouldn't we just set that incorrectly right back to true then if we did this at line 458?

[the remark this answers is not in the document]
rubencm: What about just replace remove method by

```php
	/**
	 * Remove file
	 */
	public function remove($storage)
	{
		$storage->delete($this->destination_file);
	}

```

If not feel free to fix it in another ticket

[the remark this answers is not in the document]
iMattPro: I think that would work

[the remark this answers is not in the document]
rubencm: I removed `file_moved`, I hope I have not break more things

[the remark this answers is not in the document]
rubencm: It doesn't work, it works, but the tests dont pass, you can try to fix it in another pr if you want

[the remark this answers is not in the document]
iMattPro: Why in another PR? From my tests it works as it should when adding–or moving–(I don't understand all the moving parts going on here) `$this->file_moved = true;` after `$storage->write($this->destination_file, $fp);` here on line 449.

marc1706 (approved):

PR #6758, opened 2024-12-05
[ticket/17381] Topic view count validation

battye:
The user mentioned in the ticket that they were getting an error `SQL ERROR [ mysqli ] Out of range value for column 'topic_views' at row 1 [1264]` and I was able to replicate this by forcing the topic_views for a topic to to 16777215 and trying to load the page.

This fix validates that the incremented topic_views value won't exceed that mediumint maximum figure.

Note: I was hesitant to use a magic number for this, but that's how we've previously handled this exact use case in two other places:

https://github.com/phpbb/phpbb/blob/259d086e456cf642aca1b822dc63323512dead34/phpBB/phpbb/session.php#L147 https://github.com/phpbb/phpbb/blob/259d086e456cf642aca1b822dc63323512dead34/phpBB/includes/acp/acp_groups.php#L464

PHPBB-17381

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/projects/PHPBB/issues/PHPBB-17381

### DIFF
--- a/phpBB/phpbb/db/migration/data/v33x/topic_views_update.php
+++ b/phpBB/phpbb/db/migration/data/v33x/topic_views_update.php
@@ -1,0 +1,47 @@
+<?php
+/**
+*
+* This file is part of the phpBB Forum Software package.
+*
+* @copyright (c) phpBB Limited <https://www.phpbb.com>
+* @license GNU General Public License, version 2 (GPL-2.0)
+*
+* For full copyright and license information, please see
+* the docs/CREDITS.txt file.
+*
+*/
+
+namespace phpbb\db\migration\data\v33x;
+
+class topic_views_update extends \phpbb\db\migration\migration
+{
+	public static function depends_on()
+	{
+		return [
+			'\phpbb\db\migration\data\v33x\v3314',
+		];
+	}
+
+	public function update_schema(): array
+	{
+		// This extends the topic view count field so we can support much larger values.
+		return [
+			'change_columns' => [
+				$this->table_prefix . 'topics' => [
+					'topic_views'  => ['ULINT', 0],
+				],
+			]
+		];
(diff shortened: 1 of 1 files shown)


rxu:
Then probably also change `topic_views` from `mediumint` to something longer? f.e. `ULINT`.

phpBB/viewtopic.php
```diff
@@ -2263,9 +2263,22 @@
...
+	if ($incremented_topic_views <= 16777215)
+	{
+		$topic_sql['topic_views'] = $incremented_topic_views;
+	}
+
 	$sql = 'UPDATE ' . TOPICS_TABLE . '
-		SET topic_views = topic_views + 1, topic_last_view_time = ' . time() . "
+		SET ' . $db->sql_build_array('UPDATE', $topic_sql) . "
 		WHERE topic_id = $topic_id";
+
```
kaileymsnay, line 2281: Just curious, why the space?

marc1706:
@battye I agree with @rxu, we should migrate this to a bigger column type.

Paullht:
And welcome all new members to the community happy to hear you all in the community ok LFG 🚀🚀 and don't forget to get a bag 🛍️🛍️ ok fam LFG 🚀🚀🚀

rxu:
Is viewtopic.php change still needed?

battye:
I don't think the original logic is needed any more @rxu now there's a data type change. Thanks also @marc1706 and @kaileymsnay.

I've verified the migration with `php phpBB/bin/phpbbcli.php db:list` and `php phpBB/bin/phpbbcli.php db:migrate`:

Screenshot 2025-02-28 at 11 24 41 AM

marc1706 (approved):

PR #6766, opened 2024-12-23
[ticket/17447] Add http range requests to attachments downloads

rubencm:
Checklist:

- [ ] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-12345


phpBB/phpbb/storage/controller/attachment.php
```diff
@@ -247,67 +255,126 @@ public function handle_attachment(int $id, string $filename): Response
...
 			);
 		}
 
+		if ($this->config['storage\\attachment\\provider'] === local::class)
+		{
+			$response = new BinaryFileResponse($this->config['storage\\attachment\\config\\path'] . '/' . $attachment['physical_filename']);
+		}
+		else
+		{
+			$response = new StreamedResponse();
```
marc1706, line 290: I assume this is only done for cloud providers then and physical_filename will be the cloud provider url?

phpBB/phpbb/storage/controller/attachment.php
```diff
@@ -247,67 +255,126 @@ public function handle_attachment(int $id, string $filename): Response
...
+		// Don't cache attachments on proxies
+		$response->setPrivate();
+
+		// Content-type and content-disposition headers
+		$response->headers->set('Content-Type', $attachment['mimetype']);
 		$response->headers->set('Content-Disposition', $disposition);
 
-		// Set expires header for browser cache
-		$time = new \DateTime();
-		$response->setExpires($time->modify('+1 year'));
```
marc1706, line 288: You're no longer setting expires to +1 year?

phpBB/phpbb/storage/controller/attachment.php
rubencm: This is for cloud providers yes, and filename will be real_filename, set in disposition header.
Physical filename is never exposed
All files are downloaded through this controller, who read the file from the provider, the provider url is not exposed.

phpBB/phpbb/storage/controller/attachment.php
rubencm: After a year I dont remember why I made this decision.
Probably because of when you set `$response->isPrivate`  that header was ignored

marc1706: Ok, fine with me

PR #6774, opened 2025-01-27
Bug-Fix: Edit profile with deactivated Modul Edit profile

nice42q:
Removes "&mode=profile_info" `&amp;mode=profile_info`
If the module (Edit profile) is deactivated in ACP -> System -> Module management -> Profile, the ‘Edit profile’ function no longer works. This change corrects the error.

Checklist:

- [x] Tests pass


marc1706:
Can you clarify a bit of what you are trying to fix? If one disables a core module, this can indeed result in some unexpected side effects. Apparently the link not working properly is one of them. However, this seems like a misconfiguration and the change would seemingly break the link for everyone else that does not have this misconfiguration.

iMattPro:
The proper solution to this I think would be to not display the "Edit Profile" link if the module is disabled.

PR #6779, opened 2025-02-02
[ticket/17465] Add unit tests for handling of web push notifications

marc1706:
Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-17465

### DIFF
--- a/phpBB/phpbb/ucp/controller/webpush.php
+++ b/phpBB/phpbb/ucp/controller/webpush.php
@@ -165,6 +165,11 @@ class webpush
 		$notification_data = $this->db->sql_fetchfield('push_data');
 		$this->db->sql_freeresult($result);
 
+		if (!$notification_data)
+		{
+			throw new http_exception(Response::HTTP_BAD_REQUEST, 'AJAX_ERROR_TEXT');
+		}
+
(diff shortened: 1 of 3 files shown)


phpBB/phpbb/ucp/controller/webpush.php
```diff
@@ -135,7 +135,6 @@ public function notification(): JsonResponse
 
 		// Decode and return data if everything is fine
 		$data = json_decode($notification_data, true);
-		$data['url'] = isset($data['url']) ? $this->path_helper->update_web_root_path($data['url']) : '';
```
marc1706, line 138: Still need to figure out if this is / was needed.

iMattPro: I'm just glad I wasn't who came up with this line so I can't be directly blamed for it's existence 😂

iMattPro: By the way, if we don't need that line, there's no need to decode the JSON to an array. We can just send the `$notification_data` in the `JsonResponse` using the `$json` argument:
```php
return new JsonResponse($notification_data(), 200, [], true);
```

PR #6780, opened 2025-02-16
[ticket/17470] Feed enable config not enforced

Neo-CTC:
The `feed_enable` config is never checked before generating the feeds. The effective result is that feeds are always enabled, regardless of the setting.

PHPBB-17470

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-17470


marc1706:
First of thanks for your PR. I have modified it to not check inside a constructor since that's really not what constructors are meant for in a controller. I've also changed this to use the already existing language string and added a simple functional test.

PR #6784, opened 2025-02-28
fix: DOM text reinterpreted as HTML Improper Neutralization of Input Cross-site Scripting

odaysec:
https://github.com/phpbb/phpbb/blob/3584ef7ecab35fdec2890b07b7fae79aced51eb6/phpBB/styles/prosilver/template/forum_fn.js#L602-L602

Fix the problem, we need to ensure that any text content derived from the DOM and used in HTML is properly escaped to prevent XSS attacks. The best way to fix this issue is to use a function that escapes HTML special characters before concatenating the text into the HTML string.

A webpage with this vulnerability reads text from the DOM, and afterwards adds the text as HTML to the DOM. Using text from the DOM as HTML effectively unescapes the text, and thereby invalidates any escaping done on the text. If an attacker is able to control the safe sanitized text, then this vulnerability can be exploited to perform a cross-site scripting attack.

## POC
The following vulnerable shows a webpage using a `data-target` attribute to select and manipulate a DOM element using the JQuery library. In the `data-target` attribute is read into the `target` variable, and the `$` function is then supposed to use the `target` variable as a CSS selector to determine which element should be manipulated.
```js
$("button").click(function () {
    var target = $(this).attr("data-target");
    $(target).hide();
});
```
However, if an attacker can control the `data-target` attribute, then the value of `target` can be used to cause the `$` function to execute arbitrary JavaScript.

The above vulnerability can be fixed by using `$.find` instead of `$`. The `$.find` function will only interpret `target` as a CSS selector and never as HTML, thereby preventing an XSS attack.
```js
$("button").click(function () {
    var target = $(this).attr("data-target");
	$.find(target).hide();
});
```
## References
- [DOM based XSS Prevention Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/DOM_based_XSS_Prevention_Cheat_Sheet.html).
- [XSS (Cross Site Scripting) Prevention Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html).
- [DOM Based XSS](https://owasp.org/www-community/attacks/DOM_Based_XSS).
- [Types of Cross-Site Scripting](https://owasp.org/www-community/Types_of_Cross-Site_Scripting).
- [Cross-site scripting](http://en.wikipedia.org/wiki/Cross-site_scripting).
- [CWE-79](https://cwe.mitre.org/data/definitions/79.html).
- [CWE-116](https://cwe.mitre.org/data/definitions/116.html).

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-12345

### DIFF
--- a/phpBB/styles/prosilver/template/forum_fn.js
+++ b/phpBB/styles/prosilver/template/forum_fn.js
@@ -5,6 +5,20 @@
 */
 
 /**
+* Escape HTML special characters
+*/
+function escapeHtml(text) {
+	var map = {
+		'&': '&amp;',
+		'<': '&lt;',
+		'>': '&gt;',
+		'"': '&quot;',
+		"'": '&#039;'
+	};
+	return text.replace(/[&<>"']/g, function(m) { return map[m]; });
+}
+
+/**
 * Find a member
 */
 function find_username(url) {
@@ -599,7 +613,7 @@ function parseDocument($container) {
 				}
 
 				if ((text.length && text !== '-') || cell.children().length) {
-					cell.prepend('<dfn style="display: none;">' + headers[column] + '</dfn>');
+					cell.prepend('<dfn style="display: none;">' + escapeHtml(headers[column]) + '</dfn>');
 				} else {
 					cell.addClass('empty');
 				}


marc1706:
Please use the appropriate channels for creating reports about potential security issues. We have also added a securities page now that should be a good start for reporting potential issues:
https://github.com/phpbb/phpbb/security/policy

PR #6803, opened 2025-04-15
[ticket/17496] Fix Implicitly marking parameters as nullable PHP deprecations

rxu:
Checklist:

- [ ] Tests pass

Tracker ticket:

<a href="https://tracker.phpbb.com/browse/PHPBB-17496">PHPBB-17496</a>.

### DIFF
--- a/phpBB/composer.json
+++ b/phpBB/composer.json
@@ -39,7 +39,7 @@
 		"composer/package-versions-deprecated": "^1.11",
 		"doctrine/dbal": "~3.3.6",
 		"google/recaptcha": "~1.1",
-		"guzzlehttp/guzzle": "~6.3",
+		"guzzlehttp/guzzle": " ^7.0",
 		"marc1706/fast-image-size": "^1.1",
 		"minishlink/web-push": "^8.0",
 		"s9e/text-formatter": "^2.0",
--- a/phpBB/phpbb/ban/type/email.php
+++ b/phpBB/phpbb/ban/type/email.php
@@ -28,7 +28,7 @@ class email extends base
 	/**
 	 * {@inheritDoc}
 	 */
-	public function get_user_column(): ?string
+	public function get_user_column(): string|null
 	{
 		return 'user_email';
 	}
--- a/build/code_sniffer/phpbb/Sniffs/Namespaces/UnusedUseSniff.php
+++ b/build/code_sniffer/phpbb/Sniffs/Namespaces/UnusedUseSniff.php
@@ -49,11 +49,20 @@ class phpbb_Sniffs_Namespaces_UnusedUseSniff implements Sniff
 			$phpcsFile->addError($error, $stack_pointer, 'FullName');
 		}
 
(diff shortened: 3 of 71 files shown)


phpBB/composer.json
```diff
@@ -39,7 +39,7 @@
 		"composer/package-versions-deprecated": "^1.11",
 		"doctrine/dbal": "~3.3.6",
 		"google/recaptcha": "~1.1",
-		"guzzlehttp/guzzle": "~6.3",
+		"guzzlehttp/guzzle": " ^7.0",
```
marc1706, line 42: I take it guzzle had to be updated?

phpBB/phpbb/ban/type/email.php
```diff
@@ -28,7 +28,7 @@ public function get_type(): string
 	/**
 	 * {@inheritDoc}
 	 */
-	public function get_user_column(): ?string
+	public function get_user_column(): string|null
```
marc1706, line 31: I guess it's more readable this way but was there another reason you went with this format?

iMattPro: I would kind-of prefer the `?string` over the `string|null` version just because it better resembles how we deal with this in the config.yml too.

Also a question, don't any of these null params that appear in class constructors need to be done in the config.yml files too? For example in `phpBB/phpbb/composer/installer.php` if `config|null $config = null)` then shouldn't `'@config'` in this class's definition in the config yaml be `'@?config'`

phpBB/composer.json
rxu: Yeah, it's been throwing these depreciations too in background.

phpBB/phpbb/ban/type/email.php
rxu: > I guess it's more readable this way but was there another reason you went with this format?

Nope, also it's just for consistency as different syntaxes were used in different places.

rxu: >  Or am I just confusing this nullable stuff with optional args?

Probably yes, it can be passed as null explicitly.

iMattPro:
@marc1706 

waiting-wait

phpBB/phpbb/ban/type/email.php
marc1706: I have proposed this change to the coding guidelines for 4.0:
https://github.com/phpbb/documentation/pull/299

rxu: Then code sniffer should check that probably.

rxu: > Then code sniffer should check that probably.

May be in another ticket/PR? This requires more work to dig into custom sniffs, couldn't find ready ones.

marc1706: Sure, fine with me

rxu:
Requires #6810 tests fix.

marc1706 (approved):

PR #6838, opened 2025-07-18
[ticket/17535] Upgrade PHPUnit to version 10

rxu:
Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-17535


tests/bbcode/parser_test.php
```diff
@@ -249,10 +249,18 @@ public function test_bbcode_firstpass($description, $message, $expected, $incomp
 			$this->markTestIncomplete($incomplete);
 		}
 
-		global $user, $request, $symfony_request;
+		global $user, $request, $symfony_request, $phpbb_dispatcher, $config, $phpEx, $request, $symfony_request;
```
iMattPro: should be:
global $user, $request, $symfony_request, $phpbb_dispatcher, $config, $phpEx;

tests/bbcode/url_bbcode_test.php
```diff
@@ -52,10 +52,18 @@ public function url_bbcode_test_data()
 	*/
 	public function test_url($description, $message, $expected)
 	{
-		global $user, $request, $symfony_request;
+		global $user, $request, $symfony_request, $phpbb_dispatcher, $config, $phpEx, $request, $symfony_request;
```
iMattPro: should be:
global $user, $request, $symfony_request, $phpbb_dispatcher, $config, $phpEx;

tests/bbcode/parser_test.php
rxu: I've looked into the chain of methods called in deep and had to add more global vars.

tests/bbcode/url_bbcode_test.php
rxu: Same as above.

tests/help/manager_test.php
```diff
@@ -78,25 +78,27 @@ public function test_add_block($block_name, $switch, $questions, $switch_expecte
...
-			);
+			->willReturnCallback(function ($arg1, $arg2) use ($matcher, $block_name, $switch_expected, $template_call_args) {
+				$callNr = $matcher->numberOfInvocations();
+				match (true) {
+					$callNr == 1 => $this->assertEquals([$arg1, $arg2], ['faq_block', [
+							'BLOCK_TITLE' => strtoupper($block_name),
+							'SWITCH_COLUMN' => $switch_expected,
+						]]),
+					$callNr > 1 => $this->assertEquals([$arg1, $arg2], $template_call_args[$callNr - 2]),
+				};});
```
iMattPro: This can also be written like:
```php
$expected_calls = [
	['faq_block', [
		'BLOCK_TITLE' => strtoupper($block_name),
		'SWITCH_COLUMN' => $switch_expected,
	]],
	...$template_call_args
];

$this->template->expects($this->exactly(count($questions) + 1))
	->method('assign_block_vars')
	->willReturnCallback(function($blockName, $blockVars) use (&$expected_calls) {
		static $call = 0;
		$this->assertEquals($expected_calls[$call][0], $blockName);
		$this->assertEquals($expected_calls[$call][1], $blockVars);
		$call++;
	});
```
I think this is cleaner, does not rely on setting conditions based on call being equal to or greater than 1. This would also apply to other places in this PR

tests/bbcode/parser_test.php
iMattPro: You've duplicated `$request, $symfony_request`

tests/bbcode/url_bbcode_test.php
iMattPro: duplication

tests/bbcode/parser_test.php
rxu: Oh. Okay, that's to be double-sure lol.
Thanks.

tests/help/manager_test.php
rxu: I don't fully understand what change to what. `$template_call_args` to `$expected_calls`?
I'd rather leave it as is as far as it just working.

rxu: @iMattPro Thanks for the PR.

phpBB/phpbb/cache/driver/memory.php
```diff
@@ -45,6 +45,30 @@ function __construct()
 		}
 	}
 
+	/**
+	* {@inheritDoc}
+	*/
+	function purge()
```
marc1706, line 51: This change should result in this bug again as regression:
https://tracker.phpbb.com/browse/PHPBB-12789

tests/auth/provider_apache_test.php
```diff
@@ -74,8 +74,10 @@ public function test_login()
 			->will($this->returnValue(true));
 		$this->request->expects($this->exactly(2))
 			->method('server')
-			->withConsecutive(['PHP_AUTH_USER'], ['PHP_AUTH_PW'])
-			->will($this->onConsecutiveCalls($this->returnValue('foobar'), $this->returnValue('example')));
+			->willReturnCallback(fn(string $arg) => match(true) {
```
marc1706: How about https://docs.phpunit.de/en/10.5/test-doubles.html#willreturnmap instead?

tests/dbal/connect_test.php
```diff
@@ -31,15 +31,23 @@ public function test_failing_connect()
 		// Failure to connect results in a trigger_error call in dbal.
 		// phpunit converts triggered errors to exceptions.
 		// In particular there should be no fatals here.
-		try
+		
+
+		if ($db->get_sql_layer() === 'mysqli')
+		{
+				$this->setExpectedTriggerError(E_WARNING);
```
marc1706: Why two indents?

tests/functional/extension_acp_test.php
```diff
@@ -255,6 +255,7 @@ public function test_extensions_catalog()
 
 		$form = $crawler->selectButton('Submit')->form();
 		$form['minimum_stability']->select('dev');
+		$form['enable_packagist']->select('1');
```
marc1706: Is this required for the test? Might cause other tests to be running slower IMHO.

tests/text_processing/message_parser_test.php
```diff
@@ -60,14 +60,15 @@ protected function prepare_s9e_services($setup = null)
...
-		$user->lang = array(
-			'NO_POLL_TITLE' => 'You have to enter a poll title.',
-			'POLL_TITLE_TOO_LONG' => 'The poll title must contain fewer than 100 characters.',
-			'POLL_TITLE_COMP_TOO_LONG' => 'The parsed size of your poll title is too large, consider removing BBCodes or smilies.',
-			'TOO_FEW_POLL_OPTIONS' => 'You must enter at least two poll options.',
-			'TOO_MANY_POLL_OPTIONS' => 'You have tried to enter too many poll options.',
-			'TOO_MANY_USER_OPTIONS' => 'You cannot specify more options per user than existing poll options.',
-		);
+		$user->expects($this->any())
+			-> method('__get')->with('lang')->willReturn([
```
marc1706:
```suggestion
			->method('__get')->with('lang')->willReturn([
```

tests/functional/extension_acp_test.php
rxu: The issue here was that `satis.phpbb.com` sometimes returns an empty extensions list for some reason which causes tests to fail.

phpBB/phpbb/cache/driver/memory.php
rxu: Tests fail without this because `driver/base` doesn't reach to unsetting cache vars in case if cached directories don't exist which is caused by this code snippet https://github.com/phpbb/phpbb/blob/e3b837f1b78877351fae2e33d2dc62b495f4ea3d/phpBB/phpbb/cache/driver/base.php#L35-L38
Alternatively, that `return` could be removed in the base class probably.

rxu: Just recalled, another issue with the base class `purge()` method is that it defines `$this->is_modified = false;` which prevents purged cache vars from being saved in memory driver.

https://github.com/phpbb/phpbb/blob/e3b837f1b78877351fae2e33d2dc62b495f4ea3d/phpBB/phpbb/cache/driver/base.php#L75


That causes tests to fail too.

tests/dbal/connect_test.php
rxu: Typo.

phpBB/phpbb/cache/driver/memory.php
rxu: Addressed the cache issue another way and it seems to be working too.

tests/dbal/connect_test.php
```diff
@@ -31,15 +31,23 @@ public function test_failing_connect()
 		// Failure to connect results in a trigger_error call in dbal.
 		// phpunit converts triggered errors to exceptions.
 		// In particular there should be no fatals here.
-		try
+
```
marc1706: Any reason for the two white lines? Otherwise I'd remove them :)

PR #6846, opened 2025-09-11
[ticket/17540] Use windows 2025 runners and limit to single PHP version

marc1706:
PHPBB-17540

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-17540

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -145,6 +145,10 @@ jobs:
                       db: "mysql:8.0"
                     - php: '8.4'
                       db: "mariadb:10.3"
+                    - php: '8.5'
+                      db: "mysql:8.0"
+                    - php: '8.5'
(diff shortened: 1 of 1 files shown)


phpBB/phpbb/db/migration/data/v400/rename_duplicated_index_names.php
```diff
@@ -27,7 +27,7 @@ class rename_duplicated_index_names extends migration
 	/**
 	 * @var array
 	 */
-	protected static $rename_index;
+	protected static $rename_index = [];
```
rxu: This way `!isset(self::$rename_index)` below will always return false and rename index array will never be filled with actual data.
Probably it would be better adding `self::$rename_index = [];` right after `if (!isset(self::$rename_index)) {`, around line 43.

PR #6848, opened 2025-09-16
[ticket/17542] Mock event dispatcher override early exits

iMattPro:
PHPBB-17542

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-17542

### DIFF
--- a/tests/mock/event_dispatcher.php
+++ b/tests/mock/event_dispatcher.php
@@ -22,8 +22,20 @@ class phpbb_mock_event_dispatcher extends \phpbb\event\dispatcher
 	{
 	}
 
-	public function trigger_event($eventName, $data = array())
+	public function trigger_event($eventName, $data = array()): array
 	{
-		return array();
+		// Ensure tests never hard-exit when phpBB calls exit_handler()
+		if ($eventName === 'core.exit_handler')
+		{
+			// Set the override flag so exit_handler() returns instead of exit;
+			if (is_array($data))
+			{
+				$data['exit_handler_override'] = true;
+			}
+			return (array) $data;
+		}
+
+		// Default behaviour of the mock: return the input data unchanged
+		return (array) $data;
 	}
 }
--- a/tests/mock/event_dispatcher.php
+++ b/tests/mock/event_dispatcher.php
@@ -24,18 +24,13 @@ class phpbb_mock_event_dispatcher extends \phpbb\event\dispatcher
 
 	public function trigger_event($eventName, $data = array()): array
 	{
-		// Ensure tests never hard-exit when phpBB calls exit_handler()
+		$data = (array) $data;
+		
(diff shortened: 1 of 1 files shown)


tests/mock/event_dispatcher.php
```diff
@@ -22,8 +22,20 @@ public function __construct()
...
-		return array();
+		// Ensure tests never hard-exit when phpBB calls exit_handler()
+		if ($eventName === 'core.exit_handler')
+		{
+			// Set the override flag so exit_handler() returns instead of exit;
+			if (is_array($data))
+			{
+				$data['exit_handler_override'] = true;
+			}
+			return (array) $data;
```
toxyy: I think this line (line 35) could be safely removed, unless there's a reason to have two returns, for debug purposes maybe?

iMattPro: Every tool at my disposal says both returns are needed. I had it explained to me by my IDE:

> The early return is needed and should be kept. Here's why:
> 
> The method has two distinct behaviors:
> 1. Special handling for core.exit_handler: Sets the override flag and returns
> 2. Default behavior: Returns data unchanged for all other events
> 
> Without the early return, the `core.exit_handler` case would fall through and execute the default behavior, potentially overwriting the `exit_handler_override` flag that was just set.
> 
> The current structure is correct and minimal - it handles the special case first, then provides the default behavior for everything else.

toxyy: Strictly looking at just this function, I disagree with that LLM's analysis. If it's worth anything, I can link to [Gemini 2.5 Pro's analysis](https://gist.github.com/toxyy/92fff74ff44ea93af06c9382f61930b4), or [ChatGPT 5's](https://gist.github.com/toxyy/b045d6eb6de8fb8113f665d6a8af64f9).

Forgetting LLM wars, I still disagree with your IDE's analysis:
`Without the early return, the core.exit_handler case would fall through and execute the default behavior, potentially overwriting the exit_handler_override flag that was just set.`
This is wholly untrue. If we walk through the function:
- default behavior is to return `(array) $data`
- If `$eventName` is `core.exit_handler`, and `$data` is an array, then `$data['exit_handler_override']` is set to `true`.
- `$data` will always be modified "if `$eventName` is `core.exit_handler`, and `$data` is an array".
- Therefore, no matter if $eventName is set to `core.exit_handler`, or if `$data` is an array or not, `(array) $data` is always returned.

See this test example I made. As much as I've been able to test this function, if the first `return (array) $data` were deleted, all results are equal to the previous function in all cases:
https://gist.github.com/toxyy/9e88736c9f536c246e2c19fc969558bf

iMattPro: It should be even better now

toxyy: Looks great to me

PR #6852, opened 2025-09-24
[ticket/17543] Optimize functional tests running time

rxu:
Slight functional tests framework optimization by:

- using persistent database connection
- closing database connection only after the last test got finished  in a testsuite
- only performing extensions setup before the 1st test in a testsuite
- only performing extensions uninstall after the last test in a testsuite
- skip unsupported search backends tests before the `setUp()` method call
- decreasing pausing time in some tests

This seems to save 1-2 minutes for Unix-based tests and 8-9 minutes for Windows functional tests.

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-17543

### DIFF
--- a/tests/functional/extension_module_test.php
+++ b/tests/functional/extension_module_test.php
@@ -23,34 +23,22 @@ class phpbb_functional_extension_module_test extends phpbb_functional_test_case
 		'./',
 	);
 
-	static public function setUpBeforeClass(): void
+	public static function setUpBeforeClass(): void
 	{
 		parent::setUpBeforeClass();
 
 		self::$helper = new phpbb_test_case_helpers(__CLASS__);
 		self::$helper->copy_ext_fixtures(__DIR__ . '/fixtures/ext/', self::$fixtures);
+
+		self::install_ext('foo/bar');
 	}
 
-	static public function tearDownAfterClass(): void
+	public static function tearDownAfterClass(): void
 	{
 		parent::tearDownAfterClass();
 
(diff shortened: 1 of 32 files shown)


tests/functional/extension_module_test.php
```diff
@@ -40,14 +40,17 @@ static public function tearDownAfterClass(): void
 
 	protected function tearDown(): void
 	{
-		$this->uninstall_ext('foo/bar');
+		if (self::$tests_count == 1)
```
Derky: I think the variable name `$tests_count` is too ambiguous to determine if extensions need to be uninstalled

rxu: It's to determine if the last test in the queue has been executed (1 means exactly that), and then extensions get uninstalled.
What name would suite for that purpose?

Derky: Can you perhaps use `tearDownAfterClass()` for this use case? https://docs.phpunit.de/en/10.5/fixtures.html#sharing-fixture

rxu: That's not possible with the current tests framework codebase as it's not allowed to use `$this` keyword in static methods.
But I can look if making `install_ext()`/`uninstall_ext()` static is viable.

Derky: Yeah sounds good, thanks!

marc1706 (approved):

PR #6864, opened 2025-09-27
[ticket/17469] Automatic updater front end

CHItA:
PHPBB3-17469

Checklist:

- [ ] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-17469

### DIFF
--- a/phpBB/adm/style/admin.js
+++ b/phpBB/adm/style/admin.js
@@ -246,6 +246,57 @@ function parse_document(container)
 	});
 }
 
+async function send_ajax_request(elem)
+{
+	const method = (elem.hasAttribute('data-method')) ? elem.getAttribute('data-method') : 'get';
+	const params = new URLSearchParams();
+	if (method === 'post')
+	{
+		for (const attr of elem.attributes) 
+		{
+			if (attr.name.startsWith('data-post-'))
+			{
+				params.append(attr.name.substring(10).replace('-', '_'), attr.value);
+			}
+		}
+	}
+	
+	const response = await fetch(elem.getAttribute('data-url'), {
+		method: method,
+		body: params,
+	});
+	if (!response.ok)
+	{
+		phpbb.alert(response.status, response.statusText);
+		return;
+	}
+	
+	const resp = await response.json();
+	switch (resp.type)
+	{
+		case 'error':
+			phpbb.alert(resp.status, resp.error);
+			return;
+		case 'resubmit':
+			await send_ajax_request(elem);
+			return;
+		case 'message':
+			phpbb.alert(resp.status, resp.msg);
+			return;
+	}
+}
+
+function attach_ajax() 
+{
+	const elems = document.querySelectorAll('a[data-url]');
+	for (const elem of elems) 
+	{
+		elem.addEventListener('click', async() => {
+			await send_ajax_request(elem);
+		});
+	}
+}
+
 /**
 * Run onload functions
 */
@@ -260,6 +311,7 @@ function parse_document(container)
 		});
 
 		parse_document($('body'));
+		attach_ajax();
 
 		$('#questionnaire-form').css('display', 'none');
 		var $triggerConfiglist = $('#trigger-configlist');
--- a/phpBB/config/default/container/services_updater.yml
+++ b/phpBB/config/default/container/services_updater.yml
@@ -11,4 +11,10 @@ services:
         arguments:
             - '@filesystem'
             - '@updater.get_updates'
+            - '@language'
+            - '@user'
+            - '@config'
             - '%core.root_path%'
+
+    updater.url_provider:
+        class: phpbb\update\update_info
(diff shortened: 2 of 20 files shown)


phpBB/adm/style/admin.js
```diff
@@ -246,6 +246,57 @@ function parse_document(container)
...
+{
+	const method = (elem.hasAttribute('data-method')) ? elem.getAttribute('data-method') : 'get';
+	const params = new URLSearchParams();
+	if (method === 'post')
+	{
+		for (const attr of elem.attributes) 
+		{
+			if (attr.name.startsWith('data-post-'))
+			{
+				params.append(attr.name.substr(10).replace('-', '_'), attr.value);
```
marc1706: Good news, substr() is deprecated:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr

phpBB/adm/style/admin.js
```diff
@@ -246,6 +246,57 @@ function parse_document(container)
...
+		}
+	}
+	
+	const response = await fetch(elem.getAttribute('data-url'), {
+		method: method,
+		body: params,
+	});
+	if (!response.ok)
+	{
+		phpbb.alert(response.status, response.statusText);
```
marc1706, line 270: Was the idea to just output the status id and the text without any language parsing?

phpBB/adm/style/admin.js
```diff
@@ -246,6 +246,57 @@ function parse_document(container)
...
+		return;
+	}
+	
+	const resp = await response.json();
+	switch (resp.type)
+	{
+		case 'error':
+			phpbb.alert(resp.status, resp.error);
+			return;
```
marc1706, line 279: I guess you don't like break?

phpBB/config/default/container/services_updater.yml
```diff
@@ -11,4 +11,10 @@ services:
         arguments:
             - '@filesystem'
             - '@updater.get_updates'
+            - '@language'
```
marc1706, line 14: we typically try to order these alphabetically

phpBB/includes/acp/acp_update.php
```diff
@@ -39,6 +39,9 @@ function main($id, $mode)
 		{
 			$recheck = $request->variable('versioncheck_force', false);
 			$do_update = $request->variable('do_update', false);
+			$token = $request->variable('form_token', false);
```
marc1706, line 42: Since token is a string shouldn't we kind of default to ''?

phpBB/phpbb/update/controller.php
```diff
@@ -59,8 +74,13 @@ public function __construct(
 	 */
 	public function handle(string $download): array
 	{
+		if ($this->user->data['user_type'] != USER_FOUNDER)
+		{
+			throw new http_exception(403);
```
marc1706, line 79: You might want to add a language string for this:
`throw new http_exception(403, 'INSTALL_PHPBB_NOT_INSTALLED');`

phpBB/phpbb/update/update_info.php
marc1706, line 1: I wonder if we should maybe have these strings in a container parameter file

phpBB/phpbb/update/update_info_provider_interface.php
marc1706, line 1: also not sure if this is actually interface worthy :D

phpBB/phpbb/update/controller.php
CHItA: Maybe, but it should be inaccessible for other users anyway - as in link hidden, so probably you called this thing by hand, in which case no information for you.

phpBB/phpbb/update/update_info.php
CHItA: i think that would make it harder to mock? the purpose here is to have them modifiable for test cases.

phpBB/phpbb/update/update_info_provider_interface.php
CHItA: it is not, but if you want to change them in the tests, this seems reasonable. although we could do mock objects as well i guess.

phpBB/adm/style/admin.js
CHItA: yeah, pretty much.

phpBB/adm/style/admin.js
CHItA: i have no problem with it, but you may or may not want further code to execute based on the status, hence return seemed more fitting.

phpBB/config/default/container/services_updater.yml
CHItA: no

phpBB/includes/acp/acp_update.php
CHItA: it is easier to do the check if the default is false, does this break any parsing/validation?

PR #6875, opened 2025-10-02
[ticket/17562] Fix prosilver base font-size

cabot:
Checklist:

- [ ] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-17562

### DIFF
--- a/phpBB/styles/prosilver/theme/common.css
+++ b/phpBB/styles/prosilver/theme/common.css
@@ -5,8 +5,6 @@
 /* stylelint-disable selector-no-qualifying-type */
 
 html {
-	font-size: 100%;
-
 	/* Always show a scrollbar for short pages - stops the jump when the scrollbar appears. non-IE browsers */
 	height: 101%;
 }


phpBB/styles/prosilver/theme/common.css
```diff
@@ -5,8 +5,6 @@
 /* stylelint-disable selector-no-qualifying-type */
 
 html {
-	font-size: 100%;
```
iMattPro, line 8: Removing this doesn't change anything. But setting it to **85%** rescales the base font size for the whole forum to _normal_ sizing

cabot: As explained in the [ticket](https://tracker.phpbb.com/browse/PHPBB-17562):
> The 100% value on the root element takes the default size of the user agent (usually 16px, unless configured differently). Consequently, the final size applied will be that provided by the agent (100%) and not the value declared in base.css. The rem units will be calculated from this final size.
> 
> Reproduction of the bug:
> declare a different value on --ps-font-base => no change in font size or derived values is applied.

cabot: I realised this problem when porting one of my styles to phpBB4. When I re-declared a different font size on the `--ps-font-base` var, it had no effect because the declaration in common.css overrode the variable's value.
```
:root {
	--ps-font-base: clamp(0.875rem, 0.8315rem + 0.2174vw, 1rem);
	...
}
```

cabot: So, of course, it is possible to redefine a font size on `html`, but it would seem more logical to be able to use the `--ps-font-base` variable directly.

PS: Not really a bug, but rather CSS logic.

PR #6876, opened 2025-10-04
[ticket/15085] Add back http authentication support for feeds

marc1706:
Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-15085

### DIFF
--- a/phpBB/phpbb/feed/event/http_auth_subscriber.php
+++ b/phpBB/phpbb/feed/event/http_auth_subscriber.php
@@ -1,0 +1,242 @@
+<?php
+/**
+*
+* This file is part of the phpBB Forum Software package.
+*
+* @copyright (c) phpBB Limited <https://www.phpbb.com>
+* @license GNU General Public License, version 2 (GPL-2.0)
+*
+* For full copyright and license information, please see
+* the docs/CREDITS.txt file.
+*
+*/
+
+namespace phpbb\feed\event;
+
+use phpbb\auth\auth;
+use phpbb\config\config;
+use phpbb\request\request_interface;
+use phpbb\user;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+/**
+ * Event subscriber for HTTP authentication on feed routes
+ */
+class http_auth_subscriber implements EventSubscriberInterface
+{
+	/** @var auth */
+	protected $auth;
+
+	/** @var config */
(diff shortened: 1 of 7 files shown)


phpBB/phpbb/feed/event/http_auth_subscriber.php
```diff
@@ -0,0 +1,211 @@
...
+		$route = $request->attributes->get('_route');
+
+		// Only apply to feed routes
+		if (strpos($route, 'phpbb_feed_') !== 0)
+		{
+			return;
+		}
+
+		// Only allow HTTP authentication in secure context (HTTPS)
+		if (!$request->isSecure())
```
Derky, line 88: Shouldn't we connect this to the HTTPS config for the board? Because with this hard coded check it's harder to test locally and login data is also not encrypted on login forms on boards that don't have HTTPS configured.

phpBB/phpbb/feed/event/http_auth_subscriber.php
```diff
@@ -0,0 +1,211 @@
...
+		if ($auth_result['status'] == LOGIN_SUCCESS)
+		{
+			// Reload ACL for the newly logged-in user
+			$this->auth->acl($this->user->data);
+			return;
+		}
+		else if ($auth_result['status'] == LOGIN_ERROR_ATTEMPTS)
+		{
+			// Too many login attempts
+			$response = new Response($this->language->lang('NOT_AUTHORISED'), Response::HTTP_UNAUTHORIZED);
```
Derky: I think you need a different language key, like `LOGIN_ERROR_ATTEMPTS`. That also states you need to solve a CAPTCHA which I think still needs to be done right? (However not possible through basic auth login)

marc1706: Yeah, that states that you'll need to solve a captcha which does not work for this. Hence why we're only showing the unauthorized message instead.

phpBB/phpbb/feed/event/http_auth_subscriber.php
marc1706: I think this might actually be the safer bet. Also, you should be able to test this with `X-Forwarded-Proto` set to `https`

phpBB/phpbb/feed/event/http_auth_subscriber.php
Derky: But it's actually correct that they need to solve a captcha. If your RSS reader suddenly gets an NOT_AUTHORISED out of the blue it can be really hard to find the cause.

marc1706: Replaced the output with the error message `LOGIN_ERROR_ATTEMPTS` now as requested.

PR #6927, opened 2026-01-25
[ticket/17609] Remove jumpbox

rubencm:
PHPBB-17609

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-17609

### DIFF
--- a/phpBB/docs/events.md
+++ b/phpBB/docs/events.md
@@ -2348,13 +2348,6 @@ search_results_header_before
 * Since: 3.1.4-RC1
 * Purpose: Add content before the header of the search results.
 
-search_results_jumpbox_before
-===
-* Locations:
-    + styles/prosilver/template/search_results.html
-* Since: 3.3.4-RC1
-* Purpose: Add content before the jumpbox of the search results.
-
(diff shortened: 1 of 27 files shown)


marc1706:
> > Can you also add the deprecated tag to the make_jumpbox function?
> 
> Is not, is still used by mcp

Ah, missed that part. Thanks.

bennybernaer:
I understand the intention to simplify the interface, but the jumpbox does not provide exactly the same functionality as the breadcrumbs.

Breadcrumbs only allow users to navigate back through the current forum hierarchy, while the jumpbox allows them to move directly to any other forum. This can still be very useful on boards with many forums or long topics.

Would it be possible to keep the jumpbox as an optional feature, disabled by default, instead of removing the ACP setting and regular-page support completely? This would keep the default interface clean while still allowing board administrators and style authors to enable it when needed.

rubencm:
It shouldn't be hard to create an extension for this.

Also, everyone wants to have big changes with phpbb4, and every time I try to change something there is people against it

PR #6955, opened 2026-04-27
[ticket/17604] Prevent disabling teamsecurity extension

iMattPro:
PHPBB-17604

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-17604

### DIFF
--- a/phpBB/phpbb/db/migration/data/v400/disable_extensions.php
+++ b/phpBB/phpbb/db/migration/data/v400/disable_extensions.php
@@ -18,10 +18,11 @@ use phpbb\db\migration\container_aware_migration;
 class disable_extensions extends container_aware_migration
 {
 	/**
-	 * @var array List of extensions included with phpBB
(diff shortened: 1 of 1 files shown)


iMattPro:
The issue - the **teamsecurity** extension cannot be disabled without removing the `ext.php` from the filesystem. That file calls a `trigger_error` during the disable step. That causes an infinite loop when this disable-extensions-migration runs.

This is likely the cleanest fix - whitelist **teamsecurity** like we already do with **viglink**, in this migration.

Otherwise, the only other real fix I have found is more substantial (requiring changes in this file, the ACP, and the extension) :
- Replace the `trigger_error` in the **teamsecurity** extension with a thrown Exception
- Catch the Exception in this migration and silently continue
- Catch the Exception in the ACP and display the exception message

marc1706 (approved):

PR #6960, opened 2026-05-18
[ticket/17656] Add shorter guest session length and new AI bots

marc1706:
Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-17656


phpBB/language/en/acp/board.php
```diff
@@ -329,20 +352,26 @@
...
+	'COOKIE_SECURE_EXPLAIN'			=> 'If your server is running via SSL set this to enabled else leave as disabled. Having this enabled and not running via SSL will result in server errors during redirects.',
+	'ONLINE_LENGTH'					=> 'View online time span',
+	'ONLINE_LENGTH_EXPLAIN'			=> 'Number of minutes after which inactive users will not appear in “Who is online” listings. The higher this value the greater is the processing required to generate the listing.',
+	'SESSION_LENGTH'				=> 'Session length',
+	'SESSION_LENGTH_EXPLAIN'		=> 'Sessions will expire after this time, in seconds.',
+	'SESSION_GC'					=> 'Session garbage collection time',
+	'SESSION_GC_EXPLAIN'			=> 'Sessions will be garbage collected after this time, in seconds.',
+	'SESSION_GUEST_LENGTH'			=> 'Guest session length',
+	'SESSION_GUEST_LENGTH_EXPLAIN'	=> 'Guest sessions will expire after this time, in seconds.',
+	'SESSION_GUEST_GC'				=> 'Guest session garbage collection time',
```
Derky: Garbage collection is not really known by most people. Maybe say something about what is happening, like guests will lose their session after this time.

phpBB/phpbb/db/migration/data/v33x/ai_crawler_bots.php
```diff
@@ -0,0 +1,153 @@
...
+	{
+		return [
+			['custom', [[$this, 'add_bots']]],
+		];
+	}
+
+	public function add_bots(): void
+	{
+		$bots = [
+			// 'BotName' => 'Agent partial name',
```
Derky, line 41: Hmm, opposed to search bots leading traffic to my site, I don't think I want to have these bots in the same category since they copy contents, train their datasets and consume resources without contributing to anyrhing useful.

phpBB/includes/acp/acp_board.php
```diff
@@ -349,11 +333,14 @@ function main($id, $mode)
...
 					'title'	=> 'ACP_LOAD_SETTINGS',
 					'vars'	=> array(
-						'legend1'			=> 'GENERAL_SETTINGS',
-						'limit_load'		=> array('lang' => 'LIMIT_LOAD',		'validate' => 'int:0:9999',	'type' => 'number:0:9999', 'explain' => true),
-						'session_length'	=> array('lang' => 'SESSION_LENGTH',	'validate' => 'int:60:9999999999',	'type' => 'number:60:9999999999', 'explain' => true, 'append' => ' ' . $user->lang['SECONDS']),
-						'active_sessions'	=> array('lang' => 'LIMIT_SESSIONS',	'validate' => 'int:0:9999',	'type' => 'number:0:9999', 'explain' => true),
-						'load_online_time'	=> array('lang' => 'ONLINE_LENGTH',		'validate' => 'int:0:999',	'type' => 'number:0:999', 'explain' => true, 'append' => ' ' . $user->lang['MINUTES']),
+						'legend1'				=> 'GENERAL_SETTINGS',
+						'limit_load'			=> array('lang' => 'LIMIT_LOAD',			'validate' => 'int:0:9999',	'type' => 'number:0:9999', 'explain' => true),
+						'session_length'		=> array('lang' => 'SESSION_LENGTH',		'validate' => 'int:60:9999999999',	'type' => 'number:60:9999999999', 'explain' => true, 'append' => ' ' . $user->lang['SECONDS']),
```
Derky: Maybe use fool proof max values here.

phpBB/language/en/acp/board.php
marc1706: "Guest session purge time"?

phpBB/phpbb/db/migration/data/v33x/ai_crawler_bots.php
marc1706: What do you mean with leading traffic to your site? This will cause these to show up as Bots visiting your site and the bots will see less user-specific links and therefore should be less likely to crawl useless links or sites.

phpBB/includes/acp/acp_board.php
marc1706: Default is 3600, not sure what a sensible max length would be :)

phpBB/phpbb/cron/task/core/tidy_guest_sessions.php
```diff
@@ -0,0 +1,57 @@
...
+	*/
+	public function run()
+	{
+		$this->user->session_guest_gc();
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function should_run(): bool
```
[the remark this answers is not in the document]
marc1706: Completely not true, thanks for hallunicating. This is perfectly valid in PHP's type inheritance.

phpBB/language/en/acp/board.php
```diff
@@ -329,20 +352,26 @@
...
+	'COOKIE_NOTICE_EXPLAIN'			=> 'If enabled a cookie notice will be displayed to users when visiting your board. This might be required by law depending on the content of your board and enabled extensions.',
+	'COOKIE_PATH'					=> 'Cookie path',
+	'COOKIE_PATH_EXPLAIN'			=> 'This will usually be the same as your script path or simply a slash to make the cookie accessible across the site domain.',
+	'COOKIE_SECURE'					=> 'Cookie secure',
+	'COOKIE_SECURE_EXPLAIN'			=> 'If your server is running via SSL set this to enabled else leave as disabled. Having this enabled and not running via SSL will result in server errors during redirects.',
+	'ONLINE_LENGTH'					=> 'View online time span',
+	'ONLINE_LENGTH_EXPLAIN'			=> 'Number of minutes after which inactive users will not appear in “Who is online” listings. The higher this value the greater is the processing required to generate the listing.',
+	'SESSION_LENGTH'				=> 'Session length',
+	'SESSION_LENGTH_EXPLAIN'		=> 'Sessions will expire after this time, in seconds.',
+	'SESSION_GC'					=> 'Session garbage collection time',
```
Crizz0: Same change as SESSION_GUEST_GC necessary

phpBB/language/en/acp/board.php
Crizz0: or "Guest session cleanup interval"?

marc1706: Changed it to the version proposed by Crizz0

phpBB/language/en/acp/board.php
marc1706: Done

phpBB/includes/acp/acp_board.php
Crizz0: In other settings we use 99.999 minutes e.g. for time span for post edits. So roughly 70 days. 

I would use here like a maximum of 1 day, ca. 80k to 99k seconds.

phpBB/phpbb/db/migration/data/v33x/ai_crawler_bots.php
marc1706: You could even ban them if you don't like them based on this :)

phpBB/includes/acp/acp_board.php
marc1706: Updated

phpBB/install/schemas/schema_data.sql
```diff
@@ -285,6 +285,8 @@ INSERT INTO phpbb_config (config_name, config_value) VALUES ('server_port', '');
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('server_protocol', '');
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('session_gc', '3600');
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('session_length', '3600');
+INSERT INTO phpbb_config (config_name, config_value) VALUES ('session_guest_length', '300');
```
Derky: Could these 5 minutes lead to problems if you first click around on a forum and then start the registration process? E.g. session token invalid errors.

phpBB/phpbb/db/migration/data/v33x/add_ai_crawlers_group.php
```diff
@@ -0,0 +1,73 @@
...
+		}
+
+		if (!function_exists('group_create'))
+		{
+			include($this->phpbb_root_path . 'includes/functions_user.' . $this->php_ext);
+		}
+
+		$group_id = 0;
+		$group_attributes = [
+			'colour'			=> '7D9EAE',
```
Derky:
```
Fatal error: SQL ERROR [ mysqli ]<br /><br />Unknown column 'colour' in 'INSERT INTO' [1054]<br /><br />An SQL error occurred while fetching this page. 
```

You need `group_colour`, `group_legend`, `group_max_recipients` etc.

Derky (changes requested):

phpBB/phpbb/db/migration/data/v33x/add_ai_crawlers_group.php
```diff
@@ -0,0 +1,73 @@
...
+			'legend'			=> 0,
+			'max_recipients'	=> 5,
+			'founder_manage'	=> 0,
```
ECYaz: Reproduced Derky's fatal on a 3.3.17 board (MySQL 8.0) by running `db:migrate` with this PR's migrations:

```
Fatal error: SQL ERROR [ mysqli ] Unknown column 'colour' in 'field list' [1054]
```

`group_create()` merges `$group_attributes` verbatim into the column list of the `INSERT` into the groups table (`includes/functions_user.php`), so the keys need to be the real column names:

```suggestion
			'group_colour'			=> '7D9EAE',
			'group_legend'			=> 0,
			'group_max_recipients'	=> 5,
			'group_founder_manage'	=> 0,
```

With just that change, `db:migrate` completes on 3.3.17 and the result matches the fresh-install path in `schema_data.sql`: group row created as `group_type` 3, `group_colour` 7D9EAE, `group_legend` 0, `group_max_recipients` 5, and `ai_crawler_bots` then registers all 26 crawlers linked to the group.

phpBB/install/schemas/schema_data.sql
ECYaz: I tested this on a 3.3.17 board with the PR's runtime changes applied (guest GC cron wired up, `session_guest_length`/`session_guest_gc` at the default 300), driving registration over HTTP as a real guest. Two scenarios:

**Browse, go idle > 5 minutes, then start registration — works fine.** Loading the registration page creates a fresh guest session, and the agreement + form tokens are issued against it. Registration completes normally.

**Start registration, then take > 5 minutes on the form — breaks, and silently.** If the guest GC runs between loading the registration form and pressing Submit, the guest's session row is deleted. With the default `form_token_sid_guests = 1`, guest form tokens are keyed to the session id (`check_form_key()` in `includes/functions.php`), so the submit — even with a correct CAPTCHA answer — fails the token check, `$agreed` resets to false, and the user is bounced back to the agreement page with everything they typed discarded. No error message is shown at all. I verified the same submit succeeds when the session survives, so the token/session loss is the only cause.

The same window exists today with the normal hourly GC and `session_length` = 3600 — the PR doesn't introduce the failure mode, but it shrinks the time to trigger it from ~1 hour to ~5 minutes, which is well within the time a slower user needs to read the agreement and fill in the form. It also isn't registration-specific: any session-bound guest form (guest posting where permitted, the contact form) has the same exposure.

A few possible directions, in case it helps the discussion: a larger default for `session_guest_length` (e.g. 15–30 min) would keep most of the table-size benefit while making mid-form expiry unlikely; alternatively the guest GC could skip sessions that have a pending row in the confirm table (i.e. a guest mid-CAPTCHA-flow), which targets exactly the sessions that are actually in the middle of something.

One small observation from the same test: `session_guest_gc()` deletes the session rows but doesn't run the captcha factory's garbage collection like `session_gc()` does, so confirm rows for purged guest sessions linger until the hourly GC sweeps them. Harmless, but worth knowing it's intentional.

marc1706: I've extended this to 15 minutes now

phpBB/includes/acp/acp_bots.php
```diff
@@ -291,6 +298,40 @@ function main($id, $mode)
 								user_update_name($row['bot_name'], $bot_row['bot_name']);
 							}
 
+							// Update group if it contains valid value
+							$bot_group_name = $request->variable('bot_group', 'BOTS');
+							if ($bot_group_name === 'BOTS' || $bot_group_name === 'AI_CRAWLERS')
```
Derky: If you create an array of these bot names, you can reuse it in the query and form validation and it's easier to expand later on with new options.

Derky (approved):

PR #6972, opened 2026-06-28
[ticket/17550] Log uncaught server errors to the PHP error log

ECYaz:
Uncaught exceptions that result in an HTTP 500 response were rendered as the generic error page but were never logged anywhere, so they could not be diagnosed (PHPBB-17550).

This logs server errors (HTTP 5xx) from `kernel_exception_subscriber` to the PHP error log, including the exception message and stack trace. Client errors such as 404/403 are intentionally not logged, to avoid noise. `error_log()` is used rather than the database-backed log because the database may itself be unavailable during a 500.

Note: the same gap exists in the analogous 3.3.x exception subscriber. I've targeted master because the ticket is filed against 4.0.0-a1 and the symptom is the Symfony 4.x error page, but I'm happy to provide a 3.3.x backport if the team prefers.

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-17550

### DIFF
--- a/phpBB/phpbb/event/kernel_exception_subscriber.php
+++ b/phpBB/phpbb/event/kernel_exception_subscriber.php
@@ -142,6 +142,20 @@ class kernel_exception_subscriber implements EventSubscriberInterface
 			$response->headers->add($exception->getHeaders());
 		}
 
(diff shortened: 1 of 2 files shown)


rubencm:
#6361

PR #7001, opened 2026-07-22
[ticket/17627] Carry topic views over when merging posts deletes the topic

ECYaz:
Merging topics from the MCP forum view (merge_topics) has kept the highest view count of the merged topics since PHPBB-14795. Merging a whole topic into another one via the MCP topic view or the quickmod "Merge topic" option (merge_posts) never received the same treatment: when all of a topic's posts are moved and the emptied topic is deleted, its view count is silently lost.

This applies the established behaviour to merge_posts(): the destination topic takes the maximum of both topics' view counts, and only when the source topic was fully merged away. Partially merged topics keep their counts untouched. Using the maximum rather than the sum follows the explicit decision made in PHPBB-14795.

Covered by a new functional test that performs the full merge through the MCP and asserts the destination topic inherits the higher view count once the source topic is gone.

Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-17627

### DIFF
--- a/phpBB/includes/mcp/mcp_topic.php
+++ b/phpBB/includes/mcp/mcp_topic.php
@@ -817,9 +817,11 @@ function merge_posts($topic_id, $to_topic_id)
 	}
 
 	$sync_forums = array();
+	$topic_views = 0;
 	foreach ($topic_data as $data)
 	{
 		$sync_forums[$data['forum_id']] = $data['forum_id'];
+		$topic_views = max($topic_views, $data['topic_views']);
 	}
 
 	$topic_data = $topic_data[$to_topic_id];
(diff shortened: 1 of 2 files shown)


Neustradamus:
@ECYaz: Thanks for your PR, I have not seen before!

It follows:
- https://www.phpbb.com/community/viewtopic.php?t=2669962
- https://tracker.phpbb.com/browse/PHPBB-17627

So 1000 views for the first and 500 views for the second, after merging, it will be 1500 views?

ECYaz:
> @ECYaz: Thanks for your PR, I have not seen before!
> 
> It follows:
> 
> * https://www.phpbb.com/community/viewtopic.php?t=2669962
> * https://tracker.phpbb.com/browse/PHPBB-17627
> 
> So 1000 views for the first and 500 views for the second, after merging, it will be 1500 views?

Sorry for the confusion but no. It will be 1000. The patch takes the higher of the two counts rather than the sum, matching what merge_topics() already does. 

The team had made a decision and that behaviour was settled in PHPBB-14795: many views come from the same readers visiting both topics, so adding the counters would overstate the real number.

Neustradamus:
@ECYaz: So, it is not a patch for my ticket.

The request is to have a merging of number of views.

It is for this, if the solution is not solved by default, I have proposed to add a checkbox to fix/merge the number of views.

PR #7007, opened 2026-07-24
[ticket/17668] Prevent managing special groups from the UCP

ECYaz:
Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-17668

Group leaders could manage special groups from the UCP without any ACP permission check, so a leader of the Administrators group was able to add new admins. Special groups now have to be managed from the ACP and are no longer listed on the UCP manage groups page. The existing UCP group tests managed the Administrators group directly, they now use regular groups led by the admin user instead.


marc1706:
Is there a reason why this is targeting master and not 3.3.x?

PR #7012, opened 2026-07-27
[ticket/17665] Add automatic admin notifications on security updates

marc1706:
Checklist:

- [x] Tests pass

Tracker ticket:

https://tracker.phpbb.com/browse/PHPBB-17665


ECYaz:
Some findings from looking into the red CI, verified locally against this branch (phpunit 9.6.35, MySQL 8 strict mode, PostgreSQL 16):

**1. The whole matrix is red from one failure.** `migrations_check_config_added_test` fails because the two new config options are missing from `phpBB/install/schemas/schema_data.sql`, and fail-fast then cancels the other jobs. Since the cron rewrites `version_check_last_cron` on every run, it should also be dynamic like the other cron timestamps (`cron_lock`, the `*_last_gc` ones), otherwise each run purges the config cache:

    INSERT INTO phpbb_config (config_name, config_value) VALUES ('version_check_interval', '60');
    INSERT INTO phpbb_config (config_name, config_value, is_dynamic) VALUES ('version_check_last_cron', '0', 1);

with the matching third argument in the migration:

    ['config.add', ['version_check_last_cron', 0, true]],

**2. The generated `item_id` does not fit the column on any supported DBMS.** In `version_check::notify_admins()`:

    'item_id' => (int) sprintf('%u', crc32($template . $update_data['current'])),

`notifications.item_id` is `UINT`, which is `mediumint` (max 16777215) on MySQL and `INT4` (max 2147483647) on PostgreSQL. Nearly every crc32 value exceeds mediumint, so the insert fails with "Out of range value" under MySQL strict mode, and about half also exceed INT4 on PostgreSQL. The very next security notification on this branch would produce `crc32('update_security' . '3.3.18') = 3943736323`, which I verified fails the insert on both MySQL 8 and PostgreSQL 16. Masking the hash to 24 bits keeps it inside every mapping, including Oracle's `number(8)`:

    'item_id' => crc32($template . $update_data['current']) & 0xFFFFFF,

The hardcoded `1089886753` in `test_run_with_updates` becomes `16144929` with the mask. The tests do not catch this today because they only use small fixed ids.

**3. The migration has no `depends_on()`**, so the migrator is free to run it at any point in the tree:

    public static function depends_on(): array
    {
        return ['\phpbb\db\migration\data\v33x\v3317'];
    }

**4. Failed checks retry on every cron trigger.** `run()` only sets `version_check_last_cron` on success, and `get_update_on_branch(true)` bypasses the versioncheck cache, so while the version server is unreachable every eligible cron trigger performs a blocking remote fetch. Setting the timestamp in the catch block as well avoids hammering both ends (`test_run_with_exception` would then assert greater than 0 instead of 0). There is also a leftover `// @todo` in that catch.

Minor: `$db` and `$user` are injected into the cron task but not used.

phpBB/install/schemas/schema_data.sql
```diff
@@ -320,6 +318,8 @@ INSERT INTO phpbb_config (config_name, config_value) VALUES ('upload_icons_path'
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('upload_path', 'files');
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('use_system_cron', '0');
 INSERT INTO phpbb_config (config_name, config_value) VALUES ('version', '3.3.18-dev');
+INSERT INTO phpbb_config (config_name, config_value) VALUES ('version_check_interval', '60');
```
Derky, line 321: Do you happen to know how often the version is currently cached after someone visits the ACP? This change may lead to an unwanted DDoS attack on our version check servers.

(Config setting also in phpBB/phpbb/db/migration/data/v33x/add_version_check_cron.php)

phpBB/language/en/email/update_urgent.txt
```diff
@@ -0,0 +1,14 @@
+Subject: "{SITENAME}" - phpBB urgent update required
```
Derky: I think "critical update" will cover the urgency better and makes it also more clear it's even higher than a security update.

phpBB/language/en/email/update_maintenance.txt
```diff
@@ -0,0 +1,14 @@
+Subject: "{SITENAME}" - phpBB update available
```
Derky: I'd suggest to add {NEW_VERSION} in the subject line, like:
`Subject: "{SITENAME}" - phpBB {NEW_VERSION} update available`

Also for the other two emails.

phpBB/language/en/email/update_maintenance.txt
```diff
@@ -0,0 +1,14 @@
...
+
+Hello {USERNAME},
+
+This is a notification to inform you that a new update is available for your phpBB installation.
```
Derky, line 5: I think we should add an intro why this user is receiving this email, like;
`You're receiving this notification because you are an administrator of "{SITENAME}".`

(Also for the other two emails)

phpBB/language/en/email/update_maintenance.txt
```diff
@@ -0,0 +1,14 @@
...
+
+Your current version is {CURRENT_VERSION}, and the latest available version is {NEW_VERSION}. It is highly recommended that you keep your forum up to date to ensure the highest level of security, stability, and performance.
+
+You can view the release details in the Admin Control Panel of your phpBB installation or by visiting the official phpBB website.
```
Derky: I think if we don't link to the phpBB website we should also not mention it. I think referring to the ACP is fine, it will provide a link to phpbb.com there as well.

(Also for the other two emails)

phpBB/phpbb/notification/type/update_maintenance.php
```diff
@@ -0,0 +1,156 @@
...
+	/**
+	* {@inheritdoc}
+	*/
+	public function get_title()
+	{
+		$template = $this->get_data('template');
+		$new_version = $this->get_data('new_version');
+		$current_version = $this->get_data('current_version');
+
+		if ($template === 'update_security')
```
Derky: Reusable constants would be nice for these.

phpBB/phpbb/notification/type/update_maintenance.php
```diff
@@ -0,0 +1,156 @@
...
+		{
+			return $this->language->lang('NOTIFICATION_UPDATE_SECURITY', $current_version, $new_version);
+		}
+
+		if ($template === 'update_urgent')
+		{
+			return $this->language->lang('NOTIFICATION_UPDATE_URGENT', $current_version, $new_version);
+		}
+
+		return $this->language->lang('NOTIFICATION_UPDATE_MAINTENANCE', $current_version, $new_version);
```
Derky, line 112: Maybe just set a `$title` variable in the if statement and only have the actual return on this line.

A switch case may even look nicer.
```
switch ($template)
{
	case 'update_security':
		$title = 'NOTIFICATION_UPDATE_SECURITY';
		break;

	case 'update_urgent':
		$title = 'NOTIFICATION_UPDATE_URGENT';
		break;

	default:
		$title = 'NOTIFICATION_UPDATE_MAINTENANCE';
		break;
}

return $this->language->lang($title, $current_version, $new_version);
```


Too bad we can't use match() yet because of the minimum supported PHP version.

tests/cron/version_check_test.php
```diff
@@ -0,0 +1,295 @@
...
+	{
+		$task = $this->get_task();
+		$this->assertTrue($task->is_runnable());
+	}
+
+	public function should_run_data()
+	{
+		return [
+			'never_run_before' => [0, 1440, true],
+			'just_run' => [time(), 1440, false],
```
Derky, line 93: Why is the third parameter `$expected` set to `false` if it should "_just run_"?

phpBB/install/schemas/schema_data.sql
marc1706: This should currently be 24 hours whereas this is set to 60 minutes so once an hour. We could also increase this to a higher number.

tests/cron/version_check_test.php
marc1706: I think that might have been confusing, it just "ran". Last run was now so it shouldn't run now.

phpBB/language/en/email/update_urgent.txt
marc1706: Done, renamed to critical

phpBB/language/en/email/update_maintenance.txt
marc1706: Removed

phpBB/phpbb/notification/type/update_maintenance.php
marc1706: added those

phpBB/language/en/email/update_maintenance.txt
marc1706: Honestly I find that a bit weird. You're an admin of said forum, you should know why you're getting this mail.

phpBB/phpbb/notification/type/update_maintenance.php
marc1706: I think it's tomato tomato, not really much difference here. Maybe the if is faster? :D

phpBB/install/schemas/schema_data.sql
Derky: I've talked with Paul and he expects the version check servers can handle this without a problem.

phpBB/language/en/email/update_maintenance.txt
Derky: Ok leave it out then.

phpBB/language/en/common.php
```diff
@@ -510,6 +509,9 @@
 	'NOTIFICATION_TOPIC_IN_QUEUE'		=> '<strong>Topic approval</strong> request by %1$s:',
 	'NOTIFICATION_TYPE_NOT_EXIST'		=> 'The notification type "%s" is missing from the file system.',
 	'NOTIFICATION_ADMIN_ACTIVATE_USER'	=> '<strong>Activation required</strong> for deactivated or newly registered user: “%1$s”',
+	'NOTIFICATION_UPDATE_CRITICAL'		=> '<strong>phpBB critical update required</strong>: Your board is running %1$s and a newer secure version %2$s is available.',
```
Derky: Maybe also add the urgency in the explanation.

 "Your board is running %1$s and a newer critical version %2$s is available Please update as soon as possible to keep your board secure."

tests/cron/version_check_test.php
Derky: Oh... that wasn't clear for me when reading the code.

Derky:
I'm having troubles testing this, current feedback
- Editing the`version_check_interval` doesn't seem to help. Tried it on one minute and deleted cache multiple times.
- It only returns a notifcation, no mail by default
- After enabling an email notification it still doesn't send out a mail
- The content of the notification is not escaped, I can add HTML to the version that gets parsed

marc1706:
Updated the PR. Cron is working fine now and the default admin will receive the notification via mail by default. When running the migration, all admins with appropriate permissions should get the same setting. I also double checked and the version check data gets verified against the schema and no HTML can be added to inject HTML.

phpBB/language/en/common.php
marc1706: Extended the text

CHItA (approved):

Derky (approved):

